PR #381, opened 2017-03-24
CRLF -> LF for everything

fpoussin:
_find . -type f -print0 | xargs -0 dos2unix_

I converted everything from CRLF to LF.
You are likely to break the merge if you push any commit to master before this.


rusefi:
I am trying to run this on a per-folder basis on my end, wonder if it reduces the number of changed files on your end?

rusefi:
just commited more of find_cygwin . -type f -print0 | xargs -0 dos2unix

let me know how that looks from your end, I will keep hardware as the last step

for the record - I've created a tag midway through changing this

git tag -a 20170334_mid_through_eol_migration -m "20170334_mid_through_eol_migration"
git push origin 20170334_mid_through_eol_migration

fpoussin:
Let's not duplicate efforts here...

I'm gonna close it since it's not possible to merge anymore, you can just run the command to fix the files like I did.

rusefi:
Right, but please confirm that what I am doing is helping? My concern is that I am executing all this on Windows and who knows how this goes with all the pieces of the local and remote magic.

So do you see progress?

rusefi:
I think I've converted everything

See also https://github.com/rusefi/rusefi/blob/master/misc/dos2unix_all_job.bat

PR #389, opened 2017-03-31
Fixed Posix simulator.

fpoussin:
Just the simulator fix this time.


rusefi:
-static in rules is needed by windows, the .exe does not run for me without it. Is it breaking things on unix?

fpoussin:
Yes, It doesn't work on linux with -static.

fpoussin:
That should do it0

PR #411, opened 2017-05-09
Odometer for RX8 cluster

thirstyone:
tested a bit today - seems to be working fine


rusefi:
Please re-implement using vssDebugEventCounter

Issue #421, opened 2017-05-23
convert serial port to DMA


rusefi:
In order to get reliable serial port we should migrate to DMA mode

We cannot afford one IRQ per byte while engine is running, we cannot even afford one light DMA IRQ per byte - we need DMA buffers, and looks like we would need to part ChibiOS to get those

rusefi:
See http://rusefi.com/forum/viewtopic.php?f=15&t=1233&start=30

See http://www.chibios.com/forum/viewtopic.php?f=25&t=3441&start=10

rusefi:
https://github.com/rusefi/ChibiOS/pull/1

rusefi:
https://github.com/rusefi/rusefi/pull/422/

rusefi:
![image](https://cloud.githubusercontent.com/assets/7377949/26381191/fda1ff24-3ff0-11e7-8416-29a5205e34e0.png)
@andreika-git ops

andreika-git:
A conflict of USART3 and SPI2. (STM32_SPI_SPI2_RX_DMA_STREAM == STM32_UART_USART3_TX_DMA_STREAM).

rusefi:
https://github.com/rusefi/rusefi/pull/425

andreika-git:
It seems that the conflict is generally unsolvable for STM32F4:
http://www.chibios.com/forum/viewtopic.php?t=336#p2843
http://www.chibios.com/forum/viewtopic.php?t=1240#p11334
It's a hardware limitation (see "DMA1 request mapping" in RM0090 reference manual).

The good news is it's USART3_TX alone that's causing the conflict, not Rx. And we mostly need DMA for receiving. So the only way I see in this situation - is to use DMA only for USART3-Rx, and 'Serial' interrupts for USART3-Tx. It's like a hybrid of Serial and Uart drivers - Rx half from Uart and Tx half from Serial... Maybe we need to create a separate UART driver with our new DMA code, which would have two switchable modes - full-dma and 'hybrid'. 
As for now, users will have to choose between SPI2 (the default) and USART3-DMA in their configs...

rusefi:
#define STM32_SPI2_RX_DMA_MSK               STM32_DMA_STREAM_ID_MSK(1, 3)
#define STM32_SPI2_TX_DMA_MSK               STM32_DMA_STREAM_ID_MSK(1, 4)


[...]

PR #425, opened 2017-05-24
Swap USART2 & USART3 DMA channels

andreika-git:


rusefi:
image

andreika-git:
We need to roll back this commit. The problem is far worse than I thought... See my comment on #421

PR #432, opened 2017-05-30
Sr5 read data timeout

andreika-git:


firmware/console/binary/tunerstudio_io.h
```diff
@@ -101,6 +101,7 @@ void sr5WriteData(ts_channel_s *tsChannel, const uint8_t * buffer, int size);
 void sr5WriteCrcPacket(ts_channel_s *tsChannel, const uint8_t responseCode, const void *buf, const uint16_t size);
 void sr5SendResponse(ts_channel_s *tsChannel, ts_response_format_e mode, const uint8_t * buffer, int size);
 int sr5ReadData(ts_channel_s *tsChannel, uint8_t * buffer, int size);
+int sr5ReadDataTimeout(ts_channel_s *tsChannel, uint8_t * buffer, int size, int timeout);
```
rusefi: are there any usages of this new method? why do we need this method?

andreika-git: Yes, it's used by the bootloader code (we need to control the timeout value depending on the state of the bootloader). I'll post those PRs soon.

PR #434, opened 2017-05-30
Bootloader

andreika-git:


rusefi:
and now we need a few kilobytes of documentation :)

readme.txt somewhere? links to DFU utility? workflow user documentation.

and implementation overview - how does it work, how does it enter DFU mode, why two Makefiles etc etc

please understand that I need to understand what's going on

firmware/bootloader/src/rusefi_stubs.cpp
```diff
@@ -0,0 +1,29 @@
+#include "main.h"
+#include "hardware.h"
+#include "efiGpio.h"
+#include "global.h"
+
```
rusefi: this file needs an explanation - why do we need stubs? which methods need stubs?

andreika-git: We need only a small portion of code from rusEFI codebase in the bootloader. Mostly it's tunerstudio_io.cpp. And other files like efiGpio.cpp etc. needed only to make it work. And stubs needed just to settle down compiler errors. 
The whole idea of bootloader is to make it as small as possible. And reasonably independent.

andreika-git:
I'll write readme.txt tomorrow.

andreika-git:
The bootloader works as follows:
- init ChibiOS and UART/Serial driver using tunerstudio_io code;
- create a thread to listen to UART (thBootloaderSerial), using dfuStartLoop();
- The PC 'stm32-flasher' software sends its request byte only once, so we don't wait for it - the bootloader sends an answer as soon as it starts (both the request & answer bytes are known consts);
- if the next command doesn't come immediately (<100 ms), we abort the bootloader dfu loop and run the application code - calling dfuJumpToApp() in main().
- otherwise, if at least one command is received, we stay in the bootloader mode and process commands.

andreika-git:
Added more info about using STMFlashLoader and comprehensive screenshots.

andreika-git:
Done?

PR #446, opened 2017-06-21
EFI_MAP_AVERAGING_ITB

andreika-git:


rusefi:
Can we please migrate to mapMinBufferLength?

andreika-git:
Maybe enable EFI_MAP_AVERAGING_ITB by default?

rusefi:
I am not sure it's really about EFI_MAP_AVERAGING_ITB (probably not specific to ITB), and I am reallly not excited about too many artificial compile-type parameters

I'd rather see everything as a dynamic parameter and rather go with reasonably universal

so instead of two branches of code with two options - min(1) and min(count_cylinders), it would be better to have one branch of code with min(x) where X could be 1, or count_cylinder, or 3, or 2...

firmware/controllers/map_averaging.cpp
```diff
@@ -193,6 +213,17 @@ static void mapAveragingCallback(trigger_event_e ckpEventType,
 		return;
 	}
 
+	if (boardConfiguration->mapMinBufferLength != mapMinBufferLength) {
+		// check range
+		mapMinBufferLength = maxI(minI(boardConfiguration->mapMinBufferLength, MAX_MAP_BUFFER_LENGTH), 0);
```
rusefi: how about maxI(,,,,, 1)? as long as mapMinBufferLength is at least 1, we would not need "mapMinBufferLengt > 1" checks since mapMinBufferLengt = 1 would work same exact was as mapMinBufferLengt = 2, and there would be no possibility for mapMinBufferLengt = 0?

this case we have a bit less code

firmware/controllers/map_averaging.cpp
```diff
@@ -193,6 +213,17 @@ static void mapAveragingCallback(trigger_event_e ckpEventType,
...
+		// reset index
+		averagedMapBufIdx = 0;
+		// fill with maximum values
+		for (int i = 0; i < MAX_MAP_BUFFER_LENGTH; i++) {
```
rusefi: technically could be "i < mapMinBufferLength " but not sure if that's any better, up to you :)

firmware/controllers/map_averaging.cpp
```diff
@@ -193,6 +213,17 @@ static void mapAveragingCallback(trigger_event_e ckpEventType,
...
+			averagedMapRunningBuffer[i] = CONFIG(mapErrorDetectionTooHigh);
```
rusefi: technically could be just SOME_HUGE_VALUE just to make code simpler by reducing dependency on "CONFIG(mapErrorDetectionTooHigh)"

andreika-git:
All done!

firmware/controllers/map_averaging.cpp
```diff
@@ -170,6 +178,18 @@ static void endAveraging(void *arg) {
 #if EFI_PROD_CODE || defined(__DOXYGEN__)
 	v_averagedMapValue = adcToVoltsDivided(
 			mapAccumulator / mapMeasurementsCounter);
+	if (mapMinBufferLength > 1) {
```
rusefi: I believev this line is not needed, now that 'mapMinBufferLength ' is always at least 1 I would prefer that the same code is executed even if mapMinBufferLength  == 1

firmware/controllers/map_averaging.cpp
```diff
@@ -240,8 +271,10 @@ float getMap(void) {
 #if EFI_ANALOG_SENSORS || defined(__DOXYGEN__)
 	if (!isValidRpm(engine->rpmCalculator.rpmValue))
 		return validateMap(getRawMap()); // maybe return NaN in case of stopped engine?
-	return validateMap(getMapByVoltage(v_averagedMapValue));
-#else
+	if (mapMinBufferLength > 1)
```
rusefi: I believe this if-statement is not needed, we can simplify the code by throwing both if-statements away...

firmware/controllers/map_averaging.cpp
andreika-git: Please be aware that this code is inside lockAnyContext()! It's all about optimization (the feature is rarely used).

andreika-git: I'll remove it if you insist.

rusefi: yes, you are right about one less % and one less for loop with zero invocation count

but in my opinion the benefits is in complexity reduction, I am sure that this code is not the bottleneck, we definitely have MUCH bigger ones somewhere else. So yes, let's make is simpler and more reliable as a consequence.

firmware/controllers/map_averaging.cpp
```diff
@@ -240,7 +269,7 @@ float getMap(void) {
 #if EFI_ANALOG_SENSORS || defined(__DOXYGEN__)
 	if (!isValidRpm(engine->rpmCalculator.rpmValue))
 		return validateMap(getRawMap()); // maybe return NaN in case of stopped engine?
-	return validateMap(getMapByVoltage(v_averagedMapValue));
+	return validateMap(minMapPressure);
```
rusefi: here 'minMapPressure' name is really just too much information, and in reality it;s 'min averaged' pressure

how about we rename to 'currentPressure' and add a one-line comment next to the declaration explaining that this is 'minimal averaged' MAP?

firmware/controllers/map_averaging.cpp
```diff
@@ -170,6 +178,16 @@ static void endAveraging(void *arg) {
 #if EFI_PROD_CODE || defined(__DOXYGEN__)
 	v_averagedMapValue = adcToVoltsDivided(
 			mapAccumulator / mapMeasurementsCounter);
+	// todo: move out of locked context?
+	averagedMapRunningBuffer[averagedMapBufIdx] = getMapByVoltage(v_averagedMapValue);
+	// increment circular running buffer index
+	averagedMapBufIdx = (averagedMapBufIdx + 1) % mapMinBufferLength;
+	// find min. value (only works for pressure values, not raw voltages!)
+	minMapPressure = averagedMapRunningBuffer[0];
```
rusefi: Do not we need to run this loop with a local variable? otherwise would not we have a chance to read the un-minimalized value once in a while as long as we read not under a lock?

we probably never read not under a lock in reality so we can either extract a local variable or add a comment clarifying that we are aware of this concurn but choose not to care since we believe we should not care.

Issue #442, opened 2017-06-12
ADC_VCC should be configurable


rusefi:
ADC_VCC could be either 3.0 or 3.3v depending on the regulator on your actual board

this should become configurable

rusefi:
See engineConfiguration->adcVcc

PR #478, opened 2017-08-30
Fix adc vcc

mck1117:
Fix issue #442 

Tested on a running car. Works properly.

What's the right solution for board_test.cpp? It doesn't have the concept of engineConfiguration, so should it just use a default of 3.0v or 3.3v?


rusefi:
See #442

PR #479, opened 2017-08-30
Fix adc vcc

mck1117:
Fix issue #442

Tested on a running car. Works properly.

What's the right solution for board_test.cpp? It doesn't have the concept of engineConfiguration, so should it just use a default of 3.0v like the previous setup?


rusefi:
because this is an incompatible change I cannot merge it at this point, probably one day but not sure when

mck1117:
Is it actually incompatible? Unless you change the voltage from the default (3.0v, same as the #define), this won't change behavior if flashed over the top of an older version.

The only risk is that the adcVcc field gets a bogus (probably a zero) because it used to be an unused field.  This could be mitigated by checking if adcVcc is a reasonable value, and if not, set it to 3.0v (the current default).

Something like:
`if(engineConfiguration->adcVcc > 4.0f || engineConfiguration->adcVcc < 1.4f) {	engineConfiguration->adcVcc = 3.0f;
}`

rusefi:
I've just commited something similar under #442

PR #489, opened 2017-11-09
Implement oil pressure sensor

mck1117:
Implement oil pressure sensor support.  Currently supports a simple linear-type voltage sensor, like the current wideband O2 sensor support.

Adds a configuration page in TunerStudio, similar to the one for EGO.  Adds an oil pressure gauge, and datalog column.

Shouldn't break configuration as it uses spare config space, and zeros are safe for all new config fields.


rusefi:
Can we please split this PR into at least two PRs? Looks like fixing the C/F symbols is unrelated from oil pressure.

The reason for that would be to make the review simpler. I am also hesitant about adding oil pressure as a top level feature, I am not sure if it justifies the custom code. I know we also have unfinished feature of linear temperature sensors - the whole linear part would be a similarity.

I need some time to think about it :) My concern is scope crepe. On the one hand this is simple code, on the other hand how many people need it and how often is it linear? Myself I was using FSIO for aux analog inputs.

mck1117:
The C/F symbol stuff isn't actually a diff.  My editor screwed it up in a previous commit, and I fixed it, so it claims it's a change, but if you look at the delta it's not actually changing anything.  I'm not sure how to get git to ignore those changes.

Plenty of cars and factory ECUs have analog oil pressure sensors, and they're generally the three-pin linear analog style (ground/power/signal).  I know that tons of GM vehicles use the same one, including most of their V6s, and the LS architecture engines.

Does FSIO have support for gauges, and/or logging of FSIO inputs/outputs? It looks like there aren't currently gauges for them in TunerStudio.

rusefi:
So if these lines are not real changes would it be nicer to have a PR without the artefacts of your editor screw up to reduce the chance of a screw up in master branch? :)

See also https://rusefi.com/forum/viewtopic.php?f=13&t=325 - can you please print & sign & email to rusefillc gmail address?

mck1117:
Cleaned up in #491.

PR #491, opened 2017-11-15
Implement oil pressure sensor

mck1117:
This is a cleanup of #489.

Implement oil pressure sensor support. Currently supports a simple linear-type voltage sensor, like the current wideband O2 sensor support.

Adds a configuration page in TunerStudio, similar to the one for EGO. Adds an oil pressure gauge, and datalog column.

Shouldn't break configuration as it uses spare config space, and zeros are safe for all new config fields.


rusefi:
@mck1117 please sign https://rusefi.com/forum/viewtopic.php?f=13&t=325

PR #499, opened 2017-11-25
Fix accel enrichment for negative fuel pulse

mck1117:
If the target fuel was very small, and there was lots of fuel on the wall, you could get an injector pulse of <0.  This will inject no fuel (like a pulse of 0ms would), but will more importantly screw up the estimate for how much fuel is on the wall.  If the pulse went negative, the model would update as though the injector had sucked fuel back in to the tank.


mck1117:
I don't have a printer until Monday, so I'll sign/send the contribution agreement then

PR #506, opened 2017-12-03
Implement Cranking to Running Adv.Angle Interpolation for faster start

andreika-git:
Tested.


andreika-git:
> I could be wrong but it looks like minRpm value always stays zero? there is a bug
No, minRpm is updated in getAdvance() only if isCranking(), thus only if rpm > 0.
I guess it's better to change the name to minCrankingRpm, not actual.

andreika-git:
Done!

Issue #435, opened 2017-06-03
PID logic: min/max handling


rusefi:
open question if current PID implementation is right in terms of limiting iTerm into the minValue/maxValue range

https://github.com/rusefi/rusefi/blob/master/firmware/controllers/math/pid.cpp

rusefi:
http://brettbeauregard.com/blog/2011/04/improving-the-beginner%E2%80%99s-pid-reset-windup/

rusefi:
See also https://sourceforge.net/p/rusefi/tickets/353/

rusefillc:
Too much was improved in this area since 2017

PR #519, opened 2017-12-22
New PID with CIC integrator

andreika-git:
See https://rusefi.com/forum/viewtopic.php?f=9&t=1315

### DIFF
--- a/firmware/controllers/math/pid.cpp
+++ b/firmware/controllers/math/pid.cpp
@@ -38,29 +38,23 @@ float Pid::getValue(float target, float input) {
 	return getValue(target, input, 1);
 }
 
-float Pid::getValue(float target, float input, float dTime) {
+float Pid::getRawValue(float target, float input, float dTime) {
 	float error = target - input;
 	prevTarget = target;
 	prevInput = input;
 
 	float pTerm = pid->pFactor * error;
-	iTerm += pid->iFactor * dTime * error;
+	updateITerm(pid->iFactor * dTime * error);
 	dTerm = pid->dFactor / dTime * (error - prevError);
 
 	prevError = error;
 
-	/**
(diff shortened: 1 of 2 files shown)


andreika-git:
Solves #435 ?

firmware/controllers/math/pid.cpp
```diff
@@ -151,3 +145,61 @@ void Pid::showPidStatus(Logging *logging, const char*msg) {
...
+
+
+PidCic::PidCic() {
+	// call our derived reset()
+	reset();
+}
+
+PidCic::PidCic(pid_s *pid) : Pid(pid) {
+	// call our derived reset()
+	reset();
```
rusefi: looks like parent constructor is invokeing reset()  already

firmware/controllers/math/pid.cpp
```diff
@@ -151,3 +145,61 @@ void Pid::showPidStatus(Logging *logging, const char*msg) {
...
+		iTerm = pid->maxValue * 100;
+
+	// this is kind of a hack. a proper fix would be having separate additional settings 'maxIValue' and 'minIValye'
+	if (iTerm < -pid->maxValue * 100)
+		iTerm = -pid->maxValue * 100;
+}
+
+
+PidCic::PidCic() {
+	// call our derived reset()
```
rusefi: would it be nicer to invoke parent constructor?

rusefi:
It could be amazing to have a unit test for this new class

firmware/controllers/math/pid.cpp
andreika-git: Unfortunately, in C++ the  polymorphic inheritence is limited, unlike Java/C#. 
The parent constructor doesn't know how to call the virtual reset() method, because vtbl is not created jet...
https://isocpp.org/wiki/faq/strange-inheritance#calling-virtuals-from-ctors

PR #525, opened 2017-12-31
Fix VSS init/deinit

andreika-git:
Tested.


rusefi:
I believe this cannot work - there is no "stop" method.

So at the moment changing some other settings without changing VSS would give us an error. Also new listener is added every time which should get to overflow pretty soon?

andreika-git:
What do you mean? There IS a "stop" method:
void stopVSSPins(void) {
	stopWaveAnalyzerDriver("VSS", activeConfiguration.bc.vehicleSpeedSensorInputPin);
}
And it's called from applyNewHardwareSettings() every time, regardless of which settings are changed.
So on the contrary, the problem was that the 'stop' method was called, and there wasn't any 'start' method.

rusefi:
You are right, the code was actually good it's just hard to understand even for the author :)

I've renamed methods and added two lines of comments to improve this a bit under https://github.com/rusefi/rusefi/commit/1c08f2ea8a1644d8dc3987089a65369a0823c5a3

andreika-git:
Ok, thanx

PR #528, opened 2018-01-02
Impl. idlePid RPM dead-zone and upper-limit

andreika-git:
Tested.

### DIFF
--- a/firmware/controllers/idle_thread.cpp
+++ b/firmware/controllers/idle_thread.cpp
@@ -163,7 +163,7 @@ percent_t getIdlePosition(void) {
 	return currentIdlePosition;
 }
 
-static float autoIdle(float cltCorrection) {
+static float autoIdle(float cltCorrection, float manualIdlePosition) {
 	percent_t tpsPos = getTPS(PASS_ENGINE_PARAMETER_SIGNATURE);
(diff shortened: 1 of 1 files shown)


firmware/controllers/idle_thread.cpp
```diff
@@ -255,11 +268,12 @@ static msg_t ivThread(int param) {
...
 			baseIdlePosition = iacPosition;
 		} else {
+			float manualIdlePosition = manualIdleController(cltCorrection);
 			if (engineConfiguration->idleMode == IM_MANUAL) {
 				// let's re-apply CLT correction
-				iacPosition = manualIdleController(cltCorrection);
+				iacPosition = manualIdlePosition;
 			} else {
-				iacPosition = autoIdle(cltCorrection);
+				iacPosition = autoIdle(cltCorrection, manualIdlePosition);
```
rusefi: I am a bit confused here. I would think that idle is either MANUAL or AUTO. With the change, it looks like manualIdleController(cltCorrection) is used in both MANUAL and AUTO modes, this is a bit unexpected?

andreika-git: I suggest it only as a temporary solution. 
See https://rusefi.com/forum/viewtopic.php?p=27880#p27880

PR #545, opened 2018-01-22
Impl. narrow EGO averaging (alpha)

andreika-git:
Alpha-version. Partially tested.
Disabled by default (requires #define EFI_NARROW_EGO_AVERAGING)


firmware/controllers/sensors/ego.cpp
```diff
@@ -15,6 +15,78 @@
...
 EXTERN_ENGINE;
 
+#ifdef EFI_NARROW_EGO_AVERAGING
+// Needed by narrow EGOs (see updateEgoAverage()).
+// getAfr() is called at ~50Hz, so we store at most (1<<3)*32 EGO values for ~5 secs.
+#define EGO_AVG_SHIFT 3
+#define EGO_AVG_BUF_SIZE 32 // 32*sizeof(float)
+
+static bool useAveraging = false;
+// Circular running-average buffer, used by CIC-like averaging filter
```
rusefi: is there a chance to reuse cyclic_buffer.h - would this reduce the code here?

andreika-git:
Now cyclic_buffer is used as a generic container for CIC buffer.
But we cannot use it as a full replacement of CIC because it would require 1kb of RAM (32*8*sizeof(float)) instead of just 128 bytes (32*float). 
Do you think it makes sense to spare some RAM in this case?

PR #548, opened 2018-01-22
Fix currentVE display

andreika-git:
We need to know the real current VE used in formulas.

### DIFF
--- a/firmware/console/status_loop.cpp
+++ b/firmware/console/status_loop.cpp
@@ -253,7 +253,7 @@ static void printSensors(Logging *log, bool fileFormat) {
 
 		reportSensorF(log, fileFormat, GAUGE_NAME_TCHARGE, "K", engine->engineState.tChargeK, 2); // log column #8
 		if (hasMapSensor(PASS_ENGINE_PARAMETER_SIGNATURE)) {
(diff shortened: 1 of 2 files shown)


firmware/console/status_loop.cpp
```diff
@@ -696,7 +696,7 @@ void updateTunerStudioState(TunerStudioOutputChannels *tsOutputChannels DECLARE_
 
 	if (hasMapSensor(PASS_ENGINE_PARAMETER_SIGNATURE)) {
 		float mapValue = getMap();
-		tsOutputChannels->veValue = veMap.getValue(rpm, mapValue);
+		tsOutputChannels->veValue = engine->engineState.currentVE * 100.0f;//veMap.getValue(rpm, mapValue);
```
rusefi: We probably do not need dead "//veMap.getValue(rpm, mapValue);" here

also kind of in the direction of https://refactoring.com/catalog/replaceMagicNumberWithSymbolicConstant.html - can we maybe help a helper macro with a clear name? I assume we convett from 0..1 to human-readable 0..100% here?

maybe this should be done not in the firmware but in TS file? I am not sure if there is a solution for logging.

andreika-git:
Done!

PR #554, opened 2018-01-26
Add a new MAP sensor type (MPX4250A)

andreika-git:


andreika-git:
Needs calling enum2string.jar

PR #559, opened 2018-01-28
Reduce step skipping for stepper motor

andreika-git:


firmware/hw_layer/stepper.cpp
```diff
@@ -48,20 +48,23 @@ static msg_t stThread(StepperMotor *motor) {
 		 *
 		 * I believe it's safer to retract the valve for parking - at least on a bench I've seen valves
 		 * disassembling themselves while pushing too far out.
+		 *
+		 * Use 10% more steps to compensate step skipping by some old motors.
 		 */
-		for (int i = 0; i < motor->totalSteps; i++) {
+		int numParkingSteps = (int)efiRound(1.1f * motor->totalSteps, 1.0f);
```
rusefi: 1.1f is some dark magic. maybe a setting?

PR #564, opened 2018-02-02
Fix globalConfigurationVersion++ old bug

andreika-git:
See https://rusefi.com/forum/viewtopic.php?p=26520#p26520


andreika-git:
Please check if simulator is ok

firmware/controllers/trigger/trigger_emulator_algo.cpp
```diff
@@ -116,7 +117,11 @@ void setTriggerEmulatorRPM(int rpm DECLARE_ENGINE_PARAMETER_SUFFIX) {
 }
 
 static void updateTriggerShapeIfNeeded(PwmConfig *state) {
+#if EFI_PROD_CODE || EFI_SIMULATOR || defined(__DOXYGEN__)
+	if (emulatorConfigVersion.isOld() && isEmulatorConfigChanged) {
```
rusefi: looks like with this change we cannot change simulation trigger shape unless we change simulation RPM?

PR #572, opened 2018-03-18
ini01 Sensors submenu

andreika-git:


firmware/tunerstudio/rusefi.input
```diff
@@ -69,10 +69,6 @@ enable2ndByteCanID = false
 ; CONFIG_DEFINITION_END
 	
 
-[SettingContextHelp]
-
-	fuelAlgorithm = "See also http://rusefi.com/fuel"
```
rusefi: why are we removing this at all? why are we removing this as part of "Sensors submenu"?

PR #573, opened 2018-03-18
ini02 Remove '3D Map Tuning','Racing Features','Board' and add 'FSIO'

andreika-git:


firmware/tunerstudio/rusefi.input
```diff
@@ -69,10 +69,6 @@ enable2ndByteCanID = false
 ; CONFIG_DEFINITION_END
 	
 
-[SettingContextHelp]
-
-	fuelAlgorithm = "See also http://rusefi.com/fuel"
```
rusefi: (I've modified the other PR but same question) why are we removing SettingContextHelp?

firmware/tunerstudio/rusefi.input
```diff
@@ -987,23 +971,13 @@ gaugeCategory = Fuel Data
 		subMenu = std_separator
 		subMenu = std_realtime,    "&Realtime Display",              0
 		subMenu = tChargeSettings, "tCharge Settings"
+		subMenu = std_separator
+		subMenu = smLaunchControl,           "Launch Control"
 		
-	menu = "3D Map Tuning"
-;		subMenu = std_realtime,				"Realtime Display"
-;		subMenu = std_warmup,				"Warmup Wizard"
-		subMenu = fuelTableELMap,			"Fuel Table", 0, {fuelAlgorithm != 3}
```
rusefi: what is the plan for 3D map tuning? are we just removing it without moving anywhere?

firmware/tunerstudio/rusefi.input
andreika-git: Redundancy:
image

The new version is more explanatory:
image

Besides, we've mastered an integrated help system in TunerStudio, and we can add any info or web-links there also:
image

firmware/tunerstudio/rusefi.input
andreika-git: We're just removing the menu item, but 3D-view is still available as before:
image
If a user likes to work with 3D-maps, there's no need to constantly check this option - it's saved as a preference.

PR #574, opened 2018-03-19
ini03 Add Controller menu and Move items between Engine/Tuning/Controller

andreika-git:


firmware/tunerstudio/rusefi.input
```diff
@@ -1716,7 +1718,7 @@ cmd_stop_engine              = "w\x00\x99\x00\x00"
 		field = "revolution to speed mult",				vehicleSpeedCoef
 
 ;			Engine->Other inputs	
-	dialog = analogInputs, "Analog inputs"
+	dialog = analogInputSettings, "Analog Input Settings"
```
andreika-git: We'll remove all pins from this panel later

PR #575, opened 2018-03-19
ini04 Add 'Sensor Inputs' panel

andreika-git:


rusefi:
it says it has conflict :(

PR #578, opened 2018-03-20
ini05 Add 'Device Outputs' and move some pins there

andreika-git:
Attempt to recteate PR using "git add -p" split-staging technique...


firmware/tunerstudio/rusefi.input
```diff
@@ -1565,28 +1565,21 @@ cmd_stop_engine              = "w\x00\x99\x00\x00"
 	dialog = fanSetting, "Fan Settings"
 		field = "Fan On Temperature",					fanOnTemperature
 		field = "Fan Off Temperature",					fanOffTemperature
-		field = ""
+		
+;			Controller->Actuator Outputs
+	dialog = deviceOutputSettings, "Actuator Outputs"
```
rusefi: changing name a bit here

PR #589, opened 2018-03-29
Idle PID Coasting & Amplification Improvements

andreika-git:
Tested on a real car.
What's new:
1) A new table "Coasting IAC Position for Auto-Idle" (replaces old rejected PR with 'Manual' table used for PID)
2) A new setting "Use separate IAC Table For Coasting" (Idle Settings)
3) A new setting "PID Extra for low RPM" (reasonable tested value is 4%)
4) Implement idlePidRpmDeadZone and idlePidRpmUpperLimit (already existed in settings)
5) Implement idlePositionSensitivityThreshold
6) Small fixes

[2018-03-29_12.54.41.msl.zip](https://github.com/rusefi/rusefi/files/1860271/2018-03-29_12.54.41.msl.zip)


firmware/controllers/math/pid.cpp
```diff
@@ -39,7 +39,7 @@ float Pid::getValue(float target, float input) {
 }
 
 float Pid::getRawValue(float target, float input, float dTime) {
-	float error = target - input;
+	float error = (target - input) * errorAmplificationCoef;
```
andreika-git: This is needed for Idle-IAC PID regulator.
When rpm < targetRpm, there's a risk of dropping RPM too low - and the engine dies out.
So PID reaction should be somewhat increased by adding extra percent to PID-error (a guessed value is ~4% more).
This is a way to 'amplify' the error by a given coefficient. Normally it's 1.0, so nothing happens. 
But for low RPMs we set this coef to the user-defined value (ex. 1.04) - then PID thinks that RPM is lower than it is, and controls IAC more aggressively.

andreika-git: This can be potentially useful for other PID applications as well - like in ETB (still needs testing), when some external force (ex. a spring) helps to move in one side and resists in the opposite direction.

firmware/controllers/math/pid.cpp
```diff
@@ -101,6 +102,10 @@ float Pid::getOffset(void) {
 	return pid->offset;
 }
 
+void Pid::setErrorAmplification(float coef) {
+	errorAmplificationCoef = coef;
+}
+
```
andreika-git: A setter to control the error magnitude.

firmware/integration/rusefi_config.txt
```diff
@@ -903,7 +903,13 @@ float[CRANKING_ADVANCE_CURVE_SIZE] crankingAdvance    ;Optional timing advance t
 	int16_t coastingFuelCutTps;percent between 0 and 100;"%",        1,     0,  0,    100,  0
 	int16_t coastingFuelCutClt;Fuel cutoff is deactivated if CLT<threshold;"C",        1,     0,  -100,    100,  0
 	
-	int[774] unusedEnd;
+	int16_t pidExtraForLowRpm;+Increases PID reaction for RPM<target by adding extra percent to PID-error;"%",        1,     0,  0,    100,  0
+	int16_t unusedInt16;
+	
+float[CLT_CURVE_SIZE] iacCoastingBins;CLT-based idle position for coasting (used in Auto-PID Idle mode);"C",        1,     0,   -100.0,    250.0,  2
+float[CLT_CURVE_SIZE] iacCoasting;    CLT-based idle position for coasting (used in Auto-PID Idle mode);"%",      1,     0,      0.0,    100.0,   2
+
```
andreika-git: We add a new table for Auto-Idle PID regulator instead of 'Manual' warmup multiplier.
But this time we set IAC positions (%) directly, not just multipliers. In the future, we might want to fill-in this table automatically & store real working values.
Also, there's a good reason to set a little oversized values in this table (i.e. greater than your normal 'Idle' values) just for safety...

firmware/integration/rusefi_config.txt
```diff
@@ -934,8 +940,8 @@ float[IAT_CURVE_SIZE] iatFuelCorr;;"%",      1,     0,      0.0,    500.0,   2
 	float[CRANKING_CURVE_SIZE] crankingCycleCoef;;"%",      1,     0,      0.0,    500.0,   2
 	float[CRANKING_CURVE_SIZE] crankingCycleBins;;"counter",        1,     0,    -80.0,    170.0,  2
 
-float[CLT_CURVE_SIZE] cltIdleCorrBins;CLT-based idle position multiplier for simple manual idle controller, or target RPM multiplier for PID-based idle;"C",        1,     0,   -100.0,    250.0,  2
-float[CLT_CURVE_SIZE] cltIdleCorr;    CLT-based idle position multiplier for simple manual idle controller, or target RPM multiplier for PID-based idle;"%",      1,     0,      0.0,    1000.0,   2
+float[CLT_CURVE_SIZE] cltIdleCorrBins;CLT-based idle position multiplier for simple manual idle controller;"C",        1,     0,   -100.0,    250.0,  2
+float[CLT_CURVE_SIZE] cltIdleCorr;    CLT-based idle position multiplier for simple manual idle controller;"%",      1,     0,      0.0,    1000.0,   2
```
andreika-git: A little fix of description: we don't use cltIdleCorr for PID-regulator anymore... Instead we have a target-RPM table.

firmware/tunerstudio/rusefi.input
```diff
@@ -431,6 +431,14 @@ fileVersion = { 20171101 }
...
       gauge       = CLTGauge
 
+	curve = iacCoastingCurve, "Coasting IAC Position for Auto-Idle"
+      columnLabel = "Coolant", "Multiplier"
+      xAxis       = -40, 120, 10
+      yAxis       =  0,  100, 10
+      xBins       = iacCoastingBins, coolant
+      yBins       = iacCoasting
+      gauge       = idleAirValvePositionGauge
+
```
andreika-git: We set gauge to IAC position for convenience - it helps to grab some real IAC values from the idling engine and put into the table.

firmware/tunerstudio/rusefi.input
```diff
@@ -1578,15 +1587,17 @@ cmd_stop_engine              = "w\x00\x99\x00\x00"
 		field = "Min",									idleRpmPid_minValue
 		field = "Max",									idleRpmPid_maxValue
 		field = "period",								idleRpmPid_period
-		field = "RPM dead zone to deactivate IAC pid",	idlePidRpmDeadZone, { idleMode == 0}
-		field = "RPM upper limit to deactivate IAC pid",idlePidRpmUpperLimit, { idleMode == 0}
+		field = "RPM dead zone to deactivate IAC pid",	idlePidRpmDeadZone
+		field = "RPM upper limit to deactivate IAC pid",idlePidRpmUpperLimit
+		field = "PID Extra for low RPM",				pidExtraForLowRpm
```
andreika-git: See https://github.com/rusefi/rusefi/pull/589/commits/589c9d5382474b02d60faf9ce39d94088173b91b

firmware/controllers/idle_thread.cpp
```diff
@@ -276,7 +280,8 @@ static msg_t ivThread(int param) {
 					engine->rpmCalculator.getRevolutionCounterSinceStart());
 		}
 
-		if (absF(iacPosition - currentIdlePosition) < 1) {
+		// The threshold is dependent on IAC type (see initIdleHardware())
```
andreika-git: This greatly improves PID accuracy for steppers with a small number of steps

firmware/controllers/idle_thread.cpp
```diff
@@ -170,6 +172,13 @@ percent_t getIdlePosition(void) {
 static float autoIdle() {
 	percent_t tpsPos = getTPS(PASS_ENGINE_PARAMETER_SIGNATURE);
 	if (tpsPos > boardConfiguration->idlePidDeactivationTpsThreshold) {
+		// Don't store old I and D terms if PID doesn't work anymore.
+		// Otherwise they will affect the idle position much later, when the throttle is closed.
+		if (mightResetPid) {
+			mightResetPid = false;
+			shouldResetPid = true;
```
andreika-git: This is a main idea of 'mightResetPid': We need to reset PID when the throttle is opened and PID deactivated.
Otherwise they will affect the idle position much later, when the throttle is closed.
BUT. We can't just set shouldResetPid immediately - or it will be reseted every time, every cycle.
Instead we wait until at least 1 PID iteration works - and set mightResetPid, a "permission" to do a reset.

firmware/controllers/idle_thread.cpp
```diff
@@ -184,7 +193,23 @@ static float autoIdle() {
 		targetRpm = interpolate2d("cltRpm", clt, CONFIG(cltIdleRpmBins), CONFIG(cltIdleRpm), CLT_CURVE_SIZE);
 	}
 
-	percent_t newValue = idlePid.getValue(targetRpm, getRpmE(engine), engineConfiguration->idleRpmPid.period);
+	// check if within the dead zone
+	int rpm = getRpmE(engine);
+	if (absI(rpm - targetRpm) <= CONFIG(idlePidRpmDeadZone))
+		return baseIdlePosition;
```
andreika-git: This is a RPM dead zone implementation, rather straightforward.

firmware/controllers/idle_thread.cpp
```diff
@@ -184,7 +193,23 @@ static float autoIdle() {
...
+
+	// When rpm < targetRpm, there's a risk of dropping RPM too low - and the engine dies out.
+	// So PID reaction should be increased by adding extra percent to PID-error:
+	percent_t errorAmpCoef = 1.0f;
+	if (rpm < targetRpm)
+		errorAmpCoef += (float)CONFIG(pidExtraForLowRpm) / PERCENT_MULT;
+	// If errorAmpCoef > 1.0, then PID thinks that RPM is lower than it is, and controls IAC more aggressively
+	idlePid.setErrorAmplification(errorAmpCoef);
+
```
andreika-git: And this is an error amplification coef. implementation.
See https://github.com/rusefi/rusefi/pull/589/commits/016084a76c8d2c4c9bb7bef0f21f637c316c5524

PR #596, opened 2018-06-22
Coasting Fuel Cut Fixes

mck1117:
This PR solves two problems:
1) When fuel is cut, getInjectionDuration returns zero duration, so that no fuel is injected (as desired).  However, when handleFuelInjectionEvent goes to schedule the opening and closing of the injectors schedules two events very close together (sometimes exactly on top of each other).  These events sometimes get swapped in order of execution by the scheduler, so the injector will STICK OPEN FLOODING THE ENGINE.  I discovered this when driving, and the coasting fuel cut caused my O2 sensors to read full scale rich.

image

Here's an example of that condition while running on the bench.  Note that it sometimes sticks open, and sometimes doesn't because it's a tossup between whether the opening or closing event gets scheduled first.

This fix for this problem adds a check in handleFuelInjectionEvent to skip scheduling injections with very small (< 50us) duration.

2) An additional check is added for a maximum MAP pressure at which fuel is cut during coasting.  On my car the first ~3 percent of the throttle travel does quite a bit, so a threshold for manifold pressure is required so that fuel isn't cut under light throttle, only when actually coasting.  This also provides flexibility so that a throttle position sensor isn't strictly required to use the feature.

ps: Andrey check your email for my signed contributor agreement

### DIFF
--- a/firmware/controllers/algo/engine_configuration.cpp
+++ b/firmware/controllers/algo/engine_configuration.cpp
@@ -492,6 +492,7 @@ static void setDefaultFuelCutParameters(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 	engineConfiguration->coastingFuelCutRpmLow = 1300;
 	engineConfiguration->coastingFuelCutRpmHigh = 1500;
 	engineConfiguration->coastingFuelCutTps = 2;
+	engineConfiguration->coastingFuelCupMap = 30;
 	engineConfiguration->coastingFuelCutClt = 30;
 }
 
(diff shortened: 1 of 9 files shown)


rusefi:
The 50us defect, is it on the hardware timer side or software queue side or else? Can we fix the root cause or closer to root cause? I wonder if similar issue would be affecting PWM or coils or else?

mck1117:
The 50us number in use is a guess.  It's much smaller than the smallest legitimate fuel pulse that should ever be used, but still a reasonable amount more than the smallest "safe" interval.  50us could probably be a shorter constant, but it's so far below the shortest real pulse that there isn't a reason to figure out where the actual threshold is for unsafe.

The problem is that the scheduler isn't deterministic.  When you schedule two events at the same time, it makes no guarantees about their order of execution.

Anyway, it never makes sense to schedule two events on top of each other to turn a pin on then immediately back off.  In that logic analyzer trace, the correct-polarity pulses were only ~2.5us positive pulse width, which won't actually open the injector at all.

It's probably worth investigating a similar check when scheduling PWM events.  At less than 1% or 0.5% duty or something the pulse is going to be so narrow that it's not actually doing anything, and at very low duty cycle it likely risks running in to this problem.

PR #602, opened 2018-08-02
Cj125 Improvements

mck1117:
1) Rework state machine to function correctly, and greatly improve readability.
2) Add missing cases to switch statements
3) Add configuration for LSU 4.2 vs 4.9.  Sets default PID constants, selects correct lookup table.

Note that this does not fix whatever weird SPI issues I was having, which requires further investigation.


rusefi:
I am starting to look at this PR and I will be taking a weird crazy man approach, please forgive me for that in advance.

As a first step I've merged my version of the new setting which has introduced some conflicts. Can you please resolve the conflict?

Once again, thank you for you patience.

mck1117:
Why take the crazy man approach? The point of a PR is that you can merge the whole thing in one shot, without any effort...  If you want to try the change out, you can locally merge this branch in to master (without sharing that merge anywhere).

It seems counterproductive to manually cherry pick bits and pieces of the PR one at a time, introducing merge conflicts that I have to fix.  Why make more work for everybody?

mck1117:
I completely understand the limited time.  I think we all do.  However, if time is the limitation, I don't think manually merging little bits at a time is very productive.  It creates more work for you, as you have to manually pick out pieces of code to commit one at a time, and more work for a contributor as now there are merge conflicts.

A more useful solution may be merging my branch with master locally (just your machine, don't push anywhere) so you can mess around and test there, then merge the PR when you're satisfied.  It's fine if it takes longer for the whole thing to be merged in, because it doesn't pollute master with half-complete features and requires less total work from everybody.

As a side note, this feature is already well guarded behind chicken bits, and the config changes only replace extant but unused fields.  So long as it builds and doesn't cause crashes with the feature disabled, it's probably safe.  I think exactly one person regularly drives with the cj125 enabled, and that's me :)

Lastly, manually cherry-picking corrupts the git history.  If there's a question or bug found in a feature down the road, git-blame is a useful tool to go find out who touched that code last so you can go poke them to fix/modify/improve/whatever it.

darxfame:
@mck1117 hello, today i updated firmware and connect my LCU4.2 to Prometheus board and got zero value at AFR gauge, when i returned firmware to previous step i got full work AFR Gauge and CJ125. 
Yours pull request brought any troubles to code.

rusefi:
@darxfame please note that this PR is not merged yet and not part of official rusefi right now. It could not have affected you unless you run this fork of code?

mck1117:
@darxfame Oh I assumed you'd checked out this branch to check it out before merge, this hasn't hit mainline yet.

darxfame:
@rusefi maybe @andreika-git merget this PR, but i'm not sure. But all the same i have a trouble. Maybe you @rusefi and @mck1117  can help me with my trouble.
[old_release.zip](https://github.com/rusefi/rusefi/files/2271916/old_release.zip)
[new_release.zip](https://github.com/rusefi/rusefi/files/2271917/new_release.zip)

darxfame:
excuse me if I confused you

mck1117:
No, nothing has been merged yet.  Can you grab a datalog from tunerstudio with debug mode set to cj125? It provides a bit higher resolution information about the state machine and input voltages.

darxfame:
@mck1117 maybe tomorrow. today i'm very buzy

firmware/hw_layer/sensors/CJ125.cpp
```diff
@@ -290,8 +290,19 @@ static void cjStart(void) {
...
+	// Software reset cj125
+	for (int i = 0; i < 2; i++)
+	{
+		cjWriteRegister(INIT_REG2_WR, CJ125_INIT2_RESET)
+		chThdSleepMilliseconds(100);
+	}
+
 	cjIdentify();
 
+	// default values
```
rusefi: please add code comment clarifying when would these values be used? looks like mandatory calibration follows this code?

firmware/hw_layer/sensors/CJ125.cpp
```diff
@@ -290,8 +290,19 @@ static void cjStart(void) {
 		return;
 	}
 	
+	// Software reset cj125
+	for (int i = 0; i < 2; i++)
```
rusefi: can you please add code comment clarifying this voodoo?

mck1117: I don't fully understand this voodoo either

mck1117: I need to go actually diagnose whatever my weird SPI problem is, which will explain this voodoo (or obviate its existence)

rusefi: why did you add it? was it helping with something? can you add a comment on what is seems to affect?

mck1117: The problem I was trying to get around here is that the SPI controller seemed to be sending a garbage first byte, which confused the cj125.  The first reset comes out as garbage, and the second resets it and puts it in a predictable (ie normal) state.  I should sort that out more completely before this is merged.

rusefi:
Next mad micro-step - proper cherry-pick, done in a better way. 
git remote add mck1117 https://github.com/mck1117/rusefi
git fetch mck1117
# post status
git cherry-pick ca4251a35c0d87bd66fcd09ee10c5972d8c5dd2c
git push


Please resolve at least the existing conflict.

firmware/hw_layer/sensors/CJ125.h
```diff
@@ -100,7 +100,7 @@
 // See http://www.waltech.com/wideband-files/boschsensordatasheet.htm
 #define CJ125_HEATER_IDLE_RATE			0.15f	// for a very cold sensor (presumably), we allow 15% duty max.
 #define CJ125_HEATER_PREHEAT_PERIOD 	300		// 300 ms
-#define CJ125_HEATER_CONTROL_PERIOD		180		// 180 ms
+#define CJ125_HEATER_CONTROL_PERIOD		100		// 180 ms
```
rusefi: why this change? comment does not match value - 100 vs 180

rusefi:
I am looking at the huge change machine change and there is no chance for me to digest it. I did another manual merge to consume 4.9 logic

@andreika-git do you want to join the discussion about the state machine re-work?

mck1117:
The state machine change is quite a bit easier to read if you switch from the "Unified" view to the "split" view, since the whole thing got trashed and redone which makes the diff impossible to read.

Here's the short version of what's going on:
In the first switch statement, all that happens is decisions about which state we should be in.  For each state we check the conditions for any outgoing edges, and change state if met.  (no output changes, only state changes)

In the second switch statement, all that happens it output controls based upon the current state.  If we're in the open loop warmup period, run the warm up ramp, and if we're in closed loop, manage the PID state.  If we're in a failure case, turn everything off.  This one is pretty much unmodified. (output changes only, no state changes)

mck1117:
Any chance this can get merged?

rusefi:
we did not have a chance to agree on the strategy here

PR #610, opened 2018-09-25
Prepare for TPS-based Advance and tCharge Modes + Fixes

andreika-git:
Please run gen_config.bat and merge this before other PRs.
Also contains some minor fixes (see commits below).

PR #611, opened 2018-09-25
Tps-based Advance Table

andreika-git:
Please merge #610 first!

### DIFF
--- a/firmware/controllers/algo/advance_map.cpp
+++ b/firmware/controllers/algo/advance_map.cpp
@@ -32,6 +32,7 @@ extern TunerStudioOutputChannels tsOutputChannels;
 #endif
 
 static ign_Map3D_t advanceMap("advance");
+static ign_tps_Map3D_t advanceTpsMap("advanceTps", 1.0 / ADVANCE_TPS_STORAGE_MULT);
 static ign_Map3D_t iatAdvanceCorrectionMap("iat corr");
 
 static int minCrankingRpm = 0;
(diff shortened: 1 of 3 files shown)


firmware/controllers/algo/advance_map.cpp
```diff
@@ -32,6 +32,7 @@ extern TunerStudioOutputChannels tsOutputChannels;
 #endif
 
 static ign_Map3D_t advanceMap("advance");
+static ign_tps_Map3D_t advanceTpsMap("advanceTps", 1.0 / ADVANCE_TPS_STORAGE_MULT);
```
andreika-git: This coeff in ctor parameter is sufficient for int16<->float conversion!

firmware/controllers/algo/advance_map.cpp
```diff
@@ -169,6 +176,8 @@ void setDefaultIatTimingCorrection(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 void prepareTimingMap(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 	advanceMap.init(config->ignitionTable, config->ignitionLoadBins,
 			config->ignitionRpmBins);
+	advanceTpsMap.init(CONFIG(ignitionTpsTable), CONFIG(ignitionTpsBins),
+			config->ignitionRpmBins);
```
andreika-git: We init both tables in RAM because here we're at a very early stage, with no config settings loaded.

PR #614, opened 2018-10-20
Gm24x

mck1117:
Implement gm 24x trigger wheel in a way that looks a little bit less like a pile of random numbers, and properly uses CRANK_MODE_MULTIPLIER.

### DIFF
--- a/firmware/controllers/trigger/decoders/trigger_gm.cpp
+++ b/firmware/controllers/trigger/decoders/trigger_gm.cpp
@@ -39,68 +39,67 @@ void configureGmTriggerShape(TriggerShape *s DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	s->setTriggerSynchronizationGap(6);
 }
 
+static int gm_tooth_pair(float startAngle, bool isLongShort, TriggerShape* s, int mult DECLARE_ENGINE_PARAMETER_SUFFIX)
+{
+	int window = (isLongShort ? 12 : 3) * mult;
+	int end = startAngle + mult * 15;
+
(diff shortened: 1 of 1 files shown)


rusefi:
@mck1117 can u please confirm that unit tests were executed and they pass? Hopefully there is a test for this trigger

firmware/controllers/trigger/decoders/trigger_gm.cpp
```diff
@@ -39,68 +39,67 @@ void configureGmTriggerShape(TriggerShape *s DECLARE_ENGINE_PARAMETER_SUFFIX) {
...
-	s->addEvent2(634.6, ch, TV_RISE PASS_ENGINE_PARAMETER_SUFFIX);
-	s->addEvent2(659.5, ch, TV_FALL PASS_ENGINE_PARAMETER_SUFFIX);
-	s->addEvent2(683.1133333333333, ch, TV_RISE PASS_ENGINE_PARAMETER_SUFFIX);
-	s->addEvent2(689.77, ch, TV_FALL PASS_ENGINE_PARAMETER_SUFFIX);
-	s->addEvent2(695.0666666666667, ch, TV_RISE PASS_ENGINE_PARAMETER_SUFFIX);
-	s->addEvent2(720.0, ch, TV_FALL PASS_ENGINE_PARAMETER_SUFFIX);
+
+	uint32_t code = 0x0A33BE;
+
+	int angle = 0;
```
rusefi: This is elegant but a bit voodoo magic. Shall we at least present code as binary not hex value?

PR #620, opened 2018-11-15
STM32 Nucleo144-F767ZI board: Initial support

andreika-git:
Tested on a real Nucleo-767 bare board (no shields). The firmware starts and connects to PC (USB and USART3 currently supported).
No default pin configuration, no peripherals (SPI/I2C/CAN etc).
This is a test version.

Also added support for more flexible "efifeatures.h" configurations.
Most new STM32F7 files are taken from ChibiOS4 code for RT-STM32F767ZI-NUCLEO144 and STM32Cube_FW_F7 official drivers (flash support).


firmware/config/stm32f7ems/efifeatures.h
```diff
@@ -6,262 +6,75 @@
  * @date Aug 29, 2013
  * @author Andrey Belomutskiy, (c) 2012-2017
  */
+ 
+#include "../stm32f4ems/efifeatures.h"
```
andreika-git: STM32F7 config is now inherited from STM32F4. This is more consistent way to support these config 'branches' and add new features.

firmware/config/engines/custom_engine.cpp
```diff
@@ -182,9 +182,9 @@ void setCustomEngineConfiguration(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 	// todo: 8.2 or 10k?
 	engineConfiguration->vbattDividerCoeff = ((float) (10 + 33)) / 10 * 2;
 
-#if EFI_PROD_CODE
+#if (EFI_PROD_CODE && EFI_CAN_SUPPORT) || defined(__DOXYGEN__)
```
rusefi: I think just EFI_CAN_SUPPORT would be enough here, why would we need both (EFI_PROD_CODE and EFI_CAN_SUPPORT?

firmware/controllers/trigger/trigger_decoder.cpp
```diff
@@ -65,7 +65,9 @@ float actualSynchGap;
 #endif /* ! EFI_PROD_CODE */
 
 #if ! EFI_UNIT_TEST || defined(__DOXYGEN__)
+#if EFI_TUNER_STUDIO || defined(__DOXYGEN__)
```
rusefi: I believe we should not be adding second condition here but replacing current "! EFI_UNIT_TEST" with EFI_TUNER_STUDIO, same below

firmware/config/stm32f7ems/efifeatures.h
```diff
@@ -6,262 +6,75 @@
...
 
-#define EFI_GPIO_HARDWARE TRUE
+// Warning! This is a test config!
 
-#define EFI_FSIO TRUE
-
-#define EFI_PWM_TESTER FALSE
-
-#define HAL_USE_USB_MSD FALSE
-
```
rusefi: weird how PR page shows a major removal here, but I guess that's just because you've copied & edited file?

rusefi:
Wow great stuff!

In my dream world that would be two commits - one for making everything configurable, and the other for F7 changes. but one can only dream!

I can either merge as is or shall we fix these minor comments about double-conditions which do not seem necessary?

andreika-git:
Double-conditions fixed!

firmware/controllers/trigger/trigger_decoder.cpp
andreika-git: I've added some comments inside efifeatures.h on this matter.

firmware/config/stm32f7ems/efifeatures.h
andreika-git: I've added some comments inside efifeatures.h on this matter.

PR #634, opened 2018-12-25
New implementation of 2d table

mck1117:
New implementation of 2d tables, containing everything inside a nice class.

Implementation notes:
- Currently uses linear search instead of binary search for lookup
- No uses yet, but throttle body uses soon!
- Unit tests 🙂
- Also moved the existing interoplate3d function from the header to the source, and used explicit template instantiations to speed up compile and header include complexity a bit.


firmware/controllers/core/interpolation.cpp
```diff
@@ -10,10 +10,15 @@
...
 
+#include "interpolation.h"
+
 #include "global.h"
 
+#include <math.h>
+#include "efilib.h"
 #include "efilib2.h"
-#include "interpolation.h"
+#include "obd_error_codes.h"
```
rusefi: Aren't these included from global.h?

mck1117: They probably are-but global.h causes an explosion of the compiler having to parse things that are never used.  It makes unit testing nearly impossible.  If we can make each component as small and independent as possible, it reduces coupling between parts of the system, which helps make future changes cheaper to implement.

mck1117: Not including it forces you to think carefully about what you're doing, which helps drive up code quality.

rusefi:
I do not see a need for this change and this change is too big for me to be comfortable with it. I would rather prefer a series of atomic changes, each having one clear goal.

What is the relationship between this implementation and existing table_helper.h? Is there a chance to incrementally improve existing table_helper.h?

I am not comfortable with linear search instead of binary search, even while on our small tables sizes there is a good chance that linear is either not slower or maybe even faster.

I believe we need to talk over the phone a figure out a better way to collaborate, a way where I would not be caught off-guard with large changes I am not sure about :(

firmware/controllers/core/interpolation.cpp
rusefi: these are fair arguments, can we have this as an independent PR with includes removed from global.h and added everywhere where they are needed?
So,
1) complete change - not a half change
2) not mixing multiple changes into one mixed change

PR #644, opened 2019-01-04
Fixed Board design

stefanst:
Added Pin descriptions on silk screen and improved ground plane


hardware/TLE8110ED_breakout_board/TLE8110ED.bak
```diff
@@ -0,0 +1,394 @@
+EESchema Schematic File Version 2
```
rusefi: we ususally do not need .bak files - let's remove next time we make changes, or I will just remove locally later

PR #645, opened 2019-01-04
Implement common for controllers

mck1117:
This change implements base classes for a few flavors of controllers.  It removes lots of the boilerplate around creating threads, managing their execution, and running discrete-time controllers.

- PeriodicController: My controller needs to do some work periodically, with the specified period.
- ThreadController: My controller needs to run on its own thread so it can do some work.
- ControllerBase: My controller does something else.

It also replaces the old-style thread in microsecond_timer.cpp with the new implementation, as an example.


firmware/controllers/system/PeriodicController.h
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @brief Called periodically.  Override this method to do work for your controller.
+     */
+    virtual void PeriodicTask(efitime_t nowNt) = 0;
+
+    /**
+     * Called when the controller is stopped.  Optionally override this method to clean up.
+     */
+    virtual void OnStopped() {};
```
rusefi: is there one single usage for this in rusEfi at the moment?
I know there are two ways of doing OOP - you can either develop nice API regardless of usages and focus on API niceness, or you can make sure to never have a line of dead code?

mck1117: I'm not sure--is there anywhere that something has to get reset or reinitialized at run time?

mck1117: If there's not I'm happy to remove it until such a need arises (or doesn't).

firmware/controllers/system/ThreadController.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include "ControllerBase.h"
+
+#include "ch.h"
+
+template <int TStackSize>
```
mck1117: @rusefi Explanation of why this is templated: Because we want to statically allocate the thread's stack (but different controllers might need different size stacks), we have to let the compiler know how big it should be.

rusefi: this part I like. also if you feel to comment on something in PR this a sign that you need a code comment :)

mck1117: yep, writing in-source docs are in the cards

firmware/controllers/system/PeriodicController.h
rusefi: @mck1117 I have a strong preference against implementing new unused stuff for the future. Unless you have a usage, please remove

mck1117: yep, removed

firmware/controllers/system/ThreadController.h
mck1117: docs added

mck1117:
I think I'm ready for merge now, btw.

rusefi:
Hope to review this tonight, have to leave how now.

PR #649, opened 2019-01-05
TunerStudio Cleanup

mck1117:
Remove unused tune interface bloat.

The omitted calls from handlePlainCommand are never called.  TS only calls `F` and `S` without a CRC.


firmware/console/binary/tunerstudio.cpp
```diff
@@ -698,78 +680,32 @@ static void handleExecuteCommand(ts_channel_s *tsChannel, char *data, int incomi
  * @return true if legacy command was processed, false otherwise
  */
 bool handlePlainCommand(ts_channel_s *tsChannel, uint8_t command) {
-	if (command == TS_HELLO_COMMAND || command == TS_HELLO_COMMAND_DEPRECATED) {
+	// Bail fast if guaranteed not to be a plain command
+	if(command == 0)
```
rusefi: I would prefer to see a space between "if" and "("

PR #650, opened 2019-01-06
Improve wall wetting AE to use time constant

mck1117:
Previously, AE used a constant for "amount of fuel sucked from wall in one cycle" when in fact that's not how gasoline works.  There's an amount of gasoline sucked off the wall PER TIME, not PER CYCLE.  This change replaces the per-cycle parameter with a time constant (tau) parameter instead.

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -30,7 +30,7 @@ endif
 
 # C++ specific options here (added to USE_OPT).
 ifeq ($(USE_CPPOPT),)
-  USE_CPPOPT = -std=c++11 -fno-rtti -fno-exceptions -fno-use-cxa-atexit -Werror=write-strings -Werror=type-limits
+  USE_CPPOPT = -std=c++17 -Wno-register -fno-rtti -fno-exceptions -fno-use-cxa-atexit -Werror=write-strings -Werror=type-limits
 endif
 
 # Enable this if you want the linker to remove unused code and data
--- a/firmware/config/engines/dodge_neon.cpp
+++ b/firmware/config/engines/dodge_neon.cpp
@@ -502,11 +502,11 @@ void setDodgeNeonNGCEngineConfiguration(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 	engineConfiguration->tpsAccelLength = 12;
 	engineConfiguration->tpsAccelEnrichmentThreshold = 10;
 
-	engineConfiguration->suckedOffCoef = 0.05;
-	engineConfiguration->addedToWallCoef = 0.40;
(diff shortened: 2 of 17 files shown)


firmware/config/engines/dodge_neon.cpp
```diff
@@ -502,11 +502,11 @@ void setDodgeNeonNGCEngineConfiguration(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 	engineConfiguration->tpsAccelLength = 12;
 	engineConfiguration->tpsAccelEnrichmentThreshold = 10;
 
-	engineConfiguration->suckedOffCoef = 0.05;
-	engineConfiguration->addedToWallCoef = 0.40;
+	engineConfiguration->wwaeTau = 1.0f;
```
mck1117: this is an approximate conversion of the old value, at sensible engine speeds.  It's also zeroed out 3 lines later, so I'm not worried.

firmware/Makefile
```diff
@@ -30,7 +30,7 @@ endif
 
 # C++ specific options here (added to USE_OPT).
 ifeq ($(USE_CPPOPT),)
-  USE_CPPOPT = -std=c++11 -fno-rtti -fno-exceptions -fno-use-cxa-atexit -Werror=write-strings -Werror=type-limits
+  USE_CPPOPT = -std=c++17 -Wno-register -fno-rtti -fno-exceptions -fno-use-cxa-atexit -Werror=write-strings -Werror=type-limits
```
rusefi: is this really part of "Improve wall wetting AE to use time constant"? why are we changing this? is this intentional?

firmware/controllers/algo/accel_enrichment.cpp
```diff
@@ -47,34 +48,93 @@ void WallFuel::reset() {
...
+	}
+
+	float alpha = expf_taylor(-120 / (rpm * tau));
+	float beta = CONFIG(wwaeBeta);
+
+	// If beta is larger than alpha, the system is underdamped.
+	// For reasonable values {tau, beta}, this should only be possible
+	// at extremely low engine speeds (<300rpm ish)
+	// Clamp beta to less than alpha.
+	if(beta > alpha)
```
rusefi: one day we will have automated code formatting, for now I will just have to accept your code style I guess

firmware/Makefile
mck1117: It's required for the `constexpr float expf_taylor_impl` to work because there's a loop in a `constexpr` function.  I think it actually only requires c++14 to work, but if I'm going to poke it I'm going all the way to 17.

PR #664, opened 2019-01-11
Implement ST-recommended workaround for #663

### LINKED ISSUE #663: SPI reception always reads the last bit as 0 on STM32f405/7/15/17
image

Writing then reading back a config register should give the same value.  And the logic analyzer says it does.  The debugger says that the MCU read back the same thing, save for the last bit, which was read as a 0.  [And the ST silicon errata agrees, in section 2.5.2.](https://www.st.com/content/ccc/resource/technical/document/errata_sheet/0a/98/58/84/86/b6/47/a2/DM00037591.pdf/files/DM00037591.pdf/jcr:content/translations/en.DM00037591.pdf)

> In receive transaction, in both I2S and SPI Master modes, the last bit of the transacted frame is not aptured when the signal provided by internal feedback loop from the SCK pin exceeds a critical delay. he lastly transacted bit of the stored data then keeps the value from the pattern received previously. As a consequence, the last receive data bit may be wrong

mck1117:
Setting SPI pins to high speed mode fixes #663.  This isn't required on non-f405/407/415/417 parts, but the vast majority of MCUs running rusEfi are f407.  No negative consequence on chips that don't require the workaround.

Info on the silicon defect can be found in this document, section 2.5.2:
https://www.st.com/content/ccc/resource/technical/document/errata_sheet/0a/98/58/84/86/b6/47/a2/DM00037591.pdf/files/DM00037591.pdf/jcr:content/translations/en.DM00037591.pdf

### DIFF
--- a/firmware/hw_layer/stm32f4/mpu_util.cpp
+++ b/firmware/hw_layer/stm32f4/mpu_util.cpp
@@ -325,10 +325,10 @@ void initSpiModule(SPIDriver *driver, brain_pin_e sck, brain_pin_e miso,
 		int mosiMode,
 		int misoMode) {
 
-	efiSetPadMode("SPI clock", sck,	PAL_MODE_ALTERNATE(getSpiAf(driver)) + sckMode);
+	efiSetPadMode("SPI clock", sck,	PAL_MODE_ALTERNATE(getSpiAf(driver)) + sckMode | PAL_STM32_OSPEED_HIGHEST);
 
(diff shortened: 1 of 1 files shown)


rusefi:
Wow this is some great work! Really appreciate it.

I've added some of the comments from here into the code - I believe we always need comments in the code, and in these special weird cases we absolutely need them. With refactoring happening all the time, just git history would not be enough to find this ticket if anyone would be looking at the code two years from now.

FYI @mck1117 

https://github.com/rusefi/rusefi/commit/db0266927810dd20298f2d411ca544f9e8c4973b

PR #673, opened 2019-01-14
Fix unit-tests (timeNowUs=0 & isInitializingTrigger)

andreika-git:
Affected testFasterEngineSpinningUp

### DIFF
--- a/unit_tests/engine_test_helper.h
+++ b/unit_tests/engine_test_helper.h
@@ -13,11 +13,16 @@
 #include "main_trigger_callback.h"
 #include "unit_test_framework.h"
 
+class EngineTestHelperBase
+{
+public:
+	EngineTestHelperBase();
+};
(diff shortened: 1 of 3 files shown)


unit_tests/engine_test_helper.h
```diff
@@ -13,11 +13,16 @@
 #include "main_trigger_callback.h"
 #include "unit_test_framework.h"
 
+class EngineTestHelperBase
+{
+public:
+	EngineTestHelperBase();
```
rusefi: what is this and how is this helping to fix the test? Was it supposed to be part of the PR?

andreika-git: Yes, the idea is to set timeNowUs=0 before calling the EngineTestHelper ctor (and, more important, the Engine ctor).

PR #686, opened 2019-01-31
405v mini tests

andreika-git:


andreika-git:
The progress so far:
1) stm32_clock_init() doesn't work, temporary replaced with setSysClock()
2) RTC doesn't work, currently needs: -DHAL_USE_RTC=FALSE
3) Still doesn't work, the conn-LED is constantly ON (bright). Perhaps, hasFirmwareError() == true

PR #701, opened 2019-03-10
tle6240 and external gpio chips API - preliminary

dron0gus:
Preliminary.
Please take a look and say if you like this style of API and driver.
Will fully test it when assemble test bench.


rusefi:
Can we please-please-please split this PR into a series of smaller PR?

What is the first one to begin with? Maybe core.c?

core.c needs a unit test in my opinion. So how about we add a unit test into core.c and merge core.c with it's unit test first?

dron0gus:
On Mon, Mar 11, 2019 at 2:03 AM rusefi <notifications@github.com> wrote:

> Can we please-please-please split this PR into a series of smaller PR?
>
> What is the first one to begin with? Maybe core.c?
>
Yes. If you ok with coding style and API.

> core.c needs a unit test in my opinion. So how about we add a unit test
> into core.c and merge core.c with it's unit test first?
>
Ok. I'll try to do some unit test for it. Never done before.

firmware/hw_layer/drivers/gpio/tle6240.c
```diff
@@ -0,0 +1,498 @@
...
+	/* Acquire ownership of the bus. */
+	spiAcquireBus(spi);
+	/* Setup transfer parameters. */
+	spiStart(spi, &chip->cfg->spi_config);
+	/* Slave Select assertion. */
+	spiSelect(spi);
+	/* Atomic transfer operations. */
+	/* TODO: check why spiExchange transfers invalid data on STM32F7xx */
+	//spiExchange(spi, 2, txb, rxb);
+	rxb[0] = spiPolledExchange(spi, txb[0]);
```
andreika-git: SPIv1/v2 frame is uint16_t

andreika-git: (See SPI_CR1_DFF flag for 16-bit mode)

firmware/hw_layer/drivers/gpio/tle6240.h
```diff
@@ -0,0 +1,45 @@
+/*
+ * tle6240.h
+ *
+ * TLE6240GP Smart 16-Channel Low-Side Switch
+ *
+ * @date Dec 29, 2018
+ * @author Andrey Belomutskiy, (c) 2012-2018
```
rusefi: "drivers/gpio" folder idea makes total sense. Just moving tle6240 file into it and new .mk file could be another small pull request in order to move forward

PR #704, opened 2019-03-13
tle6240 gpio driver implementation

dron0gus:
For now without SPI DMA and IRQ.


firmware/hw_layer/drivers/gpio/tle6240.c
```diff
@@ -0,0 +1,499 @@
+/*
+ * tle6240.cpp
```
andreika-git: tle6240.cpp --> tle6240.c

dron0gus: Online fixed.

PR #713, opened 2019-03-22
Idle timing PID control & TPS-based VE table

andreika-git:
Tested.


andreika-git:
image
image
image

andreika-git:
Please run get_config.bat

firmware/tunerstudio/rusefi.input
```diff
@@ -7,15 +7,20 @@
 ; this should stop TS from looking for the CAN ID in the 2nd byte location and allow the page reads to work correctly.
 enable2ndByteCanID = false
 
-#unset tuneByMAF
+;#unset tuneByMAF
```
rusefi: did we intend to comment  #unset tuneByMAF? why?

firmware/tunerstudio/rusefi.input
```diff
@@ -7,15 +7,20 @@
...
 
-;[SettingGroups]
+[SettingGroups]
```
rusefi: did we intend to un-comment [SettingGroups]? why?

firmware/tunerstudio/rusefi.input
andreika-git: #unset instructed TS to create a project setting, and we don't need it anymore because now we define it explicitly in [SettingGroups] section (see below).

firmware/tunerstudio/rusefi.input
andreika-git: We need [SettingGroups] because we're adding more options to it - see "settingGroup = tuneVeMode". We use the new option "tuneByTPS" to switch between table definitions (see all "#if tuneByTPS" below in the code)

PR #743, opened 2019-04-11
trigger input: add PAL support

dron0gus:
Use EXTI for trigger inputs.
Tested on custom stm32f765 board using trigger simulation.

Set in halconfig.h
+#define HAL_USE_PAL                            TRUE
+#define PAL_USE_CALLBACKS              TRUE


rusefi:
both PAL and ICU has hardware limitations. i believe that in order to get most flexibility both modes should be supported at runtime, with same binary using for example ICU for crank and PAL for cam, or vise versa.

having this as compile time switch creates ab explosion of binary versions

PR #754, opened 2019-04-14
Add more portability to smart_gpio & SPI

andreika-git:


andreika-git:
There's a fix for SPI_CR1_16BIT_MODE, please don't miss it!

PR #789, opened 2019-05-05
Fix ETB control for 2 pin H-bridge

mck1117:
Fix ETB control for two pin H bridges.

Now 2 and 3 pin h-bridges are properly supported, without whacky behavior at low duty cycle in 2-pin mode.

Here's changing direction on two pin mode, to prove it works!
image


mck1117:
Hold for a minute-clarifying some naming

rusefi:
Sorry missed the "Hold for a minute-clarifying some naming" :(

PR #824, opened 2019-06-04
Remove GPS :D

mck1117:


rusefi:
i would keep gps, i have sentimental attachement to it and would like to make a statement that things like tgat should be agreed on in advance

mck1117:
"sentimental attachment" isn't how engineering works.

This PR is a forum for that discussion.

The GPS code parses NMEA input, but does nothing with it.  Even the compiler figured this out--it optimizes most of it out!

Even if we ignore the part where it didn't actually work, is an ECU the right place for GPS parsing? What's the proposed consumer of GPS data inside rusEfi?

mck1117:
And if that consumer is a different device in the car that isn't the ECU, then GPS should be a separate CAN bus device.  I would be all over a small box that does GPS, accelerometer, gyro, etc. as a "motion processor" that interfaces with a data system.

rusefi:
we have too many communication channels. i believe forum would be the better space to start this holy war.

PR #828, opened 2019-06-07
Static analysis pass

mck1117:
Ran the cppcheck static analysis tool against the firmware.
- Lots of `const`
- Fixed a potential buffer overflow
- Fixed a scary typo where two parameters with the same type were swapped in the header/source
- Fixed a copy-paste mistake where the same array element was set repeatedly
- `explicit` on single-parameter constructors

### DIFF
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -300,17 +300,19 @@ private:
 #else
 		bool isRunning = false;
 #endif /* EFI_SHAFT_POSITION_INPUT */
+
 		// cltCorrection is used only for cranking or running in manual mode
-		float cltCorrection;
-		if (cisnan(clt))
-			cltCorrection = 1.0f;
-		// Use separate CLT correction table for cranking
-		else if (engineConfiguration->overrideCrankingIacSetting && !isRunning) {
-			cltCorrection = interpolate2d("cltCrankingT", clt, config->cltCrankingCorrBins, config->cltCrankingCorr, CLT_CRANKING_CURVE_SIZE) / PERCENT_MULT;
-		} else {
-			// this value would be ignored if running in AUTO mode
-			// but we need it while cranking in AUTO mode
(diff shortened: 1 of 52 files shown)


firmware/controllers/actuators/idle_thread.cpp
```diff
@@ -300,17 +300,19 @@ class IdleController : public PeriodicController<UTILITY_THREAD_STACK_SIZE> {
...
-			cltCorrection = 1.0f;
-		// Use separate CLT correction table for cranking
-		else if (engineConfiguration->overrideCrankingIacSetting && !isRunning) {
-			cltCorrection = interpolate2d("cltCrankingT", clt, config->cltCrankingCorrBins, config->cltCrankingCorr, CLT_CRANKING_CURVE_SIZE) / PERCENT_MULT;
-		} else {
-			// this value would be ignored if running in AUTO mode
-			// but we need it while cranking in AUTO mode
-			cltCorrection = interpolate2d("cltT", clt, config->cltIdleCorrBins, config->cltIdleCorr, CLT_CURVE_SIZE) / PERCENT_MULT;
+		float cltCorrection = 1.0f;
+		if (!cisnan(clt))
```
rusefi: i believe this is a bad change

firmware/controllers/math/pid_auto_tune.cpp
```diff
@@ -86,7 +86,7 @@ void PID_AutoTune::SetLookbackSec(int value)
 
 double inline PID_AutoTune::fastArcTan(double x)
 {
-  // source: �Efficient approximations for the arctangent function�, Rajan, S. Sichun Wang Inkol, R. Joyal, A., May 2006
+  // source: �Efficient approximations for the arctangent function�, Rajan, S. Sichun Wang Inkol, R. Joyal, A., May 2006
```
rusefi: this does not looks right at least on the web

Looks like I will be closing this PR, let's please have a separate commit with ONLY const?

mck1117: Agh! This source we imported from elsewhere uses a bunch of non-ascii characters, like those quotes.

rusefi:
this change is too big to manage. It mixes mechanical and subjective changes.

Please separate large mechanical changes like const from smaller more focused and sometimes more subjective changes.

PR #837, opened 2019-06-12
microrusefi config

mck1117:


rusefi:

rusefi:
Weird, I do not remember writing the comment above. And the text makes no sense. @mck1117 just to confirm "Can one of the admins verify this patch?" is not your comment right?

rusefi:
Anyway. regarding this PR. I do not understand the need for any of these new files, can you please elaborate on why do we need to add so many new files?

This actually brings the point that I should probably remove some of the existing copy-pasted files since rusEfi overrides pin modes or should be overriding pin modes for everything.

mck1117:
This is a copy paste & modification of the nucleo f767 files.  The reason they all have to be copied is that there's no way to replace _some_ of them for a particular board.

`efifeatures.h` is the only one with any actual changes in it that aren't copy-paste.

mck1117:
And I have no idea what that "can one of the admins..." comment is.  It was made less than a minute after I created the PR.  Is that some artifact of trying to set up the build gate?

rusefi:
I am concerned that this copy-pasting is getting out of hand, I think now is the time to check what could be done to improve things.

1) board.c files are now taken from ChibiOS/boards folder, I believe we do not add anything custom to those files

2) https://github.com/rusefi/rusefi/commit/c8f7790a54f39a797103dac8adea70630ecf6012 is an example of how existing .ld file could be reused, I do not think that we need to copies of that file

3) I've tried removing board.h files but there is a little problem there, would be great if you can research this futher

4) efifeatures.h the alternative approach is to override not completely copy the file, the way prometheus does it. I have mixed feelings about overring and undefining but I also definitely have mixed feelings about copy pasting files

rusefi:
@mck1117 has been a few days, please share your plans for this PR. No rush obviously, just want to be aware of the plan :)

rusefi:
Zero responses from PR starter

Issue #773, opened 2019-04-23
All corrections should be ratio, 1.0 = no correction 


OrchardPerformance:
"IAT fuel", "Cranking TPS" and "Cranking CLT" are decimal multipliers while "Warmup Fuel Manual multiplier" and "Warmup Idle multiplier" are % values.

% values need changing to ratio so that "no correction" is always "1.0"

rusefi:
TODO at least one curve displays 100 instead of 1 for ratio

![image](https://user-images.githubusercontent.com/7377949/56622300-74aae980-65fd-11e9-8116-747464ec8925.png)


FYI @mck1117

rusefi:
@mck1117 poke :)

rusefi:
See also #811

rusefi:
@mck1117 I would like to state for the record that a month ago I've poked and I was hoping to get at least _some_ response within a month.

rusefi:
Thank you @mck1117  for the patience, great change!

mck1117:
I think this can close

rusefi:
@mck1117 says
```
The conclusion with ratio vs. % was that the generator tool needs to be able to emit things like:
```#if USE_PERCENT
    cltFuelCorr                        = array, F32,    10336,    [16],    "%", 100, 0, 0.0, 500.0, 0
#else
    cltFuelCorr                        = array, F32,    10336,    [16],    "ratio", 1, 0, 0.0, 5.0, 2
#endif```
```

PR #838, opened 2019-06-13
Fix tunerstudio ratio vs. percent

mck1117:
There was a mystery mix of:
- ratio (example 1.2)
- percent (example 120%)
- percent extra (example +20%)

These have all been consolidated down to use only the "ratio" option.


rusefi:

mck1117:
This commit includes a revert-ish of be926cb.  That commit took the numbers being spit out as canon, and as a result, missed the fact that cranking fuel pulse widths were 10x what they're supposed to be.  This change both fixes the actual bug (wrong cranking clt multiplier) and the tests.

firmware/controllers/algo/engine_configuration.cpp
```diff
@@ -632,11 +632,11 @@ static void setDefaultCrankingSettings(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
...
 	engineConfiguration->crankingAdvanceBins[0] = 0;
 	engineConfiguration->crankingAdvance[1] = 0;
-	engineConfiguration->crankingAdvanceBins[0] = 200;
+	engineConfiguration->crankingAdvanceBins[1] = 200;
 	engineConfiguration->crankingAdvance[2] = 0;
-	engineConfiguration->crankingAdvanceBins[0] = 400;
+	engineConfiguration->crankingAdvanceBins[2] = 400;
 	engineConfiguration->crankingAdvance[3] = 0;
-	engineConfiguration->crankingAdvanceBins[0] = 1000;
+	engineConfiguration->crankingAdvanceBins[3] = 1000;
```
mck1117: These were just at typo.

mck1117:
Oh, and this should close #773.

Issue #695, opened 2019-02-23
Why do we use so much stack?


rusefi:
why is UTILITY_THREAD_STACK_SIZE currently 400 and not 200 or 800?

why is PORT_INT_REQUIRED_STACK so high at 768?

rusefi:
CH_DBG_FILL_THREADS

rusefi:
interpolate2d() invocation cost = 40 bytes
interpolateMsg() invocation cost = 72 bytes

rusefi:
Actually F7 chconf does NOT define PORT_IDLE_THREAD_STACK_SIZE and PORT_INT_REQUIRED_STACK and it looks like F7 is running with less stack?

FYI @andreika-git @mck1117

mck1117:
this got fixed - PORT_IDLE_THREAD_STACK_SIZE and PORT_INT_REQUIRED_STACK were reduced to the chibios recommended values

mck1117:
this got fixed - PORT_IDLE_THREAD_STACK_SIZE and PORT_INT_REQUIRED_STACK were reduced to the chibios recommended values

PR #839, opened 2019-06-13
Fix threadinfo command #695

mck1117:
`threadinfo` command now spits out stack base, stack free (in bytes), and ticks consumed by the thread.  Stack free space is included for the dedicated interrupt stack.

Here's some example data:

image

The learning is that `SingleTimerExecutor::scheduleForLater` goes _very_ deep, but the compiler inlines it to hell and back with `-O2` turned on (the giveaway here is that the only thing the servo thread does is schedule stuff, yet it uses a bunch of stack).


rusefi:

Issue #794, opened 2019-05-07
PC simulator: TS over HW VSP wants lower blockingFactor


rusefi:
At the moment PC simulator does not work with Tuner Studio Lite via https://www.hw-group.com/software/hw-vsp3-virtual-serial-port

a work around for this is to use smaller blockingFactor which at the moment is not working as desired  ochBlockSize command

rusefi:
Looking at TS documentation and I am not getting it
"11.2.1 Supporting Un-Optimized" is not what we have
and
rusEfi project seems to be closer to 11.2.2 Optimized Blocking ?

But if that's the case why do we have the error? 
```
1 Errors:
Error: blockingFactor smaller than ochBlockSize, but ochCommand does not support blocking.
```

rusefi:
I got a word from Tuner Studio people that they will adjust Tuner Studio so that our gauges would load with our "blockingFactor = 200" workaround. Right now TunerStudioMS Beta is v3.0.60.54 - sounds like we need to wait for the NEXT beta and try again.

https://www.efianalytics.com/TunerStudio/beta/

rusefi:
The root cause of the instability between TS and simulator over virtual serial port is not know, but we know that with lower blockingFactor or ochBlockSize things are better.

TS bug prevents us from using a smaller block size as is, waiting for TS to fix the defect for us to use variable length packet.

rusefi:
Status update: still looking for release AFTER  3.0.28
Have no checked if latest TS BETA has the fix, at the moment when this ticket was reported TS beta was v3.0.60.54

I believe the issue is that TS does not realize that rusEfi ochGetCommand supports runtime blocking.

rusefi:
New TS releases, time to try this again

rusefi:
Progress https://rusefi.com/forum/viewtopic.php?f=5&t=1433&p=35422#p35422

Issue #807, opened 2019-05-27
Annotations in C++ code to produce formulas in rusEfi console


rusefi:
In order to increase usability it would be nice for rusEfi console to show more formulas

We need more automation in this area, just coding java in parallel to C++ does not go well. Maybe code annotations?

rusefi:
@mck1117  says "I have concerns about inheritance vs. composition here

the compiler is MUCH more aggressive about changing the layout of things when you use inheritance
there's no guarantee from the c++ spec (or even requirement) about how inheritance should be laid out in memory"

rusefi:
@mck1117, that's a totally legit concur but it looks like we are good with GCC at the moment. There is definitely an implicit dependency on lack of virtual table.

But I am fine accepting this. If/When this becomes an issue I will deal with it. If vtable would be needed I would skip 8 bytes. If/When I see memory layout totally messed up, I will start generating getters and composing packet dynamically. This would not change the implementation fundamentally.

While things seems to be working I will depend on de-facto compiler implementation. If things change it should not be a huge deal in my opinion.

PR #847, opened 2019-06-17
Annotations in C++ code to produce formulas in rusEfi console #807

rusefi:
firmware part of the change separately in order to reduce confusion

this is what developers would have to deal with


rusefi:

rusefi:
@mck1117 @andreika-git @dron0gus FYI

rusefi:
https://github.com/rusefi/rusefi/commit/1f8072b034ba640413d72e3ad4acbea08ec61546.patch is the whole thing :)

firmware/console/binary/tunerstudio.cpp
```diff
@@ -277,6 +277,28 @@ static void onlineApplyWorkingCopyBytes(int currentPageId, uint32_t offset, int
...
+/**
+ * Read internal structure for Live Doc
+ * This is somewhat similar to read page and somewhat similar to read outputs
+ * We can later consider combining this functionality
+ */
+static void handleGetStructContent(ts_channel_s *tsChannel, int structId, int size) {
+	tsState.readPageCommandsCounter++;
+
+	const void *addr = NULL;
+	if (structId == LDS_CLT_INDEX) {
```
mck1117: switch instead of `if`? (switch can give far better performance when checking consecutive numbers, it will use a lookup table)

firmware/console/binary/tunerstudio.cpp
```diff
@@ -743,6 +766,10 @@ int tunerStudioHandleCrcCommand(ts_channel_s *tsChannel, char *data, int incomin
 	} else if (command == TS_PAGE_COMMAND) {
 		uint16_t page = *(uint16_t *) data;
 		handlePageSelectCommand(tsChannel, TS_CRC, page);
+	} else if (command == TS_GET_STRUCT) {
```
mck1117: Could this use the `getOutputChannels` or reading pages? Do we need to duplicate this, or can the 'normal' output channels just be one case for GET_STRUCT?

firmware/controllers/math/speed_density.cpp
```diff
@@ -29,42 +29,48 @@ baroCorr_Map3D_t baroCorrMap("baro");
...
 float getTCharge(int rpm, float tps, float coolantTemp, float airTemp DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	if (cisnan(coolantTemp) || cisnan(airTemp)) {
 		warning(CUSTOM_ERR_NAN_TCHARGE, "t-getTCharge NaN");
 		return coolantTemp;
 	}
 
-	float Tcharge_coff;
 
-	if (CONFIG(tChargeMode) == TCHARGE_MODE_AIR_INTERP) {
+	if ((engine->engineState./*DISPLAY_IF*/isTChargeAirModel = (CONFIG(tChargeMode) == TCHARGE_MODE_AIR_INTERP)) == true) {
```
mck1117: `== true` should never be typed - just omit it and you'll get the same behavior

firmware/console/binary/tunerstudio.cpp
rusefi: Maybe probably. There is a chance this would be code-generated later or soon.

firmware/console/binary/tunerstudio.cpp
rusefi: there is actually a code comment somewhere along these lines.

I am waiting on Tuner Studio to fix https://github.com/rusefi/rusefi/issues/794 - once they fix there is a chance I will make _all_ these data structures part of dynamically generated virtual getOutputChannels  mega-page. This case we would get logging of any of those if needed

for now it's easier to copy-paste a bit to keep code simpler.

firmware/console/binary/tunerstudio.cpp
```diff
@@ -277,6 +277,28 @@ static void onlineApplyWorkingCopyBytes(int currentPageId, uint32_t offset, int
...
+ * Read internal structure for Live Doc
+ * This is somewhat similar to read page and somewhat similar to read outputs
+ * We can later consider combining this functionality
+ */
+static void handleGetStructContent(ts_channel_s *tsChannel, int structId, int size) {
+	tsState.readPageCommandsCounter++;
+
+	const void *addr = NULL;
+	if (structId == LDS_CLT_INDEX) {
+		addr = &engine->engineState.cltCurve;
```
mck1117: the type-safe way to do this is
`addr = static_cast<thermistor_state_s*>(&engine->engineState.cltCurve)`
That cast will automagically fix any vtable-related woes.

firmware/controllers/math/speed_density.cpp
rusefi: You are right, this was part of a wrong "error: suggest parentheses around assignment used as truth value -Werror=parentheses" fix. Will change now.

mck1117: oh - let's also avoid assignment in an if statement then :smile:

firmware/console/binary/tunerstudio.cpp
rusefi: Done, thank you!

firmware/controllers/math/speed_density.cpp
rusefi: need to save condition check result for live docs. Did not want to move condition logic from if statement, it would get ugly if we have chained if/else if

PR #850, opened 2019-06-18
config

960:


rusefi:

rusefi:
me7_pnp not a great name for an official board

I would need a real board in my hands to debug it :(

PR #853, opened 2019-06-21
Fix mappings for microRusefi

mck1117:
Fix a few mappings that were wrong, and align naming with the corresponding wiring docs.

Question: do we want pin numbers in the TS names, like this?
image

This PR doesn't currently have pin numbers like that, but I could add them.


rusefi:
The pin thing is an interesting idea!

mck1117:
I think this one is ready to go (and is now tested on a real board).

PR #855, opened 2019-06-23
Fix F7 ADC #854

### LINKED ISSUE #854: ADC issue on F7
@mck1117 believes that the ADC issues @960 has originally reported are related to F7 memory caching mechanism:

@mck1117 says here's how the ADC works:
1) timer triggers ISR which starts the ADC, writing to DMA
2) ADC generates samples, DMA controller copies them to SRAM
3) "ADC is done" interrupt fires, CPU reads values from SRAM, averages, and writes back in to sample buffer
on an f7, there's an extra half step
step 2b) is that the DMA controller DOES NOT notify the CPU cache that it should invalidate those cache lines
so in step 3), if you get cache hits, you read old numbers

mck1117:
On ARM Cortex M7, only the CPU can invalidate the CPU cache.  So when DMA writes to SRAM, the CPU doesn't know it should re-read.  This change adds an invalidate to the ADC's sample buffer before reading it, so we force a read from SRAM, actually loading the new values.

Fixes #854

### DIFF
--- a/firmware/hw_layer/AdcConfiguration.h
+++ b/firmware/hw_layer/AdcConfiguration.h
@@ -22,8 +22,9 @@ public:
 	int conversionCount;
 	int errorsCount;
 	int getAdcValueByIndex(int internalIndex) const;
+	void invalidateSamplesCache();
 
-	adcsample_t samples[ADC_MAX_CHANNELS_COUNT * MAX_ADC_GRP_BUF_DEPTH];
+	__ALIGNED(32) adcsample_t samples[ADC_MAX_CHANNELS_COUNT * MAX_ADC_GRP_BUF_DEPTH];
 
 	int getAdcValueByHwChannel(int hwChannel) const;
 
--- a/firmware/hw_layer/AdcConfiguration.h
+++ b/firmware/hw_layer/AdcConfiguration.h
@@ -24,7 +24,12 @@ public:
 	int getAdcValueByIndex(int internalIndex) const;
 	void invalidateSamplesCache();
 
+	// This must be aligned on a 32-byte boundary, and be a multiple of 32 bytes long.
+	// When we invalidate the cache line(s) for ADC samples, we don't want to nuke any
+	// adjacent data
 	__ALIGNED(32) adcsample_t samples[ADC_MAX_CHANNELS_COUNT * MAX_ADC_GRP_BUF_DEPTH];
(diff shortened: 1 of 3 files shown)


mck1117:
This also required aligning the ADC buffers to the width of a cache line - always 32 bytes on ARM M7.

firmware/hw_layer/adc_inputs.cpp
```diff
@@ -148,7 +148,7 @@ ADC_TwoSamplingDelay_20Cycles,   // cr1
 		0  // Conversion group sequence 1...6
 		};
 
-AdcDevice slowAdc(&adcgrpcfgSlow);
+__ALIGNED(32) AdcDevice slowAdc(&adcgrpcfgSlow);
```
rusefi: why do we align this?

firmware/hw_layer/AdcConfiguration.h
```diff
@@ -22,8 +22,9 @@ class AdcDevice {
 	int conversionCount;
 	int errorsCount;
 	int getAdcValueByIndex(int internalIndex) const;
+	void invalidateSamplesCache();
 
-	adcsample_t samples[ADC_MAX_CHANNELS_COUNT * MAX_ADC_GRP_BUF_DEPTH];
+	__ALIGNED(32) adcsample_t samples[ADC_MAX_CHANNELS_COUNT * MAX_ADC_GRP_BUF_DEPTH];
```
rusefi: I believe that this code is vague. Quality code would self-document WHY do we suddenly have to do an unusual thing. I believe that either a human-readable comment referring to F7 caches or MAGIC_F7_DMA_FLASH mentioning is needed here

firmware/hw_layer/adc_inputs.cpp
```diff
@@ -256,10 +256,10 @@ static void pwmpcb_fast(PWMDriver *pwmp) {
 float getMCUInternalTemperature(void) {
 #if defined(ADC_CHANNEL_SENSOR)
 	float TemperatureValue = adcToVolts(slowAdc.getAdcValueByHwChannel(ADC_CHANNEL_SENSOR));
-	TemperatureValue -= 0.760; // Subtract the reference voltage at 25�C
+	TemperatureValue -= 0.760; // Subtract the reference voltage at 25�C
```
rusefi: Why are we touching this line here?

mck1117: fixed as discussed

PR #859, opened 2019-06-24
Relocate pstack/mstack to DTCM

mck1117:
Move stacks (and heap - which we don't use) to DTCM - this certainly won't hurt performance, and will reduce bus contention (not that we had a problem with that).


rusefi:
@mck1117 was this tested on real hardware, does rusEfi start with this change?

mck1117:
Yep, tested on MRE hardware, works fine.

PR #860, opened 2019-06-26
Update

960:


rusefi:
@mck1117 what do you think? Some things are questionable

960:
I know :-)

960:
But which of it?

firmware/tunerstudio/rusefi.input
```diff
@@ -1220,6 +1222,15 @@ menuDialog = main
 		subMenu = oilPressureSensor,		"Oil pressure"
 		subMenu = egtInputs,				"EGT" @@if_ts_show_egt
 
+menu = "&E-Throttle"
+
+        subMenu = etbDialog,				"E-Throttle Settings" @@if_ts_show_etb
```
mck1117: Use the full word - "Electronic Throttle Body"

firmware/tunerstudio/rusefi.input
```diff
@@ -1220,6 +1222,15 @@ menuDialog = main
...
+		subMenu = std_separator
+		subMenu = etbTpsBiasCurve,			"FeedForward Map" @@if_ts_show_etb
```
mck1117: either lowercase "f" like `Feedforward` or lowercase and a space `Feed forward`

mck1117:
And one more thing - let's match casing with other items, which is "Sentence case like this" not "Every Word Caps Like This"

firmware/tunerstudio/rusefi.input
960: I can compromise with Electronic Throttle ?

960:
Why dont you want uppercase for every word? It looks much better and others do it as well..

firmware/tunerstudio/rusefi.input
```diff
@@ -1139,21 +1139,23 @@ menuDialog = main
 		subMenu = cltTimingCorrCurve,		"CLT correction", 0, {isIgnitionEnabled == 1}
 		subMenu = ignitionIatCorrTableTbl,	"IAT correction", 0, {isIgnitionEnabled == 1}
 		
-	menu = "&Cranking"
-		subMenu = crankingDialog,			"Cranking settings"
+
+		
+	menu = "&Idle and Cranking"
```
mck1117: Order in the menu title should match the order in the menu - if cranking is first in the title, it should be first in the menu

rusefi:
I am unable to review this change as is, you would have to split it into many smaller changes.

Review process is very manual and requires some effort. This is a pretty important file. Here we have mixed two changes - we have mixed trivial mechanical Case Renaming with actual meaningful dialog rearrangements.

I see too many issues here and this makes me thing that major effort would be needed from me to review.

1) I am very upset about crankingTpsCurve being simply removed.

2) CLT to ECT rename is questionable and worth a separate discussion

3) fuelLevelSensor is not digital input


I would suggest that instead of having so many changes squashed together a series of much smaller changes is made, one by one. I understand that this requires effort from the contributor - but the review also requires effort from reviewer.

mck1117:
Agreed on all counts.
1. Yeah - this one serves a purpose, so we can't remove it.
2. CLT is the standard - we (and many others) use it everywhere else.  Let's not rename things without a good reason to.
3. Yep.

I have more:

4. FSIO isn't an input, why is it in the inputs menu?
5. "Other Sensor Inputs" -> "Digital Inputs" In fact, flex, fuel level, and oil pressure are all analog.

960:
Yeah, I saw CLT in maxxecu now

I cant recall seeing other than ECT, Engine Coolant Temperature before
Just like IAT, Intake Air Temperature

But I can change it back.

I'll fix those other things as well

Do you use TPS Enrichment for cranking?
I'll add it back, but we need to discuss the cranking/cold start settings.

Please look at the diagram found in links help file.

We need to be able to end up with that curve somehow.
Here in north with ethanol, we have to spend a lot of time with these settings to get it start

960:
What do you mean FSIO isnt input?

It says FSIO Inputs?

mck1117:
"diagram found in the links help file"
Can you share that here for context?

rusefi:
One change - one PR, one discussion. Let's please not mix ethanol with FSIO.

FSIO is flexible input AND output

https://rusefi.com/wiki/index.php?title=Manual:Flexible_Logic

mck1117:
Oh - I see what happened, FSIO input/output split between different menus.  Let's not do that.

PR #862, opened 2019-06-26
Careful start

960:


firmware/tunerstudio/rusefi.input
```diff
@@ -1220,6 +1223,15 @@ menuDialog = main
 		subMenu = oilPressureSensor,		"Oil pressure"
 		subMenu = egtInputs,				"EGT" @@if_ts_show_egt
 
+menu = "&Electronic Throttle"
+
+		subMenu = etbDialog,				"Electronic Throttle Settings" @@if_ts_show_etb
+		subMenu = std_separator
```
rusefi: Please note the "@@if_ts_show_etb" conditional thing. I believe that for configurations which do not show ETB we will have a nasty artifact with a en empty menu.

Let me remind that changes which benefit you should not negatively affect other configurations. Prometheus would be an example which ts_show_etb false. Please invoke gen_config and test how would rusefi_prometheus.ini look

rusefi:
You are adding one top menu and changing the other top menu name. It looks like 1100 pixels wide screen was needed to display all menus without your changes so that all words would fit. How much screen width is now required to display all the top menus completely? Please attach a screen shot with your changes

960:
javaw_mSPnpfsxtG

960:
This is with 25% overscaling at screen

rusefi:
Oh, so it does line breaks and things look nice. Thank you, so not a problem with screen with.

960:
I have OCD and keeps on forever with the smallest details to get it perfect, and never get completly satisfied.                    
So, it's painful :-)

960:
Have also been working a little with the dialogs, testing different approaches.

Am getting closer, but still lots to do

YnUGVsPLvb

960:
As I have split it up in small PR's you dont get to see the full "result".
But you can have the full files if you want to have a look?

rusefi:
i have my priorities and little time slots. i would bot be able to review comolete changes

960:
I understand.

Was thinking more if you wanted to try them out to see if you like any of it.

960:
javaw_xqy36lT5ZR

mck1117:
Try splitting out the FSIO in to one channel per window.  The window you have a screenshot of is massive, which is very hard to process for a user.  Too much information on one page is overwhelming.

rusefi:
this PR would be closed the moment it has any third unrelated change. at the moment there are two unrelated changes here issues are already introduced

mck1117:
A good place to start is normalizing the case of menu items.  Get that one out of the way, and do ONLY that in one PR.

960:
The other changes than this file was far from finished. Just been testing different methods and map-types.

But I think we can just close this anyway.

Issue #861, opened 2019-06-26
Use static_assert where possible


mck1117:
There are a number of places where we do things like:

```
#define SOME_PARAMETER 25
#define ANOTHER_PARAMETER 40
... later ...

if(SOME_PARAMETER > ANOTHER_PARAMETER)
    firmwareError("bad parameter");

or

if(sizeof(some_struct_s) != EXPECTED_STRUCT_SIZE)) ...;
```

C++ has just the feature for this.  If we instead write:
`static_assert(SOME_PARAMETER <= ANOTHER_PARAMETER);`
or
`static_assert(sizeof(some_struct_s) == EXPECTED_STRUCT_SIZE);`
a compiler error will be generated if the condition is unmet (instead of a runtime error).

Here's the reference for `static_assert`: https://en.cppreference.com/w/cpp/language/static_assert

PR #871, opened 2019-07-05
Tunerstudio.cpp cleanup

mck1117:
Kill some dead code, convert the main TS handler to a switch, and use `static_assert`.

### DIFF
--- a/firmware/console/binary/tunerstudio.cpp
+++ b/firmware/console/binary/tunerstudio.cpp
@@ -192,49 +192,18 @@ void tunerStudioDebug(const char *msg) {
 char *getWorkingPageAddr(int pageIndex) {
 	switch (pageIndex) {
 	case 0:
-#if !defined(EFI_NO_CONFIG_WORKING_COPY)
+#ifndef EFI_NO_CONFIG_WORKING_COPY
 		return (char*) &configWorkingCopy.engineConfiguration;
 #else
 		return (char*) engineConfiguration;
 #endif /* EFI_NO_CONFIG_WORKING_COPY */
-//	case 1:
-//		return (char*) &configWorkingCopy.ve2Table;
-//	case 2:
-//		return (char*) &configWorkingCopy.fuelTable;
-//	case 3:
-//		return (char*) &configWorkingCopy.ignitionTable;
-//	case 4:
-//		return (char*) &configWorkingCopy.veTable;
-//	case 5:
-//		return (char*) &configWorkingCopy.afrTable;
-//	case 6:
-//		return (char*) &configWorkingCopy.injectionPhase;
-//	case 2: // fuelTable
-//	case 3: // ignitionTable
-//	case 4: // veTable
-//	case 5: // afrTable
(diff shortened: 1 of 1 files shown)


firmware/console/binary/tunerstudio.cpp
```diff
@@ -263,7 +232,7 @@ void handlePageSelectCommand(ts_channel_s *tsChannel, ts_response_format_e mode,
  */
 static void onlineApplyWorkingCopyBytes(int currentPageId, uint32_t offset, int count) {
 	UNUSED(currentPageId);
-	if (offset > sizeof(engine_configuration_s)) {
+	if (offset >= sizeof(engine_configuration_s)) {
```
mck1117: This was a potential off-by-one buffer overflow.

firmware/console/binary/tunerstudio.cpp
```diff
@@ -751,85 +720,88 @@ int tunerStudioHandleCrcCommand(ts_channel_s *tsChannel, char *data, int incomin
...
 
-	// Output command first since it's popular
-	if (command == TS_OUTPUT_COMMAND) {
-		uint16_t offset = *(uint16_t *) (data);
-		uint16_t count = *(uint16_t *) (data + 2);
-		handleOutputChannelsCommand(tsChannel, TS_CRC, offset, count);
-	} else if (command == TS_HELLO_COMMAND) {
+	const uint16_t* data16 = reinterpret_cast<uint16_t*>(data);
+
+	switch(command)
```
mck1117: Changing to a switch has a few benefits:  Not only does it make it clearer to the reader that this is a "pick one" scenario, it also gives the compiler more leeway about optimization.  `if else` statements are required to be tested in order, but `switch` isn't.  This means the compiler can do smart things like replace it with a jump table (which it does with `-O2` enabled), so that this is an O(1) lookup instead of an O(n) lookup.

mck1117: (looking at `rusefi.list`, it uses the super cool `tbh` instruction: http://www.keil.com/support/man/docs/armasm/armasm_dom1361289912399.htm)

firmware/console/binary/tunerstudio.cpp
```diff
@@ -751,85 +720,88 @@ int tunerStudioHandleCrcCommand(ts_channel_s *tsChannel, char *data, int incomin
...
 	return true;
 }
 
 void startTunerStudioConnectivity(void) {
-	if (sizeof(persistent_config_s) != getTunerStudioPageSize(0))
-		firmwareError(CUSTOM_OBD_TS_PAGE_MISMATCH, "TS page size mismatch: %d/%d", sizeof(persistent_config_s), getTunerStudioPageSize(0));
-
-	if (sizeof(TunerStudioOutputChannels) != TS_OUTPUT_SIZE)
-		firmwareError(CUSTOM_OBD_TS_OUTPUT_MISMATCH, "TS outputs size mismatch: %d/%d", sizeof(TunerStudioOutputChannels), TS_OUTPUT_SIZE);
+	// Assert tune & output channel struct sizes
```
mck1117: The result of these comparisons is known at compile time (ie, it won't even compile a branch here) - so we can throw the error at compile time. (#861)

PR #875, opened 2019-07-08
Periodic timer controller

rusefi:
The end goal is to shift away from "one thread per one task" paradigm currently used by alternator/aux pid/ETB/etb

As a first step I am extracting existing pattern used for fast and slow and plan to use same helper class for the above mentioned current tasks which would become not tasks.

FYI @mck1117 @andreika-git

### DIFF
--- a/firmware/controllers/system/periodic_timer_controller.h
+++ b/firmware/controllers/system/periodic_timer_controller.h
@@ -13,12 +13,18 @@ class PeriodicTimerController;
 
 void runAndScheduleNext(PeriodicTimerController *controller);
 
+/**
+ * this is an intermediate implementation - we should probably move from using virtual_timer_t which works on interrupts
+ * into sharing one thread with potentially lower priority instead
+ */
+
 class PeriodicTimerController {
 
 public:
+#if !EFI_UNIT_TEST
 	virtual_timer_t timer;
+#endif /* EFI_UNIT_TEST */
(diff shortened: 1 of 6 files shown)


firmware/controllers/system/periodic_timer_controller.h
```diff
@@ -0,0 +1,42 @@
...
+
+class PeriodicTimerController {
+
+public:
+#if !EFI_UNIT_TEST
+	virtual_timer_t timer;
+#endif /* EFI_UNIT_TEST */
+	int periodMs;
+
+	virtual void PeriodicTask() = 0;
```
mck1117: One option is to have the task's implementation return the period (change the signature here to `virtual uint32_t PeriodicTask() = 0`).  Since we can't actually do anything about period changes at any time other than immediately after the task runs, returning it helps convey that limitation (it also saves having to store the period somewhere).

firmware/controllers/system/system.mk
```diff
@@ -3,6 +3,7 @@ SYSTEMSRC =
 
 SYSTEMSRC_CPP =	\
 	$(PROJECT_DIR)/controllers/system/efi_gpio.cpp \
+	$(PROJECT_DIR)/controllers/system/periodic_timer_controller.cpp \
```
mck1117: rename file to `periodic_task.cpp`?

PR #879, opened 2019-07-09
Remove warmup afr pid

mck1117:
Remove the warmup afr pid - it does the same thing as the normal short-term fuel PID, plus the normal warmup fuel curve.

Also saves 576 bytes of code space at level `-O2`, a handful of branches, and 56 bytes + 1 bit of config space.

(requires a config re-gen to compile)


mck1117:
poke

rusefi:
Why now? Is anything driving this change?

I am jet lagged now, I would prefer to think about this in a couple days later once/if I gain my brain back.

mck1117:
There's not a particular rush, no.  It's just that this functionality was causing bloat, unnecessary complexity in `periodicFastCallback`, and potential confusion for users about a feature that overlaps partially (in a sort of confusing way) with others.  Plus it's 56 bytes of config we can reclaim.

mck1117:
After further thought - this is actually broken, and maybe dangerous.  This function tries to adjust the fueling output to match the AFR in the `warmupTargetAfr` table (a function of coolant temperature), effectively **ignoring** the values from the normal AFR target table.

If you have a value of 14.7 (stoichiometric) in this table, and stand on the gas, the ECU will try to correct BACK to 14.7 even if the open-loop AFR target is something different.  Running at 14.7 under boost WILL overheat pistons, even when the engine is cold.

rusefi:
@mck1117 can you please rebase to resolve the merge conflict? or what is the smart way for me to handle it?

mck1117:
yes - I can resolve

mck1117:
fixed - ready for merge

PR #880, opened 2019-07-10
Fix some maf logic

mck1117:
- Remove maf decoding cache.  It isn't on a particularly hot path, so the negligable performance gains aren't worth it.  Also, chopping down to 0.2 volt increments loses so much resolution that MAF wouldn't actually work very well.
- `getMafVoltage` now gets the voltage at the maf pin
- `getRealMaf` gets the mass flow rate, in kg/hr (TODO: should we do kg/s or g/s instead?

This change makes a negligible change to compiled image size, but saves 1KB (!) of RAM.

### DIFF
--- a/firmware/controllers/sensors/maf.h
+++ b/firmware/controllers/sensors/maf.h
@@ -2,7 +2,7 @@
  * @file    maf.h
  * @brief
  *
- * by the way 2.081989116 kg/h = 1 ft³/m
+ * by the way 2.081989116 kg/h = 1 ftï¿½/m
  *
  *
  * @date Nov 15, 2013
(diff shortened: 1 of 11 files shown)


firmware/controllers/sensors/maf.h
```diff
@@ -2,7 +2,7 @@
  * @file    maf.h
  * @brief
  *
- * by the way 2.081989116 kg/h = 1 ft�/m
+ * by the way 2.081989116 kg/h = 1 ft�/m
```
rusefi: here we change risky but working symbol into something not working?

mck1117: ah geez this is the same thing as the degree symbols

mck1117: Hey look, there's even a bug filed for it on vscode!
https://github.com/Microsoft/vscode/issues/52642

mck1117: now we change risky but working symbol in to not-risky but working two symbols

OrchardPerformance: Could always abbreviate that to CFM, its a pretty common abbreviation in the aftermarket as its used for cylinder head flows all the time.

PR #895, opened 2019-07-31
microRusefi Default Configuration

mck1117:
It's the moment you've all been waiting for, you can now run an MRE build, wipe the board, flash firmware, and it's configured properly!


firmware/config/boards/microrusefi/!compile-mre.bat
```diff
@@ -3,16 +3,8 @@ cd ../../..
...
+set PROJECT_BOARD=microrusefi
 set PROJECT_CPU=ARCH_STM32F7
-set EXTRA_PARAMS=-DDUMMY ^
- -DSTM32F767xx ^
- -DEFI_USE_OSC=TRUE ^
- -DFIRMWARE_ID=\"microRusEfi\" ^
- -DEFI_COMMUNICATION_PIN=GPIOE_1 ^
- -DEFI_FATAL_ERROR_PIN=GPIOE_2 ^
- -DEFI_RUNNING_PIN=GPIOE_3 ^
- -DEFI_WARNING_PIN=GPIOE_4
```
mck1117: in general we should set this stuff in the `.mk` when we can - it's far sturdier and more portable

firmware/config/boards/microrusefi/board_configuration.cpp
```diff
@@ -0,0 +1,203 @@
...
+ */
+void setBoardConfigurationOverrides(void) {
+	engineConfiguration->baroSensor.hwChannel = EFI_ADC_NONE;
+	engineConfiguration->throttlePedalPositionAdcChannel = EFI_ADC_NONE;
+
+	// not used
+	engineConfiguration->dizzySparkOutputPin = GPIO_UNASSIGNED;
+	engineConfiguration->externalKnockSenseAdc = EFI_ADC_NONE;
+	engineConfiguration->displayMode = DM_NONE;
+	boardConfiguration->HD44780_rs = GPIO_UNASSIGNED;
```
rusefi: Do we need any of GPIO_UNASSIGNED in master? is not master NOT applying Frankenso configs? shall we consider the fact that we have all zeros and GPIO_UNASSIGNED is zero?

PR #902, opened 2019-08-17
Attempt at Boost/VVT control. Not working btw, so if you can have a look?

960:


rusefi:
This is https://en.wikipedia.org/wiki/Copy-and-paste_programming - that's a dangerous practice and should be avoided. See also https://en.wikipedia.org/wiki/Duplicate_code

960:
There are no copy paste in the TS part at all.

The boost controller.cpp was copied from alternator_controller and tried edited. I wrote it's not working. So that part can be skipped. For the TS part, I dont think there are much to complain about. This whole BS are just way too childish :-)

rusefi:
Unfortunately I do not have time to skip and filter. Please increase the quality of your potential contribution - there should be a clear description of the problem statement, and focus on one problem at a time.

PR #903, opened 2019-08-18
TS Boost settings + maps for boost and vvt

960:
Have a new look and a new feedback how bad it is :-)


rusefi:
It is my opinion that boost, same as VVT, should be implemented on top of FSIO.

I am not ready to do the analysis of what is boost and how exactly it should be implemented using FSIO.

It is my opinion that firmware changes have to happen before it would make sense to spend time on the TS change. TS is only the graphical view for firmware logic - if there is no firmware it is not yet time to spend time on TS.

I humbly ask to leave me alone at least for the next 48 hours.

960:
Thanks for giving me a good laugh :-)

960:
Now I have to get someone to write the code for it just for the next entertainment :-)

960:
The code landed at around 80 lines, and works just fine. 
Will PR you this nice contribution next time I have friends at visit.

PR #920, opened 2019-09-07
TS update 

960:


rusefi:
Please elaborate on what are the areas change in what way?

960:
It's only some small changes.

It lets you hide/show the controller menu in project settings

0G0cyKhjkp

I changed the "FSIO" menu to "Advanced", and added a initial vvt setup (Auxpid).

I also added this ETB setup

javaw_XjpjArZBFs

mck1117:
Does the VVT page actually do anything, or is it dead UI?

960:
It's the VVT settings we have so far together with Aux-pid and fsiotable1

So it works the same as doing those separate.

It puts out pwm, and seems to work as supposed.

I have four flanks cam trigger wheel, so I cant get it tested in car.

rusefi:
Not sure what is the exact improvement here. Look like a couple unrelated changes, not sure if these changes are positive or negative

960:
I will have to agree with that.

This crap didn't add anything of use.

960:
I dont see any value in any of this "Refactoring" thing.

The setup we have works just fine as is.

Can we agree to stop messing with this BS, and focus on important things instead?

rusefi:
No we cannot, it's not a democracy. Feel free to discuss this further on #strategy_without_drama slack room.

PR #934, opened 2019-09-14
Store table ratio at compile time

mck1117:
Since the ratio from stored value to "real" value in a table can't change at runtime, we should store and compute it all at compile time.  This change uses `std::ratio` to embed the ratio in the type, instead of expecting consumers to remember to use it.

Also moves implementation up in to the declaration, to prevent having to re-write the full templated type every single time (see old file, line 110 for an example).

### DIFF
--- a/firmware/controllers/algo/engine2.cpp
+++ b/firmware/controllers/algo/engine2.cpp
@@ -24,6 +24,7 @@
 #endif
 
 extern fuel_Map3D_t veMap;
+extern fuel_Map3D_t veTpsMap;
 extern afr_Map3D_t afrMap;
 
 EXTERN_ENGINE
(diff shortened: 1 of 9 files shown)


firmware/controllers/algo/engine2.cpp
```diff
@@ -203,10 +204,11 @@ void EngineState::periodicFastCallback(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 		 */
 		if (CONFIGB(useTPSBasedVeTable)) {
 			// todo: should we have 'veTpsMap' fuel_Map3D_t variable here?
-			currentRawVE = interpolate3d<float, float>(tps, CONFIG(ignitionTpsBins), IGN_TPS_COUNT, rpm, config->veRpmBins, FUEL_RPM_COUNT, veMap.pointers);
```
mck1117: Why does this use `ignitionTpsBins` for the Y axis?

rusefi:
@mck1117 what is the state of this PR? does it pass unit tests? Do we believe it to work as intended and not break anything?

rusefi:
@mck1117 what is the state of this PR? does it pass unit tests? Do we believe it to work as intended and not break anything?

rusefi:
closing due to lack of communication from contiutor

PR #939, opened 2019-09-17
Reduce PID significant figures from 7 -> 4

mck1117:
After we fixed PID to properly use `dt`, we don't need so many digits that the string doesn't fit in the text box.

Would you be able to tell that the P-factor is set to 1.5?
image

### DIFF
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -147,9 +147,9 @@ custom angle_table_t	4*@@IGN_RPM_COUNT@@x@@IGN_LOAD_COUNT@@ array,   F32,   @OFF
 custom pedal_to_tps_t	@@PEDAL_TO_TPS_SIZE@@x@@PEDAL_TO_TPS_SIZE@@ array,   U08,   @OFFSET@, [@@PEDAL_TO_TPS_SIZE@@x@@PEDAL_TO_TPS_SIZE@@],"deg",	   1,     0,     -720,  720,    2
 
 struct pid_s
-	float pFactor;;"",      1,      0,       -10000, 10000,      3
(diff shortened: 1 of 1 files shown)


rusefi:
how about 7 > 4 to begin with?

mck1117:
I already thought about 2, but decided we might need a third.  Do you know of an actual use of >2?

rusefi:
```
	// VVT closed loop
	engineConfiguration->auxPid[0].pFactor = 2;
	engineConfiguration->auxPid[0].iFactor = 0.005;
	engineConfiguration->auxPid[0].dFactor = 0;
	engineConfiguration->auxPid[0].offset = 33;
```

```
	// below 20% this valve seems to be opening for fail-safe idle air
	engineConfiguration->idleRpmPid.minValue = 20;
	engineConfiguration->idleRpmPid.pFactor = 0.01;
	engineConfiguration->idleRpmPid.iFactor = 0.00001;
	engineConfiguration->idleRpmPid.dFactor = 0.0001;
	engineConfiguration->idleRpmPid.periodMs = 100;

```

stefanst:
My idle "p" factor is 0.005 - so I would very much like to keep at least 3

mck1117:
Yes - 3 or 4 is sensible, as that's in the range of sensible idle control stuff.  100 RPM error with a 0.001 I-term would make a 0.1%/second change, which is probably the lower limit of useful.

@rusefi why is your I factor 0.00001? That means 0.00001% per RPM error per second.  So a 100 rpm error for 10 seconds would only change the idle position by 0.01%.

rusefi:
I would not pay too much attention to my data, I would value @stefanst 's input way more

mck1117:
Upped 3 -> 4, still fits okay on PC and mac.

PR #940, opened 2019-09-18
Kill Dead Code 2: Electric Boogaloo

mck1117:


rusefi:
I have mixed feelings about this. Yes, that's an empty skeleton at this point but also we might agree that we will need these files eventually?

What do we gain today by removing this code?

mck1117:
It's going to be replaced by the new sensor logic anyway.  Right now, it's adding compile cost (and user confusion when they see the option in the UI) for zero functionality that actually runs.

rusefi:
Whatever. I believe it's a distraction to bother but whatever.

PR #946, opened 2019-09-22
hook up rpm acceleration

mck1117:
We already had a field for it - and the math to do it.  There's even a gauge!


rusefi:
```tsOutputChannels->rpmAcceleration engine->rpmCalculator.getRpmAcceleration();``` wow

PR #947, opened 2019-09-22
Kill dead code 3: Return of the Jedi

mck1117:
Kill even more dead code.

ADC filtering needs further thought at some point in the future, but probably a median filter instead of biquad.


rusefi:
It i my subjective opinion that this is not a useful change.

PR #956, opened 2019-09-25
Fix sensor mocking bug

mck1117:
Actually really reset the registry when you ask it to.

### DIFF
--- a/firmware/controllers/sensors/sensor.cpp
+++ b/firmware/controllers/sensors/sensor.cpp
@@ -35,6 +35,7 @@ bool Sensor::Register() {
 		auto &entry = s_sensorRegistry[i];
 
 		entry.sensor = nullptr;
+		entry.useMock = false;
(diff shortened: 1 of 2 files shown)


rusefi:
how come tests did not reveal issue? do we need changes in tests?

mck1117:
The issue was that state was leaking between tests.  Since the test run order isn't guaranteed, this one is pretty hard to test.  Let me see if I can come up with something, though.

mck1117:
I suppose I can at least test that `resetRegistry` is really fully resetting it.

rusefi:
unit tests have failed - reverting commit

[----------] Global test environment tear-down
[==========] 100 tests from 27 test cases ran. (34 ms total)
[  PASSED  ] 99 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] SensorMocking.ResetWithResetRegistry

PR #963, opened 2019-10-03
fix gm 24x decoder

mck1117:
Documentation online said that the teeth are 3 or 12 degrees long.  My car and an oscilloscope says that they're 5 and 10 degrees long (which actually makes more sense).

### DIFF
--- a/firmware/config/engines/chevrolet_camaro_4.cpp
+++ b/firmware/config/engines/chevrolet_camaro_4.cpp
@@ -43,8 +43,8 @@ void setCamaro4(DECLARE_CONFIG_PARAMETER_SIGNATURE) {
 	engineConfiguration->tps1_1AdcChannel = EFI_ADC_2;
 
 	// todo: move this into trigger definition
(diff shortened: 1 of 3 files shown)


mck1117:
hmm - looks like this trips a fatal error when trying to calculate the sync point

rusefi:
what is the state of this PR? does it pass unit tests? do simulator and firmware compile?

mck1117:
It compiles, but tests (and firmware) fail because the sync ratios need tweaking.

mck1117:
This is now actually fixed - some of the logic was flipped too.

Verified with unit tests, and by getting a real ECU to sync with a real engine:

https://www.youtube.com/watch?v=t8rsThX2rtE

(pardon the crappy video, this was shot at night, thru the gap in the open hood, with my laptop sitting on the slam panel up front)

mck1117:
fyi we're syncing on these three rising edges, sync ratio gaps 2-0.5-2, ending at 310 degrees:

image

Issue #639, opened 2018-12-29
ICU is not really needed, EXTI should be enough for digital input interrupt


rusefi:
See top of digital_input_hw.cpp file for details

rusefi:
@mck1117 says 
what is to be gained by removing it?
1) Surely all of you guys have wanted to add a feature to the firmware, but looked at it and said "damn, I wish that didn't look so complicated."  I've heard that exact feedback from a few people who wanted to contribute, but found the firmware too daunting to comprehend, so they gave up on the project.
   If there's a desire for more hands to help out on firmware, then there's immeasurable value in making it easier to look at the firmware and understand what it's doing.
2) Using the ICU is preventing us from switching to tickless mode.  Right now the OS is firing an interrupt every 1ms that takes a global lock, contributing to jitter and CPU load.
3) It saves 17k of flash.  That's real value, since it takes pressure off of the cache/flash prefetcher, and branch predictor.

rusefi:
@mck1117 please elaborate on "ing the ICU is preventing us from switching to tickless mode. Right now the OS is firing an interrupt every 1ms that takes a global lock, contributing to jitter and CPU load." - where is the tick coming from / how is this related to tickless

1. and 3. are great reasons to move forward with this PR I guess

rusefi:
TIM2 is needed by tickless, that's the connection.

Issue #630, opened 2018-12-09
migrate slow ADC to software timer so that TIM8 is also available for input capture


rusefi:
We would like to change
#define STM32_ICU_USE_TIM8                  FALSE
to
#define STM32_ICU_USE_TIM8                  FALSE
which means that

#define STM32_PWM_USE_TIM8                  TRUE
should be changed to
#define STM32_PWM_USE_TIM8                  FALSE

which means that 
#define EFI_INTERNAL_SLOW_ADC_PWM	&PWMD8
has to go

PR #971, opened 2019-10-10
Remove ICU

mck1117:
Remove ICU completely, allowing only EXTI for trigger input (and COMP on Kinetis 😉), closing #639 and #630.

This change removes the ICU code, disables it in `mcuconf.h` and `halconf.h`, and fixes up usages of the ICU to use EXTI instead (frequency MAP and vehicle speed).  Trigger, VSS, and freq MAP have all been tested and work great.

Somehow this change also manages to save 17 kilobytes of flash space on a release build. Yep. 17k.


rusefi:
did we run any performance comparison between icu and exti?
i have an interesting mental block about removing old code.

rusefi:
did we run any performance comparison between icu and exti?
i have an interesting mental block about removing old code. 

@andreika-git what do you think?

firmware/development/wave_analyzer.cpp
```diff
@@ -203,49 +178,45 @@ static float getSignalPeriodMs(int index) {
...
-	if (readers[index].hw->started) {
-//	int counter = getEventCounter(index);
-//	debugInt2(logging, "ev", index, counter);
 
-		float dwellMs = getSignalOnTime(index);
-		float periodMs = getSignalPeriodMs(index);
-
-		appendPrintf(logging, "duty%d%s", index, DELIMETER);
-		appendFloat(logging, 100.0f * dwellMs / periodMs, 2);
+	float dwellMs = getSignalOnTime(index);
```
mck1117: I'm not sure why it thought this was a diff, all I did was unwrap it from an `if (...)`

firmware/development/wave_analyzer.cpp
```diff
@@ -266,12 +237,12 @@ void initWaveAnalyzer(Logging *sharedLogger) {
 #if EFI_WAVE_ANALYZER
 	initWave(PROTOCOL_WA_CHANNEL_1, 0);
 	initWave(PROTOCOL_WA_CHANNEL_2, 1);
+	initWave(PROTOCOL_WA_CHANNEL_3, 2);
+	initWave(PROTOCOL_WA_CHANNEL_4, 3);
```
mck1117: looks like these were just missing before?

rusefi:
pins 0 thru 4 get their own interrupt

5-9 share one
10-15 share one

@mck1117 says
all the code has to do is check which mask bits are set in the IMR register (ie, bit 7 would be set if channel 7 caused the interrupt)
I looked at the disassembly and it's two cycles per channel
effectively
```
skip_next_if_bit_not_set(pr,5);
handle_exti_5();
skip_next_if_bit_not_set(pr,6);
handle_exti_6();
skip_next_if_bit_not_set(pr,7);
handle_exti_7();
```
...

rusefi:
I would like to take a different approach on this one:

first, I would like to unify ICU and PAL implementations for those to provide same internal API, then I would like to keep ICU via compile-time defines. I know there might be different opinions on this but hopefully we can agree that this is not too important.

Issue #973, opened 2019-10-13
[MOST IMPORTANT TICKET] Both HandleFuel and PeriodicFastCallback call GetBaseFuel


mck1117:
Why do both of these compute the same thing? Should one be storing it for the other to use?

![image](https://user-images.githubusercontent.com/568254/66712449-a05b9080-ed51-11e9-9ff5-5420855f2194.png)

Top row is the trigger interrupt.  Bottom row is the periodic timer running PeriodicFastCallback.

mck1117:
Some napkin math: the trigger interrupt invocation, on a 60-2 wheel, at 6000 RPM, gets called 5800 times per second, representing around 14% CPU usage.

rusefi:
![image](https://user-images.githubusercontent.com/7377949/66803112-b18cd480-eeec-11e9-98a0-067f78801eb2.png)

I can only think of acceleration enrichment as a reason?

mck1117:
A reasonable compromise may be:
1) remove it from `handleFuel`
2) increase `PeriodicFastCallback` to 200hz

mck1117:
That will only use up around 2% CPU for the `PeriodicFastCallback`, and will run it at a lower priority than it currently runs, to prevent starvation of the trigger decoder interrupt.

rusefi:
Prior to merging existing PR I would like to add more unit tests to ensure that functionally we would not be affected

rusefi:
Fixed by #979

PR #979, opened 2019-10-15
Performance improvement: call GetBaseFuel less

mck1117:
As shown in #973, both `MainTriggerCallback` and `PeriodicFastCallback` call `GetInjectionDuration`, updating the stored injection duration in the engine state.

The trigger callback calls it for every tooth we actually use (generally all rising or all falling edges).  `GetInjectionDuration` takes about 32us to execute, incurring a 32us cost in every tooth decode interrupt.  On high tooth count triggers, this one call can incur significant CPU cost.  For example, on a 60-2 wheel, it's called 5800 times per second, which amounts to 19% of total CPU load on an STM32F4 at 168MHz.

This change omits the call from MainTriggerCallback, and increases the frequency of the `PeriodicFastCallback` to 200hz to make up for the removal.  Since `PeriodicFastCallback` takes around 100us, this will use a constant 2% CPU (up from 0.5% for PFC previously, plus a variable amount from MainTriggerCallback).

All of the calls made to the various acceleration enrichment algorithms made by `getInjectionDuration` are idempotent, so it's safe to call them whenever without changing the result.  Important state is updated only as needed from the trigger callbacks.

### DIFF
--- a/firmware/controllers/algo/engine.cpp
+++ b/firmware/controllers/algo/engine.cpp
@@ -368,13 +368,9 @@ void Engine::periodicFastCallback(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 
 	engine->m.beforeFuelCalc = getTimeNowLowerNt();
 	int rpm = GET_RPM();
-	/**
-	 * we have same assignment of 'getInjectionDuration' to 'injectionDuration' in handleFuel()
-	 * Open question why do we refresh that in two places?
-	 */
+
 	ENGINE(injectionDuration) = getInjectionDuration(rpm PASS_ENGINE_PARAMETER_SUFFIX);
 	engine->m.fuelCalcTime = getTimeNowLowerNt() - engine->m.beforeFuelCalc;
-
 }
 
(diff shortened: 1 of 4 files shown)


rusefi:
I am not convinced that this change would not have functional consequences. Before anything could be changed in this area I believe that some testing coverage should be added.

rusefi:
It is my expectation that every PR is merge-able by default. It seems like we shall adjust expectations :)

Since there is no automation for in-branch unit test validation and I have no interest to validate unit tests manually, I would have to revert this change.

Going forward I would expect a clear human-readable confirmation that
1) three projects compile
2) unit tests are passing

rusefi:
[  PASSED  ] 99 tests.
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] misc.testFuelMap
[  FAILED  ] misc.test1995FordInline6TriggerDecoder
[  FAILED  ] misc.testRpmCalculator
[  FAILED  ] fsio.testLogicExpressions
[  FAILED  ] fuelCut.coasting
[  FAILED  ] idle.timingPid

 6 FAILED TESTS

PR #980, opened 2019-10-17
fix wwae for multi cylinder engines

mck1117:
Previously, were sharing one state for all cylinders.  This means it would run N times too fast, making the wall wetting parameters wrong.


mck1117:
image

PR #991, opened 2019-10-31
Kinetis & TLE8888 compatibility & fixes...

andreika-git:


firmware/hw_layer/drivers/gpio/tle8888.c
```diff
@@ -40,13 +40,6 @@
 #include "pin_repository.h"
 #include "os_util.h"
 
-/* to be removed  */
```
andreika-git: This was not compatible with 'activeConfiguration_s &' because of C++-reference included from .c-file.

PR #992, opened 2019-10-31
Tps Calibration

960:
ETB Tps calibration under tools menu.

Syncs controller and TS value

I dont know how to make a command, so AC relay test are used.

So that needs to be edited

### DIFF
--- a/firmware/console/binary/tunerstudio_configuration.h
+++ b/firmware/console/binary/tunerstudio_configuration.h
@@ -176,7 +176,9 @@ typedef struct {
 	float etbTarget;		// 312
 	float etb1DutyCycle;	// 316
 	float etb1Error;		// 320
-	int unused3[8];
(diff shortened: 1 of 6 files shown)


rusefi:
I do not understand what is 'calibrateMode' and why is it needed, please add comments into the code explaining the logic

you cannot really sleep from TS commands - TS communication would be halted while you sleep and this could cause USB disconnect or TS timeout. proper implementation should be only requesting calibration from command while executing calibration from some other unrelated thread which is allowed to sleep

960:
It adds these menus under tools.

This is for running the tps calibration and maintain/sync the tpsMin and tpsMax values in TS.

It works just fine, and it's the solution we got from EfiAnalytics.

image

rusefi:
please explain problem all the way from the beginning and solution in the CODE so that three years from now other people could understand what's going on

PR #1000, opened 2019-11-07
Performance tracing

mck1117:


rusefi:
Conflicts we have here

We need some readme content somewhere on how to actually view the results?

mck1117:
Yes - needs documentation, some cleanup, and the tooling to get it off of the MCU and convert to the json format that Chrome can consume.

firmware/development/perf_trace.cpp
```diff
@@ -0,0 +1,93 @@
...
+	PE Event;
+	EPhase Phase;
+	uint8_t Data;
+	uint8_t ThreadId;
+	uint32_t Timestamp;
+};
+
+static_assert(sizeof(TraceEntry) == 8);
+
+static TraceEntry s_traceBuffer[TRACE_BUFFER_LENGTH];
```
andreika-git: We still need  #if ENABLE_PERF_TRACE because of large s_traceBuffer.

mck1117: yes - I haven't guarded the stuff in here yet with `ENABLE_PERF_TRACE`

firmware/console/binary/tunerstudio.cpp
```diff
@@ -848,6 +852,14 @@ int tunerStudioHandleCrcCommand(ts_channel_s *tsChannel, char *data, int incomin
...
 
 		break;
+	case TS_PERF_TRACE_BEGIN:
+		perfTraceEnable();
+		break;
+	case TS_PERF_TRACE_GET_BUFFER:
+		{
+			auto trace = perfTraceGetBuffer();
+			sr5SendResponse(tsChannel, TS_CRC, trace.Buffer, trace.Size);
+		}
```
rusefi: do we need a "break" here?

rusefi:
This PR was not linking on F4 due to RAM usage. https://github.com/rusefi/rusefi/blob/master/misc/SoftwareDevelopmentProcess.md now spells out F4 as official primary hardware platform.

Still trying to understand what's broken with real hardware test.

PR #1004, opened 2019-11-12
Fix

mck1117:
Don't put header includes inside an `extern "C"` block when that header has its own guards.


rusefi:

PR #1005, opened 2019-11-13
Put back static asserts, fix the mess

mck1117:
If you have a header included inside an `extern "C"` block, you need to consider very **very** carefully what you're doing.  Pretty much only function declarations should be inside an `extern "C"`.  No includes, no defines, nothing else.

This change does not alter the compiled binary.

### DIFF
--- a/firmware/config/stm32f7ems/chconf.h
+++ b/firmware/config/stm32f7ems/chconf.h
@@ -710,6 +710,9 @@
 
 #ifndef __ASSEMBLER__
 
+#ifdef __cplusplus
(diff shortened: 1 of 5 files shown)


rusefi:

rusefi:
it is unfortunate that baing aware on the risks in this area we have PR which does not compile with both firmware and unit teste

just to clarify, gate check is currently unreliable and only covers unit tests. i would like to ask for a bit more validation on the side of the contributor. the low quality of trivial PRs makes me question the quality of larger PRs

PR #1006, opened 2019-11-13
Fix f4/kinetis

mck1117:
Fix f4 and kinetis.  We have 3 identical chconf files, which is great.


rusefi:

PR #1008, opened 2019-11-17
cisnan() warnings fix

andreika-git:
We use a bit of C++ magic to fix tons of these warnings:
```
    util/efilib.h:48:35: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
    #define cisnan(f) (*(((int*) (&f))) == 0x7FC00000)
                                   ^
    controllers/error_handling.h:54:59: note: in definition of macro 'efiAssertVoid'
```
Due to the gnu-c++ optimisation, the binary code produced is practically the same as before:
```
      87              	@ aliasing-test.cpp:15: 	if (cisnan(oneDegreeUs)) {
      88 0010 20309FE5 		ldr	r3, .L6+4	@ tmp118,
      89 0014 030050E1 		cmp	r0, r3	@, tmp118
      90 0018 0200000A 		beq	.L5	@,
```
But without optimisation it's slightly larger (by 2-3 opcodes), so we leave it as it was for -O0...

### DIFF
--- a/firmware/util/math/interpolation.h
+++ b/firmware/util/math/interpolation.h
@@ -46,7 +46,8 @@ int needInterpolationLogging(void);
  */
 template<typename kType>
 int findIndexMsgExt(const char *msg, const kType array[], int size, kType value) {
-	if (cisnan(value)) {
+	float fvalue = (float)value;
+	if (cisnan(fvalue)) {
 		firmwareError(ERROR_NAN_FIND_INDEX, "NaN in findIndex%s", msg);
 		return 0;
 	}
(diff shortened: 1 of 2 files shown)


rusefi:

firmware/util/math/interpolation.h
```diff
@@ -46,7 +46,8 @@ int needInterpolationLogging(void);
  */
 template<typename kType>
 int findIndexMsgExt(const char *msg, const kType array[], int size, kType value) {
-	if (cisnan(value)) {
+	float fvalue = (float)value;
```
mck1117: If kType isn't a float, then it can't be NaN anyway.  It'd be better to do `if (std::is_floating_point_v<kType> && cisnan(value))`  That way, cisnan is never called if `kType` isn't a float

PR #1009, opened 2019-11-19
ETB Deviation error-handling

960:


rusefi:

960:
Please check if this conflicts with anything or needs some refinement.

It seems to work correct.

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -252,6 +246,17 @@ static percent_t currentEtbDuty;
 		percent_t etbIdleAddition = CONFIGB(useETBforIdleControl) ? engine->engineState.idle.etbIdleAddition : 0;
 		percent_t targetPosition = engine->engineState.targetFromTable + etbIdleAddition;
 
+
+//This prevents ETB to hit hard-stops
+
+		if (targetPosition > 99.0f) {
```
rusefi: this is not related to deviation error-handling, I am mostly afraid of 1% since it's quite a large number

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -269,6 +274,29 @@ static percent_t currentEtbDuty;
 
 		etb1.dcMotor.Set(ETB_PERCENT_TO_DUTY(currentEtbDuty));
 
+	   	float etbError = absF(targetPosition - actualThrottlePosition); // Used for I-Term limiter and Error-detection
+
+
+        // ETB Deviation monitoring
+        if(etbError < 3 ){
```
rusefi: I have to say that this starts to looks like real code!

On a more boring note, code formatting is important. Nicer way to write the same would be "if (etbError < 3) {"

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -269,6 +274,29 @@ static percent_t currentEtbDuty;
...
+          start_time = chVTGetSystemTimeX(); //Updates timestamp as long as ETB-Error stays within 3% deviation
+        }
+      else {
+        }        // ETB Deviation outside limit detected, set timestamp and monitor it for two seconds
+        if(chVTIsSystemTimeWithinX(start_time, start_time + TIME_MS2I(2000))) {
+        	  //Less than two seconds, system OK: do nothing
+            }
+       else {
+         if(rpm > 600){  //We dont want or need any of this to execute if engine are not running
```
rusefi: 600 is a magic number and I am suspisious of universal number of everyone. Maybe you want "2 * crankingRpm" or something like that?

Also https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -269,6 +274,29 @@ static percent_t currentEtbDuty;
...
+                         // It would make a huge issue while tuning ETB
+        	 etbErrorCondition = true;
+        	   warning(OBD_Throttle_Actuator_Control_Range_Performance_Bank_1, "ETB Deviation Error"); //Set Error-code
+        	   engineConfiguration->rpmHardLimit = 2000;  //Set RPM Limit to 2000 to not kill engine if ETB opening are too big
```
rusefi: Logic flow should never write into settings like that "engineConfiguration->rpmHardLimit"

if you need to use a setting and some logic, you need engine->rpmHardLimit variable for CURRENT rpmHardLimit , or, better, "if LIMPING_MODE then ONE RPM LIMIT LOGIC else NORMAL SETTING"

firmware/controllers/actuators/electronic_throttle.cpp
960: The 99% are most important, as the idle-addition on top of target position brings you over 100%

The 1 % have not made any issues for me, but I know TPS calibration varies a lot due to mechanical play. At the ETB I have here now, it varies up to 2%.

So staying away from the hard-stop area might be a good idea.

We can adjust it down if you think it's high.

firmware/controllers/actuators/electronic_throttle.cpp
960: Yes, this one could be replaced by cranking rpm

firmware/controllers/actuators/electronic_throttle.cpp
960: This is the part I haven't found the exact way how to do it, so a little help would be nice.

If the flow comes to this condition, I would like it to stay like this until a reset.

As is now, it passes on every time etberror are within the allowed area, and triggers again.

I tried with a while-loop, but that shuts the whole firmware.

I only want the rpm-limit, error message and duty = 0 to stay.

960: I was thinking having only etbErrorCondition == true; at the end.

Then have another function to define if etbError = true.

Could that be a idea?

rusefi: in order for unrelated files to share this variable you should define it next to hwTriggerInputEnabled on Engine or even better on EngineState next to engineNoiseHipLevel

I would say in this case you have

etbDeviationError = false
on start-up

and if you ever get in deviation error state you say etbDeviationError  = true


in the code which reads rpmHardLimit you would have to read etbDeviationError  as well

firmware/controllers/actuators/electronic_throttle.cpp
mck1117: 1% is pretty small - though we could make it 0.5% or something if we're worried.

firmware/controllers/actuators/electronic_throttle.cpp
mck1117: Don't directly use RPM, check that the current engine state is running

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -87,9 +87,9 @@
 #endif
 
 #define ETB_MAX_COUNT 2
-
+	static systime_t start_time;
```
mck1117: don't put these as global statics - move them in to the EtbControl class.  These are per-throttle conditions, and one instance of EtbControl will exist per throttle.

firmware/controllers/actuators/electronic_throttle.cpp
960: if(rpm > engineConfiguration->cranking.rpm) ?

firmware/controllers/actuators/electronic_throttle.cpp
mck1117: For now maybe don't worry about rev limiting - we need a more cohesive "limp home" strategy for broader categories of errors.  Implementing every possible condition that should trigger limp home is going to cause a spaghetti mess.

firmware/controllers/actuators/electronic_throttle.cpp
mck1117: I should check in my clang-format file, and we can create a gate/ci for it.

firmware/controllers/actuators/electronic_throttle.cpp
mck1117: I think we also have a `clamp()` function or something that clamps a value between two limits.  Clear intent is important, and a function called `clamp` explains that well.

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -269,6 +274,29 @@ static percent_t currentEtbDuty;
...
+      else {
+        }        // ETB Deviation outside limit detected, set timestamp and monitor it for two seconds
+        if(chVTIsSystemTimeWithinX(start_time, start_time + TIME_MS2I(2000))) {
+        	  //Less than two seconds, system OK: do nothing
+            }
+       else {
+         if(rpm > 600){  //We dont want or need any of this to execute if engine are not running
+                         // It would make a huge issue while tuning ETB
+        	 etbErrorCondition = true;
+        	   warning(OBD_Throttle_Actuator_Control_Range_Performance_Bank_1, "ETB Deviation Error"); //Set Error-code
```
mck1117: A better error message might be "Electronic throttle deviation limits exceeded"

firmware/controllers/actuators/electronic_throttle.cpp
960: I dont think we should ever operate within the 1% area, but we can change to 0.5% to be sure if you want.

firmware/controllers/actuators/electronic_throttle.cpp
960: I have "Electronic_Throttle_Position_Control_Error" in my code, but I used the one you already had in this file to not edit more files for this PR

mck1117: I mean the second parameter, currently `"ETB Deviation Error"`

960: Ah, ok. I can change that.

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -269,6 +274,29 @@ static percent_t currentEtbDuty;
 
 		etb1.dcMotor.Set(ETB_PERCENT_TO_DUTY(currentEtbDuty));
 
+	   	float etbError = absF(targetPosition - actualThrottlePosition); // Used for I-Term limiter and Error-detection
+
+
+        // ETB Deviation monitoring
+        if(etbError < 3 ){
+          start_time = chVTGetSystemTimeX(); //Updates timestamp as long as ETB-Error stays within 3% deviation
```
mck1117: please don't directly use ChibiOS's time API - we have our own

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -178,6 +176,7 @@ Pid etbPid;
 
 static percent_t directPwmValue = NAN;
 static percent_t currentEtbDuty;
+static systime_t start_time;
```
mck1117: you don't want this to be static - just write `systime_t start_time;`

firmware/controllers/actuators/electronic_throttle.cpp
960: currentTimeMillis() ?

and efitimems_t instead f systime_t?

960:
Tested, and still works..

firmware/controllers/actuators/electronic_throttle.cpp
mck1117: The high side stop can probably actually come down to 95% or something - some throttles actually even go slightly past fully parallel for exactly that reason.

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -264,11 +269,33 @@ static percent_t currentEtbDuty;
...
+        if (chVTIsSystemTimeWithinX(start_time, start_time + TIME_MS2I(2000))) {
+        	  //Less than two seconds, system OK: do nothing
+            }
+       else {
+    	   if (rpm > engineConfiguration->cranking.rpm) {  //We dont want or need any of this to execute if engine are not running
+                         // It would make a huge issue while tuning ETB
+        	   etbDeviationError = true;
+        	   warning(OBD_Throttle_Actuator_Control_Range_Performance_Bank_1, "Electronic throttle deviation limits exceeded"); //Set Error-code
+        	  // engineConfiguration->rpmHardLimit = 2000;  //Set RPM Limit to 2000 to not kill engine if ETB opening are too big
+        	   etb1.dcMotor.Set(0); // Disable ETB Duty
```
mck1117: this won't actually disable the throttle - it'll continue to operate, since just after this call we'll just reset it with the PID'd duty

rusefi: Perfect catch.

We need automated test coverage. I am sorry but I cannot rely on humans promising me that their code works :(

mck1117: let's just fire the warning for now - like I mentioned on the other comment, we need somewhat more comprehensive limp home, which needs to be designed properly instead of tacking in things like "oh and here we disable the ETB motor"

960: How long does that take?

I am having it shut down in error now, but nothing happens..

mck1117: It sort of doesn't matter how long it takes.  The ETB is safety-critical hardware, so we really shouldn't do anything half-baked or "hackey" when dealing with it.

960: That's the whole point of this function.

This is the proper way to monitor and catch all etb-issues.

I am not a professional programmer, that's why I have asked if you can help refining it

960: And the right reaction for this error are to disable etb, to not fry anything + set the rpm limit to not blow your engine.

960: A "Limp mode" would be a additional thing, but are never used in a aftermarket

mck1117: There are 2.5 sides to the desired behavior here:
1) detect a problem - this code does a good job of that
2)
    a) report the problem to the user
    b) do something safe with the engine/hardware/car/whatever

This PR does a fine job of 1 and 2a.  2b is a larger problem that requires involvement of many systems other than just the electronic throttle controller.  Let's deal with 1 and 2a in this PR, and then figure out the right way to tackle 2b.

mck1117: By the way, disabling ETB, setting a rev limit, etc are absolutely a limp mode.  Call it what you want, but everything in the realm of "impose lower limits to prevent engine damage" is limp mode.

mck1117: Logically separating those concerns helps improve code quality and reliability.  Which for this particular case means:
- Detecting an error when there is one (prevent false negatives)
- NOT detecting an error when there isn't one (prevent false positives)

Both may be equally dangerous if they can alter engine behavior.

960: In 99% of the cases, people go aftermarket to avoid the limp-mode who are a issue when tuning a OEM.
For this specific one, I would not call it limp-mode.
If you have deviation over time at the etb, something are seriously wrong, and it would probably shut down soon because of overtemp.

But no reason to discuss this.
At least the function works as it should, and does a good job. I am sure we find out the last step.

rusefi:
I am not comfortable taking a change to ETB without test automation

@960 I salute your software development skills progress but I am not ready for ETB code to be your sand ground.

Please consider looking unit_tests folder and practicing by expanding test coverage. We have coverage report online to show which code is currently covered and which is not.

firmware/controllers/actuators/electronic_throttle.cpp
960: Just tested with directPwmValue = 0; instead now, and it's stone dead even after error are corrected. 
Only way are reset, just like it should.

I used that for the Tps Calibration, and could not get it back running after calibration.
Tried everything, and the only thing that could get it moving again was directPwmValue = NAN; or reset

PR #1010, opened 2019-11-20
Perf trace cleanup, etc

mck1117:
- Properly disable via ENABLE_PERF_TRACE
- Move enable defines to `chconf_common.h`
- Trace global lock/unlock events
- Remove `EFI_CLOCK_LOCKS`: we now profile with the new system


rusefi:

rusefi:
Add to whitelist

firmware/config/chconf_common.h
```diff
@@ -12,6 +12,12 @@
 #include "obd_error_codes.h"
 #endif /* _FROM_ASM_ */
 
+
+#ifndef ENABLE_PERF_TRACE
+// looks like this value could not be defined in efifeatures.h - please define either externally or just change the value here
+ #define ENABLE_PERF_TRACE FALSE
```
mck1117: @rusefi Do we have a preference about the default here? Normal users probably don't have a need to enable it, and there is a small (but extant) performance impact to leaving it on.

rusefi: My preference is TRUE
I would hope that with our gigantic CPU we shall afford some monitoring features

firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -509,12 +509,8 @@ extern PwmConfig triggerSignal;
 extern uint32_t hipLastExecutionCount;
 extern uint32_t hwSetTimerDuration;
 
-extern uint32_t maxLockedDuration;
```
rusefi: why do we remove maxLockedDuration? in my opinion this is an important and unrelated feature

rusefi:
I am not comfortable with the size and multiple changes bundled together into one PR

this is not a perf trace cleanup since I see 1) clean-up 2) new feature 3) removal of old unrelated feature

Please close this PR and consider submitting this as a few smaller changes

rusefi:
@mck1117 can you please help this PR?

i was hesitant to merge. Maybe we can take care of https://github.com/rusefi/rusefi_documentation/issues/28 and return to this after documenting our solution?

PR #1020, opened 2019-11-21
Build prometheus with makefiles instead of batch scripts

mck1117:
Also add Prometheus to CI/gate builds


mck1117:
@andreika-git can you test that the batch scripts still work, since I don't have a windows build environment?

rusefi:
Waiting for an approval from @andreika-git

andreika-git:
> I have doubts that the "default" release build options are not optimized enough. Can you please transfer all Prometheus-specific defines (EFI_ENABLE_ASSERTS=FALSE, CH_DBG_***=FALSE) from the bat file to .mk?..

PR #1026, opened 2019-11-23
inject interfaces to ETB, renaming

mck1117:
Working towards testability of electronic throttle.

### DIFF
--- a/firmware/controllers/actuators/electronic_throttle.cpp
+++ b/firmware/controllers/actuators/electronic_throttle.cpp
@@ -105,7 +105,7 @@ static bool startupPositionError = false;
 
 #define STARTUP_NEUTRAL_POSITION_ERROR_THRESHOLD 5
 
-class EtbControl {
+class EtbHardware {
 private:
 	OutputPin m_pinEnable;
 	OutputPin m_pinDir1;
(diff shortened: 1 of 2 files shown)


firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -118,9 +118,7 @@ class EtbControl {
 	SimplePwm etbPwmUp;
 
 public:
-	DECLARE_ENGINE_PTR;
```
mck1117: bonus: we get rid of this :smile:

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -187,152 +187,157 @@ static percent_t currentEtbDuty;
...
 
 
-	int EtbController::getPeriodMs() {
-		return GET_PERIOD_LIMITED(&engineConfiguration->etb);
-	}
+int EtbController::getPeriodMs() {
+	return GET_PERIOD_LIMITED(&engineConfiguration->etb);
+}
 
-	void EtbController::PeriodicTask() {
```
mck1117: this file was also mis-tabbed

PR #1027, opened 2019-11-23
Use templated array references

mck1117:
Don't need to pass length - it's an opportunity for a bug.  Use a `TSize` template parameter to bind the length of the passed array.


rusefi:
CI does it's job but we better be still careful since many things would pass by CI

PR #1032, opened 2019-11-25
Starter relay + Hardcoded 'FSIO' + simulator fix

andreika-git:
Hardcoded FSIO was tested on Kinetis.

### DIFF
--- a/firmware/controllers/core/fsio_impl.cpp
+++ b/firmware/controllers/core/fsio_impl.cpp
@@ -102,6 +102,7 @@ static LEElement * acRelayLogic;
 static LEElement * fuelPumpLogic;
 static LEElement * radiatorFanLogic;
 static LEElement * alternatorLogic;
+static LEElement * starterRelayLogic;
 
 #if EFI_MAIN_RELAY_CONTROL
 static LEElement * mainRelayLogic;
@@ -464,6 +465,9 @@ void runFsio(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 		enginePins.mainRelay.setValue(true);
 #endif /* EFI_MAIN_RELAY_CONTROL */
 
+	if (CONFIGB(starterRelayPin) != GPIO_UNASSIGNED)
+		setPinState("starter_relay", &enginePins.starterRelay, starterRelayLogic PASS_ENGINE_PARAMETER_SUFFIX);
+
 	/**
 	 * o2 heater is off during cranking
 	 * todo: convert to FSIO?
@@ -673,6 +677,8 @@ void initFsioImpl(Logging *sharedLogger DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	if (CONFIGB(mainRelayPin) != GPIO_UNASSIGNED)
 		mainRelayLogic = sysPool.parseExpression(MAIN_RELAY_LOGIC);
 #endif /* EFI_MAIN_RELAY_CONTROL */
+	if (CONFIGB(starterRelayPin) != GPIO_UNASSIGNED)
+		starterRelayLogic = sysPool.parseExpression(STARTER_RELAY_LOGIC);
 
 #if EFI_PROD_CODE
 	for (int i = 0; i < FSIO_COMMAND_COUNT; i++) {
--- a/firmware/controllers/core/fsio_impl.cpp
+++ b/firmware/controllers/core/fsio_impl.cpp
@@ -12,13 +12,13 @@
  */
 
 #include "global.h"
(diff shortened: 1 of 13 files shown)


andreika-git:
Please run gen_config!

firmware/controllers/core/fsio_impl.cpp
```diff
@@ -724,5 +730,37 @@ void initFsioImpl(Logging *sharedLogger DECLARE_ENGINE_PARAMETER_SUFFIX) {
...
+
+EXTERN_ENGINE
+;
+extern EnginePins enginePins;
+
+// "Limp-mode" implementation for some RAM-limited configs without FSIO
+void runHardcodedFsio(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
+	// see MAIN_RELAY_LOGIC
+	if (CONFIGB(mainRelayPin) != GPIO_UNASSIGNED) {
+		enginePins.mainRelay.setValue((getTimeNowSeconds() < 2) || (getVBatt(PASS_ENGINE_PARAMETER_SIGNATURE) > 5) || engine->isInShutdownMode());
```
mck1117: why vbatt over 5? I've seen vbatt during cranking below 4.

rusefi:
image

firmware/controllers/core/fsio_impl.cpp
andreika-git: Not all MCUs can withstand less than 5V, we cannot rely on that: now we have not only 3.3V-powered MCUs, but also 5V MCUs. I hope we'll make this setting configurable one day...

mck1117: I'm not sure why that means we should turn off the main relay.  

1. There are plenty of power supply schemes where an MCU can survive a low battery voltage (either a transient or sustained).   Some ECUs have a boost converter than can run the ECU clear down to 3 volts continuously.  I've seen the 3.3v rail dip on my microRusefi during cranking, but not low enough to reset the MCU, simply because it was below spec, but not low enough to actually trigger misbehavior.

2. If the voltage gets low enough to reset the MCU, there isn't any amount of software that can alter that behavior.  Either the hardware design causes it to reset or not.

I'm mostly wondering what the intent is here, not why the particular value was selected.

andreika-git:
@rusefi Please explain what's wrong here. We do have "config/boards/kinetis/mapping.yaml", and when I call !gen_config.bat for Kinetis locally, it works fine!

firmware/controllers/core/fsio_impl.cpp
andreika-git: > I'm not sure why that means we should turn off the main relay.

Currently we don't have a lower-power standby mode in firmware. And we don't want to use an ignition switch voltage as a power source for the ECU, because we need some time for proper firmware shutdown and don't want to just instantly kill the firmware with the turn of the key. We may need time to park the stepper motor, or to keep the ECF running some extra seconds to cool down the hot engine, or just wait for engine to stop and save to flash the settings, etc.. 

So the idea behind EFI_MAIN_RELAY_CONTROL is to use the main relay output voltage as the power source for ECU instead ignition switch voltage. And then do a self switch-off when the ignition key voltage drops when a user turns the ignition key, but only after the shutdown is complete - that's why we have isInShutdownMode(). So please note that in EFI_MAIN_RELAY_CONTROL we track vBatt as the ignition key voltage, not the ECU power voltage!

mck1117: is there currently hardware that supports such a configuration?

andreika-git: First of all, it's Prometheus. It's the main power scheme for it, derived from SECU-3 ECU. 
And also Deucalion configuration uses it on my car. You can search for "EFI_MAIN_RELAY_CONTROL TRUE" in our board configs.

mck1117: Have a link to the Prometheus schematic? I'd love to wire https://github.com/mck1117/proteus like that as I'm about to run a first spin of the boards.

andreika-git: Schematic: https://github.com/andreika-git/prometheus/raw/master/hardware/prometheus-schematic.pdf
Wiring diagram: https://github.com/andreika-git/prometheus/raw/master/hardware/prometheus-wiring-diagram.pdf

andreika-git: Please don't forget that you'll need a diode from the ignition switch to the main relay for initial power-up.

PR #1034, opened 2019-11-30
Harmonize chconf

mck1117:
Start dealing with #1033 

- Turn off OS features we weren't using (and were turned off anyway on f7!)
- Match remaining different settings on f4/f7 in chconf and mcuconf


rusefi:
thank you!

PR #1042, opened 2019-12-04
Move slow ADC to thread #630

### LINKED ISSUE #630: migrate slow ADC to software timer so that TIM8 is also available for input capture

mck1117:
Move the slow ADC to its own thread.  We were previously using a hardware timer, which had a few problems.
1. It tied up the hardware timer
2. It didn't participate in the OS's priority system - so it could stomp other work being done.
3. Wasn't very portable.  Required a dedicated hardware timer, but now only relies on core OS functionality to run the ADC.

This closes #630.


rusefi:
this is related to #630

rusefi:
Oh! That's a great change!

How easy is it for you to test on a real vehicle?

mck1117:
Might be a little difficult to test on a real vehicle for a few weeks...

image

mck1117:
This code has now been used on a running, driving car!

PR #1043, opened 2019-12-04
iacPidMultTbl + EFI_IDLE_PID_CIC + kinetis_hysteresis

andreika-git:
1) Implement iacPidMultTable
2) Replace INCREMENTAL_PID_CIC with ordinary PID_CIC (we don't need it anymore since our Autopid can calculate the correct PID offset value)
3) Kinetis: Trigger Comparator Hysteresis settings (+adaptive mode draft - currently disabled)


andreika-git:
image

PR #1048, opened 2019-12-10
copyArray helper

mck1117:
Saves code and data space by storing constant arrays as arrays, then simply copying them in to place instead of hand setting every element.  The new functions are also useful as they guarantee that you can't overrun the destination array.

### DIFF
--- a/firmware/controllers/algo/engine_configuration.cpp
+++ b/firmware/controllers/algo/engine_configuration.cpp
@@ -431,20 +431,41 @@ static void setDefaultWarmupIdleCorrection(DECLARE_CONFIG_PARAMETER_SIGNATURE) {
 }
 
 static void setDefaultWarmupFuelEnrichment(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
-	initTemperatureCurve(WARMUP_CLT_EXTRA_FUEL_CURVE, 1.0);
+	static const float bins[] =
+	{
+		-40,
+		-30,
+		-20,
+		-10,
+		0,
+		10,
+		20,
+		30,
+		40,
+		50,
+		60,
+		70
+	};
 
-	setCurveValue(WARMUP_CLT_EXTRA_FUEL_CURVE, -40, 1.50);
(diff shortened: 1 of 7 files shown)


firmware/controllers/algo/engine_configuration.cpp
```diff
@@ -431,20 +431,41 @@ static void setDefaultWarmupIdleCorrection(DECLARE_CONFIG_PARAMETER_SIGNATURE) {
 }
 
 static void setDefaultWarmupFuelEnrichment(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
-	initTemperatureCurve(WARMUP_CLT_EXTRA_FUEL_CURVE, 1.0);
```
andreika-git: Are you sure you want to remove this line? You set only 12 values with copyArrayPartial() but CLT_CURVE_SIZE is 16...

firmware/util/containers/cyclic_buffer.h
```diff
@@ -81,9 +81,7 @@ cyclic_buffer<T, maxSize>::cyclic_buffer(const cyclic_buffer& cb) {
 	currentIndex = cb.currentIndex;
 	count = cb.count;
 	size = cb.size;
-	for (int i = 0; i < size; ++i) {
-		elements[i] = cb.elements[i];
-	}
+	copyArray(elements, cb.elements);
```
andreika-git: We may get a small performance penalty here because 'size' could be just 1 and 'maxSize' could be much bigger, so we don't really need to copy the entire array.

mck1117: I'm not sure why we even define the copy constructor.  The default copy constructor simply copies all the fields, which is what we're doing by hand here.

mck1117: (or the copy-assign constructor or destructor, for that matter)

mck1117:
Fun fact: on an default f4 build (with optimization on), this saves 208 bytes of flash.

PR #1049, opened 2019-12-10
ETB Encapsulation & C++ Conventions

mck1117:
No longer does any logic prod the innards of the ETB controller. Hooray!

~~No longer requires `INJECT_ENGINE_REFERENCE` for `EtbController`!~~ Nope, that was a lie.

Makes naming and conventions more in line with C++ conventions.  That means prefixing member variables with `m_`, and not using `this->` when accessing.

This also fixes the bug that multiple ETBs will overwrite each other's state in the tunerstudio outputs, making the gauges useless.

### DIFF
--- a/firmware/controllers/actuators/electronic_throttle.cpp
+++ b/firmware/controllers/actuators/electronic_throttle.cpp
@@ -180,9 +180,24 @@ static percent_t currentEtbDuty;
 // this macro clamps both positive and negative percentages from about -100% to 100%
 #define ETB_PERCENT_TO_DUTY(X) (maxF(minF((X * 0.01), ETB_DUTY_LIMIT - 0.01), 0.01 - ETB_DUTY_LIMIT))
 
-void EtbController::init(DcMotor *motor, int ownIndex) {
-	this->m_motor = motor;
-	this->ownIndex = ownIndex;
+void EtbController::init(DcMotor *motor, int ownIndex, pid_s *pidParameters) {
+	m_motor = motor;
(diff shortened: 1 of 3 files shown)


firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -555,9 +577,8 @@ void stopETBPins(void) {
 #endif /* EFI_PROD_CODE */
 
 void onConfigurationChangeElectronicThrottleCallback(engine_configuration_s *previousConfiguration) {
-	bool shouldResetPid = !etbController[0].etbPid.isSame(&previousConfiguration->etb);
-	for (int i = 0 ; i < ETB_COUNT; i++) {
-		etbController[i].shouldResetPid = shouldResetPid;
```
mck1117: Believe it or not, this used to be a race condition.  If the configuration was changed twice quickly back to back, and the ETB didn't need to update the second time, it would clear the flag even though it never reset the PID controller!

firmware/controllers/actuators/electronic_throttle.h
```diff
@@ -15,20 +15,32 @@
...
+	// Print this throttle's status.
+	void showStatus(Logging* logger);
+
+	// Used to inspect the internal PID controller's state
+	const pid_state_s* getPidState() const { return &m_pid; };
 
 private:
-	int ownIndex;
-    DcMotor *m_motor;
+	int m_myIndex;
```
mck1117: hooray, all private!

PR #1050, opened 2019-12-11
DRAFT simplify dequeue logic

mck1117:
For some reason, dequeuing events from our scheduler queue was a two phase process:
1) move all events that are due to run to their own temporary list
2) execute everything in that list.

I can't come up with a good reason for why we were doing that.

This change simply pops elements off the head one at a time, and executes them, instead of the aforementioned complicated (and slow) logic.

Unit tests pass, but pending testing on real hardware.

### DIFF
--- a/firmware/controllers/system/timer/event_queue.cpp
+++ b/firmware/controllers/system/timer/event_queue.cpp
@@ -165,10 +165,6 @@ int EventQueue::executeAll(efitime_t now) {
 #endif
 	}
 
-#if EFI_UNIT_TEST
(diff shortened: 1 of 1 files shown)


rusefi:
Sweet stuff :+1:

rusefi:
this PR looks pretty convincing, let me take a bet with it and try on my engine

rusefi:
tested on real engine - idles fine, good enough for me.

mck1117:
I just went for a short drive on it too - seems to be fine

PR #1055, opened 2019-12-12
Make stacks smaller

mck1117:
Excerpt from `ChibiOS/os/common/ports/ARMCMx/chcore_v6m.h`:
```
/**
 * @brief   Per-thread stack overhead for interrupts servicing.
 * @details This constant is used in the calculation of the correct working
 *          area size.
 * @note    In this port this value is conservatively set to 64 because the
 *          function @p chSchDoReschedule() can have a stack frame, especially
 *          with compiler optimizations disabled. The value can be reduced
 *          when compiler optimizations are enabled.
 */
#if !defined(PORT_INT_REQUIRED_STACK)
#define PORT_INT_REQUIRED_STACK         64
#endif
```

Interrupts are NOT serviced on the stack of the thread that was running when the interrupt occurred.  The only thing that happens on that thread's stack is that its registers are pushed (by hardware) when an interrupt occurs, just before swapping the stack pointer out for the main stack (currently 0x400=1024 bytes), where the ISR actually runs.

Likewise, the idle thread actually doesn't do anything, so it shouldn't be so tremendous.

This change saves a lot of RAM.  640 bytes per thread + 960 for the idle thread.

### DIFF
--- a/firmware/config/stm32f7ems/chconf.h
+++ b/firmware/config/stm32f7ems/chconf.h
@@ -41,12 +41,13 @@
  *
  */
 
-#define PORT_IDLE_THREAD_STACK_SIZE     1024
+#define PORT_IDLE_THREAD_STACK_SIZE     32
 
-// rusEfi main processing happens on IRQ so PORT_INT_REQUIRED_STACK has to be pretty large.
-// see also a strange comment about PORT_INT_REQUIRED_STACK in global_shared.h
+// PORT_INT_REQUIRED_STACK is the stack space added to every thread's stack so that even if the thread
+// has used all of its stack space, an interrupt can still occur, and the hardware can push the
+// execution context before switching to the main stack to execute the interrupt.
 // see also http://www.chibios.org/dokuwiki/doku.php?id=chibios:kb:stacks
-#define PORT_INT_REQUIRED_STACK 	768
+#define PORT_INT_REQUIRED_STACK 	128
 
 #define CHPRINTF_USE_FLOAT          	TRUE
 
--- a/firmware/config/stm32f7ems/chconf.h
+++ b/firmware/config/stm32f7ems/chconf.h
@@ -43,9 +43,7 @@
 
 #define PORT_IDLE_THREAD_STACK_SIZE     32
 
-// PORT_INT_REQUIRED_STACK is the stack space added to every thread's stack so that even if the thread
-// has used all of its stack space, an interrupt can still occur, and the hardware can push the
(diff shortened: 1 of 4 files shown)


firmware/controllers/global_shared.h
```diff
@@ -14,11 +14,6 @@
  * The following obscurantism is a hack to reduce stack usage, maybe even a questionable performance
  * optimization.
  *
- * rusEfi main processing happens on IRQ so PORT_INT_REQUIRED_STACK has to be pretty large. Problem
```
rusefi: Please replace this comment not just remove this comment

```Interrupts are NOT serviced on the stack of the thread that was running when the interrupt occurred. The only thing that happens on that thread's stack is that its registers are pushed (by hardware) when an interrupt occurs, just before swapping the stack pointer out for the main stack (currently 0x400=1024 bytes), where the ISR actually runs.``` is good content to maybe place here?

**Comment by rusefi** [comment]
Did you test with -O0 -g3 or with high levels of optimization and without debug?

File: firmware/config/stm32f7ems/chconf.h [hunk 2]
```diff
@@ -41,12 +41,13 @@
  *
  */
 
-#define PORT_IDLE_THREAD_STACK_SIZE     1024
+#define PORT_IDLE_THREAD_STACK_SIZE     32
 
-// rusEfi main processing happens on IRQ so PORT_INT_REQUIRED_STACK has to be pretty large.
-// see also a strange comment about PORT_INT_REQUIRED_STACK in global_shared.h
+// PORT_INT_REQUIRED_STACK is the stack space added to every thread's stack so that even if the thread
```
rusefi: Actually we clearly have a copy-paste of a large comment and you are now editing the copy-paste in three places? is not this a bit weird? I wonder if global_shared.h is the better SINGLE place for this comment, with other files potentially referring people to that single centralized comment?

mck1117:
This has been tested at `-O0` and `-O2` on both F4 and Kinetis.  Besides, they both use the same core, which dictates how interrupt stuff works (the F7 is technically an m7f which is a different core, but 100% compatible with the m4f in the f4/kinetis).

firmware/config/stm32f7ems/chconf.h
mck1117: yes - probably true, I'll do that

rusefi:
Exciting change, really appreciate it! Let's just improve the documentation a bit - my rule of thumb is anything you want to say on PR even before people are asking for clarification belongs in the codebase for long-term maintainability, not on the PR.

PR #1064, opened 2019-12-17
Remove sensor reporting loc

mck1117:
I had mixed feelings about doing it this way from the beginning - it's a bit weird to have magical side effects, especially when those side effects mutate state from a huge distance away.

It's easier to reason about what's going on when it's explicitly done (set state in the TS output channels struct).

### DIFF
--- a/firmware/console/status_loop.cpp
+++ b/firmware/console/status_loop.cpp
@@ -756,7 +756,7 @@ void updateTunerStudioState(TunerStudioOutputChannels *tsOutputChannels DECLARE_
 	tsOutputChannels->rpmAcceleration = engine->rpmCalculator.getRpmAcceleration();
 	// offset 108
 	// For air-interpolated tCharge mode, we calculate a decent massAirFlow approximation, so we can show it to users even without MAF sensor!
-    tsOutputChannels->massAirFlow = getAirFlowGauge(PASS_ENGINE_PARAMETER_SIGNATURE);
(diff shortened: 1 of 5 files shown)


firmware/console/status_loop.cpp
```diff
@@ -793,12 +793,14 @@ void updateTunerStudioState(TunerStudioOutputChannels *tsOutputChannels DECLARE_
...
-    // 276
-    tsOutputChannels->accelerationX = engine->sensors.accelerometer.x;
-    // 278
-    tsOutputChannels->accelerationY = engine->sensors.accelerometer.y;
-    // 288
-    tsOutputChannels->injectionOffset = engine->engineState.injectionOffset;
+	// 276
+	tsOutputChannels->accelerationX = engine->sensors.accelerometer.x;
+	// 278
+	tsOutputChannels->accelerationY = engine->sensors.accelerometer.y;
```
mck1117: this is just fixing tabs vs. spaces

firmware/console/status_loop.cpp
```diff
@@ -793,12 +793,14 @@ void updateTunerStudioState(TunerStudioOutputChannels *tsOutputChannels DECLARE_
...
+	// 280
+	tsOutputChannels->oilPressure = Sensor::get(SensorType::OilPressure).Value;
```
mck1117: 800/801 are the only new lines

PR #1067, opened 2019-12-18
don't log EGT if disabled

mck1117:
Almost nobody has EGT enabled, so we probably shouldn't log it unless the feature is actually enabled.  TS is exerting a lot of effort writing all those zeroes!

note: this is part 1 in a scheme to allow an f4 to log in excess of ~~170hz~~ 205hz


rusefi:
why is 205hz so important? i prefer to have the egt and hope that someone will use it

mck1117:
This doesn't do anything to get rid of egt. It simply omits EGT channels from the catalog if the corresponding channels aren't enabled. If you enable them, they show up again.

mck1117:
The gauges still work properly if you add one to the dash, or course.

PR #1068, opened 2019-12-18
remove accelerometer datalog

mck1117:
no boards have `EFI_MEMS` enabled - so don't log that field

note: this is part 1.5 in a scheme to allow an f4 to log in excess of 205hz


rusefi:
lets not remove accel, lets focus on finishing this feature

PR #1069, opened 2019-12-18
Shrink output channels, use scaling

mck1117:
- Implement `scaled_channel` which allows a scaled output channel (for example, storing quarter-RPM) to be used as though it was a normal float, even though it's backed by a different type.
- Reorder TS output channels in roughly reverse priority order.  Since rarely used channels are at the end, datalogging can be sped up by reading a contiguous block of channels near the beginning that we actually care about (for example, logging up thru `lastErrorCode`, 140 bytes, updates at 210-215hz while datalogging on a debug build on an stm32f4 over USB.  This previously ran at ~90hz).
- This change reduces total output channel size from 356 to 220 bytes

The modifications to `tle8888.h`/`c` were required because that `tunerstudio_configuration.h` now requires c++, so we inject only the debug channels where we need them in c-land.

This change has been tested by both @960 and myself.  I adjusted tune settings where possible to test channels, and where not possible to test on the bench, I modified the function in `status_loop.cpp` to manually set values, then confirm that they showed as expected in TS.

### DIFF
--- a/firmware/config/boards/kinetis/config/controllers/algo/rusefi_generated.h
+++ b/firmware/config/boards/kinetis/config/controllers/algo/rusefi_generated.h
@@ -1888,7 +1888,7 @@
 #define triggerSimulatorPins3_offset 738
 #define triggerSimulatorPins3_offset_hex 2e2
 #define TS_FILE_VERSION 20190701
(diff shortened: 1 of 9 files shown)


mck1117:
Should I rev the TS protocol version/signature with this change?

PR #1072, opened 2019-12-19
Support AEM X-Series Wideband (CAN)

mck1117:


firmware/hw_layer/can_hw.cpp
```diff
@@ -241,22 +241,32 @@ static void canInfoNBCBroadcast(can_nbc_e typeOfNBC) {
...
 	CANDriver *device = detectCanDevice(CONFIG(canRxPin),
 			CONFIG(canTxPin));
 	if (device == NULL) {
 		warning(CUSTOM_ERR_CAN_CONFIGURATION, "CAN configuration issue");
 		return;
 	}
-//	scheduleMsg(&logger, "Waiting for CAN");
-	msg_t result = canReceive(device, CAN_ANY_MAILBOX, &rxBuffer, TIME_MS2I(1000));
+
+	msg_t result = canReceiveTimeout(device, CAN_ANY_MAILBOX, &rxBuffer, TIME_IMMEDIATE);
```
mck1117: `canReceive` was deprecated - replaced with `canReceiveTimeout`

firmware/hw_layer/can_hw.cpp
```diff
@@ -271,12 +281,7 @@ static msg_t canThread(void *arg) {
 			writeStateToCan();
 
 		if (engineConfiguration->canReadEnabled)
-			canRead(); // todo: since this is a blocking operation, do we need a separate thread for 'write'?
-
-		if (engineConfiguration->canSleepPeriodMs < 10) {
```
mck1117: this check didn't make sense - it might only make sense for writes.  future work is to split read/write out on to their own threads so that they aren't coupled

rusefi: self-comment is a sign of code comment opportunity missed in my opinion

mck1117: not in cases of when something was removed?

mck1117: doing this future work first - closing for now

PR #1074, opened 2019-12-20
DRAFT Shrink output channels, again

mck1117:
Second try at #1069 , now with more Java!

~~I'm currently unable to build the console, but I _think_ this is at least close.~~ edit: console now builds, seems to work okay with the new scaling!

### DIFF
--- a/java_console/models/src/com/rusefi/core/Sensor.java
+++ b/java_console/models/src/com/rusefi/core/Sensor.java
@@ -17,93 +17,101 @@ import static com.rusefi.config.generated.Fields.*;
  * 2/11/13
  */
 public enum Sensor {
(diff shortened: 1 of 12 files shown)


java_console/models/src/com/rusefi/core/Sensor.java
```diff
@@ -115,32 +121,22 @@
...
         this.minValue = minValue;
         this.maxValue = maxValue;
     }
 
-    Sensor(SensorCategory category, FieldType type, int offset, BackgroundColor color, double minValue, double maxValue, String units) {
-        this(null, category, type, offset, color, minValue, maxValue, units);
-    }
-
     Sensor(String name, SensorCategory category, FieldType type, int offset, BackgroundColor color, double minValue, double maxValue) {
-        this(name, category, type, offset, color, minValue, maxValue, "n/a");
```
mck1117: some of these constructors were no longer necessary, as I added name/min/max/units to the channels that were missing it.

PR #1075, opened 2019-12-21
remove concept of TS pages

mck1117:
Since we only ever use a single jumbo page, we don't need to keep track of what page we're on. Anything other than page 0 is an error.

As a bonus, saves ~100 bytes of code size and a tiny bit of stack.


firmware/console/binary/tunerstudio.cpp
```diff
@@ -322,38 +321,38 @@ void handleWriteChunkCommand(ts_channel_s *tsChannel, ts_response_format_e mode,
...
 
-	if (offset > getTunerStudioPageSize(currentPageId)) {
+	if (offset >= getTunerStudioPageSize()) {
 		scheduleMsg(&tsLogger, "ERROR invalid offset %d", offset);
 		tunerStudioError("ERROR: out of range");
 		offset = 0;
 	}
 
-	if (count > getTunerStudioPageSize(currentPageId)) {
+	if (count >= getTunerStudioPageSize()) {
```
rusefi: i wonder about the change here, could it be a bug? in case of offset 0?

good thing we do not have such big packets. also maybe we shall compare offset+count, not just count?

rusefi:
@mck1117 poke

firmware/console/binary/tunerstudio.cpp
mck1117: We now do a full bounds check for all read/write operations!

mck1117:
Looks like the failing simulator build is a known issue 😢:
https://github.community/t5/GitHub-Actions/Errors-on-updating-apt-cache-for-azure-cli/td-p/42743

It builds correctly for me locally.

Issue #907, opened 2019-08-26
 E30 M3 Tach output solution


rusefi:
At the moment we have a total mess of dizzy output and tach output, and potentially another unique request for E30 M3:

```Orchardperformance 11:48 AM
General question for anyone that has used the Tach output...
Just trying to solve the niggles with my E30 installation so I can run it more as a daily and I need to get the Tech working, Its a 12v high low signal so it's going on U604.
The signal needs to be a 50% duty cycle pulsed 3 times a rev like the dizzy output would give. (2 pulses for a 4 pot and 3 for a 6 pot.)
Is there a way to set the dizzy output to a duty cycle? Or is there a way to set the tach output to duty cycle and multiplied x3?
I suspect I can set this up via FSIO but with all the pre-canbus BMW stuff using this setup then it might be worth baking this into the standard ECU options.
Orchardperformance 12:16 PM
Actually having the Tach output as a multiplier and duty cycle setup may make a lot of sense for many conversions. A friend just pointed out it's a nightmare for the E30 M3 as there is only a 4 cylinder dashboard unique to the car so they often need the ECU running 6 or 8 coils but with a 4 cylinder output.
I will try to take a look at the feasibility later on```

I believe this should all be simplified by leveraging FSIO.
At the moment FSIO allows duty cycle control but not programmatic frequency control. It could be that FSIO programmatic frequency control is the way to address this requirement.

shadowm60:
I have tried an implementation as follows:
Added a new input filed for the tacho that will need the number of pulses per crank rotation.
with this info and current RPM we can calculate the pulse with that needs to output.

What i did not took into consideration is the field pulse duration, if we do have more then one pulse it will actually generate a PWM 50% duty.

Should i keep the pulse duration as well? what if the configured pulse duration is not possible since the number is to high?

rusefi:
@shadowm60 I am sorry I am not sure what is the question exactly. 

I believe there could be at least two typos - while I am probably able to guess, this complicates comprehension. If you are referencing any existing configuration fields please spell out the exact names of the existing name(s) of the existing configuration fields.

To be honest I do not exactly remember what we have in the code base right now. First step could be to document the behaviors which are possible at the moment and to analyze if we shall start by simplifying/unifying existing behavior. Maybe the first PR would be just adding human-readable overview of the current status of this area into tachometer.cpp?

Maybe chatting in Slack could help maybe not. I am not sure what is the best approach here :(

shadowm60:
[...]

PR #1080, opened 2019-12-29
Update as suggested in E30 M3 Tach output solution #907

shadowm60:
https://github.com/rusefi/rusefi/issues/907

Initial implementation, my first time to genereate rusefi.ini file as well.
Tested on STM32F407 discovery and engine stimulator. More testing in the future.

### DIFF
--- a/firmware/tunerstudio/rusefi.ini
+++ b/firmware/tunerstudio/rusefi.ini
@@ -85,7 +85,7 @@ enable2ndByteCanID = false
 
 ; see PAGE_0_SIZE in C source code    
 ; CONFIG_DEFINITION_START
-; this section was generated automatically by rusEfi tool ConfigDefinition.jar based on integration\rusefi_config.txt Tue Feb 04 01:02:34 EST 2020
+; this section was generated automatically by rusEfi tool ConfigDefinition.jar based on integration\rusefi_config.txt Tue Feb 04 13:45:28 EST 2020
 
 pageSize            = 20000
 page = 1
@@ -1341,8 +1341,8 @@ fileVersion = { 20191221 }
       
 
 ; throttle, pedal
-      TPSValue        = scalar,  U16,     20,        "%",{1/100},         0
-      throttlePedalPosition = scalar,U16, 22,        "%",{1/100},         0
+      TPSValue        = scalar,  S16,     20,        "%",{1/100},         0
+      throttlePedalPosition = scalar,S16, 22,        "%",{1/100},         0
       tpsADC          = scalar,  U16,     24,      "ADC",    1,       0.0;
 
 ; air flow/mass measurments
@@ -1375,9 +1375,9 @@ fileVersion = { 20191221 }
 
 ; Corrections
       injectorLagMs   = scalar,  U16,     62,     "ms",{1/300},       0.0
-      iatCorrection   = scalar,  U16,     64,        "%",{1/100},         0
-      cltCorrection   = scalar,  U16,     66,        "%",{1/100},         0
-      baroCorrection  = scalar,  U16,     68,        "%",{1/100},         0
+      iatCorrection   = scalar,  S16,     64,        "%",{1/100},         0
+      cltCorrection   = scalar,  S16,     66,        "%",{1/100},         0
+      baroCorrection  = scalar,  S16,     68,        "%",{1/100},         0
       fuelPidCorrection=scalar,  S16,     70,     "ms",{1/300},         0
 
 ; Wall model AE
@@ -1393,7 +1393,7 @@ fileVersion = { 20191221 }
 ; Ignition
       ignitionAdvance = scalar,  S16,     84,    "deg",{1/50},       0.0
       sparkDwellValue = scalar,  U16,     86,     "ms",{1/300},       0.0
-      coilDutyCycle   = scalar,  U16,     88,      "%",{1/100},         0
+      coilDutyCycle   = scalar,  S16,     88,      "%",{1/100},         0
 
 ; Idle & ETB
       idleAirValvePosition=scalar,S16,    90,      "%",{1/100},         0
--- a/firmware/tunerstudio/rusefi_frankenso.ini
+++ b/firmware/tunerstudio/rusefi_frankenso.ini
@@ -85,7 +85,7 @@ enable2ndByteCanID = false
 
 ; see PAGE_0_SIZE in C source code    
 ; CONFIG_DEFINITION_START
-; this section was generated automatically by rusEfi tool ConfigDefinition.jar based on integration\rusefi_config.txt Tue Feb 04 01:02:41 EST 2020
+; this section was generated automatically by rusEfi tool ConfigDefinition.jar based on integration\rusefi_config.txt Tue Feb 04 13:45:35 EST 2020
 
 pageSize            = 20000
 page = 1
@@ -1341,8 +1341,8 @@ fileVersion = { 20191221 }
       
 
 ; throttle, pedal
-      TPSValue        = scalar,  U16,     20,        "%",{1/100},         0
-      throttlePedalPosition = scalar,U16, 22,        "%",{1/100},         0
+      TPSValue        = scalar,  S16,     20,        "%",{1/100},         0
+      throttlePedalPosition = scalar,S16, 22,        "%",{1/100},         0
       tpsADC          = scalar,  U16,     24,      "ADC",    1,       0.0;
 
 ; air flow/mass measurments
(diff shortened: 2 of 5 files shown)


rusefi:

rusefi:

rusefi:

rusefi:

rusefi:
add to white list

rusefi:
@shadowm60 please rebase to master - I've manually picked your config change so that PR is focused on the actual change

as is with generated and non-generated code mixed it's hard to review the PR

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -18,26 +18,71 @@
...
+
+static void toggleTachPinHigh(void);
+static void toggleTachPinLow(void);
 
 static void turnTachPinLow(void) {
 	enginePins.tachOut.setLow();
 }
 
+static void toggleTachPinLow(void)
+{
```
rusefi: code style - note how the rest of the firmware define methods

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -18,26 +18,71 @@
...
+	if(isStarted)
+	{
+		enginePins.tachOut.setLow();
+		engine->executor.scheduleForLater(&tachTurnSignalOff, (int)MS2US(halfperiodMs), (schfunc_t) &toggleTachPinHigh, NULL);
+	}
+}
+
+static void toggleTachPinHigh(void)
+{
+	if(isStarted)
```
rusefi: code style - space is missing but new line is not consistent with project code style

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -18,26 +18,71 @@
...
-		durationMs = engineConfiguration->tachPulseDuractionMs;
-	} else {
-		durationMs = engineConfiguration->tachPulseDuractionMs;
+
+	if ((1 < engineConfiguration->tachPulsePerRev))
+	{
+
+		/* reset timing counters */
+		int rpm = GET_RPM();
+		float duration = 60000/rpm;
```
rusefi: int divided by int is an int. That's scary that we have this variable defined as float :(

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -18,26 +18,71 @@
...
+		halfperiodMs = (duration/(engineConfiguration->tachPulsePerRev*2));
+        if(isStarted==0)
+        {
+        	enginePins.tachOut.setHigh();
+        	isStarted = 1;
```
rusefi: Looks like this isStarted is a boolean. Please clarify why do we use integers here.

firmware/controllers/gauges/tachometer.cpp
mck1117: If you want to actually end up with a float (which you do!), then you should do `60000.0 / rpm`

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -18,26 +18,71 @@
 EXTERN_ENGINE;
 
 static scheduling_s tachTurnSignalOff;
+static float halfperiodMs = 4; /* period in ms, used for multi pulse output, actually half period */
+static uint8_t isStarted = 0;
```
mck1117: This is a flag - should it be a `uint8_t` (hint: no)?

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -18,26 +18,71 @@
 EXTERN_ENGINE;
 
 static scheduling_s tachTurnSignalOff;
+static float halfperiodMs = 4; /* period in ms, used for multi pulse output, actually half period */
```
mck1117: Why set an initialized value here?

firmware/controllers/gauges/tachometer.cpp
shadowm60: Do you mean by coding style? use it as suggested here, https://rusefi.com/wiki/index.php?title=Development:Code_Style

firmware/controllers/gauges/tachometer.cpp
shadowm60: Sorry but i do not get it, line number pls.

firmware/config/boards/kinetis/config/controllers/algo/engine_configuration_generated_structures.h
```diff
@@ -2052,7 +2052,11 @@ struct engine_configuration_s {
 	/**
 	 * offset 1772
 	 */
-	uint8_t unusedFormerWarmupAfrPid[4];
+	uint8_t unusedFormerWarmupAfrPid[3];
```
mck1117: please put the new field at the BEGINNING of the unused block

firmware/controllers/gauges/tachometer.cpp
mck1117:
```
if (condition) {
    ...
}
```

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -18,26 +18,71 @@
...
 static void turnTachPinLow(void) {
 	enginePins.tachOut.setLow();
 }
 
+static void toggleTachPinLow(void)
+{
+	if(isStarted)
+	{
+		enginePins.tachOut.setLow();
+		engine->executor.scheduleForLater(&tachTurnSignalOff, (int)MS2US(halfperiodMs), (schfunc_t) &toggleTachPinHigh, NULL);
```
mck1117: there's a race condition here - I'd love to discuss what the correct way to implement this would be.  Can you join our Slack so we can discuss better there?  https://rusefi.com/forum/viewtopic.php?t=1198

firmware/controllers/gauges/tachometer.cpp
shadowm60: Float was from the original implementation, then i missed to change it to int.

Just checked the method scheduleForLater, dose it make sense to use float? at the end it will be casted to int anyway.

mck1117: You actually do - even though `scheduleForLater` takes an integer, it takes an integer number of microseconds, not milliseconds.  When you call `MS2US` it multiplies by 1000.

For example, if the engine was running at 6001 RPM:
60000 / 6001 (integer) = 9
Which multiplied by 1000 gives 9000us.  A period of 9000us is actually 6666 RPM - an ~11% error.

Doing the intermediate work as a float with conversion to int at the end gives 9.998ms, or 9998us.  A period of 9998us is 6001.20 RPM, a 0.1% error

shadowm60: You are right, checked!

firmware/config/boards/kinetis/config/controllers/algo/engine_configuration_generated_structures.h
rusefi: new field is already in master. BEGINNING is better but probably not a huge difference?

rusefi:
please add unit test coverage. please understand that it's not sustainable to merge code while relying only on manual testing.

i would not be able to merge without a unit test.

shadowm60:
I have spent today afternoon on the starting up the unit tests.
This is hard to write when most of the functions are static.
Hope that i will find a way to make it work.

mck1117:
See existing tests that turn the engine over, and check that the correct things got scheduled (for example rpm calculator). That makes them really integration tests, but they're better than nothing.

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -9,47 +9,118 @@
  * @date Aug 18, 2015
  * @author Andrey Belomutskiy, (c) 2012-2018
  */
-
+//#include "engine.h"
```
rusefi: why dead code? if you want to remove it - just remove it?

firmware/controllers/gauges/tachometer.cpp
```diff
@@ -9,47 +9,118 @@
...
 #include "tachometer.h"
 #include "trigger_central.h"
 
-#if !EFI_UNIT_TEST
 
 EXTERN_ENGINE;
 
-static scheduling_s tachTurnSignalOff;
+// [4 pulse/rev] * [2 edges per pulse] * [2 revs per cycle] = 16
+static scheduling_s events[16];
```
rusefi: looks like you want to say
#define MAX_TACH_PULSE 4

and use it to define array and also use it to validate user input?

what if user enters tachPulsePerRev =12?

shadowm60: You are right, can we define maximum number of pulses supported? It would make sense no more then 12 pulses, since max 12 cylinders could be controlled, or if we take into consideration that 16 pistons are the max number that i know, maybe it would make sense to limit and add comment in TS?

firmware/controllers/gauges/tachometer.cpp
shadowm60: Still in work, mck1117 helped me to start with the unit_tests, it is not the final code here

firmware/controllers/gauges/tachometer.cpp
mck1117: Of course - we can and should fire a warning if you try and use too many.  We can also set the limit in tunerstudio so that the user can't enter a number greater than 4.

rusefi:
Poke :)

rusefi:
why close? :(

shadowm60:
Rebase and ignored my changes, then with fresh base started reintegrate and
modify to comply with the new setup and run out of ram, it would not link,
i will try it again soon. It's working slow for me :(

PR #1090, opened 2020-01-09
Inject timestamps in to hwHandleShaftSignal

mck1117:
Taking the timestamp long after the interrupt has occurred gives an opportunity for not only a fixed offset error (time elapses between interrupt and timestamp), but also jitter.  The more time that elapses without taking the timestamp, the more time other interrupts could occur (or even different execution time due to branches).

The next work is to inject this timestamp all the way down to listeners - most importantly the logic that schedules ignition events, so that ignition timing is referenced to the actual trigger tooth event, not the trigger tooth event plus however long some logic takes to run (source of both offset error and jitter!).

### DIFF
--- a/firmware/controllers/trigger/trigger_central.cpp
+++ b/firmware/controllers/trigger/trigger_central.cpp
@@ -84,7 +84,7 @@ void addTriggerEventListener(ShaftPositionListener listener, const char *name, E
 	engine->triggerCentral.addEventListener(listener, name, engine);
 }
 
-void hwHandleVvtCamSignal(trigger_value_e front DECLARE_ENGINE_PARAMETER_SUFFIX) {
+void hwHandleVvtCamSignal(trigger_value_e front, efitick_t nowNt DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	TriggerCentral *tc = &engine->triggerCentral;
 	if (front == TV_RISE) {
 		tc->vvtEventRiseCounter++;
@@ -107,8 +107,6 @@ void hwHandleVvtCamSignal(trigger_value_e front DECLARE_ENGINE_PARAMETER_SUFFIX)
 	}
 
 	tc->vvtCamCounter++;
-
-	efitick_t nowNt = getTimeNowNt();
 
 	if (engineConfiguration->vvtMode == MIATA_NB2) {
 		uint32_t currentDuration = nowNt - tc->previousVvtCamTime;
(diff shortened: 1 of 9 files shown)


rusefi:
I like it, thank you for the improvement!

dron0gus:
Does trigger IRQs have higher priority over all others?

mck1117:
> Does trigger IRQs have higher priority over all others?

I think so, yes.  It's certainly higher priority than the execution timer, ADCs, USB, etc.  The data I've collected in testing suggests that there might still be a stray interrupt that's higher priority, but it's rare.  <0.1% of ignition events.

dron0gus:
> > Does trigger IRQs have higher priority over all others?
> 
> I think so, yes. It's certainly higher priority than the execution timer, ADCs, USB, etc. The data I've collected in testing suggests that there might still be a stray interrupt that's higher priority, but it's rare. <0.1% of ignition events.

Why I'm asking, I did EXTI trigger implementation some time ago. And I did not touch EXTI irq priorities.
I can check this later. My board has no timer functions on trigger pins so I'm interested in EXTI triggers.

mck1117:
Ah - I've only tested so far with the default ICU input on PC6 on an F4 Discovery. I'd like to switch the default for Proteus over to EXTI at some point (there's only one trigger input that's on an ICU pin currently), so that would be good to know.

mck1117:
I just looked in `mcuconf.h`, and it looks like the EXTI interrupts are set to the same priority as ICU.

dron0gus:
> I just looked in `mcuconf.h`, and it looks like the EXTI interrupts are set to the same priority as ICU.

Ok. But EXTI interrupts are also used for joystick (for example). So joystick interrupt can delay trigger interrupt. I think we need to set highest priority only for those EXTI channels that are used for trigger. All others should be low priority.

mck1117:
> Ok. But EXTI interrupts are also used for joystick (for example). So joystick interrupt can delay trigger interrupt. I think we need to set highest priority only for those EXTI channels that are used for trigger. All others should be low priority.

Few problems with that:
1) We can't know a priori which EXTI channels/interrupts are triggers and which might be joystick
2) How often are you actually pressing the joystick? Are you pressing it thousands of times per second at high engine speed?
3) Wait, people use the joystick?

dron0gus:
> Few problems with that:
> 
>     1. We can't know a priori which EXTI channels/interrupts are triggers and which might be joystick
Correct. I'll check if it possible to change priority at runtime (it should be) and if ChibiOS have such API. If so - we can adjust EXTI priority when allocating it for trigger.

> 
>     2. How often are you actually pressing the joystick? Are you pressing it thousands of times per second at high engine speed?
Joystick was first that comes in mind. TLE8888/TLE6240 both have IRQs, hope someday we'll start using it.
 
>     3. Wait, people use the joystick?
:) I don't know. My own board has no user interface expect reset button.

mck1117:
Anyway, that seems like an optimization that we can evaluate when we get to it.  Under current circumstances, I don't see any of the EXTI interrupts needing anything other than trigger input (ie, very high) priority.

PR #1093, opened 2020-01-10
kill dead time type

mck1117:


rusefi:
We've been here before with removal of questionable fratures. Lets just ignore them instead.

mck1117:
That's not a convincing reason to keep it.

Do you have an actually convincing reason to keep a 1% complete feature that was abandoned 3 years ago?

There's no way to deal with technical debt other than actually growing up and facing it.

rusefi:
We have a difference of opinions here. I am not interested to drbate this issues where you definitely have a legit point of view.

mck1117:
> you definitely have a legit point of view

Great, so we agree! Why not merge then?

PR #1094, opened 2020-01-11
Dedupe

mck1117:


firmware/controllers/engine_cycle/main_trigger_callback.cpp
```diff
@@ -122,11 +122,11 @@ static inline void tempTurnPinHigh(InjectorOutputPin *output) {
 	}
 }
 
-// todo: make these macro? kind of a penny optimization if compiler is not smart to inline
```
mck1117: 1. modern compilers completely ignore the `inline` keyword
2. `seTurnPinHigh` is a callback anyway - you can't inline it

Issue #1088, opened 2020-01-08
Support stepper control with a pair of H-Bridges


mck1117:
You can control a stepper with two H bridges, like the pair of TLE9201 found on Proteus.  We currently only support controlling a stepper with a real stepper IC with step/dir pins.

rusefi:
Would TLE8888 use the same two-H-bridges code once we have TLE8888-specific initialization?

rusefi:
Which part of this picture is our case @mck1117  ?

![image](https://user-images.githubusercontent.com/7377949/73611934-5ecb4b80-45b5-11ea-8881-630cbe135903.png)

rusefillc:
this works for while.

PR #1101, opened 2020-01-14
Separate stepper control logic from hardware

mck1117:
prep for #1088


mck1117:
This has now been tested on real hardware (with logic analyzer) - no change to behavior.

PR #1111, opened 2020-01-28
Use new sensors for TPS

mck1117:
We can allow both kinds of sensor decode exist along side one another for a while.

This adds the TPS (1 and 2, for dual etb) to the new sensor model.  This will make implementing redundant TPS easier, and natively allows for broken TPS detection.

I have another PR queued behind this one (pending testing) to use the new sensors in the ETB controller.


firmware/init/sensor/init_tps.cpp
```diff
@@ -0,0 +1,48 @@
...
+//FunctionalSensor tpsSens2s(SensorType::Tps2Secondary, MS2NT(10));
+
+void initTpsFunc(LinearFunc& func, FunctionalSensor& sensor, adc_channel_e channel, float closed, float open) {
+	// Only register if we have a sensor
+	if (channel == EFI_ADC_NONE) {
+		return;
+	}
+
+	func.configure(
+		closed / 200.0f, 0,
```
mck1117: todo: we should talk about this line on slack

mck1117:
Tested on real hardware!

PR #1112, opened 2020-01-29
Rescale TPS

mck1117:
Switch TPS to use 200 "adc counts" per volt, instead of actual ADC counts.  This abstracts the fake ADC counts we give to Tunerstudio away from anything to do with a real ADC.


mck1117:
Pending testing on real hardware.

mck1117:
Tested on real hardware with a potentiometer, seems to work great!

PR #1119, opened 2020-02-02
Some initial work at Launch/ALS

960:
I did a little work at these functions.

This is a huge boring job, so we just have to do a little now and then if we are real bored I think..

What's added for the launch-function now are a good pointer to show the minimum of settings needed for this to be any useful.

I also have to say I am truly sorry for some tiny unrelated changes in the .input, but there was a couple of small bugs that gave a .ini warning. My OCD won..

Also, there are something strange about the .txt, as whatever I do, I get a TS datapage mismatch, even I cant find any offset who are changed in the generated files..


rusefi:

rusefi:

rusefi:

rusefi:

rusefi:
cherry-picked

PR #1123, opened 2020-02-04
move flash addr in to mpu_util

mck1117:
tl;dr on stm32f7 we were actually trying to write both copies in to the same flash sector.  Writing the second copy was erasing the first one!  This change leaves the f4 and kinetis flash copies in place, and moves the stm32f7 copies to sectors 8 and 9, so they don't stomp each other.

Tested by manually erasing those sectors one at a time, and checking that the ECU retained its configuration when a single copy was erased.


rusefi:
Kinetis broken?

mck1117:
yeah, kinetis is broken.  Apparently I missed a few spots we were using FLASH_ADDR

PR #1129, opened 2020-02-06
Consume new tps in status_loop.cpp, settings.cpp

mck1117:


mck1117:
Now tested on real hardware, looks good!

PR #1131, opened 2020-02-08
Sensor reconfiguration while running

mck1117:
@960 noted that changing the TPS calibration doesn't take effect until a reboot occurs.  This change reconfigures sensors whenever the configuration changes.


mck1117:
Tested on hardware, good to go!

PR #1132, opened 2020-02-08
Stricter compile options

mck1117:
Turn back on a bunch of warnings-as-errors we had turned off, even though they never happen.

### DIFF
--- a/firmware/controllers/core/common_headers.h
+++ b/firmware/controllers/core/common_headers.h
@@ -88,6 +88,8 @@
 #define EXPAND_Engine \
 	    engine_configuration_s *engineConfiguration = engine->engineConfigurationPtr; \
 		persistent_config_s *config = engine->config; \
+		(void)engineConfiguration; \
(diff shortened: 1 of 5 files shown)


firmware/controllers/core/common_headers.h
```diff
@@ -88,6 +88,8 @@
 #define EXPAND_Engine \
 	    engine_configuration_s *engineConfiguration = engine->engineConfigurationPtr; \
 		persistent_config_s *config = engine->config; \
+		(void)engineConfiguration; \
```
mck1117: this saves tons of warnings about unused variable

mck1117:
Uhh this had failing CI

rusefi:
@mck1117 github CI is not really very reliable, was it a true failure?

mck1117:
It should be a little more reliable now, and yes, this was a real failure

rusefi:
Ooops :(

PR #1136, opened 2020-02-10
fix

mck1117:


mck1117:
looks like this only adds 5-6 seconds to the checkout time

PR #1149, opened 2020-02-19
Draft: Common factor for all corrections

960:
Initial start of changing all correction factors to % to reduce confusion.

Use 1d arrays for corrections for usability.


rusefi:

rusefi:

rusefi:

rusefi:

960:
Will turn them into this format:

javaw_aOpXCxG5A4

rusefi:
Add to whitelist

rusefi:
A change like this was supposed to be discussed in advance

Also I see that we are throwing multiple changes together. We need a different approach considering .input and .txt changes

960:
It's been discussed, but never came to any conclution.

You'll keep it as is.

PR #1154, opened 2020-02-23
Simplify flash logic

mck1117:
De-duplicate some code for how we save configuration, and also actually check whether both writes succeeded.


mck1117:
Tested out on f4/f7, looks good to me!

PR #1157, opened 2020-02-24
Add single flag for omitting PnP configs

mck1117:
Setting the flag to `false` saves ~35Kb of flash space.


mck1117:
Looks like the failing simulator build is an infrastructure issue, as it builds fine locally for me.

PR #1159, opened 2020-02-26
Omit canned configs from Proteus

mck1117:
Probably unwise to include configs with wholly incompatible pinouts.


mck1117:
that was brave, merging before gates were done?

rusefi:
This was within my wide boundaries of my faith in @mck1117

mck1117:
like I said, brave 😛

PR #1161, opened 2020-02-26
this flag did nothing

mck1117:
This flag did nothing, and if you turned it off, firmware didn't build.


firmware/hw_layer/hardware.cpp
```diff
@@ -54,13 +54,13 @@
 #include "mc33816.h"
 #endif /* EFI_MC33816 */
 
-#if EFI_SPEED_DENSITY
```
mck1117: this was simply just the wrong flag - map avg doesn't care about whether SD is enabled.

PR #1162, opened 2020-02-26
Support disabling most/all features

mck1117:
Not all calls were properly guarded for disabled features.  Now they are!


firmware/controllers/settings.cpp
```diff
@@ -1408,10 +1403,6 @@ void initSettings(void) {
 
 	addConsoleActionII("set_toothed_wheel", setToothedWheel);
 
-
-	// flat curve - constant injector lag regardless of voltage
-	addConsoleActionF("set_flat_injector_lag", setFlatInjectorLag);
```
rusefi: what has happened to this line?
set_flat_injector_lag

firmware/controllers/settings.cpp
```diff
@@ -1169,7 +1160,8 @@ typedef struct {
 } command_f_s;
 
 const command_f_s commandsF[] = {
-#if EFI_ENGINE_CONTROL && EFI_ENABLE_MOCK_ADC
+#if EFI_ENGINE_CONTROL
```
rusefi: this looks a bit confusing, I would prefer to separate blocks
if A&B
endif
if B
endif

PR #1177, opened 2020-03-08
fix icu trigger polarity

### LINKED ISSUE #1176: BUG HW trigger inputs got inverted...
I think I've seen stuff inverted but I just decided to ignore it.

And now we have a complaint around https://rusefi.com/forum/viewtopic.php?p=35616#p35616

mck1117:
fix #1176

### DIFF
--- a/firmware/development/logic_analyzer.cpp
+++ b/firmware/development/logic_analyzer.cpp
@@ -110,14 +110,12 @@ static void initWave(const char *name, int index) {
 	if (brainPin == GPIO_UNASSIGNED)
 		return;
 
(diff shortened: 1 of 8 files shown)


firmware/development/logic_analyzer.cpp
```diff
@@ -110,14 +110,12 @@ static void initWave(const char *name, int index) {
 	if (brainPin == GPIO_UNASSIGNED)
 		return;
 
-	bool mode = CONFIG(logicAnalyzerMode)[index];
```
mck1117: there wasn't actually any way to set this - should probably remove and save the config space.

mck1117:
Does the correct thing now:
image

PR #1121, opened 2020-02-02
Control stepper with dual H-bridges

mck1117:
#1088

PR #1178, opened 2020-03-10
Improve dual hbridge idle UI

mck1117:
Part of #1121 / #1088.

Stepper HW page now looks like this:
image


rusefi:
fresh .ini pushed

PR #1211, opened 2020-03-22
save some ram

mck1117:
This removes a bunch of stuff that the linker was removing because it wasn't actually used in a release build.

This is all fully dead code, or data that didn't actually do anything (data written, never read, sort of thing).


mck1117:
This should also fix ci in master (the debug build ran out of RAM by 8 bytes)

PR #1216, opened 2020-03-24
Enforce & unify versioning

mck1117:
We've had a lot of problems lately with incompatible configurations causing issues.  This change enforces a single version from TS, to firmware, to flash config.


rusefi:
Do you propose to make _any_ change incompatible?

I would prefer to keep to versions - compatible version and incompatible changes.

I would totally refuse to migrate every time I need to update firmware.

mck1117:
See comment:
```
! WHEN SHOULD I UPDATE THE BELOW VERSIONS?
#define TS_SIGNATURE "rusEFI v1.2020.3"
! - Update TS_SIGNATURE whenever either the flash format or TS output format changes (format: Year.Month.Sequential)
! - Update RUSEFI_FLASH_VERSION whenever the flash format is changed in a way that isn't backward compatible in a safe way (format: YearMonthDay)
```
Were you thinking something different?

rusefi:
At the moment getRusEfiVersion() should be updated on every day when any change was made. I would prefer NOT to touch TS every day.

uint32_t getRusEfiVersion() is the third version - the "runtime" version

mck1117:
Ok - so we could split runtime to a third version.  We just have to actually update the other two religiously.

rusefi:
Absolutely, a recent f%p was due to not updating one.

Actually we can maybe have the code generator help us... But not sure how - often adding new fields is compatible.

mck1117:
Ok, this is now tested, and does indeed do what we want it to.  It'll cause a breaking change for every user - so we should probably pull the trigger sooner rather than later.

firmware/integration/rusefi_config.txt
```diff
@@ -22,22 +22,21 @@
...
 
-
-#define TS_SIGNATURE "rusEFI v1.2020.3"
+! WHEN SHOULD I UPDATE THE BELOW VERSIONS?
+! - Update TS_SIGNATURE whenever either the flash format or TS output format changes (format: Year.Month.Sequential)
+! - Update RUSEFI_FIRMWARE_VERSION periodically as the firmware is updated.  Its value doesn't necessarily indicate dangerous
+!       incompatibility, but could indicate missing features.
+! - Update RUSEFI_FLASH_VERSION whenever the flash format is changed in a way that isn't backward compatible in a safe way (format: YearMonthDay)
+#define TS_SIGNATURE "rusEFI v2020.3.0"
+#define RUSEFI_FIRMWARE_VERSION 20200323
```
rusefi: at the moment in order to increase runtime version I modify one .cpp file, only one .cpp file needs to be recompiled

if we move runtime version to rusefi_config.txt - I would need to invoke generator and I will have one or multiple headers change which would require complete recompile. I would also have to push multiple files to git not just one file

this significantly increases the number of manual clicks for me to do. at this point I am not convinces that there is enough value in this change to justify the forever discomfort.

rusefi:
if part of this PR is better names for existing #define constants - can we start by only renaming #define constants?

if part of this PR is more commends - can we start by only pushing new comments?

rusefi:
@mck1117 shall we close?

PR #1224, opened 2020-03-25
Goodbye, fahrenheit

### LINKED ISSUE #1214: Can we make the TS default deg C?

mck1117:
1. our fahrenheit support is terrible
2. tunerstudio insists upon defaulting to fahrenheit
3. most of our users probably want celsius anyway (and I've already seen one screenshot where they didn't realize they were set to F)

It turns out if you don't have any `#if CELSIUS` in the ini, then TS doesn't even show the dropdown for units!

Closes #1214.

### DIFF
--- a/firmware/tunerstudio/rusefi.input
+++ b/firmware/tunerstudio/rusefi.input
@@ -135,11 +135,7 @@ enable2ndByteCanID = false
        ; filter =  Name,        "DisplayName", outputChannel, operator, defaultVal, userAdjustable
         filter = minRPMFilter, "Minimum RPM", RPMValue,           <       , 500,      , true
 
-#if CELSIUS
          filter = minCltFilter, "Minimum CLT", coolant,       <       , 60,       , true
-#else
-         filter = minCltFilter, "Minimum CLT", coolant,       <       , 160,      , true
-#endif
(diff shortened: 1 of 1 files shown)


rusefi:
this is not an inclusive approach

I totally agree that C should become the default but in my opinion we should support F as well

We should not just wipe code out, this is a lazy approach in my opinion.

rusefi:
please do not remove C and F switch

please default to C, please consider improving F support

mck1117:
TS doesn't appear to have a way to set the default to C.  I've already seen users get confused by this.  Since we're a "shipping product" with users  who can get confused, IMO it's best for us to disable F completely for now, and re-enable later when we have the resources to do it right (and make it not the default)

rusefi:
@mck1117 can you please have a glance at the magic spec document? do they say anything about this?

mck1117:
I had a look at both the magic spec, and all it had to say was: 
> For example, this can allow you to display a temperature in °F or °C based on a set user preference.

Without an actual example of how to do that.

I also had a look at an MS3 init, and it also defaults to F, without any content in the ini about anything other than `#if CELSIUS` like we had/have.

Issue #1215, opened 2020-03-24
hide kinetis comparator settings for non-kinetis


mck1117:

mck1117:
looks like this is done

PR #1226, opened 2020-03-25
Ts cleanup 2

### LINKED ISSUE #1223: Hide SD card, TS port from Connection TS page

mck1117:
Fixes #1223.  Hides SD card, CAN pins, Tunerstudio port.

Also covers some missing stuff that should've already been hidden (SPI pins).


mck1117:
...and fixes #1215 too!

firmware/config/boards/kinetis/prepend.txt
```diff
@@ -0,0 +1 @@
+#define ts_show_trigger_comparator true
```
mck1117: @rusefi @andreika-git it seems like this doesn't work, is there some trick to get the tool to pick this up?

andreika-git: Try putting it inside firmware/config/boards/kinetis/config/rusefi_config_kinetis.txt

andreika-git: Or firmware/config/boards/kinetis/config/tunerstudio/kinetis_prefix.txt

mck1117: The former suggestion seems to work - thanks!

PR #1229, opened 2020-03-26
Finally rename injector_central to bench_test

mck1117:
As alluded to by many comments strewn about the firmware, `injector_central` was completely the wrong name. This change fixes that!

Oh, and get rid of the huge pile of `extern EnginePins enginePins;`


mck1117:
`fsio_impl.cpp` even had `extern EnginePins enginePins;` twice!

PR #1230, opened 2020-03-26
Draft: AfterStart Enrichment

960:
Afterstart enrichment

Working, but probably needs some code refinements or different location..


firmware/controllers/algo/engine2.cpp
```diff
@@ -174,17 +174,30 @@ void EngineState::periodicFastCallback(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 	// Fuel cut-off isn't just 0 or 1, it can be tapered
 	fuelCutoffCorrection = getFuelCutOffCorrection(nowNt, rpm PASS_ENGINE_PARAMETER_SUFFIX);
 
-	// post-cranking fuel enrichment.
-	// for compatibility reasons, apply only if the factor is greater than zero (0.01 margin used)
-	if (engineConfiguration->postCrankingFactor > 0.01f) {
-		// convert to microsecs and then to seconds
+
+	    float afterStartEnrich;
```
mck1117: you don't need this separate block of declarations - you can declare and initialize in the same line.
example:
`float afterstartHoldTime = interpolate2d("aseHold", .....);`

firmware/controllers/algo/engine2.cpp
```diff
@@ -174,17 +174,30 @@ void EngineState::periodicFastCallback(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
...
+
+		if (runTime < afterstartHoldTime) {
+			correction = afterStartEnrich;
+		} else {
+			correction = afterstartDecayFuel;
+		}
+		if (runTime > (afterstartHoldTime + afterstartDecayTime)) {
+			correction = 1.0f;
+		}
+		running.postCrankingFuelCorrection = correction;
```
mck1117: should we move this logic out of `engine2.cpp`? This file is quite unwieldy.  Perhaps would be more appropriate to bundle this up in to a function (also so it's easier to test!) and put it in `fuel_math.cpp`? I'm happy to help on Slack.

firmware/integration/rusefi_config.txt
```diff
@@ -155,6 +155,9 @@ struct_no_prefix engine_configuration_s
 #define BARO_CORR_SIZE 4
 
 #define MAF_DECODING_COUNT 256
+#define AFTERSTART_HOLD_CURVE_SIZE 8
```
mck1117: do we actually need this much resolution? Is 6 enough?

firmware/integration/rusefi_config.txt
```diff
@@ -1191,13 +1194,23 @@ uint8_t[4] unusuedsw;
...
-	int[527] mainUnusedEnd;
+	int[479] mainUnusedEnd;
 
 ! end of engine_configuration_s
 end_struct
 
 engine_configuration_s engineConfiguration;
 
+
+float[AFTERSTART_HOLD_CURVE_SIZE] afterstartHoldTimeBins;;"C",        1,     0,   -100.0,    250.0,  0
```
mck1117: Can we share the bins between all of these? Is there a case where we need different bins for each of these tables?

firmware/integration/rusefi_config.txt
```diff
@@ -1191,13 +1194,23 @@ uint8_t[4] unusuedsw;
...
+float[AFTERSTART_HOLD_CURVE_SIZE] afterstartHoldTime;;"Seconds",      1,     0,      0,    100,   1
+
+float[AFTERSTART_DECAY_CURVE_SIZE] afterstartDecayTimeBins;;"C",        1,     0,   -100.0,    250.0,  0
+float[AFTERSTART_DECAY_CURVE_SIZE] afterstartDecayTime;;"Seconds",      1,     0,      0,    100,   1
```
mck1117: this table can be uint8_t, not float

firmware/integration/rusefi_config.txt
```diff
@@ -1191,13 +1194,23 @@ uint8_t[4] unusuedsw;
...
+
+float[AFTERSTART_ENRICH_CURVE_SIZE] afterstartEnrichBins;;"C",        1,     0,   -100.0,    250.0,  0
+float[AFTERSTART_ENRICH_CURVE_SIZE] afterstartEnrich;;"%",            100,     -1,      0,    600,   1
```
mck1117: I'm not sure the offset/scaling/limits are correct here.  This can also probably be a `uint8_t`

firmware/controllers/algo/engine2.cpp
960: Yeah, now I remember why I did it like this.
It was to get the values in variables while debugging :-)

firmware/controllers/algo/engine2.cpp
960: I dont even know what engine2 means..
Fuel_math.cpp would be a more logic place to look for it

firmware/integration/rusefi_config.txt
960: They could share bins for sure..

firmware/integration/rusefi_config.txt
960: Scaling are right I belive.
I have done 100 and -1 for most enrichments to get value in % in addition.

So that 0% are 0 enrichment

firmware/controllers/algo/engine2.cpp
mck1117: You can still inspect the values in the debugger...  Remember that the variable doesn't exist until the line AFTER it's declared, so you have to step over a line to see what it did.

firmware/controllers/algo/engine2.cpp
mck1117: probably near things like `getCltFuelCorrection` and `getIatFuelCorrection`, line 340-ish would be good.

firmware/integration/rusefi_config.txt
mck1117: Is that consistent with other fields in the firmware? Other multipliers are just that - a multiplier.  1.0 = no adjustment, 1.5 = add 50% extra.  Adding a setting for those is a separate problem, and whatever we add for now should be consistent with our other stuff.

960: You are right, it should be consistent. This is how I have it, but doesnt match master.
Will change this.

firmware/controllers/algo/engine2.cpp
960: Do I put it inside some function?

firmware/integration/rusefi_config.txt
mck1117: thanks!

firmware/integration/rusefi_config.txt
960: Changed to uint8_t

firmware/integration/rusefi_config.txt
960: Did share their bins :-)

firmware/integration/rusefi_config.txt
960: Probably not needed.
Just used to see all these curves with 8 :-)

960:
Finally.. :-)

firmware/integration/rusefi_config.txt
960: We dont have interpolation method for uint8_t.

rusefi:
@960 please clarify why close the PR? we have some relevant settings now hanging in the codebase? what is your plan? Do you plan to remove unsued?

960:
I didnt close it.. 
I had to delete fork as GIT was totally fucked, and I couldnt unfuck it.. :-)

rusefi:
We have #git_drama in Slack with a couple experts available to help you un-fuck or not fuck in the first place.

Hope to see this PR back from the new fork! Also hope to see boost clean-up PR and etc :)

PR #1233, opened 2020-03-29
Friendlier pin names

mck1117:
Since users might see these, we should probably have them be at least a little bit nice.


firmware/hw_layer/adc_inputs.cpp
```diff
@@ -498,32 +498,31 @@ static void configureInputs(void) {
 	 */
 
 	addChannel("MAP", engineConfiguration->map.sensor.hwChannel, ADC_FAST);
-	if (hasMafSensor()) {
```
mck1117: `hasMafSensor` just checks for unset channel - which is the same thing that `addChannel` does, so we don't need the check first.

PR #1238, opened 2020-03-30
Debuggability in the new sensor world

mck1117:
- Hook up each sensor printing information about itself (next step is having the functional sensor print out info about the function in question)
- Rename `initSensors` to `initNewSensors` to prevent name collision
- Fix `efilib.h` - it didn't compile if it was the first/only thing included in a file, which somehow flew under the radar for a long time.
- Remove duplicate check from `printAnalogChannelInfo` - it already happens in `printAnalogChannelInfoExt`

### DIFF
--- a/firmware/controllers/sensors/sensor.cpp
+++ b/firmware/controllers/sensors/sensor.cpp
@@ -166,14 +166,14 @@ bool Sensor::Register() {
 		const char* name = s_sensorNames[i];
 
 		if (entry.useMock) {
-			scheduleMsg(logger, "Sensor %s mocked with value %f", name, entry.mockValue);
+			scheduleMsg(logger, "Sensor \"%s\" mocked with value %.2f", name, entry.mockValue);
 		} else {
 			const auto sensor = entry.sensor;
 
 			if (sensor) {
 				sensor->showInfo(logger, name);
 			} else {
-				scheduleMsg(logger, "Sensor %s is not configured.", name);
+				scheduleMsg(logger, "Sensor \"%s\" is not configured.", name);
 			}
 		}
 	}
--- a/firmware/controllers/sensors/sensor.cpp
+++ b/firmware/controllers/sensors/sensor.cpp
@@ -1,4 +1,6 @@
 #include "sensor.h"
+#include "efilib.h"
+#include "loggingcentral.h"
 
 // This struct represents one sensor in the registry.
 // It stores whether the sensor should use a mock value,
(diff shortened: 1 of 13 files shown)


mck1117:
Here's what the output from `show_sensors` looks like:
```
2020-03-30_14_51_33_704: EngineState: Sensor "CLT" is not configured.
2020-03-30_14_51_33_704: EngineState: Sensor "IAT" is not configured.
2020-03-30_14_51_33_711: EngineState: Sensor "Oil Pressure" is not configured.
2020-03-30_14_51_33_854: EngineState: Sensor "TPS 1": Raw value: 0.98 Valid: 1 Converted value -8.03
2020-03-30_14_51_33_859: EngineState: Sensor "TPS 1 Primary" is not configured.
2020-03-30_14_51_33_862: EngineState: Sensor "TPS 1 Secondary" is not configured.
2020-03-30_14_51_33_865: EngineState: Sensor "TPS 2" is not configured.
2020-03-30_14_51_33_868: EngineState: Sensor "TPS 2 Primary" is not configured.
2020-03-30_14_51_33_873: EngineState: Sensor "TPS 2 Secondary" is not configured.
2020-03-30_14_51_33_875: EngineState: Sensor "Acc Pedal" is not configured.
2020-03-30_14_51_33_875: EngineState: Sensor "Acc Pedal Primary" is not configured.
2020-03-30_14_51_33_876: EngineState: Sensor "Acc Pedal Secondary" is not configured.
2020-03-30_14_51_33_876: EngineState: Sensor "Driver Acc Intent" proxied from sensor "TPS 1"
```

firmware/controllers/sensors/sensor.cpp
```diff
@@ -12,6 +14,30 @@ struct SensorRegistryEntry {
 
 static SensorRegistryEntry s_sensorRegistry[static_cast<size_t>(SensorType::PlaceholderLast)] = {};
 
+static const char* s_sensorNames[] = {
+	"Invalid",
```
rusefi: dirty magic index matching? kind of hacky and poorly documentad coupling between enum and this array?

mck1117: that's what the static assert is for - it at least guarantees that the sizes match

firmware/util/efilib.h
```diff
@@ -45,19 +45,6 @@ int indexOf(const char *string, char ch);
 float atoff(const char *string);
 int atoi(const char *string);
 
-#if defined(__cplusplus) && defined(__OPTIMIZE__)
-#include <type_traits>
```
rusefi: how is this related to "Debuggability in the new sensor world"? :)

mck1117: see second bullet point:
> Fix efilib.h - it didn't compile if it was the first/only thing included in a file, which somehow flew under the radar for a long time.

firmware/controllers/sensors/sensor.cpp
rusefi: that's not enough. this has to designed better or documented better IMHO

PR #1241, opened 2020-03-31
Sensor-via-CAN

mck1117:
#1212 is blocked on this.


mck1117:
now tested in loopback mode - here's evidence of a valid sensor reading via CAN:

```
....
2020-03-31_18_50_31_155: EngineState: CAN Sensor "Acc Pedal": valid: 1 value: 23.00
....
```

PR #1247, opened 2020-04-02
only correct timing in dynamic mode

### LINKED ISSUE #1246: Should timing correction be applied in "fixed timing" mode?
Even when set to fixed, we still add CLT/IAT corrections.  That's probably wrong, since fixed mode is supposed to be used for things like engine setup and trigger debugging.

mck1117:
fixes #1246


mck1117:
Also saves 132 bytes of flash at `-O2`, so that's neat.

Issue #1240, opened 2020-03-31
ETB dual signal implementation 


OrchardPerformance:
Issue raised to house the outcome of discussion on slack. Aim is to improve ETB implementation and implement basic safety checking. 

- Both dual pedal and dual TPS signals are calibrated as normal
- Both signals are translated and compared for matching pedal(APP)/TPS percentage
- Configurable threshold for the deviation allowed between signal A and B, if this threshold is broken then trigger an error
- Current high and low thresholds will apply to both signals should they go open or short circuit
- If both signals are too similar, i.e. both 0-5v, then trigger an error or disallow the calibration to prevent single signal hacks 
- Single signal mode can be activated for testing or dev work only on a debug build and by setting both a bit in the INI and activating a bench test button to prevent accidental single signal use

rusefi:
See also #1336

OrchardPerformance:
Can this one now be closed as we have dual signal as a ticket open for the deviation check? 
Seems no need for this one anymore? 

@mck1117 - Do you think this one needs to stay open for the "similar signal check", do we care about that?

OrchardPerformance:
As Matt has said he has driven on dual signal I think we can close this unless any problems pop up.

PR #1248, opened 2020-04-02
run ETB on new TPS

mck1117:
Progress toward #1240


firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -109,6 +102,13 @@ static bool startupPositionError = false;
 
 #define STARTUP_NEUTRAL_POSITION_ERROR_THRESHOLD 5
 
+static SensorType indexToTpsSensor(size_t index) {
+	switch(index) {
```
rusefi, line 106: code style - space is missing?

mck1117: you mean between `switch` and `(index)`?

mck1117: lol why'd you merge before I could fix this 🤣

rusefi: yes before "switch" and "("

PR #1251, opened 2020-04-03
fix electronic throttle RPM rounding

mck1117:
https://www.youtube.com/watch?v=9A4M3X57PCI


rusefi:
A unit test would be nice

PR #1258, opened 2020-04-04
yet more tps consumers

mck1117:
- decel fuel cut
- closed loop fuel threshold
- TPS accel enrich
- dead TPS rate-of-change logic
- remove old TPS mocking


unit_tests/tests/test_sensors.cpp
```diff
@@ -26,35 +24,6 @@ TEST(sensors, mapDecoding) {
...
 
-TEST(sensors, tps) {
-	print("************************************************** testTps\r\n");
-
-	WITH_ENGINE_TEST_HELPER(DODGE_RAM);
-
-	engineConfiguration->tpsMax = 193;
-	engineConfiguration->tpsMin = 43;
-
-	ASSERT_NEAR(49.3333, getTpsValue(0, 117 PASS_ENGINE_PARAMETER_SUFFIX), EPS4D);
```
rusefi, line 37: why do we just remove interpolation TPS test?

mck1117: Because `getTpsValue` is about to disappear completely - since the TPS runs on a linear function (`linear_func.h`), it's covered by that test (`lin_func.cpp`).  That also tests other linear sensors like oil pressure, pedal, and eventually battery voltage and MAP.

rusefi: so the other tests test "linear" but probably no test takes care of integration with these specific "tpsMax/tpsMin" settings?

mck1117: see #1264

PR #1268, opened 2020-04-06
fix oil pressure sensor

mck1117:
- Fix the bug
- Make the bug impossible via the type system: delete the copy constructor
- Add a test that catches the bug


mck1117:
This change caused the bug in the first place: https://github.com/rusefi/rusefi/commit/986b80671b6dfa7af21d429e0a47e50bd868f5c2#diff-61dcfbcfdc0f6c27c70b325d718d29d9

In the rare-but-entertaining case of two mistakes canceling each other out.  I'd ignored the parameter being passed in to the function, but that was good - it was a copy of the thing I wanted to edit!

PR #1273, opened 2020-04-09
support ETB disable pin

mck1117:
all the ETB driver chips we currently use have a DIS pin - let's support it natively


rusefi:
We have another user on the forum in desperate need of this :)

firmware/controllers/engine_controller.cpp
```diff
@@ -695,7 +695,7 @@ void initEngineContoller(Logging *sharedLogger DECLARE_ENGINE_PARAMETER_SUFFIX)
 // help to notice when RAM usage goes up - if a code change adds to RAM usage these variables would fail
 // linking process which is the way to raise the alarm
 #ifndef RAM_UNUSED_SIZE
-#define RAM_UNUSED_SIZE 13850
+#define RAM_UNUSED_SIZE 13842
```
rusefi, line 698: C'mon, please change this value by at least 100 every time when you change it

Issue #1261, opened 2020-04-04
ETB sometimes runs at 100% duty cycle while flashing


mck1117:
We need some `enableSafeMode()` / `disableSafeMode()` to enable while erasing/writing flash memory, as that blocks the CPU.  Currently if you start erasing flash while PWM is in the high part of the duty cycle, then we get stuck at 100% duty cycle. This is bad! It pins the throttle open!

rusefi:
Is this same for any pwm like iac or vvt?

mck1117:
Yes, same for those too.

mck1117:
fixed with #1283

PR #1283, opened 2020-04-12
Hardware pwm

mck1117:
On pins that support it, we can use hardware PWM instead of software.  This relieves scheduler pressure, and improves PWM accuracy.


mck1117:
As an added bonus, hardware PWM stays running during flash erase/write... #1261

mck1117:
Tested my ETB up to ~10khz and down to 50hz PWM - works great!

mck1117:
The pins configured for hardware PWM with this PR are PB6-9, PC6-9, and PD12-15.  Most importantly that covers the ETB PWM pins on both MRE and Proteus.

rusefi:
@mck1117  poke?

firmware/hw_layer/ports/stm32/stm32_common.cpp
```diff
@@ -252,6 +252,11 @@ stm32_hardware_pwm pwmChannels[] = {
 
 /*static*/ hardware_pwm* hardware_pwm::tryInitPin(const char* msg, brain_pin_e pin, float frequencyHz, float duty)
 {
+	// Slow PWM isn't worth doing on hardware - reserve that timer for something faster if needed.
+	if (frequencyHz < 100) {
```
rusefi, line 256: that's really weird to be choosing HW at runtime. Does not look right to me.

mck1117: You mean like how we already do this for ICU inputs, EXTI inputs, SPI devices, CAN devices, serial devices, etc?

rusefi: Looks a bit different to me - in ICU inputs, EXTI inputs, SPI devices, CAN devices, serial devices we select HW _directly_ based on _HW_ parameters

here we choose HW based on something not obviously connected to HW.

mck1117: comment & threshold updated to the actual limit of what this implementation can do

firmware/hw_layer/ports/kinetis/mpu_util.cpp
```diff
@@ -257,5 +257,10 @@ uintptr_t getFlashAddrSecondCopy() {
 	return 0x10008000;
 }
 
+/*static*/ hardware_pwm* hardware_pwm::tryInitPin(const char*, brain_pin_e, float, float) {
+	// TODO: implement me!
```
rusefi, line 261: what is the plan here?

mck1117: This isn't required to be implemented, it just means hardware PWM won't ever be used on Kinetis.

PR #1292, opened 2020-04-13
Add w202 canDash support

shadowm60:
added CanDash support for w202


rusefi:

rusefi:

rusefi:

rusefi:

shadowm60:
I think it can be closed, if i find something more in the dbc while developing the car i will create a new request.

rusefi:
Usually we do not even need to create a ticket if we plan to PR the complete solution right awau

mck1117:
ooh, more dashboards!

PR #1295, opened 2020-04-13
Update to Matt's MAF code for VE table tuning. 

OrchardPerformance:
PR-ing this now in the spirit of small changes. 
Built upon Matt's last PR to allow the MAF fuelling to be tuned with the VE table. 
MAF air charge fuelling now targets AFR and has correction for air flow inconsistencies like reversion through the MAF. 
Creates a value for "airChargeLoad" which is a standardised VE value and is used as Y axis for the VE table. 
VE table tunes MAF fuel by percentage. 100 = no correction, 80 = 20% less fuel, 120 = 20% more fuel. 
VE table enabled in Tuner studio for LM_REAL_MAF instead of Fuel Table.

Known issues:
Pulse width output is tested ok but does cause the build test to fail due to the change in the maths.
Y axis on VE table does not move in TS, need to change rusefi.input file to solve this for the next PR. 
For now the table works but does not show a live highlight/cursor in TS.

### DIFF
--- a/firmware/controllers/algo/fuel_math.cpp
+++ b/firmware/controllers/algo/fuel_math.cpp
@@ -142,6 +142,7 @@ floatms_t getRunningFuel(floatms_t baseFuel DECLARE_ENGINE_PARAMETER_SUFFIX) {
 /* DISPLAY_ENDIF */
 
 /**
+ * Function block now works to create a standardised load from the cylinder filling as well as tune fuel via VE table. 
  * @return total duration of fuel injection per engine cycle, in milliseconds
  */
 float getRealMafFuel(float airSpeed, int rpm DECLARE_ENGINE_PARAMETER_SUFFIX) {
--- a/firmware/controllers/algo/fuel_math.cpp
+++ b/firmware/controllers/algo/fuel_math.cpp
@@ -151,9 +151,6 @@ float getRealMafFuel(float airSpeed, int rpm DECLARE_ENGINE_PARAMETER_SUFFIX) {
 		return 0;
 	}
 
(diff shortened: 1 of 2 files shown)


rusefi:

rusefi:

rusefi:

rusefi:

OrchardPerformance:
Something wrong with what I did here. 

Will double check and re PR

OrchardPerformance:
Reopened as it looks like github actions stopped working at the same time I submitted and I had a little panic that I had failed ALL the build tests.

firmware/controllers/algo/fuel_math.cpp
```diff
@@ -150,20 +151,28 @@ float getRealMafFuel(float airSpeed, int rpm DECLARE_ENGINE_PARAMETER_SUFFIX) {
 		return 0;
 	}
 
+	//Calculation of 100% VE air mass in g/rev
+	float StandardAirCharge = CONFIG(specs.displacement) / CONFIG(specs.cylindersCount) * 1.2929; //1 cylinder filling at 1.2929g/L
```
mck1117, line 155: Might be a bit easier to read this if we move this line down to just before the `float airChargeLoad...` line.

OrchardPerformance: The comment? 
//1 cylinder filling at 1.2929g/L

OrchardPerformance: Oh you mean the whole line. 
See no problem with that.

PR #1297, opened 2020-04-14
new temperature sensor math in TS

mck1117:


mck1117:
Now tested on real hw!

PR #1300, opened 2020-04-15
dead code

mck1117:


rusefi:
Smaller version of the fix was implemented.

mck1117:
err, ok.  Not sure why we insist upon keeping around code that's dead, and hasn't been touched in 3 years.  It clearly isn't gonna get finished any time soon, and is needlessly occupying flash/ram when we're a bit strapped for the latter.

rusefi:
Noted.

PR #1304, opened 2020-04-16
Gauges for raw sensors

mck1117:
Sensors are easier to diagnose when you don't need to get out a multimeter - now we have gauges for them!


mck1117:
https://www.youtube.com/watch?v=qhLTgItwlxw

Issue #767, opened 2019-04-20
input mode control and other pin mode control


rusefi:
we still have too much pin modes hard-coded in board.h file

also recent ICU to PAL input trigger conversion might have introduced new issues in this area

rusefi:
See https://rusefi.com/forum/viewtopic.php?p=32767#p32767

rusefi:
@dron0gus we need dynamic pin configuration for EXTI inputs
please make trigger_input more consistent with joystick which 
a) configures pin for input
b) marks it used as a consequence of that

rusefillc:
description too vague and too much has changed since 2020

Issue #1308, opened 2020-04-17
pin defaults are dangerous on microRusEfi


rusefi:
pin defaults used by microRusEfi at https://github.com/rusefi/rusefi/blob/master/firmware/config/boards/st_stm32f4/board.h

for example PD4 default is pretty bad :(

rusefi:
@dron0gus @mck1117 what is the status of this issue?

mck1117:
They are no longer as dangerous as they were.  They are now safe except in cases of spurious stm32 reset (without TLE8888 reset).  Open question whether we want to improve further with board-specific `board.h` files or not.

rusefi:
spurious stm32 reset (without TLE8888 reset) would be taken care by HW changes

Issue #1054, opened 2019-12-12
board.h default state for F7 pins


rusefi:
at the moment F7 board.h has a lot of different inconsistent default boot pin states.

input-floating should be the default pin mode for everything i-o.

rusefi:
@mck1117 can you take care of the F7 board.h file?

OrchardPerformance:
@mck1117 - Does your commit mean this ticket can be closed?

rusefillc:
@mck1117 what's the state of this - do we need this open?

rusefillc:
@dron0gus i really hope this was done

Issue #1053, opened 2019-12-12
board.h default state for F4 pins


rusefi:
at the moment F4 board.h has a lot of different inconsistent default boot pin states.

input-floating should be the default pin mode for everything i-o.

rusefi:
See also #1054 for F7

rusefi:
We also have a somewhat similar #397

rusefillc:
@mck1117 what's the state of this - do we need this open?

rusefillc:
@dron0gus i really hope this was done

PR #1310, opened 2020-04-17
Simplify F4 board.h

mck1117:
Soon to be merged with F7 board.h

Deduplicate most definitions, as most ports are set to the defaults (all pins input pulldown), only doing something different where necessary.


mck1117:
#1308 and #1053 and soon #1054

mck1117:
And also fixes #767

PR #1311, opened 2020-04-18
MRE: board.h file for MRE with MRE specific settings for GPIOs

dron0gus:


rusefi:
@dron0gus what is the plan? says it's a draft, also CI is red for some reason. Please rebase on master since @mck1117 was doing something relevant in master

mck1117:
IMO if we really want custom `board.h` settings per-board, we should:
- Include the base board.h from the specific board.h
- `#undef` then `#define` the stuff that we want to change to something other than default

PR #1319, opened 2020-04-18
Improve ETB idle control

mck1117:
Here's the problem.  With ETB idle configured, it currently just scoots the whole target position up by the desired idle amount.  If your pedal target table has values of 0-100% in it, that means you can now try to move the throttle to [0 + idle, 100 + idle], which probably isn't what you wanted. 

This change moves the "virtual 0% position" to be at the idle position, and leaves 100% as-is.  That means at 50% target, the actual position target will be 50% + idle/2.


rusefi:
this seems very testable
this seems very much worth test coverage

mck1117:
This now has full test coverage!

PR #1322, opened 2020-04-19
Closed loop controller base & ETB testing

mck1117:
- Controller base frames out the structure of a basic open loop + closed loop controller.  Implementing a new controller only requires implementing the five virtual functions of the base.  These functions are often each very simple.  Each can return `unexpected` to indicate a failure has occured, which will be passed to the final one so it can put the system in a safe state.
    - `getSetpoint`: where is the controller trying to move the plant?
    - `observePlant`: where is the plant now? (this means read a sensor, usually)
    - `getOpenLoop`: get any output value that should occur as a function of only the setpoint.
    - `getClosedLoop`: get any output value that should occur as a function of the setpoint and observed state
    - `setOutput`: drive whatever output system this controller drives, making an adjustment to the plant in the physical world
- Consume new controller base with ETB
- Thorough tests for controller base
- First ETB test against new controller base: checking whether we're reading the correct TPS sensor based on the ETB's index
- Update gtest: required for the new and improved `MOCK_METHOD` macro


unit_tests/tests/test_etb.cpp
```diff
@@ -90,3 +90,23 @@ TEST(etb, testTargetTpsIsFloatBug945) {
...
+TEST(etb, etbTpsSensor) {
+	// Throw some distinct values on the TPS sensors so we can identify that we're getting the correct one
+	Sensor::setMockValue(SensorType::Tps1, 25.0f);
+	Sensor::setMockValue(SensorType::Tps2, 75.0f);
+
+	// Test first throttle
+	{
+		EtbController etb;
+		etb.init(nullptr, 0, nullptr);
+		EXPECT_EQ(etb.observePlant().Value, 25.0f);
```
mck1117, line 103: This is the kind of unit testing I like being able to do - testing one thing at a time, no extra dependencies you have to set up (like injecting an engine reference or something)

PR #1328, opened 2020-04-19
Improve tcharge error handling

mck1117:
One or both of CLT/IAT sensors could be failed, so we should handle that appropriately, without spamming the log unnecessarily.


rusefi:
on the one hand this could be a useful fix, but technically we probably need mock IAT and mock CLT in simulator by default?

PR #1329, opened 2020-04-19
Prep TS for VSS over CAN

shadowm60:
TS update so we can select input pin for VSS or CAN values:
image


rusefi:

rusefi:

rusefi:

rusefi:

PR #1341, opened 2020-04-22
allow overriding configuration

mck1117:
On hard wired boards, some pins shouldn't be configurable by the user.

For example:
- Hardware disable pins (TLE8888)
- CAN tx/rx
- SPI devices, pinout, CS pins
- Output polarities


firmware/rusefi_rules.mk
```diff
@@ -2,4 +2,4 @@
 RUSEFI_OPT = -Werror
 
 # ...except these few
-RUSEFI_OPT += -Wno-error=unused-function -Wno-error=unused-variable -Wno-error=sign-compare -Wno-error=unused-parameter
+RUSEFI_OPT += -Wno-error=unused-function -Wno-error=unused-variable -Wno-error=sign-compare -Wno-error=unused-parameter -Wno-error=missing-field-initializers
```
rusefi, line 5: "-Wno-error=missing-field-initializers" please no?

dron0gus:
Do we really need one more abstraction like ConfigOverrides struct?
This is what we can #define in individual board.h files.

mck1117:
This isn't what we can define in board.h.  We can define whatever we want, but there's no way using purely the preprocessor to change behavior if something else is defined.

There's no way to do this:
`#define CONFIG_OVERRIDE(__x__) (isdefined(OVERRIDE##__x__) ? OVERRIDE##__x__ : CONFIG(__x__)`

I'd love to do it that way, but it explicitly isn't possible with preprocessor macros.

mck1117:
Now tested on Proteus and MRE (real hardware!) and works great.

PR #1342, opened 2020-04-22
Rewrite boost control in new closed loop framework

mck1117:


rusefi:
@mck1117 poke :)

mck1117:
gave it a quick check on hardware, works fine

PR #1343, opened 2020-04-22
switch ADC to gpt

mck1117:
ADC doesn't need pin outputs, so we can switch it to the basic timer TIM6, where it isn't wasting output pins.  Prep for #1283.


mck1117:
Tested on real hardware, works fine.

PR #1347, opened 2020-04-22
remove warning message in tune

mck1117:
This was causing several issues:
- Every time the ECU reboots, TunerStudio has to do a full read of the configuration (20KB), because the CRC will never match.  If a warning fired in either the previous boot or the current one, the tune (and corresponding CRC) won't match, so it'll have to re-read from the ECU.  This takes ~20 seconds over 115k baud UART, and even longer over Bluetooth, which is not acceptable.
- When TS sends a burn command to the ECU, it then immediately checks the CRC to ensure that the write was successful, and the controller's state matches its internal state.  If a warning has fired since the last full read of the ECU, TS's internal state is now inconsistent with that in the controller, so the CRC check will fail, and TS will display this error message:
image

Many users (myself included) have been concerned that the display of this message indicated that something went wrong (which technically, it had), when in fact the burn had succeeded fine, but the ECU had changed part of the tune (the warning message) unbeknownst to TunerStudio.


rusefi:
Any ideas how can we NOT have the issue but also keep human readable text? I wonder if there are relevant magic attributes for TS.

rusefi:
a different solution for the immediate problem implemented in master

Issue #974, opened 2019-10-13
MapAveragingCallback schedules all cylinders at once


mck1117:
![image](https://user-images.githubusercontent.com/568254/66712464-ced96b80-ed51-11e9-813f-e7ffab2fc7f5.png)

Scheduling (2 * N) callbacks at once seems like a not great idea.  `MapAveragingTriggerCallback` alone took 26us on a release build, but only one per revolution.  This has significant potential to cause jitter.

mck1117:
And a corollary: should there be an API for scheduling a batch of events, without calling `DoExecute` per item (only call it once at the end)?

![image](https://user-images.githubusercontent.com/568254/66712485-195ae800-ed52-11e9-824f-28b7959a76e0.png)

(this is is a zoom-in of one of the 16 enqueue events above)

mck1117:
Some napkin math suggests that the current logic on an F4 can't run a v8 over 4000 RPM without missing interrupts.

rusefi:
Great find, thank you!

rusefi:
Nick reported hanging self-stimulation at 6K with 60/2 once he has enabled one of the two tachometer implementations or something like that.

Maybe it was about this issue really?

mck1117:
Another consideration: We can't necessarily amortize the cost across multiple teeth, since some trigger patterns may not have more teeth, forcing our hand.  However, those triggers are also at less risk of having interrupts stomp on one another.

rusefillc:
@mck1117 poke :)

rusefillc:
@FDSoftware this is done, right?

FDSoftware:
> [@FDSoftware](https://github.com/FDSoftware) this is done, right?

yup, was https://github.com/rusefi/rusefi/pull/8099 / https://github.com/rusefi/rusefi/pull/8034

PR #1352, opened 2020-04-23
MAP scheduling

mck1117:
Schedule MAP end in the MAP start.  This helps amortize the cost of scheduling all those events, instead of scheduling them all in the trigger callback.


mck1117:
Will improve issues observed in #974

mck1117:
Perf tracing essentially matches the theoretical numbers.  On F7, `MapAveragingTriggerCallback` takes 78us without this change, and 41us with this change, a savings of ~37us, saving around 25% of the total trigger handler time.

On F4, the story is similar: 113us -> 58us, delta of -55us.

PR #1358, opened 2020-04-24
trigger emulator refactoring

mck1117:
- Only flip output pins if they're configured (saves ~6us)
- loops instead of hard coding (probably saves little-to-no time, but is good practice)


firmware/controllers/trigger/trigger_emulator_algo.cpp
```diff
@@ -148,7 +149,9 @@ static void emulatorApplyPinState(int stateIndex, PwmConfig *state) /* pwm_gen_c
 	}
 
 #if EFI_PROD_CODE
-	applyPinState(stateIndex, state);
+	if (hasStimPins) {
```
rusefi, line 152: we need code comments explaining the reasons behind new conditional logic

mck1117: I figured the logic
```
if (has_pins) {
    use_pins();
}
```
was pretty straight forward, but sure

PR #1362, opened 2020-04-24
Mre flash

dron0gus:
EFL driver is now used for internal flash
W25Q128 support added to ChibiOS
STM32F407 flash support added to ChibiOS
BaseFlash API used to access both flash storage
persistentState is now in normal RAM


firmware/config/stm32f4ems/efifeatures.h
```diff
@@ -246,7 +246,7 @@
 #endif
 
 #ifndef EFI_ENGINE_SNIFFER
-#define EFI_ENGINE_SNIFFER TRUE
+#define EFI_ENGINE_SNIFFER FALSE
```
mck1117, line 249: we can't turn this off

firmware/config/stm32f4ems/efifeatures.h
```diff
@@ -231,7 +231,7 @@
  * Do we need file logging (like SD card) logic?
  */
 #ifndef EFI_FILE_LOGGING
-#define EFI_FILE_LOGGING TRUE
+#define EFI_FILE_LOGGING FALSE
```
mck1117, line 234: we can't turn this off either

firmware/config/stm32f4ems/efifeatures.h
mck1117: This stuff can probably be in CCM though - I don't think we hit it with DMA

dron0gus: Agree. This is for testing only. STM32F407 is running out of ram. This is temporally solution.

firmware/config/stm32f4ems/efifeatures.h
dron0gus: We need to turn it for MRE at leads. Until we have SD card on MRE. )

firmware/hw_layer/ports/stm32/stm32f4/mpu_util.cpp
```diff
@@ -77,6 +78,40 @@ EXTERNC int getRemainingStack(thread_t *otp) {
 
 #endif /* GNU / IAR */
 
+
+#if (EFI_FLASH_SPI == TRUE)
```
mck1117, line 82: this stuff can probably move to `stm32_common.cpp` - it should work the same on f7

dron0gus: Currently ext flash is enabled (and tested) only on F4. Later we can move it to common files.

firmware/config/boards/proteus/board.mk
```diff
@@ -25,3 +25,6 @@ endif
 
 # Override DEFAULT_ENGINE_TYPE
 DDEFS += $(MCU_DEFS) -DEFI_USE_OSC=TRUE -DEFI_FATAL_ERROR_PIN=GPIOE_3 -DFIRMWARE_ID=\"proteus\" -DDEFAULT_ENGINE_TYPE=PROTEUS -DUSE_ADC3_VBATT_HACK -DSTM32_ADC_USE_ADC3=TRUE -DEFI_INCLUDE_ENGINE_PRESETS=FALSE
+
+#Enable on-chip flash
+DDEFS += -DEFI_INTERNAL_FLASH=TRUE
```
mck1117, line 30: why is this enabled for proteus?

mck1117: oh, wait, this is internal flash.  Should internal flash be the default?

dron0gus: It was always enabled. efifeatures.h in proteus directory includes "../../stm32f4ems/efifeatures.h" where it was enabled.
Now when we have two types of flash. And two drivers for internal flash (ChibiOS backported and RusEFI own)
I see no other way to enable SPI flash for MRE while keep it disabled for all other board that re-use stm32f4ems/efifeatures.h.
So I disable both flash options in common stm32f4ems/efifeatures.h. And enable it for each board in its board.mk.
Not sure if it is correct way. But now there is total mess with configuration files. (efifeatures.h halconf.h and board.mk).

mck1117: Yeah, I agree that it's gotten a bit out of control.  I was just wondering why not have a single define that toggles it, instead of a define to enable each.

dron0gus: There is another issue with RusEFI driver vs ChibiOS driver. RusEFI's supports flash for all SMT32 and Kinetic that were ever used in RusEFI HW.
While ChibiOS supports only 407 and 437.
So we have 4 options:
1. internal flash with RusEFI driver (used if EFI_INTERNAL_FLASH=true, EFI_FLASH_SPI=false and HAL_USE_FLASH=false)
2. internal flash with ChibiOS driver (used if EFI_INTERNAL_FLASH=true, EFI_FLASH_SPI=false and HAL_USE_FLASH=true)
3. SPI flash with ChibiOS driver (used if EFI_FLASH_SPI=true and HAL_USE_FLASH=true,  EFI_INTERNAL_FLASH don't care)
4. no flash at all ( EFI_INTERNAL_FLASH=false, EFI_FLASH_SPI=false)
I was not able to find better way than introduce EFI_FLASH_SPI.
BTW, common settings read/write code in enable by EFI_CONFIGURATION_STORAGE (= (EFI_FLASH_SPI || EFI_INTERNAL_FLASH)

rusefi: can we improve ChibiOS to add support for F7 flash?

mck1117: iirc it's the same, it just needs a `__DSB` before and after the actual flash copy step

mck1117: yep, we're running the same implementation on both, with this extra on the f7:
```
	// Cortex-M7 (STM32F7/H7) can execute out order - need to force a full flush
	// so that we actually wait for the operation to complete!
#if CORTEX_MODEL == 7
	__DSB();
#endif
```

rusefillc:
@dron0gus https://github.com/rusefi/rusefi/commit/217546b02dc73760adbd317079c9d60b24b792c2

rusefillc:
@dron0gus config for legacy MRE https://github.com/rusefi/rusefi/commit/c305a572af7b7cc3baa6d3636154ae0eb4321ce7

dron0gus:
Closed because of https://github.com/rusefi/rusefi/pull/6753

PR #1364, opened 2020-04-25
Update chibios

dron0gus:
- we need backported flash support


firmware/bootloader/src/Makefile
```diff
@@ -14,6 +14,8 @@ endif
 
 # disable some modules to shrink bootloader binary
 DDEFS += -DHAL_USE_ADC=FALSE -DHAL_USE_CAN=FALSE -DHAL_USE_EXT=FALSE -DHAL_USE_GPT=FALSE -DHAL_USE_I2C=FALSE -DHAL_USE_ICU=FALSE -DHAL_USE_PWM=FALSE -DHAL_USE_RTC=FALSE -DHAL_USE_I2C=FALSE 
+#disable ChibiOS flsah driver and prevent header from include
+DDEFS += -DHAL_USE_FLASH=FASLE
```
rusefi, line 18: FASLE typo?
flsah typo?

firmware/Makefile
```diff
@@ -127,6 +127,20 @@ MAKEFLAGS += ${NUMJOBS}
...
 #
 
+BOARDS_DIR = $(PROJECT_DIR)/config/boards
+
+include $(PROJECT_DIR)/config/boards/$(PROJECT_BOARD)/board.mk
+
+# If enabled, this option makes the build process faster by not compiling
+# modules not used in the current configuration.
+# without USE_SMART_BUILD all ChibiOS (including all drivers) are builded. And all drivers includes get included.
+ifeq ($(USE_SMART_BUILD),)
```
rusefi, line 137: let's split this PR into two smaller PRs? can we take care of makefile changes & bootloader maybe separately before flash changed?

the fact that we have failures here is an indication that we have too many changes in one PR

firmware/bootloader/src/Makefile
dron0gus: I don't know how I write code with such fingers.

PR #1374, opened 2020-04-27
Move ETB control to a thread

mck1117:
- Move ETB to its own thread, because:
    - Auto TPS calibrate is greatly simplified by the use of `chThdSleep`
    - Things running on a timer (instead of a thread) don't participate it the RTOS's thread priority system, and operate essentially "first come first serve", which risks starvation.  Since ETB is a safety critical device, we need the hard RTOS guarantee that it will be scheduled over other less important tasks.
- Clean up ETB defaults
- Hard code ETB update speed.  Since this is a safety critical system with no real reason for a user to ever need to change the update rate, it's now locked to 500hz, along with the ADC.  The increase from 200 to 500 gives a considerable increase in loop bandwidth and responsiveness.


mck1117:
Ready to go, tested on real HW with a real throttle.

rusefi:
here we have a case of talking to yourself in pull requests which should have been talking to yourself in code comments. Some stuff pasted into codebase. https://github.com/rusefi/rusefi/commit/44facf11401aefa5e8449d6c568dedbe19745a72

PR #1375, opened 2020-04-27
fix master

mck1117:
This is the kind of nasty runtime pin remapping logic I want to avoid.  Nobody should ever have to change CAN pins on the fly.  It should be hard coded for a particular board, so by definition it means nothing to reconfigure.

### DIFF
--- a/firmware/hw_layer/drivers/can/can_hw.cpp
+++ b/firmware/hw_layer/drivers/can/can_hw.cpp
@@ -139,6 +139,12 @@ void postCanState(TunerStudioOutputChannels *tsOutputChannels) {
 }
 #endif /* EFI_TUNER_STUDIO */
 
+void enableFrankensoCan(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
+	CONFIG(canTxPin) = GPIOB_6;
+	CONFIG(canRxPin) = GPIOB_12;
+	engineConfiguration->canReadEnabled = false;
+}
+
 static brain_pin_e currentTxPin = GPIO_UNASSIGNED;
 static brain_pin_e currentRxPin = GPIO_UNASSIGNED;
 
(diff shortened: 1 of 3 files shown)


rusefi:
Please elaborate how is this fixing master?

I am sure that CAN remapping can and should be flexible :( for me a board restart is a waste of time. Let's get to the bottom of what is the root cause of flexibility affecting you negatively?

mck1117:
The problem was that if pins were overridden, it would try to unmap the pins passed in from `activeConfiguration`, which were set to `GPIO_UNASSIGNED`, because they were never set in the first place.  Then it would try to remap the pins, which wouldn't work, because they never got unset.

I guess the other option is to store which pins we *actually* used, then ignore whatever is in `activeConfiguration`, and unmap what we know we mapped earlier.

rusefi:
I'm sorry I do not follow :(

can you please provide steps to reproduce the problem?

is this related to the new "ConfigOverrides" feature?

mck1117:
It requires an MRE or Proteus build to repro, because it is indeed related to ConfigOverrides.

If you write the config, you get a critical error like this:

image

Note how a pin is trying to occupy itself.  In `stopCanPins`, we were trying to clear the pins set in `activeConfiguration`, but that was set to `GPIO_UNASSIGNED`, so it was a noop.  When we went to re-set the pin, it was read out of the overrides, so we tried to reallocate the pin we're already set.

The solutions are either:
- Don't reconfigure pins upon burn (as is implemented now in this PR)
- Keep track of which pin we actually set in `startCanPins`, instead of trying to clear the pin from `activeConfiguration`

rusefi:
looks like an urgent issue but killing old working can refresh due to an urgent issues caused by new code is not the way to go.

I believe we have to revert new feature (configoverride) until we can have it working without urgent issues.

mck1117:
I think the latter is a totally fine solution - it will preserve existing behavior

PR #1378, opened 2020-04-27
proteus -> exti

mck1117:
Use EXTI for trigger inputs on Proteus, since it's required for anything other than digital input pin 1 to work.


rusefi:
@mck1117 poke

mck1117:
Just ran an engine on this change. Works!

PR #1383, opened 2020-04-29
Perf Trace Improvements

mck1117:
- Store a unique ID with every thread, so that perf tracing can distinguish threads
- Trace ISR number and thread ID
- Render ISR number and thread ID in Java converter
- Update process (ISR) names, so they display in `chrome://tracing`
- Consolidate `chconf.h` thread hooks in to `chconf_common.h`

Example:
pic1


rusefi:
i think it wants unused RAM size to change?

mck1117:
and a stub for bootloader, should be good now

PR #1402, opened 2020-05-06
short term fuel trim: part 1

mck1117:


rusefi:
code with test coverage, i love it :)

mck1117:
It's even test coverage that isn't tightly coupled to anything else on the engine working right :)

PR #1404, opened 2020-05-07
Tle8888 stage 1

dron0gus:
PP code is hidden inder #if 0, so it should not creak anything.
Diagnostic code tested.
Looks safe.


rusefi:
is this change compatible with watchdog?

dron0gus:
Yes. This does not affect WD code

PR #1407, opened 2020-05-08
tle8888: on MRE 0.5.0 we can drive ING_EN and INJ_EN from STM

dron0gus:
Do it!


firmware/hw_layer/smart_gpio.cpp
```diff
@@ -131,6 +131,8 @@ struct tle8888_config tle8888_cfg = {
 		[3] = {.port = NULL,	.pad = 0,	.output = 12},
 */
 	},
+	.ign_en =  {.port = GPIOD,	.pad = 10},
```
rusefi, line 134: we need to make this less hard-codes

PR #1410, opened 2020-05-09
Update MRE pin definitions

mck1117:
This also includes a hard break to TS signature and flash config version.  ECUs will wipe their config, then get re-configured from TS using the new ini.


rusefi:
I've stolen a piece of this change into master

mck1117:
This change now also defragments ~350 bytes in to the big unused chunk, out of small chunks.  I didn't do any single bytes yet, since those require a little more thought to get the alignment right.

mck1117:
this is covered by #1892

PR #1411, opened 2020-05-09
remove TS_FILE_VERSION

mck1117:
TS_SIGNATURE covers this need - no need for both


firmware/console/status_loop.cpp
```diff
@@ -689,9 +689,6 @@ void updateTunerStudioState(TunerStudioOutputChannels *tsOutputChannels DECLARE_
 
 	float engineLoad = getEngineLoadT(PASS_ENGINE_PARAMETER_SIGNATURE);
 
-	// header
-	tsOutputChannels->tsConfigVersion = TS_FILE_VERSION;
```
rusefi, line 693: cannot do this yet due to console. need to document console dependency

mck1117: reduced to just ini changes for now

PR #1419, opened 2020-05-10
configuration for secondary (redundant) TPS sensors

mck1117:
progress on #1240


rusefi:
we had merge conflict here and I would prefer not to break master so I've picked pieces of the change into master

please rebase

PR #1421, opened 2020-05-11
Mc33 undervolt

NOx-z:
Apparently the MC33816 is very picky about it's 5v line, and we shouldn't go under 4.75v .. which is difficult when we get ~4.5v from the Discovery board, so we need to wire off a good regulated 5v source


rusefi:

rusefi:

rusefi:

rusefi:

rusefi:
add to whitelist

PR #1427, opened 2020-05-13
Redundant tps

mck1117:
Progress on #1240.

This change adds second TPS sensors (so now there's primary and secondary for each throttle).  If the resolved position from the two sensors differs by more than a threshold, the TPS is marked as failed.


mck1117:
tested on real hardware, including auto calibrate!

PR #1431, opened 2020-05-14
Fix 100% duty injection

### LINKED ISSUE #1430: Properly handle 100% injector duty
currently when near or over 100% injector duty, chaos results. 

Here's an example log:

image

We just barely tap 100%, then the AFR goes super lean, like injections are getting completely skipped.

The correct behavior is to do what all other ECUs do, which is pin the injectors fully open while the duty cycle is over 100%, returning to normal behavior when the requested duty drops.

mck1117:
If you hit 100% duty cycle, we used to skip every other injection, resulting in Major Misbehavior.

~~This change makes the behavior a bit safer.  If we are operating with very high duty cycle (>95%), don't bother closing the injectors at all. Pin them fully open, so you get everything your fuel system has to give, instead of half.~~

If duty cycle gets near 100%, we cut fuel (like a rev limit).  This is safer than trying to pin the injectors open at 100%, because it can't let you quietly go lean and melt a piston.

Fixes #1430.


mck1117:
After some discussion with @OrchardPerformance, we decided it's best to actually impose a hard rev limiter when this happens, instead of trying to limp along.  If you try to limp at 100% duty, you risk going slowly (and quietly) lean, then melting a piston or something.  If you cut fuel altogether, the driver/user/tuner will surely notice, which is what you want in this situation.

mck1117:
Validated on hardware to properly (fully) cut fuel when duty cycle is too high.

PR #1439, opened 2020-05-16
dead datalog code

mck1117:
dead code cleanup


mck1117:
I just confirmed that this change doesn't actually change the compiled binary

PR #1443, opened 2020-05-17
binary logging

mck1117:
log to binary mlg format, instead of text format: part 1, log format & test


mck1117:
Run on PC, this has generated a log file that megalogviewer opened successfully.

PR #1445, opened 2020-05-17
Mc33 vccp uv

NOx-z:
Includes better checks for faults such as if VccP is bad before and after DRIVEN.. if after the dc-dc converter fets inchip are likely dead and a new MC33 is needed.

Adds a config to re-check the driver status register on read

Adds a slightly better shutdown sequence for the MC33


firmware/hw_layer/mc33816_data.c
```diff
@@ -51,7 +51,7 @@ const unsigned short MC33816_data_RAM[128] =
 const unsigned short MC33816_main_config[29] =
 {
     0x0003, 0x1FFE, 0x0000, 0x1200, 0x0000, 0x0000, 0x0001, 0x0000, 0x001F, 0x0000,
-    0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000,
+    0x0000, 0x0000, 0x0000, 0x0000, 0x0001, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000,
```
mck1117, line 54: what does setting this bit do?

NOx-z: Reset_behavior register (0x1CE)
bit 0: Driver enable reset behavior (driver_enable_rb): if set to ‘1’ Driver_status register is reset on read.

mck1117: Ah, neat.

PR #1453, opened 2020-05-19
Mc33 timings

NOx-z:
new configs and functions for calculating actual values for the mc33 timing


rusefi:
Please merge master and your new fields should be there

PR #1458, opened 2020-05-20
Remove old thermistor implementation

mck1117:
These have been fully replaced with new sensors 🎉 🎉 🎉


firmware/console/binary/tunerstudio.cpp
```diff
@@ -274,10 +274,6 @@ static void onlineApplyWorkingCopyBytes(uint32_t offset, int count) {
 
 static const void * getStructAddr(int structId) {
 	switch (structId) {
-	case LDS_CLT_STATE_INDEX:
```
mck1117, line 277: @rusefi let's chat about this

PR #1461, opened 2020-05-21
Binary logging to SD

mck1117:
Switch from text logging to binary logging.


rusefi:
location counter backwards (from 0000000020020368 to 0000000020020000)

PR #1470, opened 2020-05-23
Redundant accelerator pedal

mck1117:
progress on #1240


mck1117:
pending merge & config generate of #1469

Issue #1299, opened 2020-04-15
Updates to fuel algorithms 


OrchardPerformance:
Creating tickets to track the improvements to the fuel algorithms 

Link to the roadmap - https://github.com/rusefi/rusefi/wiki/Roadmap_Fuel

1- Fix LM_REAL_MAF 
This means:
a) Modifying firmware so airChargeLoad is used as the value for "engine load" in TS
b) Checking the Y axis of the fuel table uses this value 
c) Checking the Ignition table axis uses this value
d) I will put some miles on it and grab some logs so we know its sorted.

2- Deprecating LM_PLAIN_MAF, 
Not sure what this needs, after discussion with Matt there is no need for it to exist once the tasks above are done. 
Belief is that at present no users are actually using this? 

3- Move AlphaN to VE fuel table.
This is a change to move from a ms pulse width fuel table to a VE table like speed density. This is how other systems work with their AlphaN mode so is a usability improvement.

OrchardPerformance:
125JDavis said - Alpha-N fueling algorithm doesn't have VE table or target AFR table. Currently the user tunes by changing target PW values in the TPS/RPM table. Useability would be much better if the same VE/AFR fueling tables from speed density were carried over to Alpha-N.
Updated equation for O2 mass = TPS * Baro Pressure * Displacement / (Gas Constant * Intake Temp)
![82359006-53215a80-99d5-11ea-8b29-fb76f35c87d7](https://user-images.githubusercontent.com/49926376/82692048-e3220700-9c56-11ea-8767-42e3daed90c7.jpg)

mck1117 said - Instead of using TPS directly, I think you want a "VE" value looked up from a table as a function of TPS and RPM. So essentially the same as the speed density math, but with pressure removed, and VE looked up using TPS instead of MAP.

OrchardPerformance:
I also think this ticket needs bumping up the order as we have PNP boards needing the MAF and more people looking at AlphaN

OrchardPerformance:
Roadmap linked above. 

Some comments on Ethanol flex fuel implementation here for #540 

Density is currently used only in:
constexpr float convertToGramsPerSecond(float ccPerMinute) {
	float ccPerSecond = ccPerMinute / 60;
	return ccPerSecond * 0.72f;	// 0.72g/cc fuel density }

Suggest this is broken out into a "fuel_density.cpp" where we can modify this by ethanol percentage and temperature. 

[...]

PR #1481, opened 2020-06-04
extract airmass -> fuel mass math

mck1117:
this needs to get reused for other modes, not just for speed density


mck1117:
progress on #1299

PR #1485, opened 2020-06-10
fuel math simplifications

mck1117:
making things easier for #1299


mck1117:
tested on hardware, ready to go!

PR #1489, opened 2020-06-13
only setup etb if primary sensor configured instead of proxy

mck1117:


rusefi:
@mck1117 unit tests not happy?

mck1117:
Should be ready to go now, pending green check marks!

PR #1491, opened 2020-06-14
spinwait for events in the near future

mck1117:
This change improves scheduling precision for events that happen near one another.  If two events are very close together, there isn't any sense in exiting the interrupt only to be immediately interrupted again.  Instead, we just spinwait until the event is ready to go.


rusefi:
@mck1117 when is this happening - is this when two different sources of signal have signals close but not at the same time, like PWM and trigger simulation?

as for the improvement, how about we change scheduling phase: if two events on the queue with times only 10ns apart, set alarm clock for SECOND time not for the SOONEST event. Would we get same result without the busy wait?

Also technically not "second" but "once gaps exceed 10ns" since theoretically we can have a series of events shushed together

Also funny thing about current implementation - do we have to handle 3rd signal 8 more ns later, i.e. does current implementation need a while (true) loop for looking a bit into the future?

firmware/controllers/system/timer/event_queue.cpp
```diff
@@ -100,10 +95,8 @@ efitime_t EventQueue::getNextEventTime(efitime_t nowX) const {
 			 *
 			 * looks like we end up here after 'writeconfig' (which freezes the firmware) - we are late
 			 * for the next scheduled event
-			 *
 			 */
-			efitime_t aBitInTheFuture = nowX + lateDelay;
-			return aBitInTheFuture;
+			return nowX + US2NT(10);
```
rusefi, line 99: no no no, event_queue.cpp operates in univeral X units not NT units

mck1117: Hmm.  In that case, the added `while (current->momentX > getTimeNowNt()) ;` is then technically invalid too.

rusefi: line 87 " * On this layer it does not matter which units are used - us, ms ot nt."

PR #1495, opened 2020-06-16
Fix #1492

### LINKED ISSUE #1492: code generator - reduce the gap between Windows and Linux
at the moment we have to manually sync gen_config.bat and gen_config.sh since all the parameters are hard-coded inside those platform-specific scripts

@ksmola got hit by that when .sh got outdated around firing order parameter

a proper fix would be to extract parameters into input file(s) so that one file could be modified and both scripts for both platforms would still work fine

chuckwagoncomputing:
All the parameters to be passed to java are now in gen_config.conf.
gen_config.bat just calls gen_config.sh.
It checks whether C:\cygwin or C:\cygwin64 exists. Are there any situation in which one of these would not be true?

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -99,7 +99,24 @@ jobs:
       working-directory: ./firmware/
       run: make -j4 PROJECT_BOARD=${{matrix.efi-board}} PROJECT_CPU=${{matrix.efi-cpu}} EXTRA_PARAMS="${{matrix.type-extra-params}} ${{matrix.target-extra-params}}" DEBUG_LEVEL_OPT='${{matrix.build-debug-level-opt}}' ${{matrix.extra-options}}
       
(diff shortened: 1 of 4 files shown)


rusefi:

rusefi:

rusefi:

rusefi:

rusefi:
add to whilelist

rusefi:
at first glance it looks like we have to unrelated changes mixed here - I see GH actions which is one improvement and then something for gen_config which is a totally separate area? Please clarify if that's the case or not

two separate changes would have to be two separate PRs

rusefi:
add to white list

rusefi:
it looks like the root cause of the issue is the fact that you are PRing your master

it looks like a proper process would be to have branches with one change per one branch. this way you PR your specific dev branch and you can keep other activities on separate branches.

chuckwagoncomputing:
Yep, that's what I did. I'll close this and start over.

PR #1496, opened 2020-06-16
Fix #1492 redux

### LINKED ISSUE #1492: code generator - reduce the gap between Windows and Linux

chuckwagoncomputing:
All the parameters to be passed to java are now in gen_config.conf.
gen_config.bat just calls gen_config.sh.
It checks whether C:\cygwin or C:\cygwin64 exists. Are there any situation in which one of these would not be true?

### DIFF
--- a/firmware/gen_config.bat
+++ b/firmware/gen_config.bat
@@ -1,63 +1,6 @@
 @echo off
-rem This batch files reads rusefi_config.txt and produses firmware persistent configuration headers
-rem the storage section of rusefi.ini is updated as well
-
-rm gen_config.log
-rm gen_config_board.log
-
-rem lazy is broken - TS input is not considered a change
-rm build/config.gen
-
-mkdir build
-
-java -DSystemOut.name=gen_config ^
- -Drusefi.generator.lazyfile.enabled=true ^
- -jar ../java_tools/ConfigDefinition.jar ^
- -definition integration\rusefi_config.txt ^
- -romraider integration ^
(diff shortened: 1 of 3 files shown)


rusefi:

rusefi:

rusefi:

rusefi:

rusefi:
I like where you are going with this!

Shall we simply assume that cygwin is on PATH? just invoke "sh.exe" without any hard-coded path drama?

rusefi:
@chuckwagoncomputing this was the result

image

rusefi:
https://github.com/rusefi/rusefi/commit/22780edcb38c1f9cd61df390310ee880a76c70b9 should help now I need to delete all .sh locally and reset from git

rusefi:
there is progress but more stuff is broken

image


@chuckwagoncomputing this is about .sh invoking "-cp ../java_tools/ConfigDefinition.jar:../java_tools/configuration_definition/lib/snakeyaml.jar" with colon not semicolor

As a work around I would propose that we package snakeyaml.jar into ConfigDefinition.jar so that we only have to access one jar file

also next tiny improvement would be to migrate gen_config_board.bat to same style

PR #1500, opened 2020-06-16
Move gcov to Github Actions

chuckwagoncomputing:


chuckwagoncomputing:
I ran this on my fork with my own FTP server, and it worked. The error message is that one of the server details secret is missing.

PR #1504, opened 2020-06-17
Workflow to write the date, once a day

### LINKED ISSUE #1415: firmware date in git - automation
we need to update firmware build date automatically - at the moment 'getRusEfiVersion' hard-coded date is modified manually

chuckwagoncomputing:
This should fix #1415


.github/workflows/set-date.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Write Date File
+      run: |
+        echo -e -n "#ifndef VCS_DATE\n#define VCS_DATE $(date "+%Y%m%d")\n#endif" > ./firmware/date_stamp.h
+        echo -e -n "#ifndef VCS_DATE\n#define VCS_DATE $(date "+%Y%m%d")\n#endif" > ./simulator/date_stamp.h
```
rusefi, line 19: why three separate files/dates? let's have one file for all consumers?

chuckwagoncomputing: It was easier.

chuckwagoncomputing: If you'd prefer to have one file, your wish is my command.

rusefi: Let's have one single file next to engine_controller.h

chuckwagoncomputing: In the same directory?

chuckwagoncomputing: Maybe the util directory would be better?

.github/workflows/set-date.yaml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Write Date File
+      run: |
+        echo -e -n "#ifndef VCS_DATE\n#define VCS_DATE $(date "+%Y%m%d")\n#endif" > ./firmware/controllers/date_stamp.h
```
rusefi, line 18: minor simplification - can we use "#pragma once" here? ifndef VCS_DATE hints at a possibility to externally define it which we do not need. "#pragma once" would close that door thus making the thing overall easier?

.github/workflows/set-date.yaml
```diff
@@ -0,0 +1,27 @@
+name: Set Date
+
+on:
+  schedule:
+    - cron: '* 0 * * *'
```
rusefi, line 5: Please add a comment into the action explaining the logic.

Is it currently just an unconditional daily change? The desire is to have it conditional on other changes in repo. I.e. while all collaboraters are on vacation and no one touches repo, automation should NOT change repo.

chuckwagoncomputing: OK, I will add a comment. Yes, that was @mck1117 's suggestion.
What are the downsides to this changing the repo without there being any changes? It will not trigger a build.

.github/workflows/set-date.yaml
rusefi: Any directory zero difference to me, as long as one file and it compiles :)

.github/workflows/set-date.yaml
rusefi: why would it not trigger a build?
it's a git push, i would expect other actions to pick it up?

chuckwagoncomputing: From https://help.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token
"When you use the repository's GITHUB_TOKEN to perform tasks on behalf of the GitHub Actions app, events triggered by the GITHUB_TOKEN will not create a new workflow run."
This does in fact apply in our case.

chuckwagoncomputing: If we run it on schedule, but only change the date stamp if the repo has changed, The date could be behind by up to a day.

We can't really try to change it the way workflows are set up now, because there is no way to control which workflows run first, so some builds will not have the updated date and will lag behind by one build.

rusefi:
Is this now functional in any way? I am comfortable to merge as is. This is definitely a great progress and we can always improve on this later if any drama.

rusefi:
PS1: what about #pragma once?
PS2: please copy-paste relevant parts of this discussion into GH action source code comments to have a readily available overview of how this works and the URL of the article explaining why no builds as result.

chuckwagoncomputing:
There ya go. Hopefully this saves you more time than you spent helping me figure out how this should be done. :)

rusefi:
merged & changed a bit

PR #1506, opened 2020-06-18
Run gen scripts and unify firmware, simulator, and unit_tests workflows

### LINKED ISSUE #1502: GH actions to invoke gen_config
At the moment we have Windows build server invoking gen_config_and_push.bat from https://github.com/rusefi/rusefi/tree/master/misc/jenkins which produces commits like https://github.com/rusefi/rusefi/commit/4460785120a4e2e3d59c7542ca1e93d34d5f134b

Where are two problems with that
1) Windows build box is under permanent stress especially after a push into GH
2) this requires manual intervention since github authentication expires or SOMETHING - current solution is not 100% reliable

Can we make this better by adding a GH action to do same?

chuckwagoncomputing:
This should fix #1502.
Putting firmware, simulator, and unit test builds all in one workflow is necessary to ensure everything happens in the correct order.


mck1117:
> ensure everything happens in the correct order.

What does this mean exactly? All builds should be a pure function of the current state of the repo, and therefore execution order can't matter.

What are we trying to solve here?

.github/workflows/build-firmware.yaml
```diff
@@ -3,8 +3,54 @@ name: Firmware CI
...
+      run: sh gen_live_documentation.sh
+
+    # Commit changes and catch the error that occurs if nothing has been changed (without catching other errors)
+    - name: Commit changes
+      if: ${{ github.event.pusher }}
+      run: |
+        echo ${{ github.event.pusher }}
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub build-firmware Action"
+        git commit -am "Auto-generated configs and docs" 2>&1 | grep -E '(nothing to commit|changed)'
```
mck1117, line 35: we cannot have builds that commit over the top of your pull request - this sort of thing should only run on push to master

.github/workflows/build-firmware.yaml
```diff
@@ -3,8 +3,54 @@ name: Firmware CI
...
+    - uses: actions/checkout@v2
+      # We can tell if this is a push rather than a pull request by testing for the existence of the 'pusher' object
+      # We have to do this on each step in this job rather than the whole job, or everything that needs this job will be skipped.
+      if: ${{ github.event.pusher }}
+      # Because we are making changes, we need to fetch all history
+      with:
+        fetch-depth: 0
+
+    - name: Generate Configs
+      if: ${{ github.event.pusher }}
```
mck1117, line 19: PRs still need to generate config, just not commit it

.github/workflows/build-firmware.yaml
chuckwagoncomputing: This will only run if it's a push.

mck1117: Yes - it will commit over the top of your PR branch when you push to it.  We need to restrict commits to only the rusefi/master branch, not mck1117/whatever-feature-27

mck1117: Overzealous auto-commits are how you pollute the commit history with a thousand merge commits when the action has added new commits while you were still working locally

.github/workflows/build-firmware.yaml
mck1117: Also if we always generate config before building, then we don't care about the execution order.  The generated files can be stale, but that's fine since they'll get generated before building.

chuckwagoncomputing: Guess that's true. Any idea how to do that? I guess the checkout would need to be shared across jobs, but I don't think that works.

mck1117: Just skip the commit/push on PR.  So remove line 19, but leave the commit/push logic as-is.

.github/workflows/build-firmware.yaml
chuckwagoncomputing: So only commit if the push is to master?

.github/workflows/build-firmware.yaml
chuckwagoncomputing: The repo is checked out again in the build jobs, so changes made in this build job won't exist in those jobs.

mck1117: unless you add a generate step as part of the firmware/test builds

.github/workflows/build-firmware.yaml
mck1117: Yep!  Though we have to check that it's the "real" master, and not a fork's master.

rusefi:
Can we use presence of GH secrets to know the context of execution? have secrets - real master, real build. No secrets - no push?

mck1117:
there's probably a nicer way to do it, but at the very least you can run `git remote -v` to show URLs for all remotes currently configured (which will show what repo it is)

chuckwagoncomputing:
All that information is provided us in the 'github' object.
The problem here is making sure firmware, simulator, and unit_tests have access to the generated config, without actually pushing the changes if it's just a pull request.

mck1117:
is there anything stopping us from doing the generate as a step at the beginning of every build?

chuckwagoncomputing:
We could certainly do that.
Maybe I have to strong of an aversion to duplicated code.
I'm the one that's new at this, so whatever you think is best, I will gladly do.

mck1117:
I think you can call a whole action as a step: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses

mck1117:
here: https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#referencing-actions-in-your-workflow

chuckwagoncomputing:
I looked into that earlier. Actions either run a shell script in a docker image, or a javascript file. A shell script could be called from the javascript.

If we run the generation for every build task, then including them all in the same workflow wouldn't be necessary for our purposes here. However, I think this will come up again, particularly when we move to creating bundles from a Github Action.

chuckwagoncomputing:
Now the workflows that build firmware, simulator, and unit tests generate the configs and live docs themselves. Some code is duplicated, but I am resigned to that as it seems like the simplest way of doing it.
Committal of the updated configs and live docs has been moved to it's own workflow which will only run on push to master.

chuckwagoncomputing:
The checks failed because apt-get could not reach the Ubuntu package server. I don't think that had anything to do with this change.

chuckwagoncomputing:
I re-ran those jobs on my fork and they completed successfully.

PR #1507, opened 2020-06-18
Remove duplicated workflow

chuckwagoncomputing:
The workflows for console with Java version 8 and the one intended for Java 11 are identical. Might as well get rid of the 11 version for now.


firmware/controllers/date_stamp.h
```diff
@@ -1,3 +1,2 @@
-#ifndef VCS_DATE
-#define VCS_DATE 20200617
-#endif
\ No newline at end of file
+#pragma once
+#define VCS_DATE 20200618
```
mck1117, line 2: is this change on purpose?

chuckwagoncomputing: nope. set_date ran and I didn't notice.

chuckwagoncomputing: Won't hurt a thing though.

mck1117: did it run on your branch? perhaps it should only run when on real rusefi/master

mck1117: yep looks like that's exactly what happened

chuckwagoncomputing: good idea. I'll take care of that.

PR #1508, opened 2020-06-18
Fix set date

chuckwagoncomputing:
This should make it so the set_date workflow only runs on the rusefi/rusefi repo, not forks


chuckwagoncomputing:
Oops messed it up again

PR #1530, opened 2020-06-22
Gpio cleanup #2

dron0gus:
- size of gpio users array is now equal to brain_pin_e (minus special values). BOARD_EXT_PINREPOPINS option removed.
- magic math to convert brain_pin_e to index is removed (anyway it was wrong as "#ifdef \<enum element\>" is not working)
- gpio chips are now registered to given gpio base. No dynamic allocation and no magic gpiochip_use_gpio_base
- some function renames to be more clear.


mck1117:
summary of changes?

dron0gus:
> summary of changes?

Added.
Still draft.

dron0gus:
Ready for review.

rusefillc:
@mck1117 @andreika-git

mck1117:
28 files changed....ouch

firmware/config/boards/hellen/cypress/rusefi_hw_enums.h
```diff
@@ -234,7 +232,10 @@ typedef enum __attribute__ ((__packed__)) {
 } brain_pin_e;
 
 /* Plase keep updating this define */
-#define BRAIN_PIN_LAST_ONCHIP	GPIOK_15
+#define BRAIN_PIN_ONCHIP_LAST	GPIOK_15
+#define BRAIN_PIN_ONCHIP_PINS	(BRAIN_PIN_ONCHIP_LAST - GPIOA_0 + 1)
+#define BRAIN_PIN_LAST 			DRV8860_PIN_16
+#define BRAIN_PIN_TOTAL_PINS	(BRAIN_PIN_LAST - GPIOA_0 + 1)
```
mck1117, line 238: can these be values within the enum? remember multiple elements can share the same value if necessary.

dron0gus: Yes, but no.
AFAIR this header is parsed and used to generate auto_generated_enums.cpp and switch case inside. So we need to modify parser to support enum elements with same value...

PR #1541, opened 2020-06-27
gen_config is no more lazy

andreika-git:


andreika-git:
Requires running gen_config :)

PR #1554, opened 2020-07-01
Convert jenkins build_extra_bundles scripts to sh

chuckwagoncomputing:


chuckwagoncomputing:
This does not include switching Github actions to using all the same scripts.

rusefi:
this change is massive I have no idea how to review it

would it be annoying to separate a subset of changes as a separate smaller change?

chuckwagoncomputing:
That would be possible, but kind of tough as everything kind of works together. Don't be intimidated by the number of commits, most of those were fixes to bugs I introduced.

PR #1566, opened 2020-07-04
fix maf gauge scaling

mck1117:
Previously only supported up to 655.35kg/h of airflow, which is only about 400hp worth of air.  Now we support 4000hp worth of air!

### DIFF
--- a/firmware/console/binary/tunerstudio_outputs.h
+++ b/firmware/console/binary/tunerstudio_outputs.h
@@ -95,7 +95,7 @@ typedef struct {
 
 	// air flow/mass measurment
 	scaled_voltage massAirFlowVoltage; // 26
-	scaled_channel<uint16_t, 10> massAirFlow; // 28
+	scaled_channel<uint16_t, PACK_MULT_MASS_FLOW> massAirFlow; // 28
 	scaled_pressure manifoldAirPressure; // 30
 	scaled_pressure baroPressure; // 32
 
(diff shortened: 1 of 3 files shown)


firmware/console/binary/tunerstudio_outputs.h
```diff
@@ -95,7 +95,7 @@ typedef struct {
 
 	// air flow/mass measurment
 	scaled_voltage massAirFlowVoltage; // 26
-	scaled_channel<uint16_t, 100> massAirFlow; // 28
+	scaled_channel<uint16_t, 10> massAirFlow; // 28
```
rusefi, line 98: can we please extract this "10" magic constant into rusefi...txt for future reuse by console?

firmware/tunerstudio/rusefi.input
```diff
@@ -233,7 +233,7 @@ enable2ndByteCanID = false
 
 ; air flow/mass measurments
       MAFValue        = scalar,  U16,     26,        "V",,{1/@@PACK_MULT_VOLTAGE@@},,         0
-      massAirFlowValue= scalar,  U16,     28,     "Kg/h", 0.01,         0
+      massAirFlowValue= scalar,  U16,     28,     "Kg/h", 0.1,         0
```
rusefi, line 236: please do 1 / MAGIC_10_CONSTANT

firmware/console/binary/tunerstudio_outputs.h
mck1117: let's see if PR auto-generate for builds works!

mck1117: ooh, it did!

PR #1568, opened 2020-07-04
 Add ign & fuel load fields

mck1117:
Beam back the values to use on the Y axes of fuel & ignition tables, progress on #1299


firmware/console/binary/tunerstudio_outputs.h
```diff
@@ -181,17 +181,22 @@ typedef struct {
 	TsCalMode calibrationMode; // 132
 	uint8_t padding[1]; // 133
 
+	// Values used for load axes for fuel/ign tables
+	// These may or may not be the same value, depending on mode
```
rusefi, line 185: i see no change to rusEFI console which at the moment has all these offsets hard-coded

would it be better to add new fields into the end of the struct?

mck1117: it's a draft pr 🙂 

These can't be too far from the beginning because TS will read them every time, and we want them in the big contiguous block at the beginning

PR #1572, opened 2020-07-05
fix perf trace

mck1117:


rusefi:
we should have added a unit test?

PR #1573, opened 2020-07-05
plumb new load channels

mck1117:
hook up the new load fields for split fuel/ignition loads, progress on #1299


mck1117:
This change doesn't yet change the behavior of those tables (either in FW or TS), but does hook up all the prerequisites.

PR #1618, opened 2020-07-19
Upload build artifacts

### LINKED ISSUE #1613: PR builds compile bundles but does not persist artifacts
PR builds like https://github.com/rusefi/rusefi/runs/884106600 could and should attach bundles as artifacts

chuckwagoncomputing:
This should fix #1613

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -135,21 +135,21 @@ jobs:
 
     - name: Upload build bin
       if: ${{ github.event_name != 'push' || github.ref != 'refs/heads/master' }}
-      uses: actions/upload-artifacts@v2
+      uses: actions/upload-artifact@v2
       with:
         name: rusefi_${{matrix.build-target}}.bin
         path: firmware/deliver/rusefi_${{matrix.build-target}}.bin
 
     - name: Upload build hex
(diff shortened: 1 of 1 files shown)


.github/workflows/build-firmware.yaml
```diff
@@ -133,6 +133,27 @@ jobs:
...
 
+    - name: Upload build bin
+      if: ${{ github.event_name != 'push' || github.ref != 'refs/heads/master' }}
+      uses: actions/upload-artifact@v2
+      with:
+        name: rusefi_${{matrix.build-target}}.bin
+        path: ./firmware/deliver/rusefi_${{matrix.build-target}}.bin
+
+    - name: Upload build hex
+      if: ${{ github.event_name != 'push' || github.ref != 'refs/heads/master' }}
```
rusefi, line 144: copy-paste? can we squash three path uploads int one?

i was thinking to upload the whole bundle as is, not just individual files since we seem to be building the bundle? i could be wrong you have better context

chuckwagoncomputing: Nope, not building the bundles on a PR. In order to build the bundles, we would have to build simulator and console as well, which we moved away from to keep from making firmware build fail when console or simulator builds failed.

PR #1625, opened 2020-07-21
Export logicdata on all tests

mck1117:
automagically export logicdata file for every test


mck1117:
As a bonus, this will still export logicdata even in case of a failed assertion or thrown exception, since destructors always run.

PR #1628, opened 2020-07-22
always generate default config

chuckwagoncomputing:
For some reason gen_config_board.sh doesn't update changes made to TunerStudioOutputChannels.
If someone else who knows more about ConfigDefinition wants to add that to gen_config_board, go ahead, but for now here is a workaround.

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -80,6 +80,10 @@ jobs:
           bash gen_config_board.sh ${{matrix.folder}} ${{matrix.build-target}}
         fi
 
+    - name: Generate Default config
+      working-directory: ./firmware/
+      run: bash gen_config_default.sh
+
     - name: Generate Live Documentation
       working-directory: ./firmware/
       run: bash gen_live_documentation.sh


rusefillc:
See https://github.com/rusefi/rusefi/commit/1460194065896de0b6decd57d4a24cbbf3781f7e

what is the problem statement here? what exactly is broken?

chuckwagoncomputing:
firmware/controllers/generated/rusefi_generated.h
doesn't get generated when running gen_config_board, which causes a static assert on line 920 of firmware/controllers/generated/rusefi_generated.h to fail.
This isn't the only file that's not generated, this is just the problem I ran into because of this.

rusefillc:
@chuckwagoncomputing please be more specific - what change where exactly does not cause regeneration?

By design code generation only happens in case of changes only to designated files.

chuckwagoncomputing:
If I add a new member to the TunerStudioOutputChannels struct in firmware/console/binary/tunerstudio_outputs.h, and update TS_OUTPUT_SIZE in firmware/integration/rusefi_config, TS_OUTPUT_SIZE is not updated in firmware/controllers/generated/rusefi_generated.h.

PR #1629, opened 2020-07-22
[DRAFT] Idle PID improvements

andreika-git:
This patch was tested **unsuccessfully** on the real car.

And this is the same patch (hopefully) but for the [older firmware version](https://github.com/rusefi/rusefi/commit/7735758ec5c18c4ff6ddcecef87db241d0e55953), which was reportedly successful:
[older_firmware_successful_patch.patch.zip](https://github.com/rusefi/rusefi/files/4960966/older_firmware_successful_patch.patch.zip)

### DIFF
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -61,6 +61,8 @@ EXTERN_ENGINE;
 	Engine *unitTestEngine;
 #endif
 
+// todo: move all static vars to engine->engineState.idle?
+
 static bool prettyClose = false;
 
 static bool shouldResetPid = false;
@@ -69,6 +71,13 @@ static bool shouldResetPid = false;
 // We set 'mightResetPid' to true only if PID was actually used (i.e. idlePid.getOutput() was called) to save some CPU resources.
 // See automaticIdleController().
 static bool mightResetPid = false;
+
+// This is needed to slowly turn on the PID back after it was reset.
+static bool wasResetPid = false;
+// This is used when the PID configuration is changed, to guarantee the reset
+static bool mustResetPid = false;
+static efitimeus_t restoreAfterPidResetTimeUs = 0;
+
 
 class PidWithOverrides : public PidIndustrial {
 public:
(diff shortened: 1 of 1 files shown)


rusefillc:
my suggestion is to merge into master and investigate master

andreika-git:
Also here's some explanation on what's going on with PID reset...
[pid_reset_explanations.txt](https://github.com/rusefi/rusefi/files/4960986/pid_reset_explanations.txt)

rusefillc:
shall we take more of this into code comments @andreika-git ?

  There's a problem with "PID reset" method that is affecting the whole situation.
  When the PID gets deactivated due to the pressed pedal or coasting (high RPM), then we basically have two options before we turn it back:
  a) to reset the PID completely and zero the I-term - that's exactly what is implemented currently in the firmware.
  b) to leave the PID as is, not touching its state.
  Now let's imagine a situation.
  	* The engine is loaded (far light is on), the RPM tend to go low and PID's I-term rise to compensate for the load. 
  	* Then the pedal is pressed, and we deactivate the PID. 
  	* Then we drive on a highway with the pressed pedal for an hour.
  	* Finally the pedal is released, the engine returns to the idle state, and we need to reactivate the PID. 
  	* And there's no far light on anymore.
  	* The engine load is gone, so we don't want to compensate using the accumulated I-term, as before. In face, we cannot trust or use the information stored in the PID state, it's not relevant anymore. 
  	* So the best we can do is to reset the PID state and zero the I-term (which would be the best for our 'no load' situation).
  	That's why the best option in this situation is the (a) option.
  Now let's see yet another situation.
    * The engine is loaded, the I-term is increased.
    * We press the pedal shortly and then release.
    * The load remains the same.
    * If we reset the PID in this situation, we'll get a big dip in RPM, because there would be no I-term fighting the load anymore.
    * So the best we can do is reactivate the PID with the same state (parameters) as before, and let it continue its work.
	That's why the best option in this situation is the (b) option.
  The real problem is that we don't know whether the load is still there or it's gone when we reactivate the PID after the forced deactivation.
  So we cannot choose correctly between the two options (a) or (b).

  So *Maybe*. Just maybe we could try a slightly different behavior.
  - if it's time to reset the PID, we check the I-term. 
  - 1) If it has accumulated a positive amount, then it means that the regulator is currently compensating some external load.
    - In this case, we *don't* reset the PID.
    - The worst case scenario would be when the pedal is released and no load anymore: we'll get a high RPM. But that's not bad actually - the RPM is high anyway, it will drop a bit slower, and that's it.
    - BUT if the load still there, we'll get a nice and smooth RPM without dropping.
  - 2) If the I-term is negative, it means that there's some external force helping the engine to run (like in the coasting mode).
    - In this case, we reset the PID.
    - If that external force is still there, which is a worst case scenario, we'll get an increased RPM, but that's ok (see above).
  - The most important is that we won't get the RPM drop in either case!

PR #1630, opened 2020-07-22
start refactor for airmass-based alpha-n

mck1117:
yet more #1299 progress

Pulls out a base for speed density to be used by both real SD and alpha-n

### DIFF
--- a/unit_tests/tests/test_fuel_math.cpp
+++ b/unit_tests/tests/test_fuel_math.cpp
@@ -10,22 +10,22 @@ TEST(FuelMath, getStandardAirCharge) {
 	CONFIG(specs.displacement) = 1.839f;
 	CONFIG(specs.cylindersCount) = 4;
 
-	EXPECT_FLOAT_EQ(0.553539f, getStandardAirCharge(PASS_ENGINE_PARAMETER_SIGNATURE));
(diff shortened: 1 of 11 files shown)


unit_tests/tests/test_fuel_math.cpp
```diff
@@ -10,22 +10,22 @@ TEST(FuelMath, getStandardAirCharge) {
...
 	CONFIG(specs.cylindersCount) = 8;
 
-	EXPECT_FLOAT_EQ(0.80171353f, getStandardAirCharge(PASS_ENGINE_PARAMETER_SIGNATURE));
+	EXPECT_FLOAT_EQ(0.80179232f, getStandardAirCharge(PASS_ENGINE_PARAMETER_SIGNATURE));
 
 	// Chainsaw - single cylinder 32cc
 	CONFIG(specs.displacement) = 0.032f;
 	CONFIG(specs.cylindersCount) = 1;
-	EXPECT_FLOAT_EQ(0.038528003, getStandardAirCharge(PASS_ENGINE_PARAMETER_SIGNATURE));
+	EXPECT_FLOAT_EQ(0.038531788f, getStandardAirCharge(PASS_ENGINE_PARAMETER_SIGNATURE));
```
mck1117, line 24: these changes are because the order of operations for floating point numbers can change their value 🙂 (in this case, only 0.01% change)

PR #1632, opened 2020-07-23
goodbye, plain maf

mck1117:
this completes part 2 of #1299


unit_tests/tests/trigger/test_trigger_decoder.cpp
```diff
@@ -593,7 +593,7 @@ static void assertInjectionEventBatch(const char *msg, InjectionEvent *ev, int i
 static void setTestBug299(EngineTestHelper *eth) {
 	setupSimpleTestEngineWithMafAndTT_ONE_trigger(eth);
 	EXPECT_CALL(eth->mockAirmass, getAirmass(_))
-		.WillRepeatedly(Return(AirmassResult{0.1008f, 50.0f}));
+		.WillRepeatedly(Return(AirmassResult{0.1008001f, 50.0f}));
```
rusefillc, line 596: Do we need nicer EPS here?

mck1117: I don't think so.  The problem isn't for float comparisons, but is instead for integer comparisons for which microsecond (integer) a particular event is getting scheduled, for example `99999us` (wrong) vs `100000us` (correct).

Besides, the next step of the fueling math revamp is to make it easier to mock out the fuel math (not airmass math), so that this crazy `0.1008001f` constant isn't required to get a nice round pulse of 3ms.

Issue #1637, opened 2020-07-25
Weird VAG: DC motor idle valve not ETB


rusefillc:
@Abricosvw has https://en.wikipedia.org/wiki/Volkswagen_Passat_(B4) which has cable driven throttle and DC motor based IAC

We would love to support this hardware configuration

According to @Abricosvw there are Toyotas the the same setup

See https://www.youtube.com/watch?v=dR5kFb1eiqw

rusefillc:
@mck1117 any interest?

Abricosvw:
![Screenshot_20200725-141718](https://user-images.githubusercontent.com/35018246/88463558-9e4a6480-ce81-11ea-9a4c-91b3118b2883.jpg)
![Screenshot_20200725-141725](https://user-images.githubusercontent.com/35018246/88463559-9ee2fb00-ce81-11ea-90ac-7a473fc9772b.jpg)

Abricosvw:
Toyota have 2 flaps
One cable
One Etb idle

Abricosvw:
![Screenshot_20200725-141944](https://user-images.githubusercontent.com/35018246/88463616-faad8400-ce81-11ea-9914-32e56e8236ff.jpg)
![Screenshot_20200725-141951](https://user-images.githubusercontent.com/35018246/88463617-fb461a80-ce81-11ea-8766-2fca97047068.jpg)
![Screenshot_20200725-142000](https://user-images.githubusercontent.com/35018246/88463618-fb461a80-ce81-11ea-89f2-47f71ec18039.jpg)

Abricosvw:
Nissan have 2 flaps
One cable
One Etb idle

rusefillc:
We shall focus on VAG only here to keep things easier

rusefillc:
https://youtu.be/fQSfJgjDjwU

mck1117:
All of the "second throttle plate" options are traction control, IIRC.  They're sprung open, and closed by the traction control computer to manage wheel slip.  I think those engines have a normal idle solenoid.

rusefillc:
![image](https://user-images.githubusercontent.com/48498823/88482514-b11b7280-cf2f-11ea-9659-998457a9a9a3.png)

rusefillc:
@Abricosvw reports that this works now

PR #1639, opened 2020-07-26
VW's Whacky ETB Idle

mck1117:
Implement support for #1637 

Using this requires configuring the ECU like this:
1. Set the primary TPS to the cable-operated throttle's sensor
2. Set the secondary TPS to the mini ETB's position sensor(s)
3. Enable the "use VW ETB idle" bit
4. Use idle control normally, 0% will fully close the mini etb, and 100% idle fully open it.


rusefillc:
"Alex Miculescu  The same dc motor like idle is in Mercedes cars as well before etb it is like before 1999"

shall we call this flag 'dcMotorIAC'? let me push just the flag into master and make some TS project progress

rusefillc:
@Abricosvw is available to test, let me force-merge

PR #1651, opened 2020-07-29
cppcheck static analysis pass

mck1117:
- dead code
- ensuring everybody is initialized (adding `= 0` in the header to class members, etc)
- using `explicit` on single-parameter constructors
- matching parameter names between declaration and definition


rusefillc:
the length of excuses you go to remove some dead code :)

PR #1667, opened 2020-08-03
remove fuelTable

mck1117:
Remove old fuelTable, now unused.

Tangential relevance to #1299.


mck1117:
Also reclaims 1152 bytes of config space - so much room for activities!

PR #1699, opened 2020-08-18
Fuel ts reorder

mck1117:
- make 2d map values more reasonable (ie, avoid crazy decimals from uneven division)
- reorder items in the fueling menu so they group in to:
    - Airflow/airmass
    - Fuel model & corrections
    - Injection settings (currently just injection phase)
    - Enrichments
- Regroup injector hardware, modes, and deadtime in to two dialogs that look like this:

image


mck1117:
also fixes #1700

Issue #1701, opened 2020-08-18
F7 firmware issues with TS connection while transferring big data blocks


shadowm60:
There are issues with packets >256 bytes.

In code the following condition is failing:
incomingPacketSize > (sizeof(tsChannel->crcReadBuffer) - CRC_WRAPPING_SIZE

The workaround to have a usable F7 firmware is to change in TS ini file:
blockingFactor = 128 ; max chunk size

Tested with    blockingFactor = 249 ; max chunk size
This means data + 7 frame bytes -> total 256 bytes.

Works fine.

mck1117:
![image](https://user-images.githubusercontent.com/568254/90566788-f103fd00-e15d-11ea-97de-76fbf5749de0.png)

960:
With these changes my TS-connection fails (F7) :-)

I probably have different settings somewhere.. No worries :-)

mck1117:
Remember, we can't help you without you using official firmware and ini.  @960, please keep it on topic in this issue.

960:
Ah, remember me complaining about beeing unable to test with official firmware and .ini because of TS comms :-)

Just remembered it was impossible to upload settings..

PR #1702, opened 2020-08-19
remove unneccessary zeroes from TS protocol

mck1117:
- remove unused zeroes (that were once page offsets)
- actually send offset/length to the CRC command (so that partial CRCs work)
- remove page parameter from all operations since it isn't used


mck1117:
tangentially related to #1701, mostly around simplifying the protocol so it's easier to debug

PR #1704, opened 2020-08-19
don't call tach from trigger, call from periodic fast

mck1117:


rusefillc:
shall we unit test?

mck1117:
there's already a tach unit test - it still works 🙂

rusefillc:
but if we change the behavior and not change the test that means test was not sufficient?

mck1117:
A sufficient test is one that tests observable behavior, not implementation details.  The test confirms that the observable behavior didn't change.

PR #1706, opened 2020-08-20
Ts error codes

mck1117:
- send correct error codes for different problems
- actually send CRC failure when there's a crc failure (!)


firmware/console/binary/tunerstudio.cpp
```diff
@@ -544,12 +542,10 @@ void runBinaryProtocolLoop(ts_channel_s *tsChannel) {
 			scheduleMsg(&tsLogger, "TunerStudio: command %c actual CRC %x/expected %x", tsChannel->crcReadBuffer[0],
 					actualCrc, expectedCrc);
 			tunerStudioError("ERROR: CRC issue");
+			sendErrorCode(tsChannel, TS_RESPONSE_CRC_FAILURE);
```
mck1117, line 545: amazingly we never ACTUALLY responded to TS with a bad CRC response when the CRC was bad

PR #1734, opened 2020-08-30
Mc33 ucode

NOx-z:
mc33816 microcodee source code update for what we got running on the r0.4 board w/ pt2001'


firmware/hw_layer/mc33816/rusefi/MicrocodeCh1/dram1.def
```diff
@@ -6,3 +6,6 @@
 #define Tbypass 5;
 #define Thold_off 6;
 #define Thold_tot 7;
+#define SCV_I_hold 9;
```
mck1117, line 9: do we want the scv stuff in here yet?

NOx-z: I thought I already stole space in the global variable chunk for it, but I may be mistaken.. The purpose of this commit is the actual code used to generate the microcode hex already in the repo, to keep consistency. Even if it is not referenced (it's not) it is nothing but a definition of where it may be in the future.

PR #1736, opened 2020-08-30
move lambda in to sensor model

mck1117:


rusefillc:
merge conflicts

rusefillc:
@mck1117 we have a valid failure here - "cannot move location counter backwards (from 0000000020020040 to 0000000020020000)"

QC build is now running with O0 gdb on purpose to validate that we are able to compile debug firmware

mck1117:
should be set to go now

PR #1742, opened 2020-09-01
Shift buttons

chuckwagoncomputing:
GearControllerBase class for deciding what gear the driver wants
Buttonshift inherits GearControllerBase and polls switches using:
ButtonDebounce class for debouncing buttons/switches. Not very broad functionality yet, just good enough for simple button presses.

TransmissionControllerBase class for controlling the trans
Tunerstudio gauges for desired and current gears

Is 1.89:1 LOC to commit count ratio a record?


firmware/console/binary/tunerstudio_outputs.h
```diff
@@ -229,12 +229,17 @@ typedef struct {
 	scaled_voltage rawOilPressure;		// 242
 
 	int16_t tuneCrc16; // 244
-    uint8_t sd_status; // 246
-	uint8_t pad;
 
-	scaled_voltage rawPpsSecondary;		// 248
+	uint8_t sd_status; // 246
```
mck1117, line 233: why'd these get moved around? these offsets are now also wrong, and need updating in `rusefi.input`

mck1117: easiest is just to put one value in place of the old `pad` (offset 247), then everything else in the unused chunk

firmware/controllers/buttonshift.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // 500 millisecond is maybe a little long?
+    debounceUp.init(500, CONFIG(tcuUpshiftButtonPin), CONFIG(tcuUpshiftButtonPinMode));
+    debounceDown.init(500, CONFIG(tcuDownshiftButtonPin), CONFIG(tcuDownshiftButtonPinMode));
+}
+
+void ButtonShiftController::update() {
+    bool upPinState = false;
+    bool downPinState = false;
+    // Don't read the pin if TCU is not enabled, or the pin is not configured
+    if (CONFIG(tcuUpshiftButtonPin) && CONFIG(tcuEnabled)) {
```
mck1117, line 25: should this "ignore if not set" logic get put inside the debouncer itself?

mck1117, line 25: if the TCU isn't enabled, we have no business here - shouldn't even call `update`

firmware/controllers/tcu.cpp
```diff
@@ -0,0 +1,45 @@
...
+    return currentGear;
+}
+
+void TransmissionControllerBase::postState() {
+#if EFI_TUNER_STUDIO
+    tsOutputChannels.tcuCurrentGear = getCurrentGear();
+#endif
+}
+
+gear_e GearControllerBase::getDesiredGear() {
```
mck1117, line 32: in c++, functions that don't permute internal state can be declared `const`

see more here: https://www.tutorialspoint.com/const-member-functions-in-cplusplus

firmware/controllers/tcu.cpp
```diff
@@ -0,0 +1,45 @@
...
+    return desiredGear;
+}
+
+gear_e GearControllerBase::setDesiredGear(gear_e gear) {
+    desiredGear = gear;
+    return getDesiredGear();
```
mck1117, line 38: is this return value ever used?

firmware/hw_layer/debounce.cpp
```diff
@@ -0,0 +1,43 @@
...
+    pin = p;
+#ifdef PAL_MODE_INPUT_PULLDOWN
+    // getInputMode converts from pin_input_mode_e to iomode_t
+    mode = getInputMode(m);
+    efiSetPadMode("Button", p, mode);
+#endif
+}
+
+bool ButtonDebounce::readPinEvent() {
+    efitimems_t timeNow = currentTimeMillis();
```
mck1117, line 22: it's cheaper to operate in units of `efitick_t`, and convert the timeout period once upon initialization, instead of every time we have to check the time

firmware/console/binary/tunerstudio_outputs.h
chuckwagoncomputing: oh dear, didn't realize I did that

rusefillc:
sorry merge conflict with an older PR

firmware/controllers/buttonshift.cpp
chuckwagoncomputing: I agree. How should this be done? Check in the slow callback? I'm not sure I even put the call to update() in remotely the right place.

firmware/controllers/tcu.cpp
chuckwagoncomputing: Not in this. I know I had a reason for making that the API, but I don't remember why. :)

firmware/controllers/buttonshift.cpp
chuckwagoncomputing: Maybe? I suppose there might be some advantage to that.

firmware/hw_layer/debounce.cpp
chuckwagoncomputing: what's the conversion? example?

chuckwagoncomputing: And how do I get the current time in efitick_t?

mck1117: `getTimeNowNt()` will get you the current time in units of `efitick_t`.

`MS2NT(...)` will convert from milliseconds to efitick_t, so you can convert the passed in threshold to use ticks internally

rusefillc: getTimeNowNt()

firmware/controllers/tcu.cpp
chuckwagoncomputing: I think I wrote that before I decided which classes should get which functionality, and I thought maybe the class that inherited GearControllerBase might decide the request wasn't doable and return the current gear.

chuckwagoncomputing: Should I change this, or is there any value in leaving the flexibility?

rusefillc: I would keep this time but not make flexible in the future

mck1117: yeah, probably doesn't hurt since it's not on a very hot code path, but also probably not necessary until we actually want to use it

rusefillc: "not necessary until we actually want to use it" is the principle I try to live by

firmware/controllers/buttonshift.cpp
chuckwagoncomputing: I'm currently calling it in doPeriodicSlowCallback() in engine_controller.cpp

PR #1747, opened 2020-09-02
[TCU] TS interface and output channels

chuckwagoncomputing:


firmware/console/binary/tunerstudio_outputs.h
```diff
@@ -229,14 +229,18 @@ typedef struct {
...
+
+	uint8_t sd_status; // 246
+
+	int8_t tcuCurrentGear; // 247
 
 	scaled_voltage rawPpsSecondary;		// 248
 
+	int8_t tcuDesiredGear; // 250
+
 	int8_t knockLevels[12];
```
rusefillc, line 241: usually the tradition is not to move other existing fields for no reason. starting an array at an odd offset is also unusual

let's work around existing fields without shifting them please

PR #1749, opened 2020-09-02
[TCU] GearController and TransmissionController base classes

chuckwagoncomputing:


firmware/hw_layer/debounce.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+bool ButtonDebounce::readPinEvent() {
+    if (!pin) {
+        return false;
+    }
+    efitick_t timeNow = getTimeNowNt();
+    // If it's been less than the threshold since we were last called
+    if ((timeNow - timeLast) < threshold) {
+        return false;
```
mck1117, line 28: do we want to return false, or return the button's last state?

chuckwagoncomputing: for catching button events like this, we want to return false.

chuckwagoncomputing: Eventually, ButtonDebounce will need other methods that behave differently, but this was all I needed and all I'd be testing.

mck1117: ok, so it's not just a debouncer, but is also an edge detector, yes?

firmware/integration/rusefi_config.txt
```diff
@@ -1373,8 +1370,12 @@ uint8_t[4] unusedsw;;"units", 1, 0, -20, 100, 0
 	uint16_t mc33_t_hold_off;;"us",   1,   0.0,        0,  10000,   0
 	uint16_t mc33_t_hold_tot;;"us",   1,   0.0,        0,  10000,   0
 
+	bit tcuEnabled
```
rusefillc, line 1373: please use one of existing unused bits, no reason to start a new bit array

chuckwagoncomputing: Where?

chuckwagoncomputing: Oh I found what you are talking about

firmware/integration/rusefi_config.txt
```diff
@@ -1373,6 +1370,9 @@ uint8_t[4] unusedsw;;"units", 1, 0, -20, 100, 0
 	uint16_t mc33_t_hold_off;;"us",   1,   0.0,        0,  10000,   0
 	uint16_t mc33_t_hold_tot;;"us",   1,   0.0,        0,  10000,   0
 
+	pin_input_mode_e tcuUpshiftButtonPinMode
+	pin_input_mode_e tcuDownshiftButtonPinMode
+
 ! just a reminder that 'int' and 'float' are 4 bytes each
 	int[371] mainUnusedEnd;;"units", 1, 0, -20, 100, 0
```
mck1117, line 1377: needs space for button pin modes removed from this unused chunk

rusefillc: @mck1117 not really since "uint8_t alignmentFill_at_4514[2];"

Note that CI does not fail

firmware/hw_layer/debounce.cpp
```diff
@@ -0,0 +1,46 @@
...
+    timeLast = 0;
+    pin = p;
+#ifdef PAL_MODE_INPUT_PULLDOWN
+    // getInputMode converts from pin_input_mode_e to iomode_t
+    mode = getInputMode(m);
+    efiSetPadMode("Button", p, mode);
+#endif
+}
+
+bool ButtonDebounce::readPinEvent() {
```
rusefillc, line 21: Please document what why this method return
/**
@returns XXX
*/

firmware/integration/rusefi_config.txt
chuckwagoncomputing: does something need to be done?

rusefillc: @chuckwagoncomputing i think that no, you are good here

there are two other open questions
1) return meaning comment
2) return false code

PR #1750, opened 2020-09-02
[TCU] Button Shifting

chuckwagoncomputing:


firmware/controllers/buttonshift.cpp
```diff
@@ -0,0 +1,82 @@
...
+    downPinState = debounceDown.readPinEvent();
+    gear_e gear = getDesiredGear();
+    // Select new gear based on current desired gear.
+    if (upPinState) {
+        switch (gear) {
+            case REVERSE:
+                setDesiredGear(NEUTRAL);
+                break;
+            case NEUTRAL:
+                setDesiredGear(GEAR_1);
```
rusefillc, line 35: we will soon have 8 gears, this would looks a bit nasty

we probably need gear_8 enum, a property for gears count, and less copy-paste next/previous implementation?

chuckwagoncomputing: yeah. also a real world car won't let the driver switch from neutral to 1st gear with the paddle shifters

rusefillc: well this is more like "return DESIRED gear" vs "current gear" 

"desired" we can blip from 2 to 5

chuckwagoncomputing: In other words, in the real world reverse, neutral, and 1st gear at least will be special cases. I could have implemented a simple increment/decrement with a gear count property, but I figured an example that could be easily modified to use in a real-world scenario would be better than one that had to be completely rewritten.

PR #1760, opened 2020-09-06
Debounce pin management

chuckwagoncomputing:
properly manages pins and modes when the configuration changes


firmware/hw_layer/debounce.h
```diff
@@ -5,17 +5,36 @@
...
+#define DEBOUNCE_INC
+
 #include "globalaccess.h"
 #include "io_pins.h"
 
+class ButtonDebounce;
+
+class PointerListNode {
+public:
+    ButtonDebounce *pointer = NULL;
```
rusefillc, line 18: @mck1117 will tell you to use nullptr which is a fancy NULL from modern C++

firmware/hw_layer/debounce.h
```diff
@@ -5,17 +5,36 @@
  * @date Aug 31, 2020
  * @author David Holdeman, (c) 2020
  */
+#ifndef DEBOUNCE_INC
```
rusefillc, line 8: we use "#pragma once" these days - that's one line instead of three.

firmware/hw_layer/debounce.h
```diff
@@ -5,17 +5,36 @@
...
+    void init(int t, DECLARE_CONFIG_POINTERS(brain_pin_e, p), DECLARE_CONFIG_POINTERS(pin_input_mode_e, m));
+    void updateConfiguration();
     bool readPinEvent();
+    PointerListNode thisNode;
 private:
     int threshold;
     efitick_t timeLast;
-    brain_pin_e pin;
-    iomode_t mode;
+    brain_pin_e *pin;
```
rusefillc, line 31: " = nullptr;" just to be nice?

firmware/hw_layer/debounce.h
```diff
@@ -5,17 +5,36 @@
...
+    brain_pin_e *active_pin;
+    brain_pin_e oldPin;
+    pin_input_mode_e *mode;
+    pin_input_mode_e *active_mode;
     bool readValue;
```
rusefillc, line 36: "= false" just to be be nice? it's nice to have default state for as many fields as reasonable?

firmware/hw_layer/hardware.cpp
```diff
@@ -264,9 +264,20 @@ void stopSpi(spi_device_e device) {
  * this method is NOT currently invoked on ECU start
  * todo: maybe start invoking this method on ECU start so that peripheral start-up initialization and restart are unified?
  */
+
 void applyNewHardwareSettings(void) {
     // all 'stop' methods need to go before we begin starting pins
 
+	PointerListNode *listItem = &buttonDebounceListHead;
```
rusefillc, line 271: an argument could be made that we already use #include "utlist.h" which gives is list utilities for example LL_FOREACH2 see event_queue.cpp

I do not have an opinion on if this should be used here or not. Probably not since we have such a trivial use-case? @mck1117 ?

firmware/hw_layer/debounce.h
```diff
@@ -28,13 +27,11 @@ class ButtonDebounce {
 private:
     int threshold;
     efitick_t timeLast;
-    brain_pin_e *pin;
-    brain_pin_e *active_pin;
+    brain_pin_e *pin = nullptr;
+    brain_pin_e *active_pin = nullptr;
     brain_pin_e oldPin;
```
rusefillc, line 32: that's a bit weird why do we need "oldPin" permanently? please add code comment on why do we have to remember it - that's different from all other similar code where we get away without remembering old setting?

chuckwagoncomputing: I added that so when brain_pin_markUnused was called, it would call the pin that *was* set, but now that I think of it, we could just use active_pin. I'll change it.

firmware/hw_layer/hardware.cpp
mck1117: I don't think we need it for this use case either.

What we should do is move this code (lines 271-279) inside a static function on `ButtonDebounce`:

```
class ButtonDebounce {
    static void updateConfigurationList();

private:
    static ButtonDebounce* s_firstDebounce = nullptr;
};
```

chuckwagoncomputing: Are you saying s_firstDebounce should replace buttonDebounceListHead? If so, why should it be a pointer?

chuckwagoncomputing: Oh wait no it's ButtonDebounce type. I was thinking PointerListNode for some reason. never mind.

chuckwagoncomputing: I could get rid of PointerListNode entirely and just store the 'next' pointer in ButtonDebounce. Is that what you're thinking?

firmware/hw_layer/debounce.cpp
```diff
@@ -6,16 +6,60 @@
...
  */
 #include "debounce.h"
+#include "pin_repository.h"
+#include "engine_configuration.h"
+#include "hardware.h"
 
-void ButtonDebounce::init (int t, brain_pin_e p, pin_input_mode_e m) {
+ButtonDebounce* ButtonDebounce::s_firstDebounce = nullptr;
+
+void ButtonDebounce::init (int t, brain_pin_e *p, pin_input_mode_e *m) {
```
mck1117, line 15: nit: name these `pin` and `mode`

chuckwagoncomputing: sure. Golang experience kicking in on me.

firmware/hw_layer/debounce.cpp
```diff
@@ -6,16 +6,60 @@
...
+            }
+            listItem->nextDebounce = this;
+        }
+    }
     threshold = MS2NT(t);
     timeLast = 0;
     pin = p;
+    active_pin = *p;
+    mode = m;
+    active_mode = *m;
```
mck1117, line 32: can probably deduplicate a little bit by just calling updateConfiguration(), but not a big deal

firmware/hw_layer/debounce.h
```diff
@@ -5,17 +5,29 @@
...
 #include "io_pins.h"
 
 class ButtonDebounce {
 public:
-    void init(int t, brain_pin_e p, pin_input_mode_e m);
+    void init(int t, brain_pin_e *p, pin_input_mode_e *m);
+    void updateConfiguration();
     bool readPinEvent();
+    static void updateConfigurationList();
+    static ButtonDebounce* s_firstDebounce;
```
mck1117, line 20: s_firstDebounce can be private

unit_tests/globalaccess.h
```diff
@@ -23,4 +23,6 @@
 #define PASS_ENGINE_PARAMETER_SUFFIX , PASS_ENGINE_PARAMETER_SIGNATURE
 
 #define CONFIG(x) engineConfiguration->x
+#define CONFIG_POINTERS(x) &CONFIG(x), &activeConfiguration.x
```
mck1117, line 26: are these used?

chuckwagoncomputing: nope. thought I had deleted that but must have missed it.

Issue #1761, opened 2020-09-06
Selectable load axes for VE, AFR, timing


mck1117:
Sometimes it's reasonable to pick a different load axis than the default - for example, speed density mode, VE table uses MAP (default), but target AFR is determined by TPS, not MAP

as requested here: https://rusefi.com/forum/viewtopic.php?f=2&p=37814

Issue #1755, opened 2020-09-04
Use TPS insted of load for VE table not working


shadowm60:
with snapshot_20200902_173926_stm32f407_rusefi this is not working, even modifying by hand the ini file, it looks like y axis changes, but ecu is still using the load for VE lookup

https://i.imgur.com/k3XUne1.png

PR #1762, opened 2020-09-06
Move ve logic

mck1117:
- Do baro correction as a correction to fueling, not VE (so that it works in all modes!)
- Move logic for idle VE in to `airmass.h`
- Use that logic in SD

prep to do #1755 and #1761


mck1117:
Tested working on a real engine!

PR #1763, opened 2020-09-06
reorder ADC channel index

### LINKED ISSUE #1708: Zero bits should mean "NONE" adc not PA0
When we add new channels we make older configurations invalid like #1707

We need to change enum definition so that zero bits means NONE

mck1117:
fix #1708


mck1117:
Need to double check that this will work properly with latest free TS.

mck1117:
Tested working with free TS, version 3.1.04!

mck1117:
note about TS version added to: https://github.com/rusefi/rusefi/wiki/Download

PR #1764, opened 2020-09-06
move AEM xseries lambda to new sensor

mck1117:


rusefillc:
@mck1117 merge conficts

mck1117:
Ok, finally got a chance to test on real HW.  Works fine!

PR #1765, opened 2020-09-06
Adc improvements

dron0gus:


rusefillc:
@mck1117 fyi

firmware/hw_layer/adc/adc_inputs.cpp
```diff
@@ -42,6 +42,13 @@
 #define ADC_BUF_DEPTH_SLOW      8
 #define ADC_BUF_DEPTH_FAST      4
 
+// on F7 this must be aligned on a 32-byte boundary, and be a multiple of 32 bytes long.
+// When we invalidate the cache line(s) for ADC samples, we don't want to nuke any
+// adjacent data.
+// F4 does not care
+static __ALIGNED(32) adcsample_t slowAdcSampleBuf[ADC_BUF_DEPTH_SLOW * ADC_MAX_CHANNELS_COUNT];
+static __ALIGNED(32) adcsample_t fastAdcSampleBuf[ADC_BUF_DEPTH_FAST * ADC_MAX_CHANNELS_COUNT];
```
mck1117, line 50: did the `static_assert` get lost?

firmware/hw_layer/adc/adc_inputs.cpp
```diff
@@ -156,11 +165,11 @@ static ADCConversionGroup adcgrpcfgSlow = {
 	.sqr3				= 0  // Conversion group sequence 1...6
 };
 
-AdcDevice slowAdc(&adcgrpcfgSlow);
+AdcDevice slowAdc(&adcgrpcfgSlow, slowAdcSampleBuf, sizeof(slowAdcSampleBuf));
```
mck1117, line 168: do you want to pass the size (in bytes) or the size in elements (`efi::size(slowAdcSampleBuf)`)?

dron0gus:
@mck1117 updated. Please review.

rusefillc:
@mck1117?

firmware/hw_layer/adc/adc_inputs.cpp
```diff
@@ -550,6 +582,19 @@ void initAdcInputs() {
 	adcStart(&ADC_FAST_DEVICE, NULL);
 	adcSTM32EnableTSVREFE(); // Internal temperature sensor
 
+	/* Enable this code only when you absolutly sure
+	 * that there is no possible errors from ADC */
+#if 0
+	/* All ADC use DMA and DMA calls end_cb from its IRQ
```
mck1117, line 588: imo an even better solution would be to trigger the fast adc with a timer, so that we collect perhaps 1 sample per degree, then compute the average exactly once at the end, instead of ~2 interrupts per sample

mck1117: but that's a problem for another PR

dron0gus: There are few strange things with STM32 ADCs:
1. all ADCs share same IRQ vector.
2. while running continues conversion with DMA ADC still generates 2 irq for each batch conversion: on start and on end of conversion. I was not able to find way to disable this IRQ sources. Only disabling ADC vector at interrupt controller. But we still interested in error/overrun events.
3. sw logic rely on DMA end of transfer IRQ. We compute averages at DMA tranfer end callback. So these two ADC irq per batch conversion are totally useless and just wasting cpu time.

rusefillc:
@andreika-git is this related to your ADC changes in Hellen branch?

@dron0gus shall we move some of these commens on the PR into the code as code comments?

andreika-git:
> 
> 
> @andreika-git is this related to your ADC changes in Hellen branch?
> 

I don't think this is directly related, but it needs accurate merging.

PR #1782, opened 2020-09-09
DebounceButton A/C switch

chuckwagoncomputing:


rusefillc (changes requested):
we need mocked digital switches please nothing fancy?

rusefillc (approved):

rusefillc:
fails multiple checks :(

mck1117 (approved):
new digital mocking looks good to me!

rusefillc:
are we expanding scope before getting a green build of an OK version?

rusefillc:
please build locally? it's not that I am bothered by emails but you itself would be more efficient if you are not waiting for CI to fail? is not it quicker locally?

firmware/hw_layer/debounce.cpp
```diff
@@ -9,6 +9,7 @@
 #include "pin_repository.h"
 #include "engine_configuration.h"
 #include "hardware.h"
+#include "hal.h"
```
rusefillc, line 12: this does not look right. no one is including hal.h why would debounce.cpp?!

rusefillc (changes requested):

firmware/hw_layer/debounce.cpp
chuckwagoncomputing: brute force programming. it's removed now

rusefillc: I am not the one to criticize brute force programming.

chuckwagoncomputing:
> please build locally? it's not that I am bothered by emails but you itself would be more efficient if you are not waiting for CI to fail? is not it quicker locally?

I am building locally. I push then pull in a different copy, so I don't accidentally push config changes etc.

chuckwagoncomputing:
> are we expanding scope before getting a green build of an OK version?

Well, we expanded the scope from using DebounceButton for the A/C switch, to adding digital pin mocking so we can test it.
Beyond that, no. My initial implementation was broken. It was in an #if EFI_PROD_CODE, and when I got it out I discovered that it didn't actually work, because it didn't have access to the 'engine' object.

PR #1788, opened 2020-09-11
logging simplification: switch bare functions to member functions

mck1117:
Lots of functions of the form `myFunc(Logging* logger)` should actually be members of `Logging`.  This change does that!

Changing the `buffer` member to `const` means that it has to be initialized at construction - so it can't be null! Removing runtime checks in favor of compile-time checks is always great.


rusefillc:
Thank you for a great change

Issue #1784, opened 2020-09-11
Nik is asking for DC motor wastegate


rusefillc:
fancy turbo is wired as an ETB in this test

https://www.youtube.com/watch?v=n7eQ2lRd47U

rusefillc:
@mck1117 believes we are done here

@Abricosvw needs to test

mck1117:
we are *probably mostly* done here, just wanted to confirm on a real engine before we close.

PR #1806, opened 2020-09-18
two throttles one thread

mck1117:
etbs are adults, they can share a single thread between them.  Besides , ChibiOS won't even preempt a thread with the same priority, so they were already sort of running on the same thread.

progress on #1784


mck1117:
oh yeah, this also saves ~1kb of RAM!

Issue #1802, opened 2020-09-17
time units mismatch around start/stop


rusefillc:
we have a mess related to time units and @chuckwagoncomputing has totally fallen into the trap with code reviewers not noticing

"int16_t startCrankingDuration;"
"int t"

which one is in seconds? which one is in ms? impossible to tell without reading the code deeper :(

rusefillc:
See #1801

we might need to review the whole "efitimems_t" and have explicit type name for more combinations of logic type and storage type

chuckwagoncomputing:
I don't know enough yet to even know where to start with this. What types would be replaced by new types?

chuckwagoncomputing:
Does efitimems_t already have its own TS field?

rusefillc:
I am pretty distracted for 8 more days.

1) please look at current 5-8 type names and confirm that making convention makes sense. There is a chance that a better more more uniform is possible

chuckwagoncomputing:
Are you asking for efitimeXX_t types in ConfigDefinition so they can be used in rusefi_config.txt?

rusefillc:
yes please try efitimeXX16_t in ConfigDefinition

chuckwagoncomputing:
I think this should be closed, and perhaps another opened to address time type issues.

chuckwagoncomputing:
@rusefillc What needs to be done on this issue? The mismatch has long since been fixed; the only remaining issue is you wanting to revamp the time type system. I think that if that is something we still want to do, an issue should be opened with a clear statement of what needs to be changed, and this one should be closed.

PR #1808, opened 2020-09-19
fix type mismatch

chuckwagoncomputing:
This fixes the title issue of #1802


rusefillc:
at the moment this PR is just the bare minimum of the fix

let's do better - let's invest a bit into quality. at least we need field comment that units are seconds
also while init method type was changed, class field was not? let's do better than just bare minimum please.

rusefillc (changes requested):
at least config type code comment, class field type

maybe also new time type for config?

chuckwagoncomputing:
Config is already efitimesec_t, not sure what more you want for this issue.

rusefillc:
config is startCrankingDuration

startCrankingDuration is not efitimesec_t

startCrankingDuration is int16_t

actually startCrankingDuration already has "Seconds" human-readable for TS units 
```
	int16_t startCrankingDuration;Maximum time to crank starter;"Seconds",      1,      0,       0, 30,      0
```


but if a developer is looking at engine_configuration_generated_structures.h
```
	/**
	 * Maximum time to crank starter
	 * offset 826
	 */
	int16_t startCrankingDuration;
```
there is no reference to seconds

we can at least change "Maximum time to crank starter" to "Maximum time to crank starter, in seconds"

or maybe ideally change int16_t to efitimesec16_t

chuckwagoncomputing:
Not sure why I thought it did.

PR #1821, opened 2020-09-27
Hellen says Cypress build fix (usbdevice.h)

andreika-git:


firmware/config/boards/hellen/cypress/!compile_cypress.bat
```diff
@@ -0,0 +1,3 @@
+@echo off
+
+bash compile_cypress.sh
```
mck1117, line 3: is this really necessary?

PR #1824, opened 2020-09-27
Hellen says fix gen_config for Cypress

andreika-git:


rusefillc:
please remove actual .h file changes from this PR

I would like to fix scripts and in order to validate we need to give scripts a chance to generate stuff, maybe with manual .txt poke

if we just manually fix configs via this PR we would have no way to validate, we would only see potential issues next time. I would prefer not to postpone validation until next time.

PR #1844, opened 2020-10-02
move scheduling_s to pool - no longer passed!

mck1117:
We allocate many more scheduling_s than we need at a time - so recycle them with a pool instead of always bringing along your own

needs more real-world testing, some test repair (some tests seem to use an inordinate number of scheduling entries - they're probably not being run correctly as the engine "spins")


rusefillc:
@mck1117 revive or close or else?

mck1117:
unit tests aren't ready for this, maybe later :disappointed:

PR #1855, opened 2020-10-05
simpilify scheduler: we don't need two loops

mck1117:


mck1117:
best I can tell it might save about a microsecond

PR #1856, opened 2020-10-05
const ref-ify trigger

mck1117:
almost entirely a find-replace of pointer -> ref, `const` as appropriate


rusefillc:
what's the motivation for pointers to ref migration?

PR #1857, opened 2020-10-05
remove virtual calls from hot path

mck1117:
near the noise floor of our measurements, saves mayyybe 1us time per trigger tooth


mck1117:
needs #1856 to merge first, btw

PR #1859, opened 2020-10-05
use expected for event queue next event

mck1117:
Compilers are bad at optimizing around magic numbers.


mck1117:
also includes formatting & comments in `microsecond_timer.cpp`

rusefillc:
merge conflict?

PR #1861, opened 2020-10-06
FSIO refactoring

mck1117:
- much clearer tests
- move similar operators out to their own functions (reduce duplication!)
- add binary literals (`true`/`false`)
- minor parser cleanup


unit_tests/tests/test_logic_expression.cpp
```diff
@@ -217,11 +263,6 @@ TEST(fsio, testLogicExpressions) {
 	testExpression("coolant 90 >", 1, sensorVals);
 	testExpression("fan not coolant 90 > and", 1, sensorVals);
 
-	testExpression("100 200 1 if", 200);
```
mck1117, line 220: the removed stuff in this test is covered by the added tests

PR #1866, opened 2020-10-08
Hellen says fix build-signature & usb

andreika-git:


firmware/config/boards/hellen/cypress/efifeatures.h
```diff
@@ -301,8 +301,9 @@
 
 #undef TS_UART_DEVICE
 #undef TS_SERIAL_DEVICE
-//#define TS_USB_DEVICE SDU1
-#undef TS_USB_DEVICE
+
+#define TS_USB_DEVICE SDU1
+//#undef TS_USB_DEVICE
```
rusefillc, line 306: should we remove the dead code?

PR #1876, opened 2020-10-12
simplify getTimeNowNt()

mck1117:
While working on #1874, I rewrote `getTimeNowNt` to be much simpler.  On ARM this implementation is actually branch free, so has the potential to be faster.

To answer the deleted comment ` // todo: why is this implementation not used? `:
In a real-time application, I think we prefer a deterministic algorithm, even if it requires a global lock.  A lock free solution has terrible worst case performance, which is what we care most about when realtime.  At -O2, we're only locked for something like 16-18 cycles - only 110 nanoseconds.  It's just not enough time to worry about.


rusefillc (changes requested):
feel free to expand the code comment here but let's absolutely keep the lock free dead code

PR #1880, opened 2020-10-14
switch to single precision constants

mck1117:
`0.25`: this is a double
`0.25f`: this is a float

Adding this compiler option means these are both actually floats.  We don't have the hardware to process doubles, so it happens instead in software.

For example, the line of code `float instantRpm = (60000000.0 / 360 * US_TO_NT_MULTIPLIER) * angleDiff / time;` compiles to this:

image

Circled are SOFTWARE EMULATED `double` math operations.  Bad! Slow! So much extra code!

With this option added, it now compiles to this:

image

As a bonus, this saves ~1k of flash space.

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -40,7 +40,7 @@ endif
 
 # Compiler options here.
 ifeq ($(USE_OPT),)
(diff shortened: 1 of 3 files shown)


rusefillc:
@mck1117 great stuff btw :)

PR #1885, opened 2020-10-16
trigger/scheduler perf improvements

mck1117:
- avoid a 64 bit divide in scheduleByTimestamp
- Disable tooth logger above 4k rpm
- only enable trigger emulator HW outputs if NOT also doing self stimulation
- slightly decrease the EventQueue lateDelay, as the perf has been improved somewhat.


mck1117:
I can get a stable-ish 29000 rpm with this in the VW test config used by HW CI 😱

PR #1886, opened 2020-10-16
Remove trigger listener list

mck1117:
the trigger listener list is a double perf killer:
- it's a loop of runtime-determined length, so it can't be unrolled
- it's a loop of double indirect calls, so they can't be branch predicted or prefetched.  Removing the list and converting to direct calls means they can be inlined, which is good for everybody.


mck1117:
not ready for merge, depends on #1885

rusefillc:
#1885 merged

mck1117:
ok, now this is ready

PR #1893, opened 2020-10-21
Everything in lambda!

mck1117:
More lambda, less AFR.  Everybody wins!


mck1117:
tested on a real engine, works great!

rusefillc:
@mck1117 conflict

PR #1894, opened 2020-10-22
no sd enable

mck1117:
Remove SD enable bit, and just use the SPI device to enable/disable SD.  ChibiOS gets REALLY unhappy when you initialize the MMC driver without an SPI device set.


rusefillc:
hold on, I do not follow

i believe we still need SD enable bit. if we need to handle invalid config better - let's handle invalid config better?

rusefillc:
there is value in having configuration but disabling it.

mck1117:
Why do we need an enable bit? Plenty of other features use the pin/channel config as the enable bit.

PR #1902, opened 2020-10-24
Trigger logger

shadowm60:
logger update  the way we send the buffer to TS and the way we store camshaft triggers.


firmware/console/binary/tooth_logger.cpp
```diff
@@ -78,23 +81,38 @@ int copyCompositeEvents(CompositeEvent *events) {
...
+	//ptr_buffer[NextIdx].injector = injector;
 
 	NextIdx++;
 
 	static_assert(sizeof(composite_logger_s) == COMPOSITE_PACKET_SIZE, "composite packet size");
 
-	// If we hit the end, loop
-	if (NextIdx >= sizeof(buffer) / sizeof(buffer[0])) {
+	//If we hit the end, loop
+	if ((firstBuffer) && (NextIdx >= (COMPOSITE_PACKET_COUNT/2))) {
```
mck1117, line 99: why don't you just track position in the current buffer, and when it comes time to swap them, swap the pointers? We shouldn't have to care about where we are in the larger allocation.

firmware/console/binary/tooth_logger.cpp
```diff
@@ -78,23 +81,38 @@ int copyCompositeEvents(CompositeEvent *events) {
...
+	
 	// TS uses big endian, grumble
 	buffer[NextIdx].timestamp = SWAP_UINT32(nowUs);
 	buffer[NextIdx].priLevel = trigger1;
 	buffer[NextIdx].secLevel = trigger2;
 	buffer[NextIdx].isTDC = isTDC;
 	buffer[NextIdx].sync = engine->triggerCentral.triggerState.shaft_is_synchronized;
-	buffer[NextIdx].coil = coil;
-	buffer[NextIdx].injector = injector;
+	//ptr_buffer[NextIdx].coil = coil;
```
rusefillc, line 91: why do we add dead code? please add code comments explaining need for dead code

rusefillc:
is this one unseparateable change or two separate changes together in same PR?

i am not in the loop of what's going on so not sure how to review. Need more details on what's going on here and why?

firmware/console/binary/tooth_logger.cpp
shadowm60: Dose it make a difference? it was simpler for me this way, so i keep the old stuff and use it as much as i could, performance vise i do not think it is much difference.

firmware/console/binary/tooth_logger.cpp
shadowm60: i forgot to uncomment those, already did it.

shadowm60:
> is this one unseparateable change or two separate changes together in same PR?
> 
> i am not in the loop of what's going on so not sure how to review. Need more details on what's going on here and why?

It could be splitted in two tickets, but it is simpler to keep it all at once. 
Basically isTDC is really trigger, and it should be 1 when ever there is a change on the secondary input. so that fix is simple.

The half buffer send ensures that TS will always get data in order, so it would display a nice graph and correct log in csv format since all the received data is usable.

since this part of TS is not documented, or at least i could not find it, what i can say is that testing it with 60-2 and one cam pulse, the logging worked for a long time without crashing.

PR #1905, opened 2020-10-25
daily builds

ksmola:
added daily builds link


rusefillc:
Users are really starting to focus on release builds not daily builds so we either need both links or maybe just https://github.com/rusefi/rusefi/wiki/Download link

let me make it https://github.com/rusefi/rusefi/wiki/Download for now

PR #1906, opened 2020-10-25
convert mass -> duration later in the pipeline

### LINKED ISSUE #1903: injector deadtime is applied twice
#1901 seems to have accidentally added deadtime twice, so everybody runs rich 😢

mck1117:
Convert to duration later in the fueling pipeline.  This means all corrections (cranking, CLT/IAT/duration, etc) are applied as a mass, not as a duration, with injector deadtime added at the very end, after all corrections.

As a side effect, this means the base cranking value is now a mass (milligrams), not a duration (milliseconds), so it's now automatically adjusted by the injector settings, and will soon be adjusted by the fuel pressure too, which will provide better startup performance for GDI, and potentially PFI even without a pump prime!

fixes #1903 by design, not with a hack like #1904 did


mck1117:
now tested on a real engine along with #1893!

PR #1907, opened 2020-10-25
use per-device SPI mutex

mck1117:
chibios already has a mutex, no reason to add our own.


mck1117:
tested on hardware, works great!

PR #1908, opened 2020-10-25
Tle8888 small update

dron0gus:
Add way to control MR - not very useful because MR is hardware enabled while KEY is high. So we can implement only some kind of power hold with this. No emergency shutdown and no MR use as general purpose output.
Add way to read KEY and WAKE inputs. Also not very useful on MRE: WAKE is not used, KEY (pin 5 "+12 IGN") is usually shorted to BAT (pin 1 "+12 Relay") input and always hi while MRE is powered.


mck1117:
oh, so you can't turn MR off in software, only keep it on?

dron0gus:
Yes.
Screenshot from 2020-10-26 12-35-28

rusefillc:
1) merge conflict

2) what are observable differences expected with MRE hardware? or are we just confirming that there are no _unexpected_ differences?

dron0gus:
1. Merge conflict was about timestamp comment in autogenerated auto_generated_enums.cpp.

PR #1920, opened 2020-10-31
log afr in addition to lambda

mck1117:
log both, because recalibrating the brain is hard


firmware/tunerstudio/rusefi.input
```diff
@@ -370,6 +370,7 @@ enable2ndByteCanID = false
 	lowFuelPressure       = scalar,  S16,     276, "kpa", {1/@@PACK_MULT_PRESSURE@@}, 0
 	highFuelPressure      = scalar,  S16,     278, "bar", {1/@@PACK_MULT_HIGH_PRESSURE@@}, 0
 	targetLambda     = scalar,  U16,     280,      "",{1/@@PACK_MULT_LAMBDA@@},       0.0
+	AFRValue        = scalar,  U16,     282,      "AFR",,{1/@@PACK_MULT_AFR@@},,       0.0
```
rusefillc, line 373: double-comma looks unusual here is that bug or a feature?

mck1117: this was copy pasted from the "removed" side of my other pr - so we were running with that before.  However, it does indeed look wrong (I'm not sure why it worked), so I'll remove

mck1117: There are a few of them!

image

PR #1921, opened 2020-11-03
convert cranking defaults to mg

### LINKED ISSUE #1915: Change canned tunes cranking base fuel
We need to change cranking base fuel in canned Miata and v12 tunes
it was milliseconds and it's now milligrams

See #1906

mck1117:
fix #1915


rusefillc:
Master is green again please rebase

PR #1922, opened 2020-11-03
fatal error + inhibit read on TS preset

### LINKED ISSUE #1810: Presets are not user friendly
Users are missing the warning on the presets dialog and still hit 'burn' and get write error. We need a better approach to preset buttons.

mck1117:
this will probably fix #1810

If we pop a fatal error with a message, we can force the user to reboot the ECU.  Likewise, we can inhibit certain functions so that TS can't accidentally overwrite: read, write, and burn will no longer work until the ECU is rebooted.


mck1117:
Here's what it looks like:

image

rusefillc:
thank you!

PR #1923, opened 2020-11-03
extract idle hardware to its own file

mck1117:
just a copy paste job to reduce the spaghetti in `idle_thread.cpp`


rusefillc:
looks like legit build failures

mck1117:
fixed!

PR #1924, opened 2020-11-03
remove idle thread; just call from periodicSlowCallback

### LINKED ISSUE #1656: Document idleTimingPid.periodMs logic
idleTimingPid.periodMs is not a true PID period is it?

``/* is this the right dTime? this period is not exactly the period at which this code is invoked*/engineConfiguration->idleTimingPid.periodMs)``

mck1117:
Idle control doesn't need its own independent thread - it can just get called from the periodic slow callback

also fixes #1656


rusefillc:
Poke poke

rusefillc:
:thumbsup:

mck1117:
tested, seems fine

PR #1934, opened 2020-11-07
unlock/relock between scheduler queue events

mck1117:
This might help with sync loss problems.  One potential issue is that a long sequence of timer events keep a lock held (by the scheduling queue), so no trigger interrupts can happen while that sequence is being executed.  This change unlocks the global lock every iteration so that an interrupt can fire, then immediately re-locks.  This is safe to do because it is equivalent to the trigger interrupt being called by a scheduled event, which is something we already do in the form of trigger self stimulation.


ptnicu:
Hi, Try to move Critical section to protect only the needed portion of code like contents inside of the  `do{}while;` loop and remove unneeded lock checking at the end. Or use multiple Critical sections.  Critical section usage should be minimal and not include loops if possible.

mck1117:
@ptnicu I agree in the general case - finer grained locking is good.  However, in this case, we know a bit more about the system than in the general purpose operating system case.

We have the interrupt priority configured like this:
     - Trigger inputs
     - Scheduling queue
     - everything else
     - normal threads

So it's totally fine for us to stay locked while processing the scheduler queue, except for when a long run can block trigger input interrupts.  We also need to stay locked to schedule the timer callback (called after `executeAllPendingActions`), so the whole thing really does have to be done under lock.  Without that lock another interrupt could sneak in after the loop (before the timer reset) and corrupt the scheduler's state.

ptnicu:
@mck1117   I see now, that there is other `while{}` in `executeOne()` of maximum of 8uS(`lateDelay`) when "stars align".
As an idea:
 If the scheduler could calculate that if events are close to each other then group those events to execute at the same time and create delay between group of events of more than 10 uS(guesstimating) in the `insertTask()` method, and remove this `while(lateDelay){}` loop. 
  Removing busy loops will reduce Critical section time so there will be no need for locking/unlocking.

mck1117:
> If the scheduler could calculate that if events are close to each other, then group those events to execute at the same time and create delay between group of events of more than 10 uS(guesstimating) in the `insertTask()` method, and remove this `while(lateDelay){}` loop.

The problem is that any event can insert any number of new events, including at the head of the list.  This is how internal trigger stimulation works - events fire that call the trigger input interrupt as if a trigger input had actually happened.

> This lateDelay for near future events may be the culprit of maximum of 16uS latency from your post. What are the odds to catch 2 consecutive events spaced exactly at less than 8uS? Did your results where consistent in multiple tests run ? Definitely the near future events case needs to be handled better and tested in simulator.

that testing was done before the spinwait was added

ptnicu:
The event table execution should be used only consume events, not to generate events in the callback. Also the callback should contain only code for setting the pin to high/low, nothing more, execute as fast as possible.
Trigger simulation should be decoupled from the event execution. With its own timer and event table.

rusefillc:
poke @mck1117 can you test this on real car? do we want this in the release? do you have doubts about this PR?

mck1117:
I think I'm actually going to abandon this PR since I don't think it's required after figuring out the `CH_FAST_IRQ` dispatch trick.

PR #1940, opened 2020-11-11
consolidate EFI_UNIT_TEST and friends

mck1117:
depends on #1937

this removes the dependency on EVERY SINGLE FILE including `efifeatures.h`


rusefillc:
@mck1117 merge conflict it says

PR #1950, opened 2020-11-16
put getTimeNowLowerNt in to mpu_util

mck1117:
laying groundwork for the magic that yielded these results: https://rusefi.com/forum/viewtopic.php?p=38501#p38501

This change moves `US_TO_NT_MULTIPLIER` and `getTimeNowLowerNt` to be determined by the port, not globally.


rusefillc:
@mck1117 merge conflict?

mck1117:
yeah I merge conflict'd myself

firmware/hw_layer/ports/cypress/port_mpu_util.h
```diff
@@ -8,6 +8,12 @@
 
 #pragma once
 
+// 168 ticks in microsecond in case of 168MHz 407
+#define US_TO_NT_MULTIPLIER (CORE_CLOCK / 1000000)
+
+// Scheduler queue GPT device
+#define GPTDEVICE GPTD1
```
rusefillc, line 15: this was configurable via .h file why not any more?

mck1117: `port_mpu_util.h` is a .h file 🙂 

`efifeatures` and `mpu_util` are meant to serve two orthogonal purposes:
- `mpu_util` determines "how do I use the hardware [CPU, peripherals] to fulfill rusEfi's needs"
- `efifeatures` determines what particular features a particular board has, orthogonal to the operation of the CPU itself (ie "do I have an LCD", "am I running an engine", "is USB wired")

Consider the Hellen case: an STM32Fx vs. Cypress vs. Kinetis variant of Hellen might want to use the same `efifeatures.h` because the board is the same, but would certainly use different `mpu_util` since they're different chips.

Since `GPTDEVICE` is a core "how does rusEfi do anything" configuration, it should have nothing to do with a particular board, and everything to do with the MCU.  This change is happening because I have another PR queued up to stop using GPT altogether on STM32 for the microsecond timer.

PR #1944, opened 2020-11-12
fix CRC errors

mck1117:
use the `crcReadBuffer` as a transmit buffer as well.  This ensures that we compute the CRC and transmit the same data - there's no opportunity for another thread to preempt and change the data between the CRC calculation and transmission.  This might actually net improve performance - it saves two calls to `sr5WriteData`.

fixes #1943

PR #1955, opened 2020-11-17
read SD card config exactly once

mck1117:
don't re-read SD card spi device after boot - it could change mid transaction (TS thread could preempt), which will cause disaster (usually multiple-unlock of the same mutex in `UNLOCK_SD_SPI`


rusefillc:
Related to #1944

PR #1972, opened 2020-11-20
Launch control unit tests update

shadowm60:
- VSS update to have mock in case of unit tests
- launch small optimization
- condition checks added in unit tests and a basic use case.


rusefillc:
@shadowm60 I have to be honest I have not reviewed the test - I just know that _any_ test is better than nothing :)

do you have more ideas for more tests or do you think it's complete?


coverage report is at https://rusefi.com/docs/unit_tests_coverage/

rusefillc:
See https://github.com/rusefi/rusefi/issues/1973

rusefillc:
https://rusefi.com/docs/unit_tests_coverage/ shows fresh stuff now!

See https://rusefi.com/docs/unit_tests_coverage/firmware/controllers/algo/launch_control.cpp.gcov.html

shadowm60:
Based on the report i could still add some tests. Let me do it today and then i think it is fine.
I still want the update function to be reorganized,

shadowm60:
@rusefillc can we close this pull request, i plan to move from thread to periodic fast callback, and in that pull request i will add the rest of the tests, this way i think it's simpler to track, if you agree with me, if not i'll continue here.

rusefillc:
@shadowm60 this was already merged into master

PR #1977, opened 2020-11-22
avoid float -> int64 conversion

mck1117:
arm32 doesn't have any native 64b instructions, and that includes conversion from float to 64 bit int.  This is a common conversion we were doing from float -> efitick_t (aka int64_t).  This change avoids that conversion as it's pretty slow to do in software.

### DIFF
--- a/firmware/controllers/algo/advance_map.cpp
+++ b/firmware/controllers/algo/advance_map.cpp
@@ -250,8 +250,8 @@ size_t getMultiSparkCount(int rpm DECLARE_ENGINE_PARAMETER_SUFFIX) {
 		floatus_t multiDelay = CONFIG(multisparkSparkDuration);
 		floatus_t multiDwell = CONFIG(multisparkDwell);
 
-		ENGINE(engineState.multispark.delay) = US2NT(multiDelay);
-		ENGINE(engineState.multispark.dwell) = US2NT(multiDwell);
+		ENGINE(engineState.multispark.delay) = (uint32_t)USF2NT(multiDelay);
+		ENGINE(engineState.multispark.dwell) = (uint32_t)USF2NT(multiDwell);
 
 		constexpr float usPerDegreeAt1Rpm = 60e6 / 360;
 		floatus_t usPerDegree = usPerDegreeAt1Rpm / rpm;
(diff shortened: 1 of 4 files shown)


rusefillc:
Is this PR adding same "#define USF2NT(us_float) ((us_float) * US_TO_NT_MULTIPLIER)" into three different files?

I know that this follows a not great example of "#define US2NT(us)" but let's please do better?

mck1117:
will consolidate those all down to one place

rusefillc:
this is a kind of change where it would probably be easier for me to pick individual parts a few times

to make this easier can you please squash & rebase? or at least rebase? merge commit of master makes it a bit harder

rusefillc:
@mck1117 please add code comments into efitime.h - as is it would be hard for the reader to guess the difference between two sets of macro

PS: sorry my thing has caused a merge conflict again :(

firmware/controllers/bench_test.cpp
```diff
@@ -65,34 +65,27 @@ bool isRunningBenchTest(void) {
...
-	}
+	int delayUs = delayMs * 1000;
+	int onTimeUs = onTimeMs * 1000;
+	int offTimeUs = offTimeMs * 1000;
+
+	delayUs = delayUs < 1000 ? 1000 : delayUs;
+	onTimeUs = onTimeUs < 1000 ? 1000 : onTimeUs;
+	offTimeUs = offTimeUs < 1000 ? 1000 : offTimeUs;
+
 	scheduleMsg(logger, "Running bench: ON_TIME=%.2f ms OFF_TIME=%.2fms Counter=%d", onTimeMs, offTimeMs, count);
```
rusefillc, line 76: we apply thresholds to parameters but print unmodified values here

also it looks like
int delayUs = maxF(1, delayMs) * 1000;
would be a shorter way to express?

to be honest "1000" bothers me as a magic constant I would prefer MS2US here

mck1117: cleaned it up a bit, how's that?

firmware/hw_layer/stepper.cpp
```diff
@@ -195,7 +195,7 @@ bool StepDirectionStepper::pulse() {
 }
 
 void StepperHw::pause() const {
-	chThdSleepMilliseconds(m_reactionTime);
+	chThdSleepMicroseconds((int)(m_reactionTime * 1000));
```
rusefillc, line 198: 1000 magic constant, MS2NS?

rusefillc:
@mck1117 poke please squash commits?

mck1117:
resurrected as #1993

PR #1983, opened 2020-11-23
switch stm32 scheduling to use PWM driver

mck1117:


firmware/hw_layer/ports/stm32/microsecond_timer_stm32.cpp
```diff
@@ -0,0 +1,65 @@
...
+};
+
+void portInitMicrosecondTimer() {
+	pwmStart(&SCHEDULER_PWM_DEVICE, &timerConfig);
+
+	// ChibiOS doesn't let you configure timers in output compare mode, only PWM mode.
+	// We want to be able to set the compare register without waiting for an update event
+	// (which would take 358 seconds at 12mhz timer speed), so we have to use normal upcounting
+	// output compare mode instead.
+	SCHEDULER_TIMER_DEVICE->CCMR1 = 0x00006810;
```
andreika-git, line 56: Maybe we need to replace this magic value with STM32_TIM_CCMR1_xxx()?

mck1117: good idea, I'll construct it out of those macros instead of magic

PR #1988, opened 2020-11-25
fix etb wastegate percent output

mck1117:
was using duty [0, 1], should've used percent [0, 100]


rusefillc:
unit test failed

how about "percent_t" instead of one "float"?

rusefillc:
unit test failed?

mck1117:
both fixed!

PR #1992, opened 2020-11-29
Launch Control refactoring

shadowm60:
Periodic function now called from Fast callback.
Removed own thread.
Bugfix so it would really stop launch control and not just active all the time.
Added unit tests.
Not yet tested on real target.


shadowm60:
Just tested with Discovery board, now it looks fine.

firmware/controllers/algo/launch_control.cpp
```diff
@@ -134,21 +127,31 @@ bool LaunchControlBase::isLaunchConditionMet(int rpm) const {
 	return speedCondition && activateSwitchCondition && rpmCondition && tpsCondition;
 }
 
-void LaunchControlBase::update() {
+//void LaunchControlBase::update() {
```
rusefillc, line 130: why "//void LaunchControlBase::update() {" dead code?

firmware/controllers/algo/launch_control.cpp
```diff
@@ -33,36 +37,24 @@ extern TunerStudioOutputChannels tsOutputChannels;
...
-};
-
 /**
  * We can have active condition from switch or from clutch.
  * In case we are dependent on VSS we just return true.
  */
 bool LaunchControlBase::isInsideSwitchCondition() const {
 	switch (CONFIG(launchActivationMode)) {
 	case SWITCH_INPUT_LAUNCH:
+		engineConfiguration->launchDisableBySpeed = 0;
```
rusefillc, line 49: that's bool time? please use 'true' and 'false' for better readability

firmware/controllers/algo/launch_control.cpp
```diff
@@ -78,6 +70,7 @@ bool LaunchControlBase::isInsideSwitchCondition() const {
 		
 	default:
 		// ALWAYS_ACTIVE_LAUNCH
+		engineConfiguration->launchDisableBySpeed = 1;
```
rusefillc, line 73: changing engineConfiguration at runtime is a bit unexpected

maybe that's NOT _configuration_ but runtime state? please move this field to "Engine" class?

firmware/controllers/algo/engine2.cpp
```diff
@@ -19,6 +19,10 @@
 #include "closed_loop_fuel.h"
 #include "sensor.h"
 
+#if EFI_LAUNCH_CONTROL
```
mck1117, line 22: you don't need to conditionally include, only conditionally call below

firmware/controllers/algo/launch_control.cpp
```diff
@@ -24,7 +24,11 @@
 #include "engine_state.h"
 #include "advance_map.h"
 
+static bool isInit = false;
 static Logging *logger;
+static efitick_t launchTimer = 0;
```
mck1117, line 29: move this state inside the launch control instance, right now it's not testable!

firmware/controllers/algo/launch_control.cpp
```diff
@@ -198,20 +199,25 @@ void setDefaultLaunchParameters(DECLARE_CONFIG_PARAMETER_SIGNATURE) {
...
+		*limitedSpark = false;
+		*limitedFuel = false;
 	}
 }
 
 void initLaunchControl(Logging *sharedLogger DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	logger = sharedLogger;
-	retardThresholdRpm = RETART_THD_CALC;
-	Launch.Start();
+	retardThresholdRpm = CONFIG(launchRpm) +\
```
mck1117, line 217: sloppy copy paste `\`?

firmware/controllers/algo/launch_control.h
```diff
@@ -29,5 +30,5 @@ class LaunchControlBase {
 	bool isLaunchConditionMet(int rpm) const;
 
 private:
-	efitick_t launchTimer = 0;
+
```
mck1117, line 33: why did this get moved out of the class...?

unit_tests/tests/test_launch.cpp
```diff
@@ -131,3 +131,64 @@ TEST(LaunchControl, CombinedCondition) {
 	EXPECT_FALSE(dut.isLaunchConditionMet(3200));
 
 }
+
+TEST(LaunchControl, CompleteRun) {
+	bool spark, fuel;
+	WITH_ENGINE_TEST_HELPER(TEST_ENGINE);
+
+	extern LaunchControlBase lc;
```
mck1117, line 139: uhhhh

firmware/controllers/algo/launch_control.cpp
mck1117: I think what he meant to do is add a new function, `void updateLaunchConditions()` that simply does this:

```
void updateLaunchConditions() {
    launchInstance.update();
}
```

firmware/controllers/algo/launch_control.cpp
mck1117: Yes, changing config at runtime is always wrong (unless it's some "set config x" command)

shadowm60: it is not, i moved inside launchControl class, since we use it as a simple variable.

firmware/controllers/algo/launch_control.cpp
shadowm60: do we have a max length accepted for a line? i wanted to keep shorter so it can be visible on the screen

unit_tests/tests/test_launch.cpp
shadowm60: do you have a better way to do this? i could not find one.

firmware/controllers/algo/launch_control.cpp
mck1117: you only need the `\` when in a macro - normal code can span multiple lines

firmware/controllers/algo/launch_control.cpp
```diff
@@ -198,20 +195,26 @@ void setDefaultLaunchParameters(DECLARE_CONFIG_PARAMETER_SIGNATURE) {
...
+
 }
 
 void applyLaunchControlLimiting(bool *limitedSpark, bool *limitedFuel DECLARE_ENGINE_PARAMETER_SUFFIX) {
-
-	if (retardThresholdRpm < GET_RPM()) {
+	if (( engine->isLaunchCondition ) && ( retardThresholdRpm < GET_RPM() )) {
 		*limitedSpark = engineConfiguration->launchSparkCutEnable;
 		*limitedFuel = engineConfiguration->launchFuelCutEnable;
+	} else {
```
mck1117, line 205: I don't think you want this else case.  This implementation will break the normal rev limiter.

mck1117: Look at how it's called:
image

firmware/controllers/algo/launch_control.cpp
```diff
@@ -198,20 +195,26 @@ void setDefaultLaunchParameters(DECLARE_CONFIG_PARAMETER_SIGNATURE) {
...
 	}
 }
 
 void initLaunchControl(Logging *sharedLogger DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	logger = sharedLogger;
-	retardThresholdRpm = RETART_THD_CALC;
-	Launch.Start();
+	INJECT_ENGINE_REFERENCE(&launchInstance);
+
+	retardThresholdRpm = CONFIG(launchRpm) + (CONFIG(enableLaunchRetard) ?
```
mck1117, line 215: this isn't necessary, we're guaranteed a periodic fast callback execution before the engine starts

firmware/controllers/algo/launch_control.h
```diff
@@ -25,9 +26,10 @@ class LaunchControlBase {
 	bool isInsideTpsCondition() const;
 	bool isInsideSwitchCondition() const;
 	bool isInsideRPMCondition(int rpm) const;
-
 	bool isLaunchConditionMet(int rpm) const;
 
 private:
-	efitick_t launchTimer = 0;
+	efitick_t launchTimer;
```
mck1117, line 32: does this need to be initialized to 0?

unit_tests/tests/test_launch.cpp
mck1117: if you make `applyLaunchControlLimiting` a member function of `LaunchControlBase`, then you can construct an instance and call it directly without having to set up everything else and use `extern`

firmware/controllers/algo/launch_control.cpp
shadowm60: You are right, i've got that because of unit_test, but i have to modify

firmware/controllers/algo/launch_control.h
shadowm60: it dose not mater, we cannot use the value without it being initialized before

unit_tests/tests/test_launch.cpp
shadowm60: i did it with inject in the init function, is that acceptable?

shadowm60:
should i rebase? i noticed a conflict, i think i did all the review points fixes.

rusefillc:
rebase or merge you need to resolve conflict

PR #2005, opened 2020-12-03
switch rpm calculator to use timer class

mck1117:


firmware/controllers/engine_cycle/rpm_calculator.cpp
```diff
@@ -348,16 +350,18 @@ void tdcMarkCallback(
...
 
 	/**
 	 * even if we use 'getOneDegreeTimeUs' macros here, it looks like the
 	 * compiler is not smart enough to figure out that "A / ( B / C)" could be optimized into
 	 * "A * C / B" in order to replace a slower division with a faster multiplication.
 	 */
 	int rpm = GET_RPM();
-	return rpm == 0 ? NAN : timeSinceZeroAngleNt / getOneDegreeTimeNt(rpm);
+
+	float oneDegreeSeconds = (60 / 360) / rpm;
```
andreika-git, line 362: Don't we need a static_cast to float here? We have all integers in rvalue...

mck1117: good catch, fixed!

mck1117:
depends on #2023

Issue #1733, opened 2020-08-29
OBD CAN sensors


rusefillc:
For TCU, BCU, piggyback and even dual rusECU use-case we shall read sensor data from CAN

At the moment we have something for this:
CanSensorBase - hopefully could be used for ODB-II sensor input
nasty canMap for custom rusEFI CAN

We need support for many/most basic sensors from RPM to timing angle

mck1117:
should we really exert effort on this when we could pretty easily support (with almost new code written) reading broadcast CAN sensors? that would cover most piggyback scenarios, and would certainly cover multi-rusefi scenarios.  The performance is also much better - you don't have to wait for a response

rusefillc:
You bring valid points for sure but I believe that OBD CAN sensors are a very cheap way to use rusEFI as flexible BCM for non-rusEFI installations, I would like to POC that

rusefillc:
bare minimum implementation works

PR #2013, opened 2020-12-05
Revert "OBD CAN sensors #1733"

mck1117:
This reverts commit fff58289ee791e9832dd963b4cd33f7afa61f7f7.

this change wasn't necessary - SensorRegistryEntry never leaves this file


rusefillc:
why revert? regardless of leaving file, explicit separation of get and set use-cases helps in my opinion

i am starting to recall the story: I was unable to comprehend the class in it's original form. this form with explicit helper methods is way easier for me to read

PR #2019, opened 2020-12-05
Fixes after clenup

dron0gus:


firmware/console/console_io.cpp
```diff
@@ -63,6 +63,10 @@ extern SERIAL_USB_DRIVER EFI_CONSOLE_USB_DEVICE;
 
 static bool isSerialConsoleStarted = false;
 
+#if (defined(EFI_CONSOLE_SERIAL_DEVICE) && ! EFI_SIMULATOR)
+static event_listener_t consoleEventListener;
```
rusefillc, line 67: please elaborate, why do we add a field?

dron0gus: Because we have:
```
#elif (defined(EFI_CONSOLE_SERIAL_DEVICE) && ! EFI_SIMULATOR)
....
		chEvtRegisterMask((event_source_t *) chnGetEventSource(EFI_CONSOLE_SERIAL_DEVICE), &**consoleEventListener**, 1);
		isSerialConsoleStarted = true;
```

dron0gus: consoleEventListener was removed during clean-up. Seem compiler reported it as defined but not used.

PR #2021, opened 2020-12-05
ADC: correctly calculate cache invalidate range for STM32F7xxx

dron0gus:
Bug was introduced in 2312b0ae5755e47b0963586f3ae717537c57da4d when
sample buffer was moved out of class. So sample become adcsample_t*
instead of adcsample_t[] and sizeof returns 4.


mck1117:
Use `efi::size` instead of macro?

PR #2023, opened 2020-12-05
round rpm instead of truncating

mck1117:


firmware/controllers/engine_cycle/rpm_calculator.cpp
```diff
@@ -133,7 +133,7 @@ bool RpmCalculator::checkIfSpinning(efitick_t nowNt) const {
 void RpmCalculator::assignRpmValue(float floatRpmValue) {
 	previousRpmValue = rpmValue;
 	// we still persist integer RPM! todo: figure out the next steps
-	rpmValue = floatRpmValue;
+	rpmValue = floatRpmValue + 0.5f;
```
rusefillc, line 136: please add a comment "We believe that rounding is better than truncating" or something along these lines?

rusefillc:
@mck1117 it's failing tests i think?

firmware/controllers/engine_cycle/rpm_calculator.cpp
mck1117: I mean, rounding is always better than truncating (when operating with real world physical quantities, at least), but comment added/modified.

PR #2025, opened 2020-12-06
switch trigger decoder to use timer class

mck1117:
depends on #2005


rusefillc:
@mck1117 poke

rusefillc:
@mck1117 poke, forward or close?

PR #2030, opened 2020-12-06
maybe fix sensor init checks

mck1117:
legacy configs had `EFI_ADC_NONE` at a high value (17 or something) instead of a low value.  This means they'd try to get configured anyway, which doesn't work very well.


mck1117:
github is broken, closing and retrying

PR #2038, opened 2020-12-07
start fuel pump as soon as trigger events happen

### LINKED ISSUE #1961: Fuel pump should turn on at the first sign of engine rotation
image

We were cranking for 0.7 second before we got valid RPM.  When we started injecting fuel, the pump hadn't caught up yet, so fuel pressure dropped almost 150kpa as the engine started.  Had the pump been turned on at the first sign of a trigger event occuring, the pump would already be up to speed by the time we started injecting fuel.  With compensation turned on I get a noticable improvement in starting quality since the pressure drop is compensated by additional pulse width.  That same benefit is available if we don't let the pressure drop in the first place.

mck1117:
fix #1961


mck1117:
Look at all that time the pump was on before injecting!

image

rusefillc:
unit test? :)

mck1117:
depends on #2050

PR #2040, opened 2020-12-07
Split ETB in to interface + implementation headers

mck1117:
reduce the header include footprint a bit for people who aren't the ETB controller cpp itself


mck1117:
also fix the broken test in master

PR #2042, opened 2020-12-08
remove efifeatures options for cheap features (boost & launch)

mck1117:
Working on an attempt to simplify efifeatures, so that boards don't need a ton of duplication of options.

These two features use neither significant memory nor cpu.  Not much reason to ever turn them off.


rusefillc:
@andreika-git might say that all features matter for smaller RAM use-cases?

rusefillc:
@andreika-git do you vote Y or N?

mck1117:
I think it's about 250 bytes of RAM total.  Most of that is because our PWM structures are sort of huge.

andreika-git:
I like "#if xxx" concept for small MCUs. I think we should keep the possibility to minimize the size of the "kernel" firmware via efifeatures. For example, we still may use smaller (and cheaper) mcus for co-processing, based on our unified firmware codebase.

mck1117:
Why not guard it behind EFI_ENGINE_CONTROL then?

andreika-git:
IMHO EFI_ENGINE_CONTROL should contain only the most essential core features for running an engine. Even idle control has its own EFI_IDLE_CONTROL... Maybe we can use some kind of a new EFI_ADVANCED_ENGINE_CONTROL which would be common for all smaller features like launch/boost...

rusefillc:
I like the idea to separate EFI_BARE_MINIMUM_ENGINE_CONTROL from EFI_RICH_ENGINE_CONTROL

rusefillc:
but really, so many things go before launch and boost on engine control pyramid. i have a feeling that we shall keep everything 100% as is to be honest

mck1117:
@andreika-git here's a crazy idea: #2090 

Only thing required to enable features this way is to compile them.  No preprocessor at all, all done by the linker.  Then adding it to a particular board is just `include packages/boost/boost.mk` in the corresponding `board.mk`, instead of making it hard on the preprocessor.

960:
Is it possible to make selective makefiles somehow?

Say you define what features you want in board config.
Then you only compile the files needed for those features.

rusefillc:
@mck1117 merge conflict and build errors

mck1117:
@960 that's exactly what #2090 is

PR #2045, opened 2020-12-08
start peeling out ts commands

mck1117:


rusefillc:
@mck1117 unit test compilation error?

mck1117:
fixed!

PR #2046, opened 2020-12-08
Prepare configurations for SW knock detection

shadowm60:
- Configurations needed for base noise
- TS usability for base noise definition

### DIFF
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -1455,8 +1455,10 @@ uint8_t[4] unusedsw;;"units", 1, 0, -20, 100, 0
 
 	pin_output_mode_e[TCU_SOLENOID_COUNT iterate] tcu_solenoid_mode;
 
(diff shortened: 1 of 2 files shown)


shadowm60:
I've got Fail on build-primary-bundle with the following error:
    [junit] Running com.rusefi.FullServerTest
    [junit] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 62.056 sec

I have no idea what to do to make it pass.

rusefillc:
@shadowm60 that's probably sporadic unrelated issue, running again just to try

mck1117:
it probably actually needs a wider range - the noise floor on a disconnected sensor is down around -60dbv

shadowm60:
I belive that in the base noise we should not consider -60 dB, but if i lt makes sens  i can change the scale.

mck1117:
It's been a while since I looked, but I seem to remember seeing a log from @960 with a noise floor down around -35 or -40 dbv

Issue #2044, opened 2020-12-08
second wideband sensor


rusefillc:
A few things should happen:

- Support a second wideband sensor
- Select which bank each cylinder is on (we can't guess because different MFGs use different cylinder numbering schemes)
- Run a separate short term fuel instance on the second bank/sensor

mck1117:
this is now done after #2661

PR #2047, opened 2020-12-09
config for second lambda

mck1117:
progress on #2044


rusefillc:
@mck1117 " "TS output channels size mismatch""

PR #2050, opened 2020-12-10
fix mock pins array

### LINKED ISSUE #2026: Test start.startStop fails at -O2 in clang
The start stop test fails when optimization is enabled, but passes when it's off.  That's a __strong__ suggestion that we're invoking undefined behavior, which is probably also the source of whatever our start stop bug is.

mck1117:
fix #2026


rusefillc:
i do not get it :( sorry

what bug does it fix how? functional or compiler unhappiness?

also something fails to compile

mck1117:
the old size of the array was `1 << sizeof(brain_pin_e)`.  This is going to be `1 << 1`, aka 2 bytes.  But we wanted 256! This change makes it `1 << 8 * sizeof(brain_pin_e)`, aka 256.

PR #2052, opened 2020-12-10
log second AFR to sd

mck1117:


mck1117:
now tested on hardware

PR #2068, opened 2020-12-12
Fsio eval check

shadowm60:


mck1117:
do we want to fatalError anyway in unit tests? we have `EXPECT_FATAL_ERROR(...)` to be able to check if a fatal error was thrown or not

shadowm60:
What you guys think is better, nullptr is that no usable command found, but additional to that we throw error to signal user.... Shall i add that or it is fine like this?

.vscode/launch.json
```diff
@@ -56,6 +56,19 @@
             "svdFile": "${workspaceRoot}/firmware/ext/cmsis-svd/data/STMicro/STM32F427.svd",
             "servertype": "openocd",
             "configFiles": ["interface/stlink.cfg", "target/stm32f4x.cfg"],
+            "runToMain": true
+		},
+		{
+            "name": "Debug F7x6",
+            "type": "cortex-debug",
```
rusefillc, line 63: is this change supposed to be in this PR?

rusefillc:
I agree with @mck1117 that unit test would be nice but I do not have ideas on how to make this happen

so tests could be a nice thing for later and maybe merge as is without JSON change?

.vscode/launch.json
shadowm60: I left it there since i used it for debugging and we had no support for f7, i ca make it in a new pr, just wanted to reduce you're load and left it there

rusefillc: oh if that's useful stuff let's keep it here

shadowm60:
JSON i can place it in a new PR since it would make my life easier.

rusefillc:
merged since 80% of the benefits for 20% of the price

mck1117:
see #2071 for a test

Issue #2065, opened 2020-12-12
MRE issue with tle8888 - GP3 does not work?


shadowm60:
with sw 20201008 in my case GP3 was used as output (GPPWM).
with SW since TLE driver update i had issues but with latest nightly i had time to investigate.
I have excluded GPPWM and FSIO from the scheme since i did the following:
assign GP3 as ignition1, and kept ignition2 for ignition, then started simualtor and logged pin activity:
This is the output:
https://i.imgur.com/z0nnFd3.png
reverted to the old SW to ensure that hw is ok and TS config as well and the putput was ok:
https://i.imgur.com/SgirIy0.png

With a short debug session i noticed that we hit periodic  reinit request from tle8888_spi_validate since we keep getting reg 0 as a response.

rusefillc:
is that F4 or F7 MRE @shadowm60 ?

rusefillc:
i think QC build reproduces the problem on F4 - GP1 & GP2, LS1 & LS2 do blink but GP3 does NOT blink!

shadowm60:
I have mre 0.3 it is f7

dron0gus:
@shadowm60 
Can you post output of "tle8888" and "pins" commands when you hit this issue?
I'm trying to reproduce this on MRE 0.5.0 with F4 with no luck yet.
Do you see this issue with engine_type 31 and self stimulation?

shadowm60:
@dron0gus  How would you like the outputs? console? or how should i prepare it?

shadowm60:
Changed engine config to 31
https://i.imgur.com/NJboKp4.png 
Ch1 is gp3 and nothing on that line, the ch2 is ignition, just to see if the ecu is running with self stim.

020-12-13_18_09_57_205: EngineState: SPI clock on PB3
2020-12-13_18_09_57_207: EngineState: SPI master out on PB5
2020-12-13_18_09_57_208: EngineState: SPI master in  on PB4
2020-12-13_18_09_57_210: EngineState: SPI clock on PC10
2020-12-13_18_09_57_211: EngineState: SPI master out on PC12
2020-12-13_18_09_57_212: EngineState: SPI master in  on PC11
2020-12-13_18_09_57_214: EngineState: tle8888 CS on PD5
2020-12-13_18_09_57_215: EngineState: SD CS on PB9
2020-12-13_18_09_57_216: EngineState: Injector 1 on ext:TLE8888.OUT6
[...]

PR #2079, opened 2020-12-14
TLE8888: temp fix of #2065

dron0gus:


mck1117:
do we need to check that at all? why not just always update the outputs, and let it be a noop if the watchdog has expired?

dron0gus:
@mck1117 
1. this is useless on WD-enabled chips because:
2. we do chip init after WD becomes happy... because on WD-enabled chips fired WD cause all setting reset...
3. donno why un-happy if we enable WD on non-WD chip and it becomes un-happy it does not reset settings and does not disable all outputs...

PR #2080, opened 2020-12-15
Logic analyzer ts report

shadowm60:
Reports measured freq and duty in TS debug channel:
f1-f4 duty cycle in %

i1-i4 frequency as interest

just a refresh of pins that could be used:
We might need to document this, but this is what i found for f4
hwPin == GPIOA_8
hwPin == GPIOA_9 
hwPin == GPIOE_9 
hwPin == GPIOE_11
hwPin == GPIOA_1
hwPin == GPIOA_5 
hwPin == GPIOA_15|
hwPin == GPIOB_3
hwPin == GPIOA_6
hwPin == GPIOA_7 
hwPin == GPIOB_4 
hwPin == GPIOB_5
hwPin == GPIOC_6
hwPin == GPIOC_7


rusefillc:
thank you!

at the moment logic_analyzer is limited to ICU but there is no reason for that to be the case. Let me create a follow-up ticket

rusefillc:
#2081 created

Issue #2093, opened 2020-12-18
More explicit handling of CPU exhaustion


rusefillc:
S2R:
build Discovery debug 
execute HighRevTest

Result: blue LED no longer blinking, USB comms down, but firmware seem to still be running

Expected result: helpful explicit critical error

PR #2097, opened 2020-12-18
force a pwm cycle start (skip cycles) if late

mck1117:
If we're far behind schedule on PWM, skip cycles to catch up instead of executing every single event we missed.  

Trying to catch up will just swamp the scheduler while generating extremely high speed PWM (I measured it at ~50-70khz) that isn't actually doing anything useful.  For example, if you were trying to generate a 60-2 wheel at 6000 rpm, that's 12000 events per second, and if the flash erase + write takes 3 seconds, that leaves you with ~36k events to try and execute once the flash is done, which will be cranked out as fast as the PWM driver can toggle a pin.


mck1117:
related to #2093

PR #2098, opened 2020-12-18
Fix gpio reinit race condition

mck1117:
Bug: if you un-initialize a pin, setValue will still work! If you initialize pin A, then set it to "none", then initialize the same pin elsewhere, both users will be able to call setValue and set the same pin!  This causes a problem in particular when reinitializing a pin that's set by a higher priority thread, like the hardware trigger emulator does.

This PR fixes both possibilities:
- Only do the hardware set if `brainPin != GPIO_UNASSIGNED`, which is now always set by `initPin`, and always cleared by `unregister`.
- Guard pin init and de-init with a critical section so that other threads couldn't even set the pin if they tried

### DIFF
--- a/firmware/controllers/algo/engine.h
+++ b/firmware/controllers/algo/engine.h
@@ -195,12 +195,6 @@ public:
 	 */
 	efitimems64_t callFromPitStopEndTime = 0;
 
-	/**
-	 * This flag indicated a big enough problem that engine control would be
-	 * prohibited if this flag is set to true.
-	 */
-	bool withError = false;
-
 	RpmCalculator rpmCalculator;
 	persistent_config_s *config = nullptr;
 	/**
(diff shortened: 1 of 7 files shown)


mck1117:
This also runs more of the GPIO init/deinit code in unit tests, which paves the way for unit testing this stuff...

firmware/controllers/algo/engine.h
```diff
@@ -195,12 +195,6 @@ class Engine final : public TriggerStateListener {
 	 */
 	efitimems64_t callFromPitStopEndTime = 0;
 
-	/**
-	 * This flag indicated a big enough problem that engine control would be
-	 * prohibited if this flag is set to true.
-	 */
-	bool withError = false;
```
mck1117, line 202: the value of `withError` was never read!

PR #2103, opened 2020-12-18
Test Idle PWM 

shadowm60:
Using existing hw jumpers we measure idle pin pwm.

Duty is not yet checked, but we do check if we have correct output frequency.

### DIFF
--- a/java_console/autotest/src/com/rusefi/PwmHardwareTest.java
+++ b/java_console/autotest/src/com/rusefi/PwmHardwareTest.java
@@ -37,7 +37,6 @@ public class PwmHardwareTest {
         ecu.sendCommand(CMD_IGNITION_PIN + " 2 none");
         ecu.sendCommand(CMD_IGNITION_PIN + " 3 none");
 
-        //ecu.sendCommand(CMD_TRIGGER_PIN + " 0 PC7");
         ecu.sendCommand(CMD_TRIGGER_PIN + " 1 PA8");
 
         /* DBG_LOGIC_ANALYZER */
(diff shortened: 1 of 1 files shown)


java_console/autotest/src/com/rusefi/PwmHardwareTest.java
```diff
@@ -0,0 +1,65 @@
...
+        /* save these for last to ensure logic is started */
+        ecu.sendCommand(CMD_LOGIC_PIN + " 0 PA5");
+        ecu.sendCommand(CMD_WRITECONFIG);
+        sleep(2 * Timeouts.SECOND);
+        ecu.sendCommand(getEnableCommand(Fields.CMD_SELF_STIMULATION));
+
+        sleep(2 * Timeouts.SECOND);
+
+        /* +-1% is still acceptable */
+        EcuTestHelper.assertEquals("Idle PWM freq", 160, SensorCentral.getInstance().getValue(Sensor.debugIntField1),0.01);
```
rusefillc, line 57: 1% of 160 integer is "1"? that would probably be just fine here.

java_console/autotest/src/com/rusefi/PwmHardwareTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+
+        if (ControllerConnectorState.firmwareVersion == null)
```
rusefillc, line 60: please remove this - this is not related to this test, one test having this is enough

java_console/autotest/src/com/rusefi/PwmHardwareTest.java
shadowm60: since it is and int at the end, yes

java_console/autotest/src/com/rusefi/PwmHardwareTest.java
rusefillc: https://github.com/rusefi/rusefi/commit/7492172722ccb3c0f3115eede34eaa169b647b40

PR #2106, opened 2020-12-18
Partition configs by board type

### LINKED ISSUE #2101: only build in config for the correct board for unit tests and MRE
we can build less configurations into each binary

mck1117:
fix #2101

This should remove one of the major opportunities for accidental ECU bricking.


rusefillc:
we have conflicts @mck1117

mck1117:
conflicts fixed!

firmware/controllers/algo/engine_configuration.cpp
```diff
@@ -1080,8 +1080,8 @@ void setDefaultFrankensoConfiguration(DECLARE_CONFIG_PARAMETER_SIGNATURE) {
 //	engineConfiguration->gps_rx_pin = GPIOB_7;
 //	engineConfiguration->gps_tx_pin = GPIOB_6;
 
-	engineConfiguration->triggerSimulatorPins[0] = GPIOD_1;
-	engineConfiguration->triggerSimulatorPins[1] = GPIOD_2;
+	engineConfiguration->triggerSimulatorPins[0] = GPIOB_14;
```
rusefillc, line 1083: why and what's that? would I need to change my HW CI jumpers? W?

rusefillc: @mck1117 ?

mck1117: you don't, that snuck in because I was testing on an MCU that doesn't actually have D1/2

PR #2111, opened 2020-12-21
simplify gpio "is init" logic

mck1117:


firmware/controllers/system/efi_gpio.cpp
```diff
@@ -358,15 +358,11 @@ OutputPin::OutputPin() {
...
 bool OutputPin::isInitialized() {
-#if EFI_GPIO_HARDWARE && EFI_PROD_CODE
-#if (BOARD_EXT_GPIOCHIPS > 0)
-	if (ext)
-		return true;
-#endif /* (BOARD_EXT_GPIOCHIPS > 0) */
-	return port != NULL;
-#else /* EFI_GPIO_HARDWARE */
+#if EFI_PROD_CODE
+	return brainPin != GPIO_UNASSIGNED;
```
rusefillc, line 362: why does this work for TLE chips? let's add a comment?

mck1117: it works because brainPin is always configured, whether it's an external chip or not

rusefillc: oh, brain pin not port

I am paranoid about change. this would require manual testing on MRE and that's not something I am excited about :(

mck1117: the same test is already performed when actually setting the pin in `setValue`:

image

mck1117: And `OutputPin::initPin` unconditionally sets `brainPin` (no matter if unit test, on chip, external chip, etc), and `OutputPin::deInit` unconditionally sets it to `GPIO_UNASSIGNED`.

rusefillc:
Q1: do we have spaghetti of all sorts?
Q2: would you bet your life that it works, or is there a possibility of unexpected?

mck1117:
well, #2098 helped de-spaghettify a bit

actually, this is testable - let me add a few tests

rusefillc:
@dron0gus @andreika-git what's your take on this?

rusefillc:
@dron0gus ?

@andreika-git ?

dron0gus:
Looks ok for me. Can test this in my custom board half of a day later.

firmware/controllers/engine_cycle/spark_logic.cpp
```diff
@@ -68,13 +68,6 @@ static void fireSparkBySettingPinLow(IgnitionEvent *event, IgnitionOutputPin *ou
 	output->setLow();
 }
 
-// todo: make this a class method?
-#define assertPinAssigned(output) { \
-		if (!output->isInitialized()) { \
```
rusefillc, line 73: I do not get it. What happens with "pin should have a pin" validation?

rusefillc:
@mck1117 poke "What happens with "pin should have a pin" validation?"

firmware/controllers/engine_cycle/spark_logic.cpp
rusefillc: @mck1117 poke. Do we want to move this PR forward?

rusefillc (changes requested):

rusefillc:
@mck1117 poke? forward or close?

rusefillc:
@mck1117 numerous pokes over numerous months - closing

Issue #2096, opened 2020-12-18
console functional tests don't work


mck1117:
freshly downloaded console from build server:

![image](https://user-images.githubusercontent.com/568254/102576339-25ef7480-40aa-11eb-83c5-4df7215e6359.png)

rusefillc:
yep, that's due to JUnit conversion :(

i think I know how to fix

rusefillc:
Maybe use ant to scan for tests and write list of tests into .jar manifest property?

like http://ant.apache.org/manual/Tasks/pathconvert.html

mck1117:
done in #2132

PR #2114, opened 2020-12-22
use junit fail instead of throw

mck1117:
we should use junit assertions instead of custom ones, and that starts with `fail()`


rusefillc:
That's related to #2096 and overall (unclear?) big picture of how we want this to work :)

mck1117:
Running junit from the command line is easy enough - it's just a few lines to invoke junit the same as ant is invoking it today.  Besides, the junit `fail` function does just throw an exception, it's just a particular one that junit is listening for so that it can provide better diagnostics than us throwing our own exception.

rusefillc:
@mck1117 we need command line fixed before we improve anything :( and I am not available to SW for another couple of weeks :(

PR #2129, opened 2020-12-24
fix race condition in CommandQueue, massively speeding up HW CI tests

mck1117:
The pattern of 
```
send request
wait for response or timeout
```
doesn't actually work, because if the "signal listening thread" action happens elsewhere prior to you starting to listen, you get stuck.  This had the effect of waiting for the timeout (10 seconds for `set engine_type`!) instead of the command actually completing!

This PR changes it to instead
```
start listening for response
send request
wait for completion or timeout
```
which means we actually return immediately upon confirmation of the command as intended.

It also removes the retry timeout, as any connectivity problem that requires a retry indicates a bug, and we should just fail hard so the issue can be diagnosed (besides, the retry mechanism was sort of broken anyway).


rusefillc:
not everyone has USB, physical TTL is kind of a use-case for retry? but yes it's probably broken and untested anyway

rusefillc:
Great stuff! 👍 👍 👍

PR #2135, opened 2020-12-25
Faster config generate

mck1117:
less stdout -> more faster

saves about 30-40% on linux, probably more on windows


rusefillc:
i think it's better to make this logging conditional or worst case comment it out but not just flat out remove.

from my experience it's really helpful while changing or troubleshooting code generator which is a quite a pile of legacy code at this point.

mck1117:
closing in favor of rewriting the config parser

Issue #1263, opened 2020-04-04
Idle controller should use open-loop config while in automatic mode


mck1117:
You lose all of the goodness when switching to automatic mode - ideally it should be more like the ETB controller, where there's some open loop (feedforward) math doing most of the heavy lifting, with closed loop just for icing on the cake.

mck1117:
done with #2745

PR #2140, opened 2020-12-26
more testable idle logic

mck1117:
logic to detect driver intent about whether we're idling or not

progress on #1263


mck1117:
note: this isn't actually called yet, but will be soonish

PR #2156, opened 2020-12-29
Idle timing PID uses phase logic

mck1117:
Refactor idle timing PID logic in to the idle controller class, using the phase determination logic to ensure idle timing and normal idle pid are engaged together


mck1117:
depends on #2152

mck1117:
tested on real hw, works great!

rusefillc:
It looks like this PR has removed some of the functionality?

It would be not nice to remove functionality under "refactor" umbrella, it is my opinion that not enough clarity was provided here. Such unilateral actions are breaking my trust in PRs, this would require more attention in the future thus more resources :(

rusefillc:
At the moment I am considering a revert since I have a major problem with how this went.

rusefillc:
https://github.com/rusefi/rusefi/commit/46583b2a199e95a3bd0aa24d2bf2a1bbf0e8f28c

Reverted since feature removal was not handled properly.

Please have more explicit announcement for feature removals, please do not remove features under "refactor" banner.

Feature removal PR has to include CHANGELOG entry

PR #2165, opened 2020-12-31
Haltech as dash option

shadowm60:


rusefillc:
@shadowm60 need your help here:

"Build Firmware / build-primary-bundle" has failed while other tasks have not failed. Looks like other tasks were invoking gen_auto while "Build Firmware / build-primary-bundle" was not? Need to look into this and fix "Build Firmware / build-primary-bundle" task so it is same smart as other tasks?

shadowm60:
I see that auto_generated_enums.cpp is not being regenerated.... should i regen locally and pus it as well?

I regenerated the file with gen_enum_to_string.sh and i have CAN_BUS_Haltech in my switch.

rusefillc:
No, this is not the best plan.

Would you be able to call me? whatsapp or slack voice?

shadowm60:
Now it is all green.

rusefillc:
Great stuff thank you @shadowm60 for fixing this the right way!

PR #2169, opened 2021-01-01
Idle timing uses phase logic, remove falloff

mck1117:
- Refactor idle timing PID logic in to the idle controller class, using the phase determination logic to ensure idle timing and normal idle pid are engaged together. (this removes the need for independent "is idling" config fields for spark)
- Removes idle timing PID falloff based on falloff region both in RPM and TPS.
- ~Likewise, removes the deadzone for idle timing pid.~

The reasoning for the last two:
Falloff has issues because it reduces the effective gain of the controller when in the falloff region, which can encourage windup when the gain suddenly increases when it makes it in to the un-falloff'd region.

~A deadzone causes the opposite problem where the engine speed changes, but the controller doesn't respond because it's still inside the deadzone.  Then suddenly the engine leaves the deadzone, the controller wakes up, but realizes it's been caught with its pants down because it did nothing to respond while inside the deadzone.  This hurts response because the controller *could* have responded while in the deadzone, but didn't.~ There is a reasonable case for deadzone, un-deleted!


mck1117:
@andreika-git FYI

andreika-git:
The idea behind the timing PID deadzone is that is should be very narrow, like 10..50 RPM. It would not affect the sudden RPM change, but it allows to set more stiff PID coeffs which otherwise would cause oscillations because of the noisy RPM fluctuations.

mck1117:
> It would not affect the sudden RPM change

But that's exactly what it does - enabling it means the controller can't catch a quick drop until it's too late.  Adding nonlinearity to try and fix an instability is not a good solution.  More often than not, it just adds a new instability in a different spot.

If you have a P gain of 0.1, that means 10 degrees per 100 RPM.  A deadband of 50 would mean that it can't apply less than 5 degrees of timing, which is a LOT of torque at idle speed on most engines.  If there was a drop you could've caught early with 2 degrees, it'll drop further then get slapped in the face with 6 degrees instead.  The spark timing should be the one handling exactly those 10..50 degree variations, with the mechanical (airflow) idle control handling anything that needs more than that.

mck1117:
I think there might be a case for deadband on the airflow controller, but not for spark.

rusefillc:
I do not know how to make the decision on merge.

Is there a chance to split this PR into two smaller PRs to split the controversial from less controversial , removal from pure refactoring?

mck1117:
this isn't blocking anything else, so I don't mind it sitting around discussing

andreika-git:
> But that's exactly what it does

If it doesn't work on you car or makes it worse - then just don't use it, set it to zero (the default value). But it works on my car, and on @darxfame's car as well. This feature didn't appear just for fun, that's the result of some testing including daily use, and in comparison to OEM ECU's behavior. 
Btw, here's the screenshot of the OEM ECU with the deadzone in the timing idle PID-regulator:
image
And here's another aftermarket ECU:
image

mck1117:
interesting - all the OEM calibrations I've seen look something like this:

image

This is the stock GM calibration for my V8 - it has no deadzone, and is actually most aggressive near zero error.  It's possible that some of the effect is from different update rates of the RPM calculation.  I wouldn't be surprised if the GM ECU was calculating RPM on every trigger tooth (or quarter of a turn) and applying it on the very next cylinder's firing, so it can respond extremely quickly to individual events, increasing the bandwidth of the system.

Anyway, that's enough data to convince me that deadzone is worth leaving in.

mck1117:
@andreika-git, you happy with the changes now?

andreika-git:
Good to go!

PR #2175, opened 2021-01-02
Compute injection duration later in the pipeline

mck1117:
progress on #2044 

This moves the mass -> duration calculation much later so that:
- per-bank trim can be applied
- wall wetting math works correctly when fuel pressure varies (for the same mass, the pulse width could change, but that change shouldn't apply a wall wetting correction!)


rusefillc:
poke @mck1117

rusefillc:
@mck1117 what's the plan here :)

mck1117:
The plan here is that PR hardware CI has caught a problem :)

mck1117:
found the bug, ready to go!

Issue #925, opened 2019-09-09
instant RPM for idle control


rusefi:
@stefanst is wondering if one of the ways to improve idle control would be to use instance RPM value instead of current per-cycle RPM value

There is a chance that per-cycle RPM takes too long to refresh and we are loosing valuable time

stefanst:
Correct. On a 1.6l Miata engine, coming off of 2000rpm, free-revving, we have 3 full crankshaft revolutions from rpm>1000 (which is the target) to a stalled engine. The "I" factor has only 2-3 true iterations to adjust.

rusefi:
Reminder that we have "INSTANT_RPM" debug mode which allows to monitor that instant RPM value vs "regular" RPM which is currently used more often

mck1117:
We use instant RPM already for closed loop idle timing correction, which is the part that really matters.

OrchardPerformance:
@mck1117 - Can this be closed? Do you feel we have resolved this one?

mck1117:
this is well past fixed, and instant rpm wasn't even the problem.

Issue #343, opened 2017-01-19
instant shaft acceleration


rusefi:
https://m.cyberleninka.ru/article/v/diagnostirovanie-dvs-po-pokazatelyam-uglovoy-skorosti-kolenchatogo-vala is another reason to imrove instant shaft acceleration calculation

See
	SC_RPM_ACCEL = 3,
	SC_DETAILED_RPM = 4,

Reported by: rusefi

mck1117:
we now calculate and log RPM rate of change in rpm/s

Issue #329, opened 2016-11-18
better RPM calculation logic


rusefi:
current RPM is calculated only once per 720 degrees, based on the duration of these 720 degrees - so RPM could be ~700 degrees late.

A precision improvement here would be to calculate RPM 
a) based on a sliding 720, not 720 cycle
or b) based on few last teeth, not full 720 cycle 

Reported by: rusefi

smonette-jive:
I'm not sure if the micro-controller has enough processing power to calculate the RPM at each crank trigger base on the time spent between the triggers. You will also need to modify the way you calculate ignition. From what I remember, the code calculates the ignition timers at the beginning of each cycle. You can fix this issue by calculating the ignition timers (modify future timers) at each crank trigger interrupt. This would also make the ignition a lot more accurate on acceleration and would also produce more power too.

OrchardPerformance:
Closing because instant RPM appears to be functional after merge #2180

PR #2180, opened 2021-01-03
fix instant rpm

mck1117:
instant RPM didn't work because multiple different systems were calling `calculateInstantRpm`, which doesn't work because that function changes the state of the instant RPM tracking.

This PR updates the instant RPM calculation from exactly one place (the rpm calculator), and just reads it everywhere else.

related to #343 and #925 and #329


mck1117:
Here's before - something that looks mostly right but has random drops to 0:

image

And here's after:

image

Red is instant rpm, white is normal "slow" RPM.  The ripples in the instant rpm are from me spinning the wheel manually with my fingers.  When running self stimulation or powered the wheel with a motor, it's totally smooth.

rusefillc:
@mck1117 did not we merge something? did we merge only first part?

mck1117:
we merged a fix to make the "dRPM" field work properly, this one is relatively unrelated

Issue #1642, opened 2020-07-26
60 pin breakout board does not say "60 pin" anywhere on the PCB


rusefillc:
it would improve usability if 60 pin board would have "60 pins" words somewhere on the surface

chuckwagoncomputing:
I fixed this too!

PR #2185, opened 2021-01-05
Fix EEC-IV

### LINKED ISSUE #1744: Ford 60 breakout inverted
Pin 60 on the board goes to pin 1 on the ford connector, 59 is actually 2, 58 is 3, and so forth.

chuckwagoncomputing:
This fixes #1744 and #1642


rusefillc:
I cannot explain it but tradition is to not delete old gerbers?

chuckwagoncomputing:
Ok done.
I should not that I am very confused about what to do with symbol and footprint libraries so that others can open them without errors. You might want to check to make sure I did it right.

PR #2186, opened 2021-01-05
generate triggers.txt from parametric test

mck1117:
now unhappy triggers show up as normal test failures:

image

Uses a Google Test parametric test to iterate over all trigger types, with a little helper to open/close the triggers.txt file before and after tests run.


mck1117:
what do you mean? that failure in the snip was failed on purpose to show what it looks like

rusefillc:
at the moment linear portion of https://rusefi.com/images/triggers/trigger_53.png does not look right

I was expecting linear section of https://rusefi.com/images/triggers/trigger_53.png to produce a fatal during generation? probably/maybe unrelated issue

mck1117:
fatals during the old trigger printout should have caused an app crash - they just throw an exception.  Gtest just catches it and prints out a nice test failure, the call from `main()` would just crash instead

rusefillc:
So it looks like whatever truncates linear section could be handled nicely enough not to cause a fatal. Open question if that's an issue or not but definitely a separate issue.

PR #2188, opened 2021-01-05
Adc isAdcChannelValid helper

dron0gus:


rusefillc:
@mck1117 @andreika-git please review

firmware/hw_layer/adc/adc_inputs.h
```diff
@@ -13,6 +13,21 @@
 
 #define SLOW_ADC_RATE 500
 
+static inline bool isAdcChannelValid(adc_channel_e hwChannel) {
```
mck1117, line 16: Why not just put this in the cpp file? We have LTO (link time optimization) enabled, so the `inline` keyword actually does nothing and the compiler will probably inline this anyway.

dron0gus: Whole adc_inputs.cpp file is covered with \#if HAL_USE_ADC. Adding this code before \#if will create more mess than inline definition inside .h file. Also I did not find any other suitable cpp for this code.
But I can move it inside adc_inputs.cpp if you think it is better.

mck1117: there shouldn't be any calls to isAdcChannelValid if HAL_USE_ADC is false though, right? Or do tests actually call it? Either way, I think the cpp is probably the right place for it so we can let the linker make the decision for us (also saves parse effort by not parsing the function and generating code every time it's included)

rusefillc:
merge quickly merge  often - let me merge since no major issues hopefully

Issue #540, opened 2018-01-22
flex fuel sensor


rusefi:
first step towards e85 EFI would be integrating with GM flex fuel sensor

OrchardPerformance:
@rusefillc @mck1117  - Has Gwendal driving around on this already? Do we feel this one is done?

CrazyStriker:
I'm driving using set stoich value of 10. need to install sensor still but I can start working on it ASAP

OrchardPerformance:
Thirsty. Would be cool to give this one a push as I don't think it will take much to get it over the line.

mck1117:
Automatic stoich compensation should work, but a real world test of it would be cool.

PR #2189, opened 2021-01-05
Implement flex sensor

mck1117:
#540


firmware/controllers/sensors/flex_sensor.cpp
```diff
@@ -0,0 +1,37 @@
...
+	// this long timeout is ok, ethanol content isn't going to change quickly
+	: StoredValueSensor(SensorType::FuelEthanolPercent, MS2NT(500))
+{
+}
+
+void FlexFuelSensor::onEdge(efitick_t nowNt) {
+	float frequency = 1 / m_edgeTimer.getElapsedSecondsAndReset(nowNt);
+
+	// TODO: are these limits reasonable?
+	if (frequency > 45 && frequency < 155) {
```
cookiesowns, line 31: Limits might be a bit broad, most people seem to settle at 49hz and 151hz for floor and ceil.

And signals above 175hz indicate a fault condition ( fuel containmination to trigger failsafe ), we can probably add it as a TODO

mck1117: Good to know. Doesn't really hurt to be a little generous here, as 0.3% is a little tight on the high end (though E100 is also rare).

cookiesowns: Yup, we should be fine with these values 👍 LGTM!

mck1117:
tested and working on real hardware!

PR #2199, opened 2021-01-07
better error checking of thermistor config

mck1117:
- check that thermistor config is actually valid (decreasing order resistance, increasing order temperature)
- check that resolved temperature is reasonable

### DIFF
--- a/firmware/controllers/sensors/converters/thermistor_func.cpp
+++ b/firmware/controllers/sensors/converters/thermistor_func.cpp
@@ -25,6 +25,12 @@ SensorResult ThermistorFunc::convert(float ohms) const {
 
 	float celsius = convertKelvinToCelcius(kelvin);
 
+	// bounds check result - please don't try to run rusEfi when colder than -50C
+	// high end limit is required as this could be an oil temp sensor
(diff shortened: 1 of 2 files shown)


firmware/controllers/sensors/converters/thermistor_func.cpp
```diff
@@ -25,6 +25,12 @@ SensorResult ThermistorFunc::convert(float ohms) const {
 
 	float celsius = convertKelvinToCelcius(kelvin);
 
+	// bounds check result - please don't try to run rusEfi when colder than -50C
```
rusefillc, line 28: s/rusEfi/rusEFI/

firmware/controllers/sensors/converters/thermistor_func.cpp
```diff
@@ -25,6 +25,12 @@ SensorResult ThermistorFunc::convert(float ohms) const {
...
+	// high end limit is required as this could be an oil temp sensor
+	if (celsius < -50 || celsius > 200) {
```
rusefillc, line 30: in an ideal world I would have those magic constants more explicitly defined on top of the file, not in the middle of random

PR #2205, opened 2021-01-09
gen enums on unittest build

chuckwagoncomputing:


rusefillc:
is it time to refactor some scripts which are often invoked together into one script?

PR #2210, opened 2021-01-09
Vss filter configuration field and param

shadowm60:
Prep work to introduce biquad filter on vss input to reduce noise on logs and for dynoview as well


rusefillc:
@mck1117 ?

mck1117:
yes, fine by me

PR #2212, opened 2021-01-09
Chibi os 20

dron0gus:


rusefillc:
ON_UNLOCK_HOOK  missing
MMC waitCounter missing

I really suggest manual review of compare https://github.com/rusefi/rusefi/blob/master/firmware/docs/chibios_rusefi_18.patch to https://github.com/rusefi/rusefi/blob/master/firmware/docs/chibios_rusefi_20.patch

ideally we shall review and fix https://github.com/rusefi/rusefi/blob/master/firmware/docs/rusEFI_custom_chibios.md

mck1117:
ChibiOS 20 is capable of rumbly v8 noisies:

https://www.youtube.com/watch?v=m-ExExuA8kQ

dron0gus:
Now compiles for Cypress. While Not tested on Cypress HW.

dron0gus:
Now should compile for Kinetis. Also not tested on real HW.

rusefillc:
@andreika-git @mck1117 how do we want to proceed?

I can test on v12 and merge.

mck1117:
I ran my engines on it the other day, so I'm happy. Go for it!

firmware/config/boards/hellen/cypress/efifeatures.h
```diff
@@ -250,7 +250,8 @@
 #define EFI_INTERNAL_FAST_ADC_GPT	&GPTD2
 #define ADC_MAX_CHANNELS_COUNT 32
 
-#define EFI_FASTER_UNIFORM_ADC TRUE
+/* TODO: Maintainer, please check this option! */
+//#define EFI_FASTER_UNIFORM_ADC TRUE
```
andreika-git, line 254: Why do we need to disable it by default? Please elaborate the problem

dron0gus: hardware.cpp:
```
#if EFI_FASTER_UNIFORM_ADC
...
void adc_callback_fast_internal(ADCDriver *adcp, adcsample_t *buffer, size_t n);
...
#define adc_callback_fast adc_callback_fast_internal
```
But adc_callback_fast_internal is not defined anywhere in RusEFI tree.

andreika-git: We have two adc_callback_fast() functions with the same name in hardware.cpp. 
The first one (from the top) is inside #if EFI_FASTER_UNIFORM_ADC, and it replaces the second default one, and calls it. In order to call the default function with the same name, we need to rename it to adc_callback_fast_internal() - that's why we have that #define. After preprocessor, the new function is adc_callback_fast(), and the old default becomes adc_callback_fast_internal().

andreika-git: > Enabling this cause undifines reference to adc_callback_fast_internal
I've just checked - the latest master commit (https://github.com/rusefi/rusefi/commit/84cacc86bd2a1487dd7becb23808eedb2c96e6ad) compiles successfully with !compile_cypress and #define EFI_FASTER_UNIFORM_ADC TRUE. So the problem seems to be in this particular pull request...

PR #2216, opened 2021-01-10
Supress annoing 'unused' warnings

dron0gus:


firmware/hw_layer/ports/stm32/stm32f7/stm32f7xx_hal_flash.c
```diff
@@ -472,7 +472,7 @@ __weak void HAL_FLASH_EndOfOperationCallback(uint32_t ReturnValue)
 __weak void HAL_FLASH_OperationErrorCallback(uint32_t ReturnValue)
 {
   /* Prevent unused argument(s) compilation warning */
-  //UNUSED(ReturnValue);
```
rusefillc, line 475: why do we move away from "UNUSED"

mck1117: yes please, `UNUSED(...)` is much clearer than `(void)...` for readability

dron0gus: UNUSED() is not defined in this file (with current includes).
UNUSED() is defined in efilib.h.
Including efilib.h because of UNUSED() will cause chained includes, function definitions and so on. All this will slow down compilation and confuse users.
Same for other files.

PR #2226, opened 2021-01-15
save memory in trigger structure

mck1117:
save 1440 bytes of memory 😮


mck1117:
btw, this was tested on real hardware along with the instant rpm PR

PR #2228, opened 2021-01-16
Save more space in the angle cache

mck1117:
Store eventAngles in a scaled_channel instead of a float, using only half the space.  With the current value of `PWM_PHASE_MAX_COUNT`, this saves 576 bytes.

This reduces the accuracy of trigger events to 1/64th of a degree, which is still more than plenty.  64 * 720 = ~46000 so it won't overflow.

### DIFF
--- a/firmware/controllers/trigger/decoders/trigger_structure.h
+++ b/firmware/controllers/trigger/decoders/trigger_structure.h
@@ -13,6 +13,7 @@
 #include "state_sequence.h"
 #include "globalaccess.h"
 #include "engine_configuration_generated_structures.h"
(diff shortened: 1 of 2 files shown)


rusefillc:
in my experience unsolicited comments in PR should be code comments

please move all valuable content into code comments

mck1117:
> unsolicited

that's not what that word means, but point taken anyway

rusefillc:
this did not go well :( https://github.com/rusefi/rusefi/commit/ef82dea2939fde529e0c213b18cb58ed8b21895d

can you fix or shall I revert?

PR #2229, opened 2021-01-16
Simplify Makefile, unify board.mk files

dron0gus:


firmware/config/boards/kinetis/board.mk
```diff
@@ -3,15 +3,20 @@
...
+BOARDCPPSRC += $(CONTROLLERS_ALGO_AUTO_GENERATED_ENUMS)
 
 # Define linker script file here
-LDSCRIPT= $(STARTUPLD)/MKE1xF512.ld
+LDSCRIPT = $(STARTUPLD)/MKE1xF512.ld
 
 CONFDIR = $(BOARD_DIR)
+
+# Shared variables
+ALLCSRC   += $(BOARDSRC)
```
mck1117, line 20: I see these lines repeated every time, why not either
- put them in `Makefile` or
- just `ALLSRC += $(BOARD_DIR)/board.c` etc

dron0gus: I tried to keep unify with board.mk from ChibiOS.
When all .mk files will be modified this way Makefile can be simplified to something like:
`CSRC =  $(ALLCSRC)
CPPSRC = $(ALLCPPSRC)
INCDIR = $(ALLINC)`

PR #2240, opened 2021-01-26
normalize configuration grammar

mck1117:
- ~~`int` -> `int32_t` to fit the pattern~~ let's solve this one later
- put "quotes" "around" "bit" "options"
- lots of missing/inconsistent commas between fields
- multiple different comment formats, all are now `!`
- changes to existing parser to support these changes


rusefillc:
it's lame but java changes require manual .jar commit

PR #2241, opened 2021-01-26
Remove whitespace

mck1117:
Remove added whitespace from TS ini output.  This will help confirm that there is zero diff between the current parser and the new parser.

depends on #2240


rusefillc:
@mck1117 rebase needed?

mck1117:
yeah nope rebase didn't work

PR #2246, opened 2021-01-27
Remove whitespace

mck1117:
Remove added whitespace from TS ini output. This will help confirm that there is zero diff between the current parser and the new parser.

try again on #2241


mck1117:
this needs jar first

mck1117:
ok, ready to go

Issue #2244, opened 2021-01-27
limp mode too good for bench testing


rusefillc:
with bench testing high RPM we have limping manager cutting stuff like fuel but there is no clear UI for that

plan to change logic of existing indicators

rusefillc:
maybe separate issue maybe same issue:

[test_screwed_up_bin.zip](https://github.com/rusefi/rusefi/files/5877854/test_screwed_up_bin.zip)

@mck1117 attached a Hellen binary which would happily run on Discovery

if self-stim it would show RPM without injection or ignition, I have no idea why. We need to improve usability of error reporting.

rusefillc:
Andrey  12:31 AM
float idealGasLaw(float volume, float pressure, float temperature) {
	return volume * pressure / (AIR_R * temperature);
}





New

Matthew Kennedy  12:32 AM
Stefan is right
```
```
```so first we have

float validateMap(float mapKPa DECLARE_ENGINE_PARAMETER_SUFFIX) {
	if (cisnan(mapKPa) || mapKPa < CONFIG(mapErrorDetectionTooLow) || mapKPa > CONFIG(mapErrorDetectionTooHigh)) {
		warning(OBD_Manifold_Absolute_Pressure_Circuit_Malfunction, "unexpected MAP value: %.2f", mapKPa);
		return 0;
	}
	return mapKPa;
}
```
then we multiple by zero

we need to integrate this into limpManager
because at the moment broken MAP kills engine in a very unclear way

stefanst:
This has been suggested before: Limp manager should use some crude form of alpha-n for limping. Maybe have a translation table of TPS -> MAP. (2D, or better 3D)?
Or, if the TPS is not available just use a fixed, programmable MAP value. Probably 60kpa or so.

mck1117:
> Maybe have a translation table of TPS -> MAP. (2D, or better 3D)?

That table has to be 3d, or if it's 2d, it has to be TPS -> mass flow rate, not pressure.  There are also some transient fuel correction models that need that table anyway, so it might get dual usage.

Issue #932, opened 2019-09-14
fail safe mode / limping mode for MAP failure


rusefi:
@stefanst was the first one to get stuck due to a broken MAP sensor hardware (it was on the board but still)

we shall automatically fall down to some TPS-based strategy if we detect an obvious MAP failure

rusefillc:
@mck1117 please CHANGELOG progress
Does anything remain here at all?

PR #2248, opened 2021-01-29
Add fallback map table

mck1117:
This change adds a table that estimates MAP based on RPM and TPS.  This isn't only useful for handling a failed MAP sensor, but could make acceleration enrichment MUCH easier to configure.

implement #932
related to #2244


firmware/controllers/algo/airmass/speed_density_airmass.cpp
```diff
@@ -41,3 +41,21 @@ AirmassResult SpeedDensityAirmass::getAirmass(int rpm) {
 		map,	// AFR/VE table Y axis
 	};
 }
+
+float SpeedDensityAirmass::getMap(int rpm) const {
+	float fallbackMap;
+	if (CONFIG(enableMapEstimationTableFallback)) {
+		// if the map estimation table is enabled, estimate map based on the TPS and RPM
+		fallbackMap = m_mapEstimationTable->getValue(rpm, 100 * Sensor::get(SensorType::Tps1).value_or(0));
```
rusefillc, line 49: "100" magic constant, can we please give it a name or introduce a helpful macro?

rusefillc:
@mck1117 poke "100" magic constant

firmware/controllers/algo/airmass/speed_density_airmass.cpp
mck1117: magic is no longer magic :heavy_check_mark:

rusefillc:
@mck1117 do we have default values for the new table?

mck1117:
it's pretty engine specific - easiest to just log some normal driving and let megalogviewer generate the table for you

Issue #2253, opened 2021-01-30
hellen128 connector yaml file


rusefillc:
hellen128 is WIP
we have photo at https://github.com/rusefi/rusefi/issues/2237

we have pinout at https://docs.google.com/spreadsheets/d/1I-lZKRajTiEGFUXdZpXEtKF2pymlOo-lPahy3cLMnl4

chuckwagoncomputing:
Should I create a hellen/hellen128/connectors directory? Or is it going to go somewhere else?

rusefillc:
yes let's do hellen/hellen128/connectors for now

chuckwagoncomputing:
Does Hellen128 have connector A? It's not in the picture, but it's listed in the spreadsheet?

chuckwagoncomputing:
Oh nevermind, I see it now. I thought it was part of B.

rusefillc:
A and B are especially weird on this one

chuckwagoncomputing:
Might as well close this

PR #2255, opened 2021-01-31
Connector images, hellen128 mapping

chuckwagoncomputing:
This is a new interactive pinout system.
GHA will generate the html pages and upload them to the /pinouts directory on the FTP server.

While I was at it, I started on a mapping for Hellen128 (#2253), but without any TS names, firmware IDs, or classes. Anyone who likes can finish it by using the other boards for an example, or I can finish it if I am provided with the information. I'm guessing that the information doesn't really exist yet, at least not the TS names.


chuckwagoncomputing:
There's another bug, don't merge this yet...

chuckwagoncomputing:
OK Should be good to go now.

PR #2265, opened 2021-02-02
Gm map sensors

hewball:
GM Map Sensors 

Data from http://www.robietherobot.com/storm/mapsensor.htm


mck1117:
Looks good to me.  Even matches the GM MAP sensor I have configured manually!

PR #2271, opened 2021-02-03
Proteus analog input ci

### LINKED ISSUE #2262: HW CI for analog inputs
One option is that if the PWM frequency is high enough, the analog input filter will make it "DC" enough to work.  Vary PWM duty, check analog voltage.

mck1117:
Test an analog input by plumbing high enough frequency PWM (5khz) in to an analog input that the input filter turns it (almost) in to DC.

Generates PWM using the idle controller, and requires a wire from ign #10 to analog volt 2 (TPS) to receive the signal as a TPS position.

depends on #2270

fixes #2262


rusefillc:
please spell out jumper wires in terms of blackX to whiteY to make it a one stop doc?

PR #2273, opened 2021-02-03
eliminate idle's custom applyPinState

mck1117:
not really a good reason to disable idle control while the engine is stopped - adds complexity and makes debugging the idle valve significantly harder as you can't move it around while looking at it taken apart.


rusefillc:
comment added into the code while I would prefer to keep it as is

also #2274 created since same applies to ETB

mck1117:
behavior is now preserved, but without requiring a custom applyPinState hack

PR #2276, opened 2021-02-04
Fix pinouts maybe

chuckwagoncomputing:
Adding the -m flag to ncftpput to create directories if necessary
Also replaced -v which is completely pointless for our uses


chuckwagoncomputing:
Also lots of debug statements

PR #2279, opened 2021-02-05
put CAN debug mode check back

mck1117:
put the debug check back - this causes serious performance problems on a busy CAN bus


mck1117:
oh wait - there's already a check when the function is called, so we were checking twice (!)

Issue #2288, opened 2021-02-06
implement LPS25HB baro sensor


mck1117:
Proteus v0.5 has a baro sensor, it needs implementation!

mck1117:
Works on proteus, done!

rusefillc:
@mck1117 please contribute default configuration code for Proteus

Issue #2281, opened 2021-02-05
implement ADS1015


mck1117:
ADS1015 looks like a good option for an external ADC - 4 channels, reasonably accurate, internal voltage reference (!), and JLC will even install them for about a buck each.

OrchardPerformance:
@mck1117 - Do you think that is enough to close this one out?

mck1117:
It's not implemented at all 😛

OrchardPerformance:
But... but.... the merge says implement!¡

mck1117:
Implement <thing that isn't ads1015> 😛

rusefillc:
@mck1117 shall we clone this ticket now?

rusefillc:
ADS126x-Q1 automotive 10 channel ADC

PR #2289, opened 2021-02-06
implement bit-banged I2C driver

mck1117:
Implement a bit-banged I2C controller for low priority I2C operations.

Bit-banged (as opposed to hardware accelerated) provides several benefits:
- No interrupts required
- No usage of microcontroller resources (particularly DMA channels)
- No operating system locks (any other thread can preempt an operation)
- Smaller code size
- No restriction on pins: can use any two GPIO pins

progress on #2288 and #2281


mck1117:
Tested with some random sparkfun accelerometer board I had in my electronics bucket.

mck1117:
...and now it's also tested with a real LPS25 pressure sensor mounted on a Proteus v0.5!

PR #2290, opened 2021-02-06
Try SSH instead of FTP

chuckwagoncomputing:


chuckwagoncomputing:
I undid the switch back to cron so we can test this without waiting until tonight

PR #2291, opened 2021-02-06
Some fixes prior to board pull request

dron0gus:
This is set of patches for common code that I need before adding my board support.

### DIFF
--- a/firmware/controllers/system/efi_gpio.cpp
+++ b/firmware/controllers/system/efi_gpio.cpp
@@ -34,7 +34,7 @@ extern WaveChart waveChart;
 EnginePins enginePins;
 static Logging* logger;
 
-static pin_output_mode_e DEFAULT_OUTPUT = OM_DEFAULT;
+pin_output_mode_e DEFAULT_OUTPUT = OM_DEFAULT;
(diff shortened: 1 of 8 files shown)


firmware/controllers/system/efi_gpio.cpp
```diff
@@ -577,7 +577,12 @@ uint8_t criticalErrorLedState;
 void initPrimaryPins(Logging *sharedLogger) {
 	logger = sharedLogger;
 #if EFI_PROD_CODE
-	enginePins.errorLedPin.initPin("led: CRITICAL status", LED_CRITICAL_ERROR_BRAIN_PIN, &(LED_ERROR_BRAIN_PIN_MODE));
+#ifdef EFI_LEDS_ACTIVE_LOW
+	static const pin_output_mode_e DEFAULT_OUTPUT_INVERTED = OM_INVERTED;
```
rusefillc, line 581: lets's reduce the conditional compilation block or even completely remove it

let's just have ERROR_PIN_OUTPUT_MODE equal either OM_INVERTED or OM_DEFAULT and always use ERROR_PIN_OUTPUT_MODE?

Issue #2292, opened 2021-02-06
Hellen72: 4S/4T 2001/2003 drama


rusefillc:
Hellen72 has two jumper configurations for differences in power distribution between 2001 and 2003 - see https://github.com/rusefi/rusefi/wiki/Mazda-Miata-2003#starting--power

meaning that for some users VBATT would be 4S and for some users VBATT would be 4T

maybe all we need is to add 
```
id: EFI_ADC_5
    class: analog_inputs
    ts_name: 4T - Alternator voltage (A5)
``` 
into 4S and rename things a bit?

rusefillc:
@chuckwagoncomputing can you help?

PR #2300, opened 2021-02-06
add info to Hellen Pin 4S

chuckwagoncomputing:
Is this what we need?
#2292


rusefillc:
Thank you!
we are still learning about new mapping way :)

YI @andreika-git

PR #2301, opened 2021-02-06
implement lps25

mck1117:
Implement the driver for the ST LPS25HB pressure sensor, as mounted on Proteus V0.5.

progress on #2288


rusefillc:
@mck1117 not green? do you expect it to get green if I merge?

mck1117:
not green is expected for now, I didn't guard it against tests/sim yet

PR #2308, opened 2021-02-07
simplify MCU selection

mck1117:
Ideally, boards and MCUs are orthogonal concerns: You can mix and match MCU and ECU PCB (which is true to some extent, for example the LQFP144 f4/f7/h7), and each board's makefiles shouldn't have to know about that.

This change starts moving makefile logic from each `board.mk` in to the corresponding `hw_ports.mk`


rusefillc:
@andreika-git please review

firmware/config/boards/hellen/hellen72/board.mk
```diff
@@ -6,19 +6,15 @@ BOARDINC = $(BOARDS_DIR)/hellen/hellen72
 
 # Target processor details
 ifeq ($(PROJECT_CPU),ARCH_STM32F4)
-  MCU_DEFS  = -DSTM32F427xx
```
andreika-git, line 9: If you remove 427 define from here, then where is it now? We cannot use 405 defs for 427 and 469, these are different chips with their own ChibiOS/os/common/ext/ST/STM32F4xx/* includes and stm32f4xx_hal_flash_ex.* support.

firmware/config/boards/prometheus/prometheus-common.mk
```diff
@@ -22,12 +22,10 @@ ifeq ($(DEBUG_LEVEL_OPT),)
 endif
 
 ifeq ($(PROMETHEUS_BOARD),405)
-	LDSCRIPT = $(PROJECT_DIR)/config/boards/prometheus/STM32F405xG.ld
-	DDEFS += -DDEFAULT_ENGINE_TYPE=PROMETHEUS_DEFAULTS -DSTM32F405xx -DFIRMWARE_ID=\"prometeus405\"
+	DDEFS += -DDEFAULT_ENGINE_TYPE=PROMETHEUS_DEFAULTS -DFIRMWARE_ID=\"prometeus405\"
 else
 	# Override DEFAULT_ENGINE_TYPE
-	LDSCRIPT = $(PROJECT_DIR)/config/boards/prometheus/STM32F469xI.ld
-	DDEFS += -DDEFAULT_ENGINE_TYPE=PROMETHEUS_DEFAULTS -DSTM32F469xx -DFIRMWARE_ID=\"prometeus469\"
```
andreika-git, line 30: If you remove 469 define from here, then where is it now? We cannot use 405 defs for 427 and 469, these are different chips with their own ChibiOS/os/common/ext/ST/STM32F4xx/* includes and stm32f4xx_hal_flash_ex.* support.

firmware/config/boards/hellen/hellen72/board.mk
mck1117: F405/7 and F427/9 have the same linker map.  The F427/9 have more RAM, but are otherwise the same layout. I've been running an F405 linker config on an F427 for a year without issue.

The flash controller is actually different, but the F42x and F46x flash controllers are fully backward compatible with the F40x.  The only difference is that the F42x _can_ be dual bank on high density devices, but we don't currently support it.  We always put the config in the same spot for all F4 chips.  As mentioned, many MRE and Proteus boards are running what's technically an F405/7 image on an F427/9.

firmware/config/boards/prometheus/prometheus-common.mk
mck1117: > We cannot use 405 defs for 427 and 469, these are different chips
This isn't true.  The F42x and F469 are strict supersets of what's supported on the F40x.  Just compare stm32f405xx.h with stm32f69xx.h - Save for the addition of features like DMA2D, camera interface, LCD controller, etc. they're the same file.  ST even claims this in their own migration guide that it's 100% software compatible to go from F405/7 to F42x and F469.

mck1117: And you can't even use the slightly faster 180MHz core clock on the 42x and 469 since that breaks USB.  The core clock has to be an integer multiple of 48mhz, which 168mhz is, and 180mhz is not.

firmware/config/boards/hellen/hellen72/board.mk
andreika-git: The whole idea of using 427 instead of 407 is for its benefits like more RAM and the 2nd flash bank (which we really need for config saving while the engine is running). Even if we are not using them now, we are definitely going to (otherwise we would solder 407s all the time because they are just cheaper and more in stock). Therefore, we need a way to distinguish between different STM32F4 families. Therefore, we need STM32F*** defines.

mck1117: There is plenty of more implementation needed before we can use dual bank at all on the F4 parts (changes in mpu_util, for example)   The purpose of this change is to start reducing the duplication, moving stuff in to hw_ports.mk instead of in board.mk.  When we need specific support for the dual bank devices, this will make it easier to make that change.

I don't want to add explicit 42x/469 support yet because:
- we don't actually use the added functionality yet
- it will add additional work while there is lots of churn happening, while simply adding it later will be much less work.

rusefillc:
This clean-up seems to not break any of the _current_ stuff.

Once we face new challenge in the future we would address those on top of the current clean-up.

Issue #2282, opened 2021-02-06
board.h F4 code duplication and F7 mess


rusefillc:
Both 'better' hellen72 and st_stm32f4 board.h have "#define EFI_DR_DEFAULT PIN_PUPDR_PULLUP" and a lot of related code duplication - that "#define EFI_DR_DEFAULT PIN_PUPDR_PULLUP" should probably be moved to a more common plance?

we still have two f7 board.h files which did not get the proper default pin state treatment? at the moment (legacy) MRE F7 and (legacy) Proteus 0.1/0.2 have weird start-up pin states?

PR #2321, opened 2021-02-08
move chconf/mcuconf/halconf in to hw_ports

### LINKED ISSUE #2305: Configuration flexibility vs Compilation time
Currently we are using USE_SMART_BUILD=yes in Makefile for all boards to speed-up build.
ChibiOS build system parses halconf.h and halconf_community.h with simple
`HALCONF := $(strip $(shell cat $(CONFDIR)/halconf.h $(CONFDIR)/halconf_community.h | egrep -e "\#define"))`
And complies only enabled drivers/modules.
So this build system is not compatible with
1. option pre-difines in files included to halconf.g (stm32f4ems has some HAL_ options defined in efifeatures.h that is included in halconf.h )
2. inheritance of configs from other boards, like it is done for many boards:
`#include "../../stm32f4ems/halconf.h"`

We should either switch back to USE_SMART_BUILD=no, or clean-up all configs.
Clean-up of configs will cause duplication of halconf.h and halconf_community.h for each board.

mck1117:
progress roughly in the direction of #2282

also fixes #2305 by turning off smart build, so we can now include files from chconf/mcuconf/halconf and it will actually work


mck1117:
@dron0gus any thoughts before I un-draft?

dron0gus:
Can we keep
```
# Shared variables
ALLCSRC   += $(BOARDSRC)
ALLCPPSRC += $(BOARDCPPSRC)
ALLINC    += $(BOARDINC)
```
at the end of each board mk file. I know that sometimes this vars are empty and it is useless.
But it corresponds to way it is done in ChibiOS.

mck1117:
I think those are now put back everywhere, ready to go!

PR #2333, opened 2021-02-12
increase SD card speed

mck1117:
SD spec says 0-25MHz for slow mode, and 0-50MHz for fast mode.

Increases slow mode to either 10.5MHz or 5.25MHz (depending on which SPI device it is), and fast mode to 42MHz or 21MHz (depending on which SPI device it is).  On an F7 that means 13.5mhz/6.75mhz slow mode and 54MHz/27MHz fast mode, which is pushing it a little bit on SPI1, but will probably be just fine.


rusefillc:
that point about amazing comments right into code? :)

PR #2342, opened 2021-02-13
RusEFI is green: put cpu in standby in idle

dron0gus:
For MRE:
This drops consumption from 0.18A@12V to 0.17A@12V.
Sorry, my lab DC power supply has 0.01A resolution.

### DIFF
--- a/firmware/config/boards/microrusefi/board.mk
+++ b/firmware/config/boards/microrusefi/board.mk
@@ -33,9 +33,6 @@ DDEFS += -DEFI_USE_OSC=TRUE -DFIRMWARE_ID=\"microRusEFI\" $(DEFAULT_ENGINE_TYPE)
 # We are running on microRusEFI hardware!
 DDEFS += -DHW_MICRO_RUSEFI=1
 
-# We are green
-DDEFS += -DCORTEX_ENABLE_WFI_IDLE=TRUE
-
 # Shared variables
 ALLCPPSRC += $(BOARDCPPSRC)
 ALLINC    += $(BOARDINC)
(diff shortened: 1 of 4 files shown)


rusefillc:
@mck1117 @andreika-git do we want this more globally?

mck1117:
datasheet on an F4 says that run vs. sleep is about 90 vs 60mA, so that math sounds about right

mck1117:
I can't find the difference in interrupt latency if asleep vs. running, @dron0gus do you know?

dron0gus:
@mck1117 look like additional 6 CPU clocks are needed to exit SLEEP mode.
Screenshot from 2021-02-13 23-23-57

mck1117:
And standard is 12 cycles for an interrupt? An extra 35ns seems like no problem at all.

rusefillc:
if we like it, this should be universal option not just MRE

mck1117:
Agree, we should just put it in chconf_common or something

dron0gus:
chcore_v6m.h and chcore_v7m.h contain port_wait_for_interrupt() implementation where CORTEX_ENABLE_WFI_IDLE is used.
Both do not include any config files.
Actually both do not include any other file.
So defining this option through mk/Makefile files and DDEFS is the only option I see.

mck1117:
Ew, that's gross.  I guess in the makefile it goes!

dron0gus:
Ok. I was wrong. chconf.h is included prior to chcore_v6m.h and chcore_v7m.h and supposed to be used for this option.
Will update PR.

mck1117:
ugh I hate it when headers don't explicitly include all their dependencies

dron0gus:
ch.h is meta-include that combines everything together.

Issue #140, opened 2015-01-21
virtual drive drivers


rusefi:
we have USB and libraries are available to emulate FAT32 device right via the USB

todo: think if we can improve useablity by having something on that drive?

drivers? console? log files?

Access to SD card logs directly via USB - this case user would not need to remove SD card

rusefi:
http://www.chibios.com/forum/search.php?st=0&sk=t&sd=d&sr=posts&author_id=571&start=10

rusefi:
[ARMCM4-STM32F407-DISCOVERY-USB-MASS-STORAGE-SPI.zip](https://github.com/rusefi/rusefi/files/1082578/ARMCM4-STM32F407-DISCOVERY-USB-MASS-STORAGE-SPI.zip)

rusefillc:
@dron0gus did you say that different USB pins/peripheral is a bit different for composite USB purposes? is that same drama for f4, f7 and h7 or are things easier on f7/h7?

rusefillc:
@dron0gus ^^^?

dron0gus:
For composite device we need:
1 EP fo enumeration (EP#0)
1 Bulk In, 1 Bulk out EPs for Mass Storage
1 Bulk In, 1 Bulk out and 1 Interrupt EPs for CDC.
(Int EP can be removed, but this is not compatible with Windows, only Linux)

STM32F427:
OTG_FS: 4 bidirectional endpoints - not enough for composite device (CDC+Mass Storage)
OTG_HS: 6 bidirectional endpoints - enough.
Bootloader is available on OTG_FS only. So no solution for Bootloader and Composite device.

STM32F765 (for example):
OTG_FS: 1 bidirectional control endpoint + 5 IN endpoints + 5 OUT endpoints - enough
OTG_HS: 8 bidirectional endpoints - enough
Bootloader is also available on OTG_FS only. OTG_FS can be used for all our dreams.

mck1117:
By my count we actually have enough endpoints:
1 for enumeration
1 bidirectional for mass storage
2 for CDC

I count only 4 total.

mck1117:
Also by my count I got it to work :slightly_smiling_face:

rusefillc:
@mck1117 please CHANGELOG

rusefillc:
@mck1117 please #2350

PR #2352, opened 2021-02-14
USB mass storage tweaking

### LINKED ISSUE #2350: enable MSD on MRE
MRE is sometimes soldered to an adapter board with SD card

we need MSD magic on MRE

mck1117:
- if USB MSD enabled, write actual timestamp to log files instead of short TS-protocol-friendly name
- enable for microRusEFI, fix #2350
- changelog
- handle "unknown command" properly in TS protocol


mck1117:
also I think this probably closes #140?

PR #2355, opened 2021-02-15
don't try to overwrite sync-wiki.yaml

chuckwagoncomputing:


rusefillc:
please add comments into the .yaml explaining what and why

chuckwagoncomputing:
There you go

rusefillc:
Well, now we explain each step but not the big picture - why do we have to handle this specific file in a unique way?

but merged

chuckwagoncomputing:
It's a workaround because I haven't figured out the actual cause yet

chuckwagoncomputing:
OK, I figured it out. It broke because you modified sync-wiki.yaml in the rusefi.wiki repo, and GHA doesn't allow modifying workflow files in a different repo from a workflow, so those changes can't be merged to the rusefi_documentation repo.
I /think/ If you modify the file in rusefi_documentation to match, then that repo's workflow merge will work, then this one will work without the workaround.

chuckwagoncomputing:
Could we just put sync-wiki.yaml in the .gitignore and completely remove it from the rusefi.wiki repo? I don't know enough about git to know if that would work.

PR #2357, opened 2021-02-15
properly put buffers in no-cache regions

mck1117:
Put buffers in a non-cached region of memory instead of doing a cache invalidate. Fixes some likely bugs on the F7.

Also cleans up defines for `CCM_OPTIONAL` and `NO_CACHE`, plus removes the option for `EFI_USE_CCM`, since that's purely a function of the MCU in use, and doesn't have any bearing on how rusEFI uses that MCU.


rusefillc:
any of this reproducible with hw ci?

Issue #1109, opened 2020-01-26
migrate cj125 to use spiPolledExchange


mck1117:
We're currently using an expensive implementation (`spiSend`/`spiReceive`), which requires three context switches, a DMA operation, and an interrupt per action.  We should switch to spiPolledExchange which does it synchronously in-thread, avoiding all that extra work to do a pair of bytes.

mck1117:
will be fixed by #2358

PR #2358, opened 2021-02-16
use spiPolledExchange for small transfers

mck1117:
Saves context switches, plus it doesn't require that buffers are in DMA-able non-cached memory.


rusefillc:
would a code comment at least for one usage be useful?

mck1117:
fixes #1109

mck1117:
I'm not sure it needs explicit commenting - it's a normal ChibiOS API just like `spiSend` or `spiReceive`, and it was already used elsewhere in the firmware, this just changes the rest of the usages.

PR #2382, opened 2021-02-20
yet more TS sanity

mck1117:
- extracts interface to TsChannelBase
- implementation in ts_channel_s for now, to be split out by underlying channel type

more continuation of #2381


rusefillc:
why refactoring this now, how did you end up in TS code?

mck1117:
I ended up in TS code while trying to simplify mcuconf/halconf/etc, and came across this monstrosity:
image
In no world is this reasonable code 🙂

rusefillc:
So a https://en.wikipedia.org/wiki/Rabbit_hole ?

mck1117:
Probably yes, but the rabbit was bored on a 6.5hr flight, so it did some digging deep in the hole.

PR #2393, opened 2021-02-24
enable CAN on H7

mck1117:


mck1117:
CAN driver seems to be broken on H7 :(

mck1117:
Currently blocked by ChibiOS as it doesn't actually support the CAN device on the H7.

see http://forum.chibios.org/viewtopic.php?f=3&t=5122&p=39536#p39537

mck1117:
woohoo, works!

Issue #2397, opened 2021-02-25
implement SDIO/SDMMC for SD card


mck1117:
SPI works, but SDIO/SDMMC works better.  It's a little CPU cheaper, and a lot faster.  Plus, it's not broken on the H7 like SPI.

rusefillc:
a better clone of #683?

mck1117:
oops, that's what I get for not searching

mck1117:
we can close #683 too

Issue #683, opened 2019-01-26
SD card in SDIO mode


rusefi:
Random boards like https://wiki.stm32duino.com/images/6/69/Vcc-gnd.com-STM32F407VET_mini-schematic.pdf have SD card wired to SDIO. While SDIO shares pins with SPI3, I believe SPI code would not work for instance because SDIO clock is PC12 but SPI clock is PC10.

We need SDIO SD card implementation.

ahmadmn:
is it useful example:
[](http://stm32f4-discovery.net/2014/07/library-21-read-sd-card-fatfs-stm32f4xx-devices/)

rusefillc:
@dron0gus @andreika-git we might need faster SD write for "smart ecu" meta epic. Open question if SDIO would help with performance or not.

rusefillc:
We now have SDIO mode as compile time option EFI_SDC_DEVICE

No official boards are using SDIO as of today

PR #2398, opened 2021-02-25
maybe hook up sdmmc

mck1117:
#683 and #2397


rusefillc:
sooo close :)

mck1117:
just tested this as working on my F769I discovery board (and modifications made to use HS USB!)

Issue #2325, opened 2021-02-09
Port-ify ADC


mck1117:
The STM32H7 has a significantly different ADC.   ADC sampling should be handled somewhere within `hw_layer/ports`, since it's going to be big spaghetti if we do it any other way.

mck1117:
let's call this done enough for now

PR #2402, opened 2021-02-25
Extract slow adc

mck1117:
Always sample all slow ADC channels (this will allow changing ADC input mappings at runtime, which is currently impossible!)

This doesn't really hurt performance, as the vast majority of the time spent in the function `readSlowAnalogInputs` is spent with the thread asleep and the ADC converting samples on its own, copying to RAM via DMA.

As a bonus this saves a little memory since now the slow ADC's config group can live in flash, instead of RAM, as it's a constant.


rusefillc:
why do we touch this? is this related to anything else?

mck1117:
relevant to #2325 since H7 needs a completely different implementation

PR #2406, opened 2021-02-26
allow disabling fast ADC

mck1117:
H7 has a different ADC structure that'll require considerably more work before fast ADC (MAP sampling) works - allow disabling the fast ADC all together for now.


mck1117:
ready to go!

PR #2414, opened 2021-02-28
hip9011 compilation fix, cosmetic clean-ups

dron0gus:
Looks like no one have been using this driver for a while.


mck1117:
why not use software knock?

dron0gus:
> why not use software knock?

Short answer: because I have HIP on my custom board.
Long answer: because I suspect there is a bug that sets HIP's chip select low even HIP driver is disabled. And my board has few devices on same SPI bus. This cause issues. So while I'm sorting out this issue I'd like to do some clean-up too.

PR #2417, opened 2021-03-03
Embed ini

mck1117:
Embed the tunerstudio ini in the firmware - it mounts as a USB mass storage device when you connect the ECU without an SD card.

This works by creating a read-only FAT filesystem embedded in the firmware image, and mounting it as a fake SD card when we don't have an SD card installed.


mck1117:
A future improvement is to mount multiple LUNs (aka multiple "drives") at the same time, so that both an SD card and the ini drive can be mounted at the same time.  I have it prototyped, but it'll require pretty significant changes to the ChibiOS-Contrib MSD driver to support, so that's a task for another time.

mck1117:
Here's what it looks like: https://youtu.be/sbxRjz48Wcc

PR #2418, opened 2021-03-04
hook up lps25 on proteus

mck1117:


mck1117:
Tested working on both F4 and H7 Proteus!

PR #2420, opened 2021-03-05
Store ini filesystem in compressed block device

mck1117:
This PR converts the ini filesystem from:
ini file -> zip file -> FAT filesystem (ini decompressed by pc)
to
ini file -> FAT filesystem -> gzip (entire filesystem decompressed on the fly, in-ECU)

This means that the filesystem appears to contain an ini file, not an ini.zip file.  With this, you can open the ini *directly from TunerStudio* instead of having to first unzip it to your PC.  As an added bonus, it consumes less flash space - currently about 75k, but presents as a full 1MB filesystem.

Unfortunately, right now it requires 32KB of RAM to store the DEFLATE dictionary, which means it doesn't fit on an F405/7 (but does fit on an F42x/F7/H7).  It may be possible to compress with a smaller dictionary, but I haven't tried yet.


mck1117:
depends on https://github.com/ChibiOS/ChibiOS-Contrib/pull/265 for MSD driver config options

mck1117:
This change also sets the priority of the MSD thread to LOWPRIO so that it doesn't get in the way of anything important we're doing.

PR #2421, opened 2021-03-05
STM32H7 Slow ADC

mck1117:
Implement slow ADC for H7.


mck1117:
I THINK after this change, an H7 can run an engine without any code mods.

PR #2422, opened 2021-03-05
don't update ramdisk image in master

mck1117:
we don't want to update the ramdisk image in master, only in builds, so add it to gitignore


rusefillc:
what about my commits to master?

mck1117:
huh?

Issue #2387, opened 2021-02-23
is gen_config invoked too often now?


rusefillc:
is looks like we now generate new signature on _any_ git push? we used to have some caching mechanism right?

![image](https://user-images.githubusercontent.com/48498823/108914448-577c4300-75f9-11eb-8c21-89fb7e20c991.png)

chuckwagoncomputing:
I thought we decided to do this so that when debugging someone that was using a git build rather than a release, we would be able to tell which version exactly they had.
I'm not here to argue that that is a good reason to do things this way, I'm just saying that if my memory isn't playing me a trick, this was done intentionally.

rusefillc:
I do not trust my memory, I do not have any recollection of that.

I am more or less OK either way.

rusefillc:
@andreika-git @mck1117 what's your take on this?

mck1117:
whether it's intentional or not, the current method is pretty bad - we're committing a new `cache.zip` after every commit, which increases the repo size by 1.1 megabytes, after EVERY SINGLE COMMIT.  Git doesn't diff binary files, it just stores the whole damn thing.

rusefillc:
@chuckwagoncomputing do you know which exact change has switched the behavior?

On the one hand we do not use the cache but also we update the cache and repo growth is a real issue. We have to move one way or another but move :)

mck1117:
We could also just diff the actual output files against the checked in files, and check if anything other than the timestamp header changed.  That's all we're using the cache for, right?

rusefillc:
if we always push _at the moment_ there is no reason to push _cache_ and use _cache_

but I am leaning towards returning it back to when signature was only touched when important files were touched. the value of artifacts having unique signature seems lower at least than the cost of constant recompile for me.

chuckwagoncomputing:
Then we need to define what files are important/some heuristic to determine whether a push to master is important.

chuckwagoncomputing:
As far as I can tell, this has been the behavior since the initial commit of gen-config.yaml

rusefillc:
@chuckwagoncomputing a while back we only had generation sometimes

![image](https://user-images.githubusercontent.com/48498823/110188343-51e8df00-7de9-11eb-84b1-b29809b476f9.png)

rusefillc:
![image](https://user-images.githubusercontent.com/48498823/110188574-33cfae80-7dea-11eb-947e-a1001f3d0680.png)

rusefillc:
[...]

PR #2423, opened 2021-03-05
Add two new V8 firing orders

slowriot:
- Adding a firing order of 1-2-3-4-5-6-7-8 for sequential testing
- Adding a firing order of 1-5-4-8-6-3-7-2 used by Audi 40v 4.2 V8 engines
- Minor tweak to a build script to use available cpu cores when building rather than a hardcoded value of 4


rusefillc:
does nproc work kn cygwin?

mck1117:
please don't commit generated files - they're generated and committed automatically

rusefillc:
we can but usually do not PR generated files. actions would invoke generator

slowriot:
> does nproc work kn cygwin?

It's part of GNU core utils, should be available in cygwin and pretty much everywhere.

> please don't commit generated files - they're generated and committed automatically

Okay, no problem - I'll revert changes to those.  Shouldn't they be in .gitignore, in that case?

firmware/integration/rusefi_config.txt
```diff
@@ -575,7 +575,7 @@ cylinders_count_t cylindersCount;
 ! FO_1_8_4_3_6_5_7_2 = 5
 ! FO_1_2_4_5_3 = 6
 
-custom firing_order_e 4 bits,    U32,    @OFFSET@, [0:4], "One Cylinder", "1-3-4-2", "1-2-4-3", "1-3-2-4", "1-5-3-6-2-4", "1-8-4-3-6-5-7-2", "1-2-4-5-3", "1-4-2-5-3-6", "1-2", "1_2_3_4_5_6", "1-2-3", "1-8-7-2-6-5-4-3", "1-5-4-2-6-3-7-8", "1-6-3-2-5-4", "1-10-9-4-3-6-5-8-7_2", "1-7-5-11-3-9-6-12-2-8-4-10", "1-7-4-10-2-8-6-12-3-9-5-11", "1-4-3-2", "1-12-5-8-3-10-6-7-2-11-4-9", "1-2-7-8-4-5-6-3", "1-3-7-2-6-5-4-8", "1-2-3-4-5-6-7-8-9", "INVALID", "1-2-3-4-5-6-7-8-9-10-11-12", "1-3-2", "INVALID", "INVALID"
+custom firing_order_e 4 bits,    U32,    @OFFSET@, [0:4], "One Cylinder", "1-3-4-2", "1-2-4-3", "1-3-2-4", "1-5-3-6-2-4", "1-8-4-3-6-5-7-2", "1-2-4-5-3", "1-4-2-5-3-6", "1-2", "1-2-3-4-5-6", "1-2-3", "1-8-7-2-6-5-4-3", "1-5-4-2-6-3-7-8", "1-6-3-2-5-4", "1-10-9-4-3-6-5-8-7_2", "1-7-5-11-3-9-6-12-2-8-4-10", "1-7-4-10-2-8-6-12-3-9-5-11", "1-4-3-2", "1-12-5-8-3-10-6-7-2-11-4-9", "1-2-7-8-4-5-6-3", "1-3-7-2-6-5-4-8", "1-2-3-4-5-6-7-8-9", "INVALID", "1-2-3-4-5-6-7-8-9-10-11-12", "1-3-2", "1-2-3-4-5-6-7-8", "1-5-4-8-6-3-7-2"
```
mck1117, line 578: The change to `-` is ok since it's an unusual firing order (really testing only), but TunerStudio saves your tune on the PC based on the string, not the binary value, so this is a no-no for any fields that we care about preserving compatibility with.

rusefillc:
in order to .gitignore we would have to integrate code generation into make

and fix #2387

mck1117:
@rusefillc I'm not sure that's true - if we completely remove the generated files from the source, they _never_ get committed to the repo, so there doesn't need to be a separate build that runs just to commit updated generated files

rusefillc:
@mck1117 do we agree that brand new clone of rusEFI has to compile with one invocation of "make"? if we do, in order for that to work without java generated in git, shall we integrate generator with Makefile?

slowriot:
You could do it a few different ways.  Just because the files are in .gitignore doesn't mean they can't actually be committed - the automated process could use `git add --force` which overrides the gitignore preference, and commit the generated files as normal.  Meanwhile, end users wouldn't see their own updates from local generation as pending stage, so these wouldn't get committed by mistake.

firmware/integration/rusefi_config.txt
slowriot: Noted, thanks.

rusefillc:
``git add --force`` is something I was not aware of. ".gitignore+git add --force" seems like a reasonable solution

mck1117:
doesn't gitignore mean that files already checked in will still show changes?

rusefillc:
Shall we tell @chuckwagoncomputing about all the amazing stuff we've discussed here?

chuckwagoncomputing:
Once the file has been added, even if it is in the gitignore, future changes will be picked up and commited. @mck1117 had this problem yesterday, and I had it a while back.

chuckwagoncomputing:
One idea I've had is to make a git pre-commit hook to remove files in the .gitignore from the commit. The --no-verify flag would bypass the hook if necessary. The problem is that everyone would have to install the git hook on their local copy. I'm not aware of any way to have it installed automatically.

chuckwagoncomputing:
Here's a magic command to list the files that have been modified, and should be ignored. (But aren't, because they're in the tree)
`comm -12 <(git ls-files -ci --exclude-standard | sort) <(git status -s | cut -d ' ' -f3 | sort)`

It would be simple to write a git hook that ran this before commit and did `git checkout -- <file>` for each result. Obviously it would only work in linux/macos/mingw/cygwin/msys. And as I mentioned above it would have to be manually installed once.

rusefillc:
manual hook is too much drama imho

PR #2425, opened 2021-03-06
update all compile scripts to use available cores rather than a hardcoded value

slowriot:
This change has already been made in some scripts - this pull request makes the same change consistently everywhere.

Hardcoded 4 core limits aren't great for those of us with 56 cores ;)


rusefillc:
do you want to sync your fork? zero priority but weird commit history on PR

anyway look at him the 1% with 56 cores

Issue #1253, opened 2020-04-03
Trigger emulator runs even if no pins are set


mck1117:
PWM is still being generated - but no output pins nor self stimulation are configured, spending a bunch of CPU cycles that do exactly nothing.

rusefi:
Those CPU cycles serve for code simplicity :)

mck1117:
![image](https://user-images.githubusercontent.com/568254/78400226-25aa5a00-75ab-11ea-8e0a-0a7ae5e3c205.png)

Issue #2331, opened 2021-02-11
Fatal error when setting Analog Volt 5 as pedal 2 on proteus


rmracingelectronics:
When setting the 2nd pedal sensor as analog volt 5 on a proteus (only hardware i had to test) there is a fatal error that bricks USB connection and stops tunerstudio form burning the configurations, if repeated a few times it bricks the ecu to a point where it needs firmware upload to unbrick.

rusefillc:
@mck1117 did you say you were able to reproduce this? what are steps for me to reproduce locally?

mck1117:
I got it to repro once, by loading [this tune](https://rusefi.com/online/view.php?msq=423) then setting the secondary pedal sensor to analog volt 5.

mck1117:
Looks like it happened again: a bin is attached to this post.
[brickedfw.zip](https://github.com/rusefi/rusefi/files/6136331/brickedfw.zip)

rmracingelectronics:
With me it happens all the time andni tried all sorts of different configurations and settings, I can't explain what triggers that because obviously I don't know the firmware well enough

mck1117:
Figured it out.  It's a buffer overflow out of the buffer where we write the error message.

PR #2443, opened 2021-03-11
stop trigger emulator PWM when disabled

mck1117:
- only start PWM running once you've actually enabled self stim
- stop PWM when you turn it off

#1253


rusefillc:
why do we care to touch this instead of for instance #2331?

mck1117:
now that bug is assigned to me so I'll see it :)

PR #2447, opened 2021-03-12
Checkout ramdisk images

### LINKED ISSUE #2436: github actions shouldn't commit ramdisk_image.h and ramdisk_image_compressed.h
These files are meant to be generated on demand for a build, and not checked in every time.  The action currently just commits the full diff every time, but needs to filter these out.

chuckwagoncomputing:
This should fix #2436


rusefillc:
please add comments explaiming magic  into code

mck1117:
you can put the original contents of those files back now, too

PR #2448, opened 2021-03-12
Try pulling first

### LINKED ISSUE #2446: false positive GHA drama
why did https://github.com/rusefi/rusefi/runs/2092138201 fail?

chuckwagoncomputing:
I *think* this should fix #2446. On the other hand I understand git poorly, so I don't have a good understanding of what issues doing it this way could cause.

### DIFF
--- a/.github/workflows/build-unit-tests.yaml
+++ b/.github/workflows/build-unit-tests.yaml
@@ -66,6 +66,7 @@ jobs:
       run: |
         git config --local user.email "action@github.com"
         git config --local user.name "GitHub build-unit-tests Action"
+        git pull
         mv triggers.txt unit_tests
         mv *.logicdata  unit_tests
         git add unit_tests/triggers.txt


rusefillc:
good change that it should be fetch and merge not pull, i think :(

we need to begin from beginning. what is image we push and why do we push so often

mck1117:
fetch + merge == pull (mostly)

I'm not in favor of this though, since it'll pollute the history with /even more/ commits (merge commits).

chuckwagoncomputing:
It'll only create a commit if there are actually changes that have been pushed while the action is running, right?

chuckwagoncomputing:
The unit test workflow only pushes unit_tests/triggers.txt.

rusefillc:
trigher is only updated if shape has changed?

mck1117:
I'm not sure this is fixed

rusefillc:
something is fixed since unit tests were finally able to commit fresh triggers

mck1117:
I suspect a change that causes a generated config commit would cause the unit test build to fail to commit, it didn't happen because the generated config commit hasn't happened

chuckwagoncomputing:
I'm confused about what's going on here. @rusefillc, did you think the issue was fixed without this PR? @mck1117, what do you think would be a proper fix for this?

rusefillc:
we had multiple different issues

most urgent issue was fixed without this PR

what @mck1117 describes is a separate less urgent scenario

at the moment master is broken for third issue around GH packages?

rusefillc:
Did not work? https://github.com/rusefi/rusefi/runs/2104185467?check_suite_focus=true

PR #2449, opened 2021-03-12
Fuel pressure sensor checks

### LINKED ISSUE #2445: fail hard if  fuel pressure compensation mode set to "sensed pressure", but you have no fuel pressure sensor configured
Let's FATAL in case of invalid configuration

let's also FATAL in case of a very broken fuel pressure, or at least default and warn?

mck1117:
- check that you actually have a pressure sensor configured if you try to use it
- handle negative pressure delta (ie, injectors are trying to flow backwards), as this suggests a failed sensor

fixes #2445


mck1117:
looks like that failing build is unrelated

PR #2450, opened 2021-03-13
use full memory on STM32F42x

mck1117:
F42x has more memory, so use it!


rusefillc:
just to make sure that we are on the same page please provide a human-readable description of the intentions of this PR

PR #2458, opened 2021-03-14
implement ads1015 driver

mck1117:
Implement the low level driver for the ADS1015 external 4 channel ADC.

related #2281


mck1117:
tested on the bench, all 4 channels working!

PR #2460, opened 2021-03-15
Logging prep - cleanup for #2439

mck1117:
cleanup before #2439 can happen


rusefillc:
what's the plan for sensor chart? why did it have to be removed?

mck1117:
Sensor chart calls `scheduleLogging` from an interrupt context, which doesn't work with #2439.  Since I can't tell that as-is the sensor chart actually does something useful, I figured it was easiest to disable it for now until refactoring can happen to make it work with that restriction.

rusefillc:
just cherry-picked some of the commits

my preference would obviously be not to disable even questionable feature even temporary since it's just easier to go forward in one go?

also need more time to think about text in warn

mck1117:
> my preference would obviously be not to disable even questionable feature even temporary since it's just easier to go forward in one go?

It will require a significant refactor of that feature, and the feature *doesn't currently do anything useful*.

> also need more time to think about text in warn

That branch was never hitting before.  I only hit it when I had another bug causing the overflow. If we hit that branch in master today, it would cause the same parsing error in the console.

mck1117:
I am happy to work on a smart-ECU-relevant "make sensor sniffer/oscilloscope/logic analyzer useful" task.

rusefillc:
I am not sure what is "same parsing error", I do not follow

I do not enjoy the hint of blackmail with the "we have to kill code for me to more forward". If you are not in the position to move forward without killing excising code please slow down until you would be able to move forward with less of a trade off.

mck1117:
> I am not sure what is "same parsing error", I do not follow

The format is comma separated between keys and values, like
`msg,Print message 1,msg,Print message 2`
which parses to
| key | value |
| --- | --- |
|msg|Print message 1|
|msg|Print message 2|

so if you tried to log a message like `test message, more things` containing a comma, printing out the warning `output overflow %s %d [%s]`, it would put a comma *inside the content of the message* and confuse the parser.

It would log as `msg,output overflow name 25 [test message, more things]` which would parse as

| key | value |
| --- | --- |
|msg|output overflow name 25 [test message|
| more things]| `blank` |

Which of course fails because ` more things]` is not a valid key.

mck1117:
> I do not enjoy the hint of blackmail with the "we have to kill code for me to more forward". If you are not in the position to move forward without killing excising code please slow down until you would be able to move forward with less of a trade off.

It is not blackmail.  The code is still there, man.  If we come up with something that will EXECUTE THE CODE instead of it sitting dormant on disk, I'm all ears (and fingers to type it in to a code editor).

rusefillc:
it's helpful to know which message has caused the overflow. I assume that while "msg" is sometimes containing comma, quite often it does not. Also note that now you have unused variable "msg" since "msg" is a pointer to "text" parameter?

PR #2465, opened 2021-03-16
Dedicated code path for USB console/TS

mck1117:
This change moves native USB serial port to its own thread, initialized as early as possible with zero dependencies on loaded configuration. The goal is to avoid any soft-brick scenarios where an invalid configuration results in a semi-functional ECU but dead USB, rendering the ECU effectively a brick.

This change (and a few after it) will also make it easier to add additional or different USB/serial/ethernet/CAN interfaces for tuning.


mck1117:
hello darkness, my old friend:
image

USB comm looked fine though, so PR is ready.

PR #2476, opened 2021-03-19
fix multi-serial-port insanity

mck1117:
- convert UART DMA to a TsChannel (now it supports multiple of them on one device, no more collisions!)
- There was no difference between `console port` and `ts port` so now it's just `TS_PRIMARY_UART` and `TS_SECONDARY_UART` (or alternatively their `_SERIAL` counterparts)


rusefillc:
Thank you, that's an epic improvement!

Issue #2464, opened 2021-03-16
flash configuration is read out twice


mck1117:
readConfiguration() is called from two places:

![image](https://user-images.githubusercontent.com/568254/111271129-03f48800-85ee-11eb-9032-113306a1c44a.png)

and then

![image](https://user-images.githubusercontent.com/568254/111271111-fb9c4d00-85ed-11eb-85b7-a7a4dd9b5178.png)

PR #2480, opened 2021-03-21
move some initialization steps earlier

mck1117:
prep related to #2464, also just general refactoring to make initialization a little less spaghetti

also removes the nearby `getBor` and `setBor` console commands for setting/setting the brownout threshold, which has been set reasonably by `mpu_util` for a long time now, and has no reason to be settable via console.


rusefillc:
looks like BOR is removed without mentioning that in description?

mck1117:
description updated

rusefillc:
let's keep unused BOR code as is, i hope that it's not blocking the rest of progress happening

Issue #2482, opened 2021-03-22
ford "barra" 3+1 cam sensor pattern


mck1117:
![image](https://user-images.githubusercontent.com/568254/111967811-bb3d4300-8ab5-11eb-8ed3-5f743cbf614f.png)

![image](https://user-images.githubusercontent.com/568254/111967838-c1cbba80-8ab5-11eb-9b39-a58ad36f8bac.png)

Two short gaps, then two long gaps of double the short gaps.

Therefore the gaps are 60-60-120-120 (cam degrees), or 120-120-240-240 crank degrees.

mck1117:
annotated:

![image](https://user-images.githubusercontent.com/568254/111968081-06efec80-8ab6-11eb-88b2-7fc47554e503.png)

PR #2483, opened 2021-03-22
barra vvt cam pattern

mck1117:
#2482


mck1117:
FW syncs on self-stim with crank set to this pattern, sniffer looks right, probably works /shrug

PR #2487, opened 2021-03-25
remove disused BOR (brown-out reset) console command

mck1117:
also removes the nearby getBor and setBor console commands for setting/setting the brownout threshold, which has been set reasonably by mpu_util for a long time now, and has no reason to be settable via console.

:slightly_smiling_face:


rusefillc:
Let's revisit this on or after Mar 23, 2022.

PR #2490, opened 2021-03-25
write flash on separate thread for H7

mck1117:
Enable writing H7 flash while the engine is running.  H7 uses dual bank flash, and this change moves the flash write to a low-priority thread (only when enabled) so as to not block any other operations on the ECU.  Behavior for F4/F7 is currently unchanged.

It also moves clearing the `needToWriteConfiguration` flag to *after* the flash write, so the indicator is extinguished in TS once the operation is actually done.


mck1117:
Tested on a real life running car! Couldn't tell a flash write was happening at the same time.

PR #2499, opened 2021-03-27
hip9011: get callback from spark on each cylinder

dron0gus:


mck1117:
@dron0gus Do you need the start timer at all any more?

dron0gus:
> @dron0gus Do you need the start timer at all any more?

Good catch. Will remove with next PR
Anyway HIP code needs deep rework.

PR #2504, opened 2021-03-28
various TS project improvements, dead config fields

mck1117:
- remove dead `fuelRailPressure` field, fix corresponding NB miata and VW defaults to use correct config for fixed rail pressure
- Shuffle around some menu options, mostly moving things out of the giant controller menu in to places where they belong
- Write tooltips for multispark
- update language on VVT PID page to match trigger input page
- a little bit of "human proofing" for after start enrichment options so you can't set it to options that makes your engine hard to start


mck1117:
image

rusefillc:
👍

PR #2525, opened 2021-04-04
Hellen72: Set the default pin states to input-pulldown instead of input-pullup

andreika-git:
This may solve some transient startup issues with Idle and VVT


rusefillc:
todo: remove most of hellen board.h while keeping the EFI_DR_DEFAULT override

PR #2527, opened 2021-04-04
hip9011 and friends

dron0gus:


firmware/hw_layer/sensors/hip9011_logic.cpp
```diff
@@ -183,3 +187,28 @@ void HIP9011::handleSettings(int rpm DEFINE_PARAM_SUFFIX(DEFINE_HIP_PARAMS)) {
...
+void HIP9011::handleChannel(DEFINE_HIP_PARAMS) {
+	int ret;
+
+	/* we did not receive any callback from spark logic with valid cylinder yet */
+	if (cylinderNumber < 0)
+		return;
+
+	/* find next firing cylinder */
+	/* MAGIC +1 -1, couse getNextFiringCylinderId expect cylinders to start from 1 */
+	expectedCylinderNumber = getNextFiringCylinderId((cylinderNumber + 1) PASS_ENGINE_PARAMETER_SUFFIX) - 1;
```
mck1117, line 205: why do we need next expected cylinder?

dron0gus: To setup hip9011 for next cylinder as soon as possible. And when times come - we only set gpio starting integration.

mck1117: ah, so we don't have to make an SPI transaction to select the correct channel at the instant we start sampling, got it

PR #2540, opened 2021-04-10
sd buffer overflow

mck1117:


rusefillc:
any code comment explain8ng size? introduce constant refactoring? how do we not have same drama in 1 year?

mck1117:
there is already a protection for this, it triggers a fatal error if you turn on SD logging

PR #2545, opened 2021-04-12
improve sensor chart flush behavior

mck1117:
prompted by #2439, this makes the sensor chart work more like the engine sniffer, flushing the buffer from the TS thread instead of from an interrupt context (which also can have performance issues!


mck1117:
In fact, I think the sensor sniffer was broken and didn't actually work, but now it actually works!

PR #2548, opened 2021-04-13
constexpr-ify log field list

mck1117:
Converting to constexpr means we can execute the constructors for this array at compile time, which will save a tiny bit of startup time, but saves ~1500 bytes of ram!  Executing those constructors at compile time means the compiler will put the array in flash instead of RAM.


firmware/console/binary_log/log_field.h
```diff
@@ -8,10 +8,10 @@ struct Writer;
 class LogField {
 public:
 	template <typename TValue, int TMult = 1>
-	LogField(const scaled_channel<TValue, TMult>& toRead, const char* name, const char* units, int8_t digits)
+	constexpr LogField(const scaled_channel<TValue, TMult>& toRead, const char* name, const char* units, int8_t digits)
 		: m_type(resolveType<TValue>())
 		, m_multiplier(TMult)
-		, m_addr(reinterpret_cast<const char*>(&toRead))
```
mck1117, line 14: this `reinterpret_cast` was the problem - all type punning is forbidden in `constexpr`

rusefillc:
Is there a way to monitor if we are still 'good' in this regard continuously? i..e now that we are 'good' how do we stay good?

PR #2557, opened 2021-04-15
gppwm zero option

mck1117:
add a Y axis option that's forced to zero, in case you want to emulate a 1d table instead of a 2d table


mck1117:
Inspired by @chuckwagoncomputing, good idea!

PR #2558, opened 2021-04-15
flush SD log 10x faster

### LINKED ISSUE #2553: Too many logs are zero size
hellen72c

I would expect a higher percentage of logs to be not empty

image

mck1117:
fix #2553


mck1117:
failed build is ok, a package failed to install on the build machine

PR #2570, opened 2021-04-19
Haltech dash support

shadowm60:
different cycletimes for messages
Haltech dash support


rusefillc:
is this one change or a number of great changed mixed together?

"5" is mentioned more than once please introduce constant and use it mire than once

@mck1117 ?

mck1117:
Lots of reformat too...?

mck1117:
yeah, I think this should be a few smaller PRs that each do one thing.

shadowm60:
I don't get it, i have smaller commits, and maybe can split this PR in 2 pull requests, on would be introduce cycle time handing in canTx, second is halltech support. it is an older branch that i have started and do have two big mergers of remote tracking branch upstream/master, maybe that is what i did not executed in the right way?

shadowm60:
> is this one change or a number of great changed mixed together?
> 
> "5" is mentioned more than once please introduce constant and use it mire than once
> 

Please point out where are those 5s, because i cannot see the forest from the tree in front of me.

mck1117:
> because i cannot see the forest from the tree in front of me.

perhaps this means the forest is too big?

firmware/controllers/can/can_tx.cpp
```diff
@@ -22,17 +22,24 @@ EXTERN_ENGINE;
...
 void CanWrite::PeriodicTask(efitime_t nowNt) {
 	UNUSED(nowNt);
+	static uint16_t cycleCount = 0;
+	uint16_t cycleMask;
+	
 
 	if (CONFIG(enableVerboseCanTx)) {
-		void sendCanVerbose();
-		sendCanVerbose();
+		uint16_t period = CONFIG(canSleepPeriodMs)/5;
```
mck1117, line 38: magic 5

mck1117: some comments here would be GREAT for what's going on

firmware/hw_layer/drivers/can/can_hw.cpp
```diff
@@ -279,7 +279,11 @@ void initCan(void) {
 
 	// fire up threads, as necessary
 	if (CONFIG(canWriteEnabled)) {
-		canWrite.setPeriod(CONFIG(canSleepPeriodMs));
+		/**
+		 * We use 5ms cycle time to ensure cyclic messages on time, user should not redefine this.
+		 * Config for periodic TX applyes only for verbose (broadcast).
+		 */
+		canWrite.setPeriod(5);
```
mck1117, line 286: magic 5

mck1117: you don't need to call this if it's set in the constructor

firmware/controllers/can/can_dash.h
```diff
@@ -7,10 +7,21 @@
...
 #pragma once
 
-void canDashboardBMW();
-void canDashboardFiat();
-void canDashboardVAG();
-void canMazdaRX8();
-void canDashboardW202();
-void canDashboardBMWE90();
-void canDashboardVagMqb();
+#define CAM_5ms         (1<<0)
```
mck1117, line 10: is `CAM` correct?

shadowm60: CAM - is short for Cyclic Activity Manager, i can change it, it cam to me natural since using same abbreviation on other projects.

mck1117: > using same abbreviation on other projects

I would be very careful mixing professional life and hobby life, if I'm reading this correctly.

rusefillc:
@shadowm60 needs rebase

shadowm60:
@rusefillc  not it should be ok

PR #2573, opened 2021-04-21
Handle can message cyclicity

shadowm60:
update of canDash to support different cyclicity of messages

Breaking down and replacing #2570

### DIFF
--- a/firmware/controllers/can/can.h
+++ b/firmware/controllers/can/can.h
@@ -17,7 +17,7 @@
 #define CAN_TIMEOUT MS2NT(100)
 
 //can tx periodic task cycle time in frequency, 200hz -> 5ms period
-#define CAN_CYCLE_FREQ		(200)
+#define CAN_CYCLE_FREQ		(200.0f)
 //can tx periodic task cycle time in ms
 #define CAN_CYCLE_PERIOD    (CH_CFG_ST_FREQUENCY / CAN_CYCLE_FREQ)
 
--- a/firmware/controllers/can/can.h
+++ b/firmware/controllers/can/can.h
@@ -16,6 +16,25 @@
 
 #define CAN_TIMEOUT MS2NT(100)
 
+//can tx periodic task cycle time in frequency, 200hz -> 5ms period
+#define CAN_CYCLE_FREQ		(200)
+//can tx periodic task cycle time in ms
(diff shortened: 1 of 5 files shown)


firmware/controllers/can/can.h
```diff
@@ -16,6 +16,25 @@
 
 #define CAN_TIMEOUT MS2NT(100)
 
+//can tx periodic task cycle time in frequency, 200hz -> 5ms period
+#define CAN_CYCLE_FREQ		(200)
+//can tx periodic task cycle time in ms
+#define CAN_CYCLE_PERIOD    (CH_CFG_ST_FREQUENCY / CAN_CYCLE_FREQ)
```
mck1117, line 22: is this actually in MS? `CH_CFG_ST_FREQUENCY` is just equal to the CPU clock.  I think you want `(1000.0f / CAN_CYCLE_FREQ)` to get ms.

firmware/hw_layer/drivers/can/can_hw.cpp
```diff
@@ -278,7 +278,8 @@ void initCan(void) {
 
 	// fire up threads, as necessary
 	if (CONFIG(canWriteEnabled)) {
-		canWrite.setPeriod(CONFIG(canSleepPeriodMs));
+		//remove user interaction to ensure we do have correct period timing
+		canWrite.setPeriod(CAN_CYCLE_PERIOD);
```
mck1117, line 282: you don't need this here if the period is set once at startup

shadowm60: i can remove it, it stayed there since it was in the old code, but yes, it makes sense not to call it any more

firmware/controllers/can/can.h
shadowm60: Freq is needed in the constructor, and the period constructor would actually reach out to a macro where CH_CFG_ST_FREQUENCY is divided by the parameter to get the needed ms. in the setPeriod method the param is ms, But yes, the freq needs to be float, i reused the macro no avoid constants.

mck1117: do we need `CAN_CYCLE_PERIOD` at all?

shadowm60:
Do i need to do something else for this ticket?

Issue #2578, opened 2021-04-21
Get unofficial eLua demo to compile with ChibiOS 20


rusefillc:
http://www.eluaproject.net/ is very promising and our friends at https://github.com/autosportlabs/RaceCapture-Pro_firmware are enjoying it https://wiki.autosportlabs.com/RaceCapturePro_Lua_Scripting with FreeRTOS

https://github.com/rusefi/chibios_with_elua is a fork of https://github.com/linvinus/chibios_with_elua

https://github.com/linvinus/chibios_with_elua is inactive for the last 6 years and it was compiled against cibios 3.0

rusefillc:
https://habr.com/ru/company/embox/blog/496146/ about lua not eLua

mck1117:
RC is not using eLua, it looks like they use "normal" Lua

mck1117:
lua works, closing

PR #2590, opened 2021-04-27
lua 5.4

mck1117:
not enabled on real hardware yet (but works), just loads a hardcoded script that prints out sequentially counting up integers


rusefillc:
can we submodule or not for some reason?

plz add a unit test ideally

plz add a brief overview of integration points/approach into some .md file

love it

mck1117:
The problem I saw with submodule was that the file `luaconf.h` is in the lua/src folder, and not really any way to replace it :(

rusefillc:
can we have our own copy of that file and rely on order of includes?

mck1117:
> rely on order of include durectories

I just tried it and unfortunately that doesn't work, since the compiler always looks in the current directory of the source file before checking the include path, and `lauconf.h` is in the same folder as most of the source c files.

rusefillc:
we have to fork them and rename file on our fork

rusefillc:
and reference our fork

rusefillc:
like we do chibios

rusefillc:
@mck1117 you have write access to https://github.com/rusefi/lua now

mck1117:
fixes #2578

PR #2600, opened 2021-04-28
Subaru board update

dron0gus:


dron0gus:
`Not true: Fatal not expected: CRITICAL error: Pin "PB10" required by "i2c" but is used by "i2c" PROTEUS_DEFAULTS 20210428@proteus`
Look like this is side issue.

Issue #2602, opened 2021-04-29
command line support for lua


rusefillc:
let's see our scripting working in rusEFI console!

mck1117:
done - there is now a console command `lua` that simply runs whatever you write after it.

For example, `lua print('hello')` will call the lua `print` function, causing it to print out `hello` to the console.

PR #2614, opened 2021-05-03
interactive lua

mck1117:
adds new console command, `lua`, that will parse & run arbitrary Lua code.

Example: `lua print('hello interactive!')`

completes #2602


mck1117:
image

mck1117:
Also enables Lua on F7.

PR #2651, opened 2021-05-08
shrink canned tables

mck1117:
- tables don't need to be using float, save 75% of the space by using uint8_t
- use type safe `copyArray` and `copyTable` instead of plain `memcpy`

saves ~10k of flash on default frankenso, and ~6k on MRE


rusefillc:
What is the problem statement - what exact use-case if pushing us into this nightmare?

If we have to go forward with this we have to make this human-readable - please add some magic macro PACK_TO_BYTE(5000) instead of magic 29

firmware/config/engines/ford_aspire.cpp
```diff
@@ -19,24 +19,24 @@
...
-/* Load 3.546667 */{	-9.866000,	-5.518000,	-2.854000,	-0.000000,	+3.022000,	+3.816000,	+6.428000,	+7.788000,	+19.426001,	+20.860001,	+19.966000,	+21.030001,	+21.396000,	+21.570000,	+21.570000,	+21.570000},
-/* Load 3.760000 */{	-9.864000,	-5.516000,	-2.772000,	+0.226000,	+2.732000,	+3.500000,	+6.798000,	+8.102000,	+8.660000,	+9.500000,	+11.788000,	+20.132000,	+20.072001,	+20.510000,	+20.510000,	+20.510000},
-/* Load 3.973333 */{	-9.864000,	-5.518000,	-2.854000,	-0.000000,	+2.880000,	+3.816000,	+6.420000,	+8.320000,	+8.426000,	+8.532000,	+11.470000,	+11.442000,	+13.610000,	+12.022000,	+12.022000,	+12.022000},
-/* Load 4.186667 */{	-9.750000,	-5.518000,	-2.604000,	-0.000000,	+2.880000,	+3.654000,	+6.050000,	+6.888000,	+8.372000,	+9.364000,	+11.764000,	+11.732000,	+11.864000,	+12.376000,	+12.376000,	+12.376000},
-/* Load 4.400000 */{	-0.350000,	-5.590000,	-0.502000,	-0.910000,	-0.864000,	-0.954000,	-1.324000,	+7.436000,	-1.170000,	-1.054000,	-2.058000,	-2.098000,	-2.636000,	+12.352000,	+12.352000,	+12.352000}
+/* Load 1.200000 */{	-0,	+7,	+16,	+23,	+29,	+31,	+32,	+30,	+30,	+26,	+24,	+25,	+25,	+25,	+25,	+25},
```
rusefillc, line 24: I do not get what's going on here. Are we simply loosing precision or are introducing some hard-coded packing/encoding?

rusefillc (changes requested):

firmware/config/engines/ford_aspire.cpp
mck1117: Simply throwing away the fractional part.  Nobody needs better than 1% VE and 1 degree timing accuracy, especially not in a base map.

mck1117:
> What is the problem statement - what exact use-case if pushing us into this nightmare?

In #2646, the MRE F4 QC build was running out of flash.  This change quarters the size of the canned tune tables, saving the flash space outlined in the description.

mck1117:
> If we have to go forward with this we have to make this human-readable - please add a macro PACK_TO_BYTE(5000) instead of magic 29

There is no magic here - not packed.  Simply tossing the fractional part and storing as an integer instead of a floating point number.

PR #2654, opened 2021-05-08
hip9011

dron0gus:
Move HIP states enum out of rusefi_enums.h
-this is internal driver stuff. No need to have it defined globaly.

Auto-generated configs and docs


dron0gus:
Human-readable names were extended and moved to firmware/hw_layer/sensors/hip9011.cpp :
```
static const char *hip_state_names[] =
{
	"Not ready/calculating",
	"Ready for integration",
	"Integrating",
	"Waiting for first ADC sample",
	"Waiting for second ADC sample"
};
```
So now all HIP debug is in one place. Next step will be to hide debug-related code under some `#ifdef EFI_HIP_9011_DEBUG`

rusefillc:
@dron0gus what you are really saying is that you want a smarter code generator which takes human messages from the header... but that's probably a zero priority.

Issue #2636, opened 2021-05-06
allow changing analog input mappings at runtime


mck1117:
$title

PR #2660, opened 2021-05-09
move ADC pin initialization

### LINKED ISSUE #2632: IAT input is being totally ignored by Frankenso
Frankenso ignores an IAT sensor...
When I attached my IAT to INP1, the "raw IAT" gauge always whows 0 volts..., but in the real life, the voltage on the PC1 changes when I heat or cold the sensor.. Then I attached the sensor to INP8 and "raw IAT" is still the same but the voltage on PA3 changes every time I heat or cold the sensor. If I vice versa versa temperature sensors pins in the TS (the Frankenso takes data from the IAT and thinks that it is CLT, the same with IAT input). I can heat and cool the IAT and the CLT gauge reacts so good and fast, but the CLT (Frankenso thought it was IAT) is ignored.
Look like this is the software bug...

Can we solve it somehow?
Thank you

mck1117:
- Move pin init to adc subscription, progress towards #2636
- Increase number of available ADC subscription entries, fixes #2632
- Fatal error if we run out of ADC subscription entries would have helped diagnose #2632

### DIFF
--- a/firmware/hw_layer/adc/adc_inputs.cpp
+++ b/firmware/hw_layer/adc/adc_inputs.cpp
@@ -463,32 +463,11 @@ static void configureInputs(void) {
 
 	addChannel("Baro Press", engineConfiguration->baroSensor.hwChannel, ADC_SLOW);
 
-	addChannel("TPS 1 Primary", engineConfiguration->tps1_1AdcChannel, ADC_SLOW);
-	addChannel("TPS 1 Secondary", engineConfiguration->tps1_2AdcChannel, ADC_SLOW);
-	addChannel("TPS 2 Primary", engineConfiguration->tps2_1AdcChannel, ADC_SLOW);
-	addChannel("TPS 2 Secondary", engineConfiguration->tps2_2AdcChannel, ADC_SLOW);
-
-	addChannel("Wastegate Position", engineConfiguration->wastegatePositionSensor, ADC_SLOW);
-	addChannel("Idle Position Sensor", engineConfiguration->idlePositionSensor, ADC_SLOW);
-
-	addChannel("Fuel Level", engineConfiguration->fuelLevelSensor, ADC_SLOW);
-	addChannel("Acc Pedal1", engineConfiguration->throttlePedalPositionAdcChannel, ADC_SLOW);
-	addChannel("Acc Pedal2", engineConfiguration->throttlePedalPositionSecondAdcChannel, ADC_SLOW);
-	addChannel("VBatt", engineConfiguration->vbattAdcChannel, ADC_SLOW);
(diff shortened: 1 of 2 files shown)


firmware/hw_layer/adc/adc_inputs.cpp
```diff
@@ -463,32 +463,11 @@ static void configureInputs(void) {
 
 	addChannel("Baro Press", engineConfiguration->baroSensor.hwChannel, ADC_SLOW);
 
-	addChannel("TPS 1 Primary", engineConfiguration->tps1_1AdcChannel, ADC_SLOW);
```
rusefillc, line 466: I do not get it - do we just remove these lines?

rusefillc (changes requested):

firmware/hw_layer/adc/adc_inputs.cpp
mck1117: They're replaced by the call added in adc_subscription.cpp. Since the slow ADC no longer uses the conversion sequence computed at runtime (it just converts every channel), there's no need to manually initialize every analog channel here.

firmware/hw_layer/adc/adc_inputs.cpp
```diff
@@ -463,32 +463,11 @@ static void configureInputs(void) {
...
-	addChannel("VBatt", engineConfiguration->vbattAdcChannel, ADC_SLOW);
 	// not currently used	addChannel("Vref", engineConfiguration->vRefAdcChannel, ADC_SLOW);
-	addChannel("CLT", engineConfiguration->clt.adcChannel, ADC_SLOW);
-	addChannel("IAT", engineConfiguration->iat.adcChannel, ADC_SLOW);
-	addChannel("AUX Temp 1", engineConfiguration->auxTempSensor1.adcChannel, ADC_SLOW);
-	addChannel("AUX Temp 2", engineConfiguration->auxTempSensor2.adcChannel, ADC_SLOW);
 
 	addChannel("AUXF#1", engineConfiguration->auxFastSensor1_adcChannel, ADC_FAST);
 
 	addChannel("AFR", engineConfiguration->afr.hwChannel, ADC_SLOW);
```
rusefillc, line 470: Do we mean to keep some of "ADC_SLOW" like AFR here and Baro above?

If we remove all ADC_SLOW shall we rename addChannel to addFastChannel and drop fast parameter?

mck1117: yes, some channels aren't yet done with the sensor model (so adc subscription doesn't happen), so they still need the call here

mck1117: making the suggested change is the goal eventually - now tracked by #2667

PR #2661, opened 2021-05-09
Fix multi bank fueling

mck1117:
use cylinder number instead of cylinder index

changelog addition of multi-closed-loop-fuel


mck1117:
Tested on a real car, actually works properly this time!

Issue #2667, opened 2021-05-09
switch remaining sensors to sensor model


mck1117:
- AFR
- baro
- MAF
- FSIO analog inputs

rusefillc:
@mck1117 with Lambda sensor do we still have anything to do here?

PR #2672, opened 2021-05-09
Maf in sensor model

mck1117:
progress #2667


rusefillc:
@mck1117 please rebase on master?

Issue #2663, opened 2021-05-09
lua: output to FSIO pwm channels


mck1117:
Add a lua hook to set PWM duty cycle and frequency.

PR #2687, opened 2021-05-12
lua pwm hooks

mck1117:
#2663


mck1117:
depends on #2725 to make bootloader happy

rusefillc:
#2725 did not make BL happy?

mck1117:
indeed it did not, let's see what's up

mck1117:
ok, now everybody is happy!

Issue #2643, opened 2021-05-07
memory: why do we need 136 bytes for SimplePwm


mck1117:
what are we doing with all those bytes?

mck1117:
dupe by #3526

PR #2689, opened 2021-05-12
simpler simple PWM

mck1117:
try again now that stateChangeCallback is gone

Cuts `SimplePwm` from 136 -> 56 bytes too, #2643.  Nets about 2k of saving.


rusefillc:
@mck1117 poke?

PR #2691, opened 2021-05-13
Adc fixes

dron0gus:


firmware/hw_layer/adc/adc_inputs.cpp
```diff
@@ -74,7 +74,7 @@ AdcDevice::AdcDevice(ADCConversionGroup* hwConfig, adcsample_t *buf, size_t buf_
 	hwConfig->sqr5 = 0;
 #endif /* ADC_MAX_CHANNELS_COUNT */
 	memset(hardwareIndexByIndernalAdcIndex, EFI_ADC_NONE, sizeof(hardwareIndexByIndernalAdcIndex));
-	memset(internalAdcIndexByHardwareIndex, 0xFFFFFFFF, sizeof(internalAdcIndexByHardwareIndex));
+	memset(internalAdcIndexByHardwareIndex, 0xFF, sizeof(internalAdcIndexByHardwareIndex));
```
mck1117, line 77: not to mention that memset actually:
`The value is passed as an int, but the function fills the block of memory using the unsigned char conversion of this value.`

mck1117: So it should have been 0xff all along 🙂

mck1117 (approved):

firmware/hw_layer/adc/adc_inputs.cpp
dron0gus: It just a bit of cosmetic.

Issue #2585, opened 2021-04-26
TS indicator idle / running


rusefillc:
Considering the importance and complexity of idle control we need a top level TS indicator "idle/running"

rusefillc:
@mck1117 what's the idea behind existing

```
tsOutputChannels.isIdleClosedLoop = phase == Phase::Idling && CONFIG(idleMode) == IM_AUTO;
```
?

It's very tempting to replace it with
```
tsOutputChannels.isIdleLoop = phase == Phase::Idling
```

since the ``CONFIG(idleMode) == IM_AUTO`` part user should see in config?

rusefillc:
@mck1117 poke

PR #2692, opened 2021-05-13
idle indicators

mck1117:
#2585

Now there are two indicators:
- one for "coasting" vs "not coasting"
- one for "idling" vs "not idling"


mck1117:
aside: I wish TS had support for enum indicators so we could show "<blank>", "idling", "coasting", etc all on the same indicator

PR #2695, opened 2021-05-14
add ability to reset Lua interpreter

mck1117:
Does a full teardown and reset of the Lua state.  Needed for development scenarios (write new script, reload lua, observe result).

Also fixes the lua memory tracking math.


mck1117:
tested working!

PR #2697, opened 2021-05-14
lua hooks for custom airmass model

mck1117:
- Get any airmass model's results from Lua
- Set the values for a custom airmass model from Lua


rusefillc:
@mck1117 progress? close? poke later?

rusefillc:
this is now two months old. close? move forward?

Issue #1753, opened 2020-09-03
De-Fragment settings 20K


rusefillc:
unusedAtOldBoardConfigurationEnd could be moved to the end, we have a bunch of random memory gaps

things like ``bit unusedHereWeHave`` could be removed

etc etc

mck1117:
this has long been defragmented and refragmented several times

PR #2701, opened 2021-05-15
save table config space, store lua script in config

mck1117:
- Remove some small unused blocks, consolidating at the end #1753
- Make many tables scaled int16 or uint16 instead of float (saves half the space!)
- Store Lua script (2.5k) at the end

### DIFF
--- a/firmware/controllers/algo/rusefi_types.h
+++ b/firmware/controllers/algo/rusefi_types.h
@@ -81,6 +81,7 @@ typedef float percent_t;
 
 typedef void (*Void)(void);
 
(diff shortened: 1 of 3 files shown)


rusefillc:
we would need a changelog here

we would need to create a release before this super incompatible change

mck1117:
super incompatible ish - revving the flash version does mean TS will be able to port across the format change

mck1117:
Hmm, this is actually a huge hassle.  Let me try to improve the world a bit and come back to this.

rusefillc:
I like improving the world :)

mck1117:
see #2702

PR #2702, opened 2021-05-15
compile gpiochips as cpp

mck1117:
Trying to get all consumers of generated config structures to compile as C++, because then we can do lovely things like encode any scaling factor in the type, instead of manually specifying it everywhere.


rusefillc:
@dron0gus you are being promoted into a C++ developer!

mck1117:
@dron0gus welcome to the land of actual type safety!

mck1117:
related #1753

dron0gus (approved):
With my low C++ skils looks ok...

dron0gus:
> @dron0gus you are being promoted into a C++ developer!

You cant teach an old dog new tricks. :)

mck1117:
> You cant teach an old dog new tricks. :)

perfectly OK to just use C++ as "C with stricter type system"

firmware/config/boards/subaru_eg33/board_configuration.cpp
```diff
@@ -262,3 +263,97 @@ void setBoardDefaultConfiguration(void) {
...
+		[3] = {.port = GPIOE, .pad = 6},	/* INJ 8 */
+		/* ignition pre-dirvers */
+		[4] = {.port = GPIOC, .pad = 14},	/* IGN 2 */
+		[5] = {.port = GPIOC, .pad = 13},	/* IGN 4 */
+		[6] = {.port = GPIOC, .pad = 15},	/* IGN 6 */
+		[7] = {.port = GPIOI, .pad = 9},	/* IGN 8 */
+	},
+	/* en shared between two chips */
+	// TODO: dron0gus, what to do here?
+	.en = {.port = GPIOI, .pad = 7}
```
mck1117, line 335: @dron0gus what to do about this one? C++ requires that all fields are initialized, so should we explicitly init this as no-pin or something?

mck1117:
everybody's favorite proteus analog fail, but everything works!

PR #2714, opened 2021-05-16
Tooltip and UI updates

mck1117:
thanks to @BeerMoneyMotorsports


rusefillc:
I love and need my console commands to be documented, how do we not wipe this stuff out?

Worst case we would need to change code generator to have some content not go into TS ui but I need this content not nuked please.

mck1117:
Putting console commands in the tooltips so that Mr. Rusefi can read them is confusing for all users, and maybe useful for one human.

To quote you on Slack:

> as in everything is valid but serves only 1% of the customers

rusefillc:
As of today tooltips are also header code comments. I agree that we shall decouple butcannot just drop comment. Please either keep or decouple but not drop.

rusefillc:
mostly merged to master

PR #2718, opened 2021-05-17
Post cranking fuel tweaks

mck1117:
- sensible defaults
- bounds check

### DIFF
--- a/firmware/controllers/algo/engine2.cpp
+++ b/firmware/controllers/algo/engine2.cpp
@@ -132,8 +132,8 @@ void EngineState::periodicFastCallback(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 	fuelCutoffCorrection = getFuelCutOffCorrection(nowNt, rpm PASS_ENGINE_PARAMETER_SUFFIX);
 
 	// post-cranking fuel enrichment.
-	// for compatibility reasons, apply only if the factor is greater than zero (0.01 margin used)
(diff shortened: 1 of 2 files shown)


rusefillc:
@mck1117 legit red?

firmware/controllers/algo/engine2.cpp
```diff
@@ -132,8 +132,8 @@ void EngineState::periodicFastCallback(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
 	fuelCutoffCorrection = getFuelCutOffCorrection(nowNt, rpm PASS_ENGINE_PARAMETER_SUFFIX);
 
 	// post-cranking fuel enrichment.
-	// for compatibility reasons, apply only if the factor is greater than zero (0.01 margin used)
-	if (engineConfiguration->postCrankingFactor > 0.01f) {
+	// for compatibility reasons, apply only if the factor is greater than unity (only allow adding fuel)
```
rusefillc, line 135: "for compatibility reasons" it has been a while shall we already drop this if condition altogether? maybe fatal in case of zero or invalid setting?

mck1117: The default, until this PR, has been for that field set to 0, so everyone upgrading a config might still have 0 there.

rusefillc: Ohh, so next time.

mck1117: Yep, maybe later we can remove the check.

PR #2724, opened 2021-05-20
convert a table to scaled_channel

mck1117:
with magic, we can now encode the scaling factor in the type of the table, instead of using it everywhere by hand

depends on #2723


rusefillc:
I like useful magic :)

mck1117:
magic is ready!

PR #2725, opened 2021-05-20
Lua on f407, try again

mck1117:
looks like problems were unrelated to #2646, let's try it again


mck1117:
failure is expected VW121 config problem, all actual builds are green

Issue #1608, opened 2020-07-15
Coasting IAC position for Auto-idle should respect AC_Idle_Offset


stefanst:
It is believed that the value for Coasting IAC position sets an absolute IAC percentage value. This value should be modified by the "AC offset extra (percent)" value.

rusefi:
@andreika-git is this something you can look into?

mck1117:
fixed with 2745

Issue #1581, opened 2020-07-08
migrate amazing idle and A/C from FSIO to less opaque TS


rusefi:
@stefanst FYI

rusefi:
@stefanst first step new "extra" properties

![image](https://user-images.githubusercontent.com/7377949/86869683-fec95b80-c0a4-11ea-8859-705968fb5a68.png)

rusefi:
FSIO can access those via

cfg_ACIDLEEXTRAOFFSET
cfg_ACIDLEEXTRAMIN

rusefi:
Just switched FSIO into "extra" not "replace" logic

my testing uses the following formulas:
```
	engineConfiguration->useFSIO12ForIdleOffset = true;
	setFsioExpression(QUOTE(MAGIC_OFFSET_FOR_IDLE_OFFSET), "ac_on_switch 0 cfg_acIdleExtraOffset if" PASS_ENGINE_PARAMETER_SUFFIX);

	engineConfiguration->useFSIO13ForIdleMinValue = true;
	setFsioExpression(QUOTE(MAGIC_OFFSET_FOR_IDLE_MIN_VALUE), "ac_on_switch 0 cfg_acIdleExtraMin if" PASS_ENGINE_PARAMETER_SUFFIX);
```

stefanst:
This seems to work with following FSIO formulas:
1: "ac_on_switch rpm 900 > time_since_ac_on_switch 0.6 > & &" for AC compressor out
2: "ac_on_switch rpm 900 > time_since_ac_on_switch 0.3 > & &" for AC fan out
12: "ac_on_switch cfg_ACIDLEEXTRAOFFSET 0 if" for extra offset
13: "ac_on_switch cfg_ACIDLEEXTRAMIN 0 if" for extra min

Old values "offset 2" and "min 2" should be removed from TS

rusefillc:
@stefanst says

AC fan and AC compressor get switched using FSIO. Look at this tune: https://rusefi.com/online/view.php?msq=172
Under "FSIO Formulas" outputs 1 and 2 are AC compressor and AC fan respectively.

rusefillc:
We have acFanRelay and acRelayPin we just have to add control logic

mck1117:
done with #2745

PR #2745, opened 2021-05-27
apply open loop idle when in automatic mode

mck1117:
- ~~simplify reset flags (it's totally fine to reset PID more than once. It's just not worth the code complexity of 3 different flags tracking PID reset)~~ now in #2776 
- ~~move IAC coasting table to `getOpenLoop` function where it belongs~~ done in another PR
- Apply open loop correction while in closed loop ("automatic") mode, using the new (but yet unused) idle open loop math #1608 #1581 #1263


mck1117:
pending @BeerMoneyMotorsports and maybe me doing testing once I'm back from CR

andreika-git:
It seems like this is yet another "stealth removal" PR...
Where is additionalAir?
Where is idlePidRpmUpperLimit?

firmware/controllers/actuators/idle_thread.cpp
```diff
@@ -241,8 +234,10 @@ float IdleController::getRunningOpenLoop(float clt, SensorResult tps) const {
...
 	running += engine->acSwitchState ? CONFIG(acIdleExtraOffset) : 0;
-	// TODO: fan idle bump needs its own config field
-	running += enginePins.fanRelay.getLogicValue() ? CONFIG(acIdleExtraOffset) : 0;
+	running += enginePins.fanRelay.getLogicValue() ? CONFIG(fan1ExtraIdle) : 0;
+
+	// TODO: once we have dual fans, enable
+	//running += enginePins.fanRelay2.getLogicValue() ? CONFIG(fan2ExtraIdle) : 0;
 
 	// Now bump it by the specified amount when the throttle is opened (if configured)
 	// nb: invalid tps will make no change, no explicit check required
```
mck1117, line 243: @andreika-git here is `iacByTpsTaper`

firmware/controllers/actuators/idle_thread.cpp
```diff
@@ -397,44 +383,26 @@ static percent_t automaticIdleController(float tpsPos, float rpm, int targetRpm,
...
 	// if tps==0 then PID just works as usual, or we completely disable it if tps>=threshold
-	newValue = interpolateClamped(0.0f, newValue, CONFIG(idlePidDeactivationTpsThreshold), engine->engineState.idle.baseIdlePosition, tpsPos);
-
-	// Interpolate to the manual position when RPM is close to the upper RPM limit (if idlePidRpmUpperLimit is set).
-	// If RPM increases and the throttle is closed, then we're in coasting mode, and we should smoothly disable auto-pid.
-	// If we just leave IAC at baseIdlePosition (as in case of TPS deactivation threshold), RPM would get stuck. 
-	// That's why there's 'useIacTableForCoasting' setting which involves a separate IAC position table for coasting (iacCoasting).
-	// Currently it's user-defined. But eventually we'll use a real calculated and stored IAC position instead.
-	int idlePidLowerRpm = targetRpm + CONFIG(idlePidRpmDeadZone);
-	if (CONFIG(idlePidRpmUpperLimit) > 0) {
```
mck1117, line 426: @andreika-git `idlePidRpmUpperLimit` is at line 215:

```
	// If rpm too high (but throttle not pressed), we're coasting
	int maximumIdleRpm = targetRpm + CONFIG(idlePidRpmUpperLimit);
	if (rpm > maximumIdleRpm) {
		return Phase::Coasting;
	}
```

The problem with this code is that it didn't actually work without the IAC coasting table (which is an open loop correction, what is it doing in the CL logic?).  This is probably related to whatever @abelom was seeing in these issues #2768 #2772

mck1117: And BTW, trying to *taper a closed loop correction* is asking for integrator windup.  Consider this: the engine speed is high, but not enough to leave the idle zone.  Suppose you're right on the edge, so that interpolate ends up knocking out 90% of the correction being made.  You've now given the integrator license to wind up 10x more than it would have without the taper!

rusefillc:
@mck1117 is this the smallest possible atomic change?
is the amount of test coverage adequate for the amount of change in this PR?

unit_tests/tests/test_idle_controller.cpp
```diff
@@ -270,3 +307,178 @@ TEST(idle_v2, openLoopRunningTaper) {
...
+
+	// Determine phase will claim we're idling
+	EXPECT_CALL(dut, determinePhase(950, 1000, expectedTps))
+		.WillOnce(Return(ICP::Idling));
+
+	// Open loop should be asked for an open loop position
+	EXPECT_CALL(dut, getOpenLoop(ICP::Idling, expectedClt, expectedTps))
+		.WillOnce(Return(13));
+
+	// getClosedLoop() should not be called!
```
mck1117, line 415: this ensures that #2768 is impossible

mck1117:
@rusefillc I can at least pull out the simpler reset logic, but the rest of it really does need to be a "rip the bandaid" change.

rusefillc:
Thank you for all your contribution here!

Need a bit of food break and will return to looking into all this

@andreika-git please review current version when you have a chance: we have code style change already merged and unit test support and failing unit test also merged into master

mck1117:
I believe this is now actually the smallest change that fixes the problem.

The observed problem of "why doesn't automatic idle do anything" is somewhere near `// Well, just leave it as is, without PID regulation...`

But that reveals that there's actually a second half to the issue - those few lines of code are trying to make an open loop correction inside a function called `getClosedLoop`, which is fishy at best, and very probably a bug in itself.

The tl;dr is that this change makes `getClosedLoop` ONLY responsible for the small +- closed loop changes, and `getOpenLoop` responsible for the "large" open loop changes (change with cranking, change with clt, etc)

mck1117:
btw depends on #2779 or it will break cranking IAC position override

rusefillc:
Damn that's a huge change for "I believe this is now actually the smallest change that fixes the problem."

rusefillc:
I just refuse to believe that this PR is a smallest possible atomic change.

I am totally unable to comprehend this PR in it's current +200 -100 form.

Please really find a way to break this PR further apart.

Please close #2776 to reduce confusion

firmware/controllers/actuators/idle_thread.cpp
andreika-git: I am not sure that iacByTpsTaper can be replaced with fan1ExtraIdle. fan1ExtraIdle is dependent on fan relay state, and iacByTpsTaper is not.

mck1117: oops, I think github got confused with where the comment was placed, that code is just after where this is tagged:

```
	// Now bump it by the specified amount when the throttle is opened (if configured)
	// nb: invalid tps will make no change, no explicit check required
	running += interpolateClamped(
		0, 0,
		CONFIG(idlePidDeactivationTpsThreshold), CONFIG(iacByTpsTaper),
		tps.value_or(0));
```

rusefillc:
@andreika-git what's your take on next steps?
@mck1117 what's your take on next steps?

mck1117:
I'm done racing this weekend, and have swapped the truck that can tow the race trailer back for my rusEFI road car, so I can actually test this properly.

andreika-git:
The proposal to refactor the code and split the open & close loop parts is good. 
getClosedLoop() becomes an incremental regulator, returning zero when disabled. Fine.

BUT:

1a) "m_lastAutomaticPosition vs 0" in getClosedLoop() is still questionable. We have already tried just to "halt" the regulator when running->idling (which roughly equals to m_lastAutomaticPosition proposal), but in practice it doesn't work as good as one might expect. The engine load conditions often change drastically (btw, that's one of the reasons why we reset PID). Often running->idle situaton is just a latent coasting on a bumpy road, when the old saved CL additive is not effective anymore. Please don't treat all roads as perfect race tracks with quasi-steady load conditions. There are also bad roads out there... That's why we used baseIdlePosition - as a "safety place" for IAC in a situation when we don't know the best position (and neither does PID), and any "bad guess" can worsen the situation. 

1b) baseIdlePosition != currentIdlePosition, i.e. != OL+CL

> "baseIdlePosition = current position without adjustments" (c) engine_state.txt 

Moreover, baseIdlePosition was not even getOpenLoop() (because it already includes some adjustments), but I think we can try to fallback directly into OL (making CL = 0). 

1c) I also understand the reasoning behind m_lastAutomaticPosition. It's worth a try on some use-cases. That's why I propose to add a new option for a new behavior:
  * we replace "return m_lastAutomaticPosition" with "return getFallbackPosition()"
  * getFallbackPosition() { return (CONFIG(idleUseLastPidPositionForFallback)) ? m_lastAutomaticPosition : 0; }
  * as far as we don't need baseIdlePosition as fallback anymore, we can just make baseIdlePosition = getOpenLoop()

 Maybe it will make all of us happy :)

2) Why useIacTableForCoasting overrides afterCrankingIACtaper inside getOpenLoop()? IMHO coasting table should not work until revsSinceStart > afterCrankingIACtaperDuration, because the cranking->idling transition is fragile, the rpm is still unstable, and we don't want the stalling because of the coasting table (which is usually more "air-covered" that the cranking IAC table).

mck1117:
1. `m_lastAutomaticPosition` was an attempt to emulate the behavior of `baseIdlePosition`, since that's what `baseIdlePosition` actually did in those cases.  It froze the idle control until you returned, at which point it was reset and regulation continued.  I agree that this is the wrong behavior, but figured you might try to defend it as it was the old way :slightly_smiling_face: I'll change it for now to reset upon *leaving* the idle zone (and return 0), then resume when you re-enter (but keeping `m_lastAutomaticPosition` around for the deadzone).
1c. Yes, that might be a future worthy change if we actually need it.  I doubt it's necessary though, as the closed loop correction should stay pretty small (<10% perhaps).  If somebody thinks they need this, they probably need to configure their open loop correction properly.  I'll believe it when I see an engine that actually doesn't tolerate resetting the CL portion when leaving the idle region.
2. Overall better behavior is required out of the cranking-to-running transition.  I'd like to do that in a separate PR, as the current change here does work well enough.  Other changes are probably worth doing like suspending closed loop control (including timing) all together until the cranking taper has concluded.

andreika-git (approved):
I think now we're fine with this one for now!
And waiting for a separate PR for cranking-to-running transition (make CL and coasting table compatible with afterCrankingIACtaper etc.)

rusefillc:
Great news on the approval!

@mck1117 what's the next step to merge this change?

mck1117:
master should be green now, this can go in pending green check mark

PR #2747, opened 2021-05-28
parse with antlr

mck1117:
This change does:
- add antlr libs/ant target
- add grammar for rusEFI config txt file
- add parse tree walker to convert parsed tree to intermediate representation (from which we can generate ts/c++ structs/etc) (mostly in `ParseState.java`)
- minor fixes to `rusefi_configuration.txt`
- Run parsing on config generator, but don't actually use result (just check that the grammar is valid for now)


mck1117:
Yaml is just hilariously verbose.

For example, the line
`angle_t[IGNITION_PIN_COUNT iterate] timing_offset_cylinder;per-cylinder timing correction;"deg", 1, 0, -720, 720, 1`
might turn in to
```
- name: timing_offset_cylinder
  type: angle_t
  array: IGNITION_PIN_COUNT
  iterate: yes
  min: -720
  max: 720
  comment: "per cylinder timing correction"
  units: "deg"
  digits: 1
  scale: 1
  offset: 0
```

Eleven lines up from one line.  If every line increased by a factor of 10 like that, the config file would be 20000 lines long.

mck1117:
Another benefit is that this grammar is actually quite a bit smarter than the old one was - it doesn't read the suffix on fields as a string, but actually parses out the numbers, and can even do math in those fields.  You could make the offset value `MY_DEFINE * 2 + 1` and it will actually evaluate the value.

rusefillc:
OK, we have an argument for current custom funny format

why do we want to change the implementation from own code to external framework?

.github/workflows/build-rusEFI-console.yaml
```diff
@@ -20,6 +20,10 @@ jobs:
       - name: Install Tools
         run: sudo apt-get install ncftp
 
+      - name: Generate Java (Antlr)
+        working-directory: ./java_tools/configuration_definition
```
rusefillc, line 24: is "./" providing any value here?

firmware/integration/rusefi_config.txt
```diff
@@ -1954,7 +1954,7 @@ end_struct
 #define ENGINE_SNIFFER_UNIT_US 10
 
 ! These commands are used exclusively by the rusEfi console
-#define TS_TEST_COMMAND 't' // 0x74
```
rusefillc, line 1957: this change removes a valuable in my opinion comment. please do not remove the valuable in my opinion comment

.github/workflows/build-rusEFI-console.yaml
mck1117: Yes.  `./` forces it to look in the current directory, instead of looking in `$PATH`.

firmware/integration/rusefi_config.txt
mck1117: we never refer to this by anything other than char value `'t'`, but ok

.github/workflows/build-rusEFI-console.yaml
rusefillc: TIL thank you

PR #2750, opened 2021-05-28
Flash fix erase check

dron0gus:


rusefillc:
It was already a lot of magic and now we are replacing it, is it worth adding more code comments?

mck1117 (approved):

PR #2765, opened 2021-05-30
smarter enum file parsing

mck1117:
don't parse in the command line decoding, keep a list and parse afterward


rusefillc:
We have no CI here shall we add CI or jar or take the risk?

PR #2776, opened 2021-05-31
boring parts of idle pr #2745

mck1117:
pull out the boring parts of #2745:

- simplify reset flags (it's totally fine to reset PID more than once. It's just not worth the code complexity of 3 different flags tracking PID reset)
- only call getIdlePid once (easier to read!)
- code style


rusefillc:
Please spell out the reasoning behind this change.

I am _very_ bothered by the fact that current implementation has no coverage and consequently no coverage is being changed here

rusefillc:
Let's close this one?

PR #2779, opened 2021-05-31
make the cranking idle implementation match what the UI claims

mck1117:
Found while testing #2745.  Makes getOpenLoop do the right thing in case `overrideCrankingIacSetting` is set to false.

nb: this code path is not currently called by master, but will be after #2745

### DIFF
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -229,9 +229,14 @@ IIdleController::Phase IdleController::determinePhase(int rpm, int targetRpm, Se
 }
 
 float IdleController::getCrankingOpenLoop(float clt) const {
-	return 
-		CONFIG(crankingIACposition)		// Base cranking position (cranking page)
-		 * interpolate2d(clt, config->cltCrankingCorrBins, config->cltCrankingCorr);
+	float mult =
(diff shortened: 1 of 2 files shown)


andreika-git:
As far as I understand, the old working implementation was as follows:

```
// Use separate CLT correction table for cranking
else if (engineConfiguration->overrideCrankingIacSetting && !isRunning) {
	cltCorrection = interpolate2d("cltCrankingT", clt, config->cltCrankingCorrBins, config->cltCrankingCorr);
else {
	// this value would be ignored if running in AUTO mode
	// but we need it while cranking in AUTO mode
	cltCorrection = interpolate2d("cltT", clt, config->cltIdleCorrBins, config->cltIdleCorr);
}
```

Then we had #2146 in Dec 2020, and the implementation has been changed:
```
// Cranking value is either its own table, or the running value if not overriden
float cranking = CONFIG(overrideCrankingIacSetting) ? getCrankingOpenLoop(clt) : running;
```

So now this PR seems to revert to the old working implementation, where overrideCrankingIacSetting is controlling only the clt correction. I'm totally fine with it!

rusefillc:
is this a chance to give 'overrideCrankingIacSetting' a better name and mention CLT?

andreika-git (approved):

Issue #2735, opened 2021-05-23
minCrankingRpm does not reset between cranking attempts


rusefillc:
At the moment minCrankingRpm keeps it's value until ECU reset, we shall reset between separate cranking attempts

chuckwagoncomputing:
```
// TODO: wow move this into engineState at least for context not to leak from test to test!
// todo: reset this between cranking attempts?! #2735
```

I think `RpmCalculator` would be a good place to track this. What do you think?

rusefillc:
@chuckwagoncomputing sounds reasonable

PR #2787, opened 2021-06-01
Cranking defaults

mck1117:
more sensible defaults for:
- IAT timing correction table
- cranking duration table
- cranking clt mult table (smooth out the weird bump in the middle)


rusefillc:
Tests are failing on windows only
@mck1117 has mentioned #2735

PR #2790, opened 2021-06-02
defaults cleanup: base engine & fuel

mck1117:
- Move defaults to their own files
- This change moves the contents of the "base engine" and "fuel" menus in TS (first two)


mck1117:
Verified with TunerStudio that all config fields match exactly!

PR #2793, opened 2021-06-03
warnings & assorted cleanup

mck1117:


mck1117:
I should really fix the hardware so that it stops failing Proteus HW CI every now and then.

PR #2807, opened 2021-06-10
Fixes

dron0gus:


firmware/hw_layer/drivers/gpio/tle6240.cpp
```diff
@@ -218,7 +218,7 @@ int Tle6240::chip_init()
 	}
 
 	/* release reset */
-	if (!cfg->reset.port) {
+	if (cfg->reset.port) {
```
mck1117, line 221: oops

dron0gus: I know that I should review and test this last weekend...

mck1117: I thought andrey was going to wait to merge until you'd tested, but then I got email this morning it had merged :joy:

PR #2811, opened 2021-06-12
Trans Range Sensor - broken draft

chuckwagoncomputing:
I can't seem to get rid of this:
```
./console/binary/tunerstudio.cpp: In function 'void startTunerStudioConnectivity()':
./console/binary/tunerstudio.cpp:864:44: error: static assertion failed: TS datapage size mismatch
  864 |  static_assert(sizeof(persistent_config_s) == TOTAL_CONFIG_SIZE, "TS datapage size mismatch");
```

### DIFF
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -1244,8 +1244,8 @@ int16_t tps2Max;Full throttle#2. tpsMax value as 10 bit ADC value. Not Voltage!\
 	brain_pin_e binarySerialRxPin;	
 	
 	brain_pin_e[AUX_DIGITAL_VALVE_COUNT iterate] auxValves
-	int unused1206
-	int unused1207
+	int unused1206;;"units", 1, 0, -20, 100, 0
+	int unused1207;;"units", 1, 0, -20, 100, 0
 
 
 	float[ENGINE_NOISE_CURVE_SIZE] knockNoise;Knock sensor output knock detection threshold depending on current RPM;"v",      1,     0,   0.0,    10,   2
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -1546,8 +1546,8 @@ tChargeMode_e tChargeMode;
 	uint16_t mc33_t_hold_off;;"us",   1,   0.0,        0,  10000,   0
 	uint16_t mc33_t_hold_tot;;"us",   1,   0.0,        0,  10000,   0
 
-	int unused1506
-	int unused1507
+	int unused1506;;"units", 1, 0, -20, 100, 0
+	int unused1507;;"units", 1, 0, -20, 100, 0
 
 	pin_input_mode_e acSwitchMode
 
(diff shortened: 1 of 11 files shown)


firmware/integration/rusefi_config.txt
```diff
@@ -1052,7 +1057,42 @@ custom maf_sensor_type_e 4 bits, S32, @OFFSET@, [0:1], @@maf_sensor_type_e_enum@
 
 	output_pin_e[LUA_PWM_COUNT iterate] luaOutputPins
 
-	int[57] unusedAtOldBoardConfigurationEnd;;"units", 1, 0, -20, 100, 0
+	#define tcu_role_e_enum "Park", "Neutral" , "Park/Neutral", "Reverse", "Drive", "Manual", "Low", "Upshift", "Downshift", "3", "2", "1"
+	custom tcu_role_e 1 bits, U08, @OFFSET@, [0:3], @@tcu_role_e_enum@@
+	tcu_role_e	tcuState0Role;
```
mck1117, line 1062: should this be an array?

firmware/integration/rusefi_config.txt
```diff
@@ -1209,9 +1249,6 @@ int16_t tps2Max;Full throttle#2. tpsMax value as 10 bit ADC value. Not Voltage!\
 	brain_pin_e binarySerialRxPin;	
 	
 	brain_pin_e[AUX_DIGITAL_VALVE_COUNT iterate] auxValves
-	switch_input_pin_e tcuUpshiftButtonPin
```
mck1117, line 1212: you can't remove stuff from the middle of the config, it'll change the address of everything after it (which breaks compatibility)

chuckwagoncomputing: but it didn't change the address of everything after, because of the automatic 4-byte alignment

chuckwagoncomputing: I was brute force programming at that point, and removing everything was what worked.

firmware/integration/rusefi_config.txt
chuckwagoncomputing: Is this not an array?

rusefillc:
I've cherry picked some portion of the change into master, please rebase

rusefillc:
please rebase since some of the changes were cherry picked

firmware/integration/rusefi_config.txt
mck1117: an array would look like `tcu_role_e [8 iterate] tcuStateRole;`

chuckwagoncomputing: oh right. No, it's not supposed to be an array.

mck1117: Why not? BTW adding the word "iterate" makes it spit out non-array for TS, but an array for the c++ part.

chuckwagoncomputing:
I'm deleting this so new changes on my branch don't run actions here.

firmware/integration/rusefi_config.txt
chuckwagoncomputing: oh I am an idiot

Issue #2810, opened 2021-06-11
gen_config, YAML: explicit failure in case of invalid pin class


rusefillc:
```
  - pin: 14a
    id: EFI_ADC_0
    class: analog_input
    ts_name: 14 - O2
    function: O2 sensor signal
```

note invalid class not analog_inputs. @chuckwagoncomputing shall we fail nicely?

chuckwagoncomputing:
This is fixed

PR #2842, opened 2021-06-21
Yaml parsing refactor

### LINKED ISSUE #2801: h121 pps1 pin conflict
in hellen121vag yaml looks like same EFI_ADC_3 was used twice: PPF1 and IN_RES2

1) short term need a fix in hellen121 yaml
2) long term open question if pinout generator should be failing faster in case of similar pin conflict?

chuckwagoncomputing:
This fixes #2801 and #2810


java_tools/configuration_definition/src/com/rusefi/ConfigDefinition.java
```diff
@@ -414,75 +405,126 @@ private static void processYamlFile(File yamlFile, ReaderState state,
...
                 }
-            } else if (idObject instanceof String ) {
-                findMatchingEnum((String) idObject, (String) pin.get("ts_name"), (String) pin.get("class"), state, listOutputs, listAnalogInputs, listEventInputs, listSwitchInputs);
+            } else if (pinId instanceof String ) {
+                Map<String, Object> thisPin = new HashMap();
+                thisPin.put("id", pinId);
+                thisPin.put("ts_name", pinName);
+                thisPin.put("class", pinClass);
+                listPins.add(thisPin);
             }
```
mck1117, line 432: we should throw in the `else { }` case if it's neither an array nor string

java_tools/configuration_definition/src/com/rusefi/ConfigDefinition.java
```diff
@@ -414,75 +405,126 @@ private static void processYamlFile(File yamlFile, ReaderState state,
...
-                list.put(kv.getValue().getIntValue(), ts_name);
-            } else if (kv.getKey().equals(nothingName)) {
-                list.put(kv.getValue().getIntValue(), "NONE");
+        for (Map.Entry<String, ArrayList<String>> kv : names.entrySet()) {
+            String outputEnumName = "";
+            String pinType = "";
+            String nothingName = "";
+            switch (kv.getKey()) {
+                case "outputs":
+                    outputEnumName = "output_pin_e_enum";
```
rusefillc, line 486: I've just pushed https://github.com/rusefi/rusefi/blob/master/java_tools/configuration_definition/src/com/rusefi/PinType.java

please refactor that enum to your liking

open question if find() method should throw exception or return null for unexpected? at the moment I am not sure why do we not handle "else" as a failure

java_tools/configuration_definition/src/com/rusefi/ConfigDefinition.java
chuckwagoncomputing: I should be able to get the class name with pinId.getClass().getCanonicalName(), right? Or do I want getSimpleName()?

chuckwagoncomputing: Oh or getName()

mck1117: something like that, yeah

java_tools/configuration_definition/src/com/rusefi/ConfigDefinition.java
chuckwagoncomputing: oh nice. I will use this. tomorrow.

chuckwagoncomputing:
I think this is ready to go. I have verified that the output matches.

Issue #2839, opened 2021-06-20
TS Burned All Data engine stop


Abricosvw:
when TS burn itself, then the engine stops ... 
i checked 3 latest firmware releases ...
![New Bitmap Image (2) (1)](https://user-images.githubusercontent.com/35018246/122683660-a9f7f580-d1ce-11eb-865c-e0e22cd5aa36.jpg)

rusefillc:
I am able to reproduce this with a discovery

1) step 1: full chip erase and flash rusefi_bundle Garbage Man Day.zip
2) connect with TS
3) enable internal stimulator
4) go make any change, hit "Save"

actual result: I get blue LED freee and write
expected result: no blue LED freeze, "unsaved changes" indicator in TS

rusefillc:
I assume we've broken it while making logic smarter defect which chips have two banks and allow hot write?

mck1117:
This is expected behavior - we're supposed to burn while self stimulating

rusefillc:
Ok, now with external stim

1) I set trigger to "one tooth"
2) I have one wire from PD1 to PC6
3) I play with "displacement" field and "save" button

Sometimes things are great. Sometimes sync is lost and RPM goes to zero. At least once I got 3k or 4k for a moment instead of 1200.

rusefillc:
Relevant console commands

"ext_stimulation"
"burnconfig"

dron0gus:
Looks like scheduling issue. Something is blocking CPU for a while.
Self stimulation. Subaru SVX trigger. First trigger is 12 pulses per rotation. With equal gaps.
![Screenshot from 2021-06-21 14-11-57](https://user-images.githubusercontent.com/28624689/122753472-fd377a00-d29a-11eb-9bf7-fcd1794c9cf9.png)

dron0gus:
1. enable self stimulation
2. go to High Speed Logger, start
3. open Ignition->Ignition cylinder trim
4. Adjust any value in this dialog
5. observe trigger errors.
Somehow TS calls flash write on any change on these values.

rusefillc:
self stimulation is a special case

to reproduce real issue you have to use a wire and external stimulation if on discovery 

```
	if (ENGINE(directSelfStimulation) || engine->rpmCalculator.isStopped()) {
		/**
		 * rusEfi usually runs on hardware which halts execution while writing to internal flash, so we
		 * postpone writes to until engine is stopped. Writes in case of self-stimulation are fine.
		 *
[...]

PR #2847, opened 2021-06-23
TS communication minor fixes

dron0gus:
Some related to https://github.com/rusefi/rusefi/issues/2839 , but does not fix it.

- do not stop "communication" blinking in case TS command executes too long. CRC calculation command previously sometimes cause communication LED to switch to slow blinking for a one period.

### DIFF
--- a/firmware/console/binary/tunerstudio.cpp
+++ b/firmware/console/binary/tunerstudio.cpp
@@ -99,6 +99,9 @@
 
 EXTERN_ENGINE;
 
+/* 1S */
+#define TS_COMMUNICATION_TIMEOUT	TIME_MS2I(1000)
+
 extern persistent_config_container_s persistentState;
 
 #if !defined(EFI_NO_CONFIG_WORKING_COPY)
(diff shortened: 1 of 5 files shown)


firmware/hw_layer/ports/stm32/flash_int.c
```diff
@@ -82,7 +82,10 @@ static bool intFlashUnlock(void) {
 /**
  * @brief Lock the flash memory for write access.
  */
-#define intFlashLock() { FLASH_CR |= FLASH_CR_LOCK; }
+static void intFlashLock(void)
```
rusefillc, line 85: why do we make this macro a function?

firmware/console/binary/ts_can_channel.cpp
```diff
@@ -10,7 +10,7 @@ class CanTsChannel : public TsChannelBase {
 
 	// TsChannelBase implementation
 	void write(const uint8_t* buffer, size_t size) override;
-	size_t readTimeout(uint8_t* buffer, size_t size, int timeout) override;
+	size_t readTimeout(uint8_t* buffer, size_t size, sysinterval_t timeout) override;
```
rusefillc, line 13: something is not right here

we change method signature only for ts_can_channel while not changing it for all implementations of ts channel

if we do a global rearch&replace we break unit tests which do not know about this ChibiOS type

since that's not important let's just undo this part of the change?

firmware/hw_layer/ports/stm32/flash_int.c
dron0gus: To be inline with intFlashUnlock
Also call of current macro cause useless semicolon:
`intFlashLock();`
Extended to
`{ FLASH_CR |= FLASH_CR_LOCK; } ;`

firmware/console/binary/ts_can_channel.cpp
dron0gus: Ups.
Anyway we pass this timeout to ChisiOS internals where `sysinterval_t` is expected. This type is unsigned 32bit.

dron0gus:
Commits
_flash: intFlashLock convert to function
console: use sysinterval_t type for timeouts_ 
Removed from this PR

firmware/hw_layer/ports/stm32/flash_int.c
mck1117: @rusefillc for simple macros like this there is no value keeping it a macro, the compiler will inline it since the body is small

PR #2853, opened 2021-06-24
don't stack overflow on early ch dbg assert

mck1117:


firmware/controllers/core/error_handling.cpp
```diff
@@ -71,10 +71,6 @@ void chDbgPanic3(const char *msg, const char * file, int line) {
 #if EFI_HD44780_LCD
 	lcdShowPanicMessage((char *) msg);
 #endif /* EFI_HD44780_LCD */
-
-	if (!main_loop_started) {
```
rusefillc, line 75: is main_loop_started still needed after this change?

mck1117: yes because I'm about to use it again

Issue #2840, opened 2021-06-21
Random discovery has RED led and no USB connectivity with rusefi_bundle Garbage Man Day.zip


rusefillc:
step 1: random discovery from pile

step 2: auto DFU to rusefi_bundle Garbage Man Day.zip

step 3: red LED, no USB comms

rusefillc:
[red_discovery.zip](https://github.com/rusefi/rusefi/files/6683665/red_discovery.zip)

rusefillc:
@mck1117 any chance you have a discovery to flash attached image?

mck1117:
will check today

mck1117:
CJ125 is trying to initialize using SPI2.

SPI2 TX uses DMA controller 1, stream 3

However, C1S3 is already used by USART3 TX

rusefillc:
why does USB not USB?

mck1117:
because it hits an `osalDbgAssert` which disables interrupts, then `while (1) { }`

rusefillc:
Do we have an opportunity to implement a smarter osalDbgAssert ?

PR #2854, opened 2021-06-25
misc threading/safety improvements

mck1117:
- ThreadController can terminate
- Fatal error on unexpected thread terminate
- Don't allow pin init if fatal error occured


mck1117:
tangentially related to #2840

PR #2859, opened 2021-06-25
more graceful chibios assertion failure

### LINKED ISSUE #2840: Random discovery has RED led and no USB connectivity with rusefi_bundle Garbage Man Day.zip

mck1117:
Store the state using `setjmp` at the start of initialization, and call `longjmp` if a failure occurs during initialization.  This way we can present chibios assertion failures (etc) to the user as fatal error messages.

fixes #2840


rusefillc:
Did you mention there was some cool magic here, some magic header? Can you please elaborate a bit in code comments?

rusefillc:
@mck1117 poke docs?

PR #2860, opened 2021-06-25
fix trigger changed logic

### LINKED ISSUE #2839: TS Burned All Data engine stop

mck1117:
You can't compare arrays by value in c++.

Fixes #2839


firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -686,28 +686,32 @@ static void resetRunningTriggerCounters() {
...
-		isConfigurationChanged(trigger.type) ||
-		isConfigurationChanged(ambiguousOperationMode) ||
-		isConfigurationChanged(useOnlyRisingEdgeForTrigger) ||
-		isConfigurationChanged(globalTriggerAngleOffset) ||
-		isConfigurationChanged(trigger.customTotalToothCount) ||
-		isConfigurationChanged(trigger.customSkippedToothCount) ||
-		isConfigurationChanged(triggerInputPins[0]) ||
-		isConfigurationChanged(triggerInputPins[1]) ||
-		isConfigurationChanged(triggerInputPins[2]) ||
-		isConfigurationChanged(vvtMode) ||
```
mck1117, line 707: This line, `isConfigurationChanged(vvtMode)`, was the problem.  If you compare arrays with identical contents, they will compare not equal since it will actually compare the _pointer_, not the values within the array.

rusefillc:
Exciting! What are we missing to have this in master? :)

mck1117:
It seems to have helped, but not 100% fixed...

PR #2870, opened 2021-06-26
fix ETB PWM

### LINKED ISSUE #2868: ETB starts extra software PWM
Why does ETB start multiple software PWM even when using hard PWM?

mck1117:
fix #2868 

- use only two PWM devices instead of three
- Only configure PWM if actually using it

depends on #2869


rusefillc:
@mck1117 close? finish? poke later?

mck1117:
Tested with logic analyzer in both modes, looks good!

PR #2873, opened 2021-06-27
hoist hwHandleShaftSignal

mck1117:
Simplify hwHandleShaftSignal logic - we don't need the logic duplicated everywhere to decide which signal maps to primary/secondary rise/fall.

As a result, we can now call that from `CsvReader::processLine`.

### DIFF
--- a/firmware/hw_layer/digital_input/trigger/trigger_input_exti.cpp
+++ b/firmware/hw_layer/digital_input/trigger/trigger_input_exti.cpp
@@ -42,23 +42,8 @@ static void shaft_callback(void *arg) {
 	// todo: start using real event time from HW event, not just software timer?
 
 	bool isPrimary = index == 0;
-	if (!isPrimary && !TRIGGER_WAVEFORM(needSecondTriggerInput)) {
-		return;
-	}
 
-	trigger_event_e signal;
-	// todo: add support for 3rd channel
-	if (rise) {
-		signal = isPrimary ?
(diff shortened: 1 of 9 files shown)


firmware/hw_layer/digital_input/trigger/trigger_input_exti.cpp
```diff
@@ -42,23 +42,8 @@ static void shaft_callback(void *arg) {
...
-					(engineConfiguration->invertPrimaryTriggerSignal ? SHAFT_PRIMARY_FALLING : SHAFT_PRIMARY_RISING) :
-					(engineConfiguration->invertSecondaryTriggerSignal ? SHAFT_SECONDARY_FALLING : SHAFT_SECONDARY_RISING);
-	} else {
-		signal = isPrimary ?
-					(engineConfiguration->invertPrimaryTriggerSignal ? SHAFT_PRIMARY_RISING : SHAFT_PRIMARY_FALLING) :
-					(engineConfiguration->invertSecondaryTriggerSignal ? SHAFT_SECONDARY_RISING : SHAFT_SECONDARY_FALLING);
-	}
 
-	hwHandleShaftSignal(signal, stamp);
+	hwHandleShaftSignal(isPrimary, rise, stamp);
```
rusefillc, line 46: "// todo: add support for 3rd channel" would not be very happy with isPrimary

mck1117: do we actually want/need a third channel? do we know of a car that needs third channel?

rusefillc: too many unrelated changes in this PR :(

Issue #2861, opened 2021-06-25
VW warnings on cranking part #1


Abricosvw:
9002 

CUSTOM_SYNC_COUNT_MISMATCH = 9002

Abricosvw:
https://rusefi.com/online/view.php?log=270

Abricosvw:
[2021-06-25_19.58.35.zip](https://github.com/rusefi/rusefi/files/6719293/2021-06-25_19.58.35.zip)

mck1117:
It's seeing an early sync point here:

![image](https://user-images.githubusercontent.com/568254/123531616-e3db6700-d6ba-11eb-8973-ccb53c8ccde7.png)

mck1117:
Then a real one here:
![image](https://user-images.githubusercontent.com/568254/123531621-ed64cf00-d6ba-11eb-9039-ef6da21f500e.png)

mck1117:
@Abricosvw was this log made with a MAX992x chip on Frankenso?

If so, swap your wires:

![image](https://user-images.githubusercontent.com/568254/123531698-a88d6800-d6bb-11eb-8614-379f5c2a4efb.png)

mck1117:
https://github.com/rusefi/rusefi/wiki/Trigger-Configuration-Guide

rusefillc:
@mck1117 you are right about the early sync point but making gap expectation smarter does not improve things

```
	s->setTriggerSynchronizationGap2(1.6, 4);
	s->setSecondTriggerSynchronizationGap(1); // redundant gap ratio to fight start-up noise
	s->setThirdTriggerSynchronizationGap(1);
```

the warning is about constant 
```
sync point: isDecodingError=1
count: cur=116 exp=58
count: cur=0 exp=0
count: cur=0 exp=0
```
which I believe is related to a bug in expected tooth calculaton

mck1117:
That warning is a product of the broken csv replay logic :)

rusefillc:
@mck1117 any specific patched you would want to contribute? :)

mck1117:
crafting a PR now!

mck1117:
with #2874, tracking two extra gaps /does/ prevent the warning!
```
s->setTriggerSynchronizationGap2(1.6, 4);
s->setSecondTriggerSynchronizationGap(1); // redundant gap ratio to fight start-up noise
s->setThirdTriggerSynchronizationGap(1);
```

Abricosvw:
[...]

PR #2874, opened 2021-06-27
fix trigger replay

mck1117:
smaller piece

- don't reinvent the wheel for how to select signal, just copy-paste logic from `trigger_input_exti.cpp` (et al)
- call `handleShaftSignal2`, doing exactly the same thing as real triggers and the engine emulator

progress to #2861: this reduces the warning count to just 1


rusefillc:
Today was a really long day for me with all the chip drama VS JLCPCB software changes VS miata at the gap deadline

Can we please NOT rename the variable in this PR to make it even smaller?

rusefillc:
I've merged a smaller fix https://github.com/rusefi/rusefi/commit/f6b003820c71c07f277d2e6aa0be40b371c9df2f

thank you for raising the flag!

PR #2875, opened 2021-06-27
Consolidate hwHandleShaftSignal

mck1117:
- Move logic to translate [index, rise] -> trigger_event_e to a single place, instead of 4 duplicates of the same logic
- No more handleShaftSignal2: everyone now calls hwHandleShaftSignal directly (including tests!)
-

### DIFF
--- a/firmware/controllers/trigger/trigger_emulator_algo.cpp
+++ b/firmware/controllers/trigger/trigger_emulator_algo.cpp
@@ -55,13 +55,8 @@ void TriggerEmulatorHelper::handleEmulatorCallback(const int size, const MultiCh
 	for (size_t i = 0; i < PWM_PHASE_MAX_WAVE_PER_PWM; i++) {
 		if (needEvent(stateIndex, size, multiChannelStateSequence, i)) {
 			pin_state_t currentValue = multiChannelStateSequence.getChannelState(/*phaseIndex*/i, stateIndex);
-			
-			constexpr trigger_event_e riseEvents[] = { SHAFT_PRIMARY_RISING, SHAFT_SECONDARY_RISING, SHAFT_3RD_RISING };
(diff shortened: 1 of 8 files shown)


firmware/controllers/trigger/trigger_emulator_algo.cpp
```diff
@@ -55,13 +55,8 @@ void TriggerEmulatorHelper::handleEmulatorCallback(const int size, const MultiCh
...
 		if (needEvent(stateIndex, size, multiChannelStateSequence, i)) {
 			pin_state_t currentValue = multiChannelStateSequence.getChannelState(/*phaseIndex*/i, stateIndex);
-			
-			constexpr trigger_event_e riseEvents[] = { SHAFT_PRIMARY_RISING, SHAFT_SECONDARY_RISING, SHAFT_3RD_RISING };
-			constexpr trigger_event_e fallEvents[] = { SHAFT_PRIMARY_FALLING, SHAFT_SECONDARY_FALLING, SHAFT_3RD_FALLING };
 
-			trigger_event_e event = (currentValue ? riseEvents : fallEvents)[i];
-
-			handleShaftSignal2(event, stamp PASS_ENGINE_PARAMETER_SUFFIX);
+			hwHandleShaftSignal(i, currentValue, stamp PASS_ENGINE_PARAMETER_SUFFIX);
```
rusefillc, line 59: this change breaks QC behavior

at the moment self-stimulation does not blink LED in VR_HW_CHECK_MODE while real HW VR does blink VR_HW_CHECK_MODE

rusefillc (changes requested):

firmware/controllers/trigger/trigger_emulator_algo.cpp
mck1117: Ahhh so that was the intended behavior.

PR #2886, opened 2021-07-01
fix instant rpm

### LINKED ISSUE #2885: movePreSynchTimestamps is called too often
movePreSynchTimestamps should be called exactly once, at the first sync event

mck1117:
fix #2885 

call movePreSynchTimestamps exactly once, on the first sync


rusefillc:
Shall we cover engine stop and engine re-start?

mck1117:
see #2888

PR #2891, opened 2021-07-01
Stricter miata sync

mck1117:


mck1117:
tested on a real car by our good friend @BeerMoneyMotorsports

Issue #2889, opened 2021-07-01
realCrankingFromFile shouldn't cause a warning


mck1117:
Here's what is happening in realCrankingFromFile:

<img width="490" alt="image" src="https://user-images.githubusercontent.com/568254/124074597-95023a00-d9f8-11eb-83d4-feade1d4ab85.png">

My suspicion is that the rapid RPM change causes the two events with sparkid 0 to get scheduled "around" the sparkid 2, and since the Miata runs wasted spark, two events can get scheduled over the top of each other that will actuate the same coil.

mck1117:
The rapid RPM change is only because the early instantRPM samples were bogus - fixed by #2892

PR #2892, opened 2021-07-01
Fix pre sync timestamp copy

mck1117:
We were copying timestamps in the wrong order, and improperly if we had more events in the buffer than the length of the trigger.


mck1117:
fixes #2889 - the wrong early RPM was causing mis-scheduled events

Issue #2813, opened 2021-06-12
inhibit closed loop idle until after the cranking -> running taper


mck1117:
Let the cranking->running idle taper happen without closed loop trying to mess with it.

stefanst:
Any progress on this? I just slammed into a problem with IAC valve dropping right after cranking.

rusefillc:
my start is now weird i wonder if my start is weird for same reason. will post logs today

rusefillc:
is https://rusefi.com/forum/viewtopic.php?p=41684#p41684 relared to this issue?

mck1117:
I think this is properly fixed now after #3051.

PR #2900, opened 2021-07-04
Disable closed loop idle during cranking->running taper

mck1117:
#2813 

fyi @andreika-git


firmware/controllers/actuators/idle_thread.cpp
```diff
@@ -228,10 +228,19 @@ IIdleController::Phase IdleController::determinePhase(int rpm, int targetRpm, Se
 		return Phase::Coasting;
 	}
 
+	// If still in the cranking taper, disable closed loop idle
+	if (crankingTaperFraction < 1) {
+		return Phase::Running;
```
andreika-git, line 233: Maybe we need a new separate phase for cranking-to-running transition state? Or just return it as Phase::Cranking here?

mck1117: yeah, I guess a separate phase is cleaner

PR #2903, opened 2021-07-04
Chibios-Contrib for stm32

dron0gus:
Chibios-Contrib contains some useful stuff. Like HW CRC calculator driver for STM32.
Include it into build and add configs for STM32 chips

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -151,7 +151,7 @@ include $(CHIBIOS)/os/license/license.mk
 include $(CPU_STARTUP_DIR)
 # HAL-OSAL files (optional).
 include $(CHIBIOS_CONTRIB)/os/hal/hal.mk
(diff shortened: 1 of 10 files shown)


mck1117:
is there a specific thing you wanted to add that's implemented in `ChibiOS-Contrib`?

dron0gus:
> is there a specific thing you wanted to add that's implemented in `ChibiOS-Contrib`?
Yes. HW CRC calculation unit driver. Gives some boost over SW implementation on blocks > 300 bytes.

mck1117:
I don't remember if it lets you configure the polynomial, does it?

dron0gus:
CRC unit at F7 and H7 allow to configure CRC bit size and polynomial.
F407 and F247 do not.

PR #2907, opened 2021-07-05
Fix bootloader build

dron0gus:


firmware/bootloader/src/Makefile
```diff
@@ -116,12 +116,12 @@ endif
...
 # Define project name here
 PROJECT = bootloader
-PROJECT_DIR = ..
+PROJECT_DIR = ../..
 
 # Imported source files and paths
 CHIBIOS = $(PROJECT_DIR)/ChibiOS
 # todo: looks like 'CHIBIOS_CONTRIB' path is universal shall we defined it only once?
-CHIBIOS_CONTRIB = $(CHIBIOS)/../ChibiOS-Contrib
+CHIBIOS_CONTRIB = $(PROJECT_DIR)/ChibiOS-Contrib
```
rusefillc, line 124: can it be $(CHIBIOS)/../ChibiOS-Contrib ?

dron0gus: > can it be $(CHIBIOS)/../ChibiOS-Contrib ?

Yes. But Chibios and Chibios-Contrib are two independent units. Why should one be related to another?

rusefillc:
on the one hand I agree that two unrelated entities

on the other hand it's nice to have contrib path defined in only one place not in 4 different places the way we have it.

PR #2909, opened 2021-07-05
Hw crc

dron0gus:
This enables CRC driver and HAL for STM32F7 and STM32H7 based boards.


mck1117:
You can just check ‘STM32F7xx` instead of with the mcuconf on the end.

rusefillc:
We have working CRC code
Looks like SW CRC would have to stay for F4 and kinetis

I am not convinced that extra complexity of HW CRC on F7 is justified. Am I missing something?

mck1117:
should we resurrect?

rusefillc:
What metric do we expect to improve? Code Complexity would go up

PR #2922, opened 2021-07-07
Update main.yaml

duhafnusa4:
Added ID, class, ts_name, type, and color to each pin.

### DIFF
--- a/firmware/config/boards/hellen/hellen128/connectors/main.yaml
+++ b/firmware/config/boards/hellen/hellen128/connectors/main.yaml
@@ -1,482 +1,1161 @@
 pins:
   - pin: A1
+    id: 
+    class: 
+    ts_name: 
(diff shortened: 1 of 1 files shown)


rusefillc:
@chuckwagoncomputing whats your prediction re empty values?

chuckwagoncomputing:
Well I expected this to catch empty ID fields:
https://github.com/rusefi/rusefi/blob/00b32798e7962eca96ffeb88fdaa1df6c9a67e2c/java_tools/configuration_definition/src/com/rusefi/ConfigDefinition.java#L438-L440

But we don't seem to be getting any error so apparently something's not working. Maybe it parses as an empty string?

duhafnusa4:
my weak attempt at PR has revealed a real issue. Glad to help haha

PR #2926, opened 2021-07-08
fix vvt indication

### LINKED ISSUE #2755: VVT position is all over the place
We had a change RE: VVT position range, we need more changes since as of today we end up with VVT position of 1200 degrees and else

image

https://rusefi.com/online/view.php?log=243

### LINKED ISSUE #2921: Matt has CUSTOM_ERR_VVT_OUT_OF_RANGE warning
See https://rusefi.com/forum/viewtopic.php?p=41486#p41486

mck1117:
- VVT output channels are signed, so they can output negative values (fixes TS)
- VVT position does indeed need to get wrapped (offset isn't enough) - if the VVT sync point moves across the crank sync point, we need to smoothly output negative -> positive values, instead of wrapping all the way around.
- Normalize TS channel names: bank 1/2 intake/exhaust
- Now that we auto-wrap VVT position, remove warning about VVT out of range

fix #2755 
fix #2921


firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -265,13 +265,18 @@ void hwHandleVvtCamSignal(trigger_value_e front, efitick_t nowNt, int index DECL
...
-	tc->vvtPosition[bankIndex][camIndex] = engineConfiguration->vvtOffsets[bankIndex * CAMS_PER_BANK + camIndex] - currentPosition;
-	if (tc->vvtPosition[bankIndex][camIndex] < -ENGINE(engineCycle) / 2 || tc->vvtPosition[bankIndex][camIndex] > ENGINE(engineCycle) / 2) {
-		warning(CUSTOM_ERR_VVT_OUT_OF_RANGE, "Please adjust vvtOffset since position %f", tc->vvtPosition);
+	auto vvtPosition = engineConfiguration->vvtOffsets[bankIndex * CAMS_PER_BANK + camIndex] - currentPosition;
+
+	// Wrap VVT position in to the range [-360, 360)
+	while (vvtPosition < -360) {
+		vvtPosition += 720;
+	}
+	while (vvtPosition > 360) {
```
rusefillc, line 274: this code seems to be doing [-360, 360]

should it be (vvtPosition >= 360) {

?

rusefillc:
"if the VVT sync point moves across the crank sync point, we need to smoothly output negative -> positive values, instead of wrapping all the way around."

please spell this a bit further. So far I am not getting it.

rusefillc:
I've picked the rename and TS part of the change please rebase

mck1117:
> "if the VVT sync point moves across the crank sync point, we need to smoothly output negative -> positive values, instead of wrapping all the way around."
> 
> please spell this a bit further. So far I am not getting it.

Suppose that you have two single tooth triggers - one at cam speed, and one for vvt.  Your "zero point" might be when the two rising edges are perfectly aligned.

If you retard the cam slightly, te duration offset will be positive, so no wrap needed.

If you advance the cam slightly, the offset would be a negative amount of time, but we can't measure negative time, so it comes out as nearly 720 degrees worth of time instead (one full cycle minus the small amount of time).

So if the cam is very near the zero point, it will be bouncing between ~0 and ~720.  No amount of setting the offset will fix that if the ECU doesn't wrap the values itself.

PR #2927, opened 2021-07-08
init loops after sensors

mck1117:
Avoids a fatal error at startup if you have ETB idle.


rusefillc:
is this a bug in current release?

please CHANGELOG this if that's a user-facing bugfix

mck1117:
bug in release no
bug in current master yes

Issue #2931, opened 2021-07-08
trailing spark outputs


mck1117:
Fire a second coil, delayed from the first by a programmable angle (table). Used for Mazda Rotary, engines with dual plugs, etc.

rusefillc:
@mck1117 I've pushed two changes
1) TS project - outputs and constant angle setting as a temp solution
2) some of the output wiring no idea how much I've missed

angle setting is totally not used yet, not planning any more changes here in the next 24 hours

rusefillc:
```
/**
 * set engine_type 107
 */
void setRotary(DECLARE_CONFIG_PARAMETER_SIGNATURE) {
```

```
2021-07-21_20_38_16_721: EngineState: pin PC9: Trail 1
2021-07-21_20_38_16_721: EngineState: pin PC10: Primary Channel TX
2021-07-21_20_38_16_721: EngineState: pin PC11: Primary Channel RX
2021-07-21_20_38_16_722: EngineState: pin PD12: led: running status
2021-07-21_20_38_16_723: EngineState: pin PD13: led: warning status
2021-07-21_20_38_16_723: EngineState: pin PD14: led: CRITICAL status
2021-07-21_20_38_16_724: EngineState: pin PD15: led: comm status
2021-07-21_20_38_16_726: EngineState: pin PE10: Trail 2

```

PR #2932, opened 2021-07-08
trailing spark scheduling

mck1117:
#2931


rusefillc:
unit test? what is missing to merge?

rusefillc:
still need pins into TS and pins GPIO init?

mck1117:
Yep, but figured I'd break it up.  There's also no way to actually set the trailing amount. Or even turn it on.

Issue #2928, opened 2021-07-08
Epic: Remove FSIO


rusefillc:
Lua is the way!

We shall migrate from FSIO to Lua

Depends on #2663

mck1117:
FSIO is gone

PR #2933, opened 2021-07-09
Fuel pump not fsio

mck1117:
#2928

### DIFF
--- a/firmware/console/status_loop.cpp
+++ b/firmware/console/status_loop.cpp
@@ -699,7 +699,6 @@ void updateTunerStudioState(TunerStudioOutputChannels *tsOutputChannels DECLARE_
 	tsOutputChannels->hasSdCard = isSdCardAlive();
 #endif /* EFI_FILE_LOGGING */
 
-	tsOutputChannels->isFuelPumpOn = enginePins.fuelPumpRelay.getLogicValue();
 	tsOutputChannels->isFanOn = enginePins.fanRelay.getLogicValue();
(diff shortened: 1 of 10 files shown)


rusefillc:
hold on, why not build-in Lua?

mck1117:
There is no convenient way to _do_ built-in Lua other than spooling up a complete separate VM.

mck1117:
I just moved AC out of fsio too

mck1117:
poke

rusefillc:
Today was the first day of me looking into rusEFI Lua

IMHO we shall really Lua system logic not C++ it

tactically I am thinking to drop half of FSIO formulas to get us that 1K of Lua script config

to be continues as I get more familiar with Lua

rusefillc:
we've got second VM since, closing this PR since looks like that's NOT the direction we are going in?

mck1117:
I guess - we have to decide what to do about MCUs that don't have much RAM.  Maybe we don't care about those MCUs, maybe we need non-lua implementation for those MCUs.

rusefillc:
legit question not sure what's the answer

many units are at least 427

for true 407 units we can offer custom rusEFI "discovery" or PCB replacement program

mck1117:
I think we can make it work on the 407.  It'll take some massaging, but it's possible.  Perhaps that means no user lua on 407.  My question was more about how much we care about kinetis/cypress/etc.

rusefillc:
I like the "no user lua" idea for 407

let's think about kinetis/cypress/etc later

@andreika-git

PR #2940, opened 2021-07-09
ConfigDefinition stuff

chuckwagoncomputing:
1. Check for empty ID string
2. I was working on checking for duplicates, but it turns out hellen121vag has multiple pins routed to the same MCU pin. I commented out the dupe checking, but left the refactoring I did to get to it. This consists of building a list for each pin before adding to the main pin list. I don't know that this has any value except when we are checking for duplicates. It's an open question whether this refactoring should stay or be rolled back.


rusefillc:
is it time to enter the magic of test driven world?

chuckwagoncomputing:
How does TDD help with duplicates?
Or do you mean have tests with inputs that are bad in various ways?

rusefillc:
Yes, i wonder if it's time for all this to have an explicit test showing good VS bad inputs. Maybe using hard-coded yaml files and bytebuffersteam / reader not recourse files.

PR #2955, opened 2021-07-13
vvt input but no mode should be a fatal

mck1117:
seen in @Abricosvw's log: https://rusefi.com/online/view.php?log=286


mck1117:
wait, on second thought, do we want this to be a warning at all? This is how you inspect VVT pattern without decoding it, which should 100% be a "normal" scenario.

rusefillc:
no mode with input is not noty mal

for engineering use case warning is appropriate imho

mck1117:
I guess.  Warnings are just hard to discover (if it's actually a problem, fatal!) for normal users, so you end up driving around with one warning per VVT tooth.

PR #2971, opened 2021-07-17
set up precompiled header

mck1117:


mck1117:
It looks like this knocks 30-40% of the build time off on linux 😳

rusefillc:
Does this change force any new conventions on us? if it does we need an ELI5 somewhere in the codebase to explain what is legal and what is now not legal

like the header changes within this PR what is driving those?

totally not insisting by maybe a couple of smaller PRs with more code comments explaining parts of changes?

mck1117:
> Does this change force any new conventions on us? if it does we need an ELI5 somewhere in the codebase to explain what is legal and what is now not legal

I don't think so, no.  The only requirement is that `#include "pch.h"` appears before any other tokens in the file if you want to use the precompiled header.  Other than that, it should behave exactly the same as a "normal" include of `pch.h`.

> like the header changes within this PR what is driving those?

those are things that worked before by luck, but wouldn't if you included a different order of things.  That was a poor example of file to include pch in, so now just `engine.cpp` includes `pch.h`.

> totally not insisting by maybe a couple of smaller PRs with more code comments explaining parts of changes?

I peeled off some of the maybe unnecessary/uninteresting bits in to #3062.  This PR is now just:

- Add `pch.h`
- Add compilation for `pch.h`
- include `pch.h` in `engine.cpp` as an example of use.

Issue #2284, opened 2021-02-06
H7 compilation


rusefillc:
We need H7 in CI

rusefillc:
current status:
SPI disabled - need more code in device_mpu_util.h

CAN disabled - differences in driver?

Flash persistence disabled

USB disabled - open question what's the state of USB in H7, ChibiOS code looked suspecious

mck1117:
H7 builds and is in CI as of #2317

ADC: disabled - different driver
CAN: disabled - different driver
USB: disabled - different driver?

mck1117:
ADC will be helped by #2325

mck1117:
Update:
ADC: disabled - different driver
CAN: minor changes needed, but it doesn't work for some reason, see #2393
Flash: disabled - different driver (but appears to only be slightly different
USB: works!

rusefillc:
DFU jump https://github.com/rusefi/rusefi/issues/2961

rusefillc:
@mck1117 do you have SD card working via SPI on H7? Or for that matter anything on SPI on H7?

mck1117:
SD works with a hacked driver on H7.

https://github.com/mck1117/atlas is H7 but uses the SDMMC peripheral instead of SPI since it's lower overhead and doesn't have the silicon bug present in the SPI controller.

andreika-git:
@mck1117 What's the current status/plans of "#define EFI_USE_FAST_ADC FALSE" for H7?

rusefillc:
@andreika-git is that a rhetorical question? 

looks like https://github.com/rusefi/rusefi/blob/master/firmware/config/stm32h7ems/efifeatures.h#L9

PR #2991, opened 2021-07-19
sample with dual simultaneous adc on STM32H7

mck1117:
Speeds up ADC sampling by using both ADCs in parallel.

Probably needed to be able to do "fast adc" MAP sampling on H7.

#2284


rusefillc:
@andreika-git

mck1117:
I don't think the chibios driver actually properly supports this - closing for now

PR #2997, opened 2021-07-20
fix led init order

### LINKED ISSUE #2960: no blue LED blinking in case of fatal
have working rusEFU

make configuration invalid by assigning same pin to camInputDebug 1 and, or probably any pin conflict

expected behavour: solid RED and blinking BLUE

actual result: solid RED and dim BLUE

mck1117:
fix #2960


rusefillc:
since i have a lot of fatal here with dbank. no LED blinking :(

Issue #2961, opened 2021-07-14
stm32h7 h7 software DFU jump


rusefillc:
https://community.st.com/s/question/0D53W00000vQEWsSAO/stm32h743-dfu-entry-doesnt-work-unless-boot0-held-high-at-poweron

PR #2999, opened 2021-07-20
H7 turn off USB before reset

mck1117:
#2961
#2284


rusefillc:
OMG we are actually working around this?!

Damn this humanity.

PR #3018, opened 2021-07-21
simplify quad vvt test

mck1117:
Use single tooth VVT instead of real patterns, as we don't need to re-test whether we can decode a vvt pattern.  Test. One. Thing. At. A. Time.  In this test, we *only* want to test that we can independently resolve the position of 4 different cams, so a single tooth trigger is more than plenty.


rusefillc:
this PR claims to be simplifying a TEST

I see this PR changing PROD code

I am very confused

mck1117:
If you `setOperationMode(engineConfiguration, FOUR_STROKE_CRANK_SENSOR);`, then we don't have the problem, so it doesn't need the trigger decoder change.

PR #3020, opened 2021-07-21
Fix ini name (hellen128mercedes drama)

chuckwagoncomputing:
In addition to the issue that turned up in the last PR, I also found, by actually testing in my master, that the .ini didn't exist in the git tree and the gen-config workflow wasn't adding it. I added a `git add` statement to that workflow to take care of that and future changes.


rusefillc:
Should we remove old file name from git? When the change goes as is would we end up with both old and new file name?

chuckwagoncomputing:
Oh hang on. I think it needs to be poked.

Issue #636, opened 2018-12-25
10KHz ADC generates 20KHz IRQs


rusefi:
At the moment each ADC measurement via PWM timer costs us two IRQs

See ```/* Note, only in the ADC_COMPLETE state because the ADC driver fires
	  an intermediate callback when the buffer is half full. /
	if (adcp->state == ADC_COMPLETE) {```

See STM32_DMA_CR_HTIE
See ADCDriver->circular

Would it be possible to eliminate the half-time callback to get free up some resources with zero consequences to the firmware?

fpoussin:
see https://github.com/ChibiOS/ChibiOS/blob/master/os/hal/ports/STM32/LLD/ADCv2/hal_adc_lld.c#L345

PR #3028, opened 2021-07-22
H7 ADC triggered by hardware timer

mck1117:
- Reduces CPU overhead by skipping adc reconfigure every time, and triggering off a hardware timer.
- ADC now samples at 10khz * 4x oversample * 16 channels = 640k total samples per second
- Slow analog sensors update at 1khz to make up for reduced hardware oversample

also deals with #636, but only on H7


mck1117:
tested on real car, works fine!

PR #3031, opened 2021-07-22
fast adc callback

mck1117:
prep for making map sampling work on H7


rusefillc:
@andreika-git does this overlap with your uncommited changes?

andreika-git:
> 
> 
> @andreika-git does this overlap with your uncommited changes?

Yes, but I did it in more quick-and-dirty way: I switched all 16 channels from slow to fast, and tweaked the sampling rate... Of course we need more proper way considering lazy Chibios drivers. I don't know the Matt's plan though...

rusefillc:
let's have this build green not red? failing to compile cypress.

PR #3032, opened 2021-07-22
use EXTI instead of ICU by default

### LINKED ISSUE #639: ICU is not really needed, EXTI should be enough for digital input interrupt

mck1117:
fix #639


rusefillc:
what's the problem statement?

with no arguments for a change i see it only as risk and mcu temp or proteus HW CI looks more important?

mck1117:
#636 is currently blocked on F4/F7 because the STM32 has a limited set of internal wires that go between timers and the ADC.

image

Timers 1, 2, 3 are used by ICU for trigger input.  Timer 4 is used for ETB PWM on Proteus, and Hellen. Timer 5 is used by the event queue (single timer executor).  Currently it isn't possible to do timer-triggered-ADC without turning off ICU. Timer 8 is used by ETB PWM on microRusEFI.

It also blocks implementing tickless mode for the same reason: https://github.com/rusefi/rusefi/issues/639#issuecomment-540712992

Not using ICU means we can use timers for tickless mode, ADC trigger, and more PWM outputs.

rusefillc:
is this abandoned?

rusefillc:
I see
1) 46 commits
2) multiple atomic changes
3) and red result in the end

in light of that shall we attempt a smaller PR here?

mck1117:
oh my goodness it's green

PR #3035, opened 2021-07-23
proteus hw ci: wait longer but be stricter

mck1117:
added an analog filter in hardware (bodge wires), we should wait longer, but it'll be more accurate now


rusefillc:
document analog filter details right in the code?

PR #3038, opened 2021-07-23
perform extra-step-prevention in the stepper driver itself

### LINKED ISSUE #3030: "prettyClose": some strange code inside idle controller?
1) We have IdleController::update() where we call:
```
float position = getIdlePosition();
applyIACposition(position PASS_ENGINE_PARAMETER_SUFFIX);
```

2) Inside IdleController::getIdlePosition() we set:
```
engine->engineState.idle.currentIdlePosition = iacPosition;
return iacPosition;

```

3) Inside applyIACposition() we check:
```
bool prettyClose = absF(position - engine->engineState.idle.currentIdlePosition) < idlePositionSensitivityThreshold;
if (prettyClose) {
	return;
}
```

Does it really work? 'prettyClose' equals true all the time, right? I cannot get past it on the bench.

mck1117:
`idle_hardware.cpp` shouldn't care about stepper-specific problems, so do that in the stepper driver itself.  Don't update setpoint unless we're at least one full step away from the last setpoint.

fix #3030


firmware/hw_layer/stepper.cpp
```diff
@@ -152,15 +155,19 @@ int StepperMotor::getTargetPosition() const {
...
 }
 
-void StepperMotor::setTargetPosition(int targetPosition) {
+void StepperMotor::setTargetPosition(float targetPosition) {
 	// we accept a new target position only if the motor is powered from the main relay
 	if (engine->isMainRelayEnabled()) {
-		m_targetPosition = targetPosition;
+		// When the IAC position value change is insignificant (lower than this threshold), leave the poor valve alone
+		// When we get a larger change, actually update the target stepper position
+		if (absF(m_targetPosition - targetPosition) > 1) {
```
andreika-git, line 163: Maybe >= 1 ?

firmware/hw_layer/stepper.cpp
```diff
@@ -152,15 +155,19 @@ int StepperMotor::getTargetPosition() const {
 	return m_targetPosition;
 }
 
-void StepperMotor::setTargetPosition(int targetPosition) {
+void StepperMotor::setTargetPosition(float targetPosition) {
```
rusefillc, line 158: Would be be better to use "percent_t" here?

firmware/hw_layer/stepper.cpp
mck1117: For a float, equality (even when you're trying) is practically impossible, so it doesn't really matter.

rusefillc: yes, but considering how code is documentation this is important to articulate the intentions more cleanly

mck1117: I thought that two line comment was the documentation, but ¯\_(ツ)_/¯

firmware/hw_layer/stepper.cpp
mck1117: name is clarified

andreika-git (approved):
Nice!

Issue #2944, opened 2021-07-10
icon paths are broken when invoking via rusefi_autoupdate


rusefillc:
S2R:
download artifacts from https://rusefi.com/build_server/rusefi_bundle.zip or build locally

expected behavior:

``java -jar rusefi_console.jar ``
![image](https://user-images.githubusercontent.com/48498823/125169778-b4783000-e179-11eb-9ebf-33483494980b.png)

broken behavior, should be same as above
``java -jar rusefi_autoupdate.jar``
![image](https://user-images.githubusercontent.com/48498823/125169801-d4a7ef00-e179-11eb-8e87-c6112fb750ba.png)

Come1LLF00:
Okay, I came up with the idea to add every URL of loaded jars including rusefi_console.jar to the system classloader when getClassLoaderByJar is invoked in AutoupdateUtil. The current solution looks like this:
`private static void addURLToSystemClassLoader( URL url ) throws IntrospectionException {
    URLClassLoader systemClassLoader = (URLClassLoader) ClassLoader.getSystemClassLoader();
    Class<URLClassLoader> classLoaderClass = URLClassLoader.class;

    try {
        Method method = classLoaderClass.getDeclaredMethod("addURL", URL.class );
        method.setAccessible(true);
        method.invoke(systemClassLoader, url );
    } catch ( Throwable t ) {
        t.printStackTrace();
        throw new IntrospectionException("Error when adding url to system ClassLoader ");
    }
}

@NotNull
public static URLClassLoader getClassLoaderByJar(String jar) throws MalformedURLException, IntrospectionException {
    final URL newURL = new File(jar).toURI().toURL();
    addURLToSystemClassLoader( newURL );
    return new URLClassLoader(
            new URL[]{ newURL },
            AutoupdateUtil.class.getClassLoader()
    );
}`

rusefillc:
@Come1LLF00 yes this looks fixed now!

PR #3043, opened 2021-07-23
hotfix 2944: broken icon paths in rusefi_autoupdate

Come1LLF00:
The problem was that none classloader of rusefi_autoupdate contains the path to `rusefi_console.jar` and URLClassLoader which is used to load `rusefi_console.jar` didn't embed in the classloaders hierarchy. Nevertheless, to load icons you use `getResource()` from `Class` this method just resolves resource name according to class package and consequently use its ClassLoader `getResource()` method. Eventually, the path to the `rusefi_console.jar` keeps nowhere after invoking the `Launcher.main` method. Thus, to solve this problem we should someway insert the path for `rusefi_console.jar`.

Solutions:
1. Insert jar's URL to system classloader using Reflection API
2. Embed classloader with jar's URL to the classloaders hierarchy. __Maybe impossible__
3. Use only our custom classloader with jar's URL to load icons. __Used in branch__

'Cause the `AutoupdateUtil.class.getResource()` was removing the leading slashes and I had no idea where these paths with slashes come from, I had to emulate this logic in my classloader.

### DIFF
--- a/java_console/.idea/ant.xml
+++ b/java_console/.idea/ant.xml
@@ -2,8 +2,6 @@
 <project version="4">
   <component name="AntConfiguration">
     <buildFile url="file://$PROJECT_DIR$/build.xml">
-      <maximumHeapSize value="1024" />
-      <maximumStackSize value="256" />
       <executeOn event="beforeCompilation" target="before_IDEA_Build" />
     </buildFile>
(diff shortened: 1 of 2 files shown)


rusefillc:
#2944

Issue #3050, opened 2021-07-24
Red test mule cranking to idling is weird


rusefillc:
See https://rusefi.com/forum/viewtopic.php?p=41684#p41684

rusefillc:
two changes:
improved TPS calibration
wider threshold (but with REO creating it now)

now idle does not fall that much down but still oscillation

https://rusefi.com/online/view.php?log=301

rusefillc:
https://github.com/rusefi/web_backend/issues/127 reported

mck1117:
related to #2813

rusefillc:
looks great! https://rusefi.com/online/view.php?log=303

PR #3051, opened 2021-07-24
adjust taper behavior

mck1117:
Put it back to the old behavior where the crank to run taper is applied even when not in closed loop mode.

#3050 

Sort of broken by #2813


rusefillc:
We need a unit test.

PR #3060, opened 2021-07-25
fix nonlinear correction threshold

mck1117:


firmware/controllers/algo/fuel/injector_model.cpp
```diff
@@ -128,7 +128,7 @@ float InjectorModel::correctShortPulse(float baseDuration) const {
 }
 
 float InjectorModel::correctInjectionPolynomial(float baseDuration) const {
-	if (baseDuration > CONFIG(applyNonlinearBelowPulse)) {
+	if (baseDuration > (CONFIG(applyNonlinearBelowPulse) * 0.001f)) {
```
rusefillc, line 131: magic constant code smell

is this seconds to MS conversion? let's find a way to make this more readable

mck1117: other way around, but yes

PR #3073, opened 2021-07-29
flash erase cache invalidate

mck1117:


mck1117:
image

mck1117:
the problem was:

1. read or write data to/from flash, fills cache
2. do erase
3. checks data after erase, gets cache hit from step 1 (but only if very lucky!)

this pr adds a step between steps 2 and 3 to invalidate the data cache lines that could have stale flash data in them.

dron0gus (approved):

Issue #936, opened 2019-09-15
misfire detection


rusefi:
Can we please detect broken injector or broken coil programmatically?

chuckwagoncomputing:
#925 #343 #2180

rusefillc:
See https://rusefi.com/forum/viewtopic.php?f=5&t=2071

chuckwagoncomputing:
FWIW, Ford's IDS software has an awesome cylinder contribution graph that has also been copied by other scan tool software. It's great that the ECU can say "cylinder n is misfiring", but being able to visualize the data is _awesome_.

rusefillc:
@chuckwagoncomputing any screenshots? :)

chuckwagoncomputing:
Even better, a video:
https://www.youtube.com/watch?v=G6WYnNlYuiQ
The video is a diesel, but the same thing exists for gas vehicles.

stefanst:
Trying to start doing this with a low-resolution Miata crank trigger may be a tall initial hurdle to jump. Wouldn't this be easier using a 60-2 or 36-2 or similar trigger?

thirstyone:
even with this config you can clearly see the difference in timing (I just looked through the second file in saleae - without injector, I believe, without a spark it would be even more obvious)

chuckwagoncomputing:
Doesn't the trigger decoder already calculate where it "expects" the tooth to be, versus where it is? A visualization of this data would be better than nothing. Perhaps as an overlay of the trigger diagram in Console?

rusefillc:
@chuckwagoncomputing we have a _proposal_ #2222 but only a proposal.

Our reference diagram at https://github.com/rusefi/rusefi/wiki/All-Supported-Triggers#mazda-miata-nb

rusefillc:
We have data and proposal at https://rusefi.com/forum/viewtopic.php?p=42922#p42922

rusefillc:
https://github.com/FOME-Tech/fome-fw/issues/575
https://github.com/FOME-Tech/fome-fw/commit/47f10725a39355ac9265aac01f41d4eab6b6a72d
https://github.com/FOME-Tech/fome-fw/commit/b6bf23e4ffefa3a4c9c7c9f7dee9419ae4dc66b0

PR #3077, opened 2021-07-30
instant rpm uses full engine cycle

### LINKED ISSUE #3015: sync nb2 vvt faster
In one log, we sync the crank after only 4 teeth, but VVT takes an additional 1.63 seconds to sync.  The culprit is probably:

```
	if (cisnan(oneDegreeUs)) {
		// todo: this code branch is slowing NB2 cranking since we require RPM sync for VVT sync!
		// todo: smarter code
		//
		// we are here if we are getting VVT position signals while engine is not running
		// for example if crank position sensor is broken :)
		return;
	}
```

### LINKED ISSUE #3070: Faster enginePins.stopPins()
at the moment we can half coils/injectors stuck open for up to half a second on loss of trigger input. in reality that usually comes with relays going off but still maybe we shall detect this condition faster for any trigger or at least consider trigger tooth count in the delay calculation

```
	if (secondsSinceTriggerEvent < 0.5f) {
		// Engine moved recently, no need to safe pins.
		return;
	}

```

mck1117:
fix #3070 

The problem was that instant RPM was trying to use `currentCycle.current_index`, which doesn't account for trigger patterns that only cover part of the engine cycle (for example, a crank only trigger).

This looks much better now!

image

As a bonus, this probably helps #3015, since now symmetrical crank instant RPM /actually/ works, instead of only calculating on one tooth per cycle, which effectively broke instant RPM on symmetrical crank triggers.

### DIFF
--- a/firmware/controllers/engine_cycle/rpm_calculator.cpp
+++ b/firmware/controllers/engine_cycle/rpm_calculator.cpp
@@ -291,7 +291,7 @@ void rpmShaftPositionCallback(trigger_event_e ckpSignalType,
 #endif /* EFI_SENSOR_CHART */
 
 	// Always update instant RPM even when not spinning up
-	engine->triggerCentral.triggerState.updateInstantRpm(&engine->triggerCentral.triggerFormDetails, nowNt PASS_ENGINE_PARAMETER_SUFFIX);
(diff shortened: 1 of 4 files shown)


mck1117:
Also helps with #936, as now we actually calculate instant RPM over the whole cycle!

Issue #3091, opened 2021-08-01
unit test for flex_sensor.cpp


rusefillc:
Similar to #3081 but could be a different :)

one day flex_sensor and VSS implementations should get much similar or maybe even same code

OrchardPerformance:
Let's get this one over the line as we want flex fuel to be part of RE1.0.
@BeerMoneyMotorsports and @CrazyStriker are able to test the sensor once the firmware is ready.

BeerMoneyMotorsports:
I WILL have data for flex this week.

BeerMoneyMotorsports:
What is the scope of the firmware? Fuel multiplication for ethanol content along with timing?

PR #3093, opened 2021-08-02
Unit test for flex sensor.cpp #3091

alxrMironov:
Unit test for flex sensor.cpp by #3091 ticket

### DIFF
--- a/unit_tests/tests/sensor/test_flex_sensor.cpp
+++ b/unit_tests/tests/sensor/test_flex_sensor.cpp
@@ -9,12 +9,6 @@ class FlexSensorForTests : public ::testing::Test
 
 protected:
 
-    FlexSensorForTests () :
-            dut ()
-    {
-    }
-    ;
-
     void
     SetUp () override
     {
--- a/unit_tests/tests/sensor/test_flex_sensor.cpp
+++ b/unit_tests/tests/sensor/test_flex_sensor.cpp
@@ -123,7 +123,7 @@ TEST_F(FlexSensorForTests, testInvalidFrequency)
  *  Sensor must take PWM input on frequency with "known output value"
  *  and generate similiar output.
  */
-TEST_F(FlexSensorForTests, testInvalidFrequency)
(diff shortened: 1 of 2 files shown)


alxrMironov:
Test

unit_tests/tests/sensor/test_flex_sensor.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+#include <gtest/gtest.h>
+
+class FlexSensorForTests : public ::testing::Test
+{
+
+protected:
+
+    FlexSensorForTests () :
+            dut ()
```
mck1117, line 13: explicit no-arg constructor is not required

mck1117: I guess no constructor is required at all, the default is plenty here.

unit_tests/tests/sensor/test_flex_sensor.cpp
```diff
@@ -0,0 +1,148 @@
+#include "flex_sensor.h"
+#include "engine_test_helper.h" // Used for timesourcing
```
mck1117, line 2: you can replace all these includes with `#include "pch.h"`, except for flex_sensor.h

unit_tests/tests/sensor/test_flex_sensor.cpp
```diff
@@ -0,0 +1,148 @@
...
+protected:
+
+    FlexSensorForTests () :
+            dut ()
+    {
+    }
+    ;
+
+    void
+    SetUp () override
```
mck1117, line 19: should be `void SetUp() override {`, similar with others

unit_tests/tests/sensor/test_flex_sensor.cpp
alxrMironov: Noted

unit_tests/tests/sensor/test_flex_sensor.cpp
alxrMironov: Noted

PR #3095, opened 2021-08-02
don't "validate instant RPM"

mck1117:
Not much reason to do this - we don't make this check at all for non-instant RPM, so there isn't a reason to do it for instant RPM either.  It could actually hinder starting for an engine that starts quickly - it will give a wrong RPM reading if the engine kicks before a full cycle has completed, increasing engine speed past the cranking threshold before a full cycle (and "normal" RPM reading) is complete.  The wrong RPM will cause events to be scheduled late, possibly hindering engine start, as it means ignition timing will be (possibly significantly) retarded from the set value.

### DIFF
--- a/firmware/controllers/engine_cycle/rpm_calculator.cpp
+++ b/firmware/controllers/engine_cycle/rpm_calculator.cpp
@@ -296,8 +296,6 @@ void rpmShaftPositionCallback(trigger_event_e ckpSignalType,
 	if (rpmState->isSpinningUp()) {
 		float instantRpm = engine->triggerCentral.triggerState.getInstantRpm();
 
-		// validate instant RPM - we shouldn't skip the cranking state
-		instantRpm = minF(instantRpm, CONFIG(cranking.rpm) - 1);
 		rpmState->assignRpmValue(instantRpm);
 #if 0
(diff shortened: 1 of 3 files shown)


rusefillc:
@andreika-git do you know anything about this check?

rusefillc:
@mck1117 how did you end up there - is this related to any other work?

mck1117:
> @mck1117 how did you end up there

I tried changing things to *always* use instant RPM instead of ever using "normal" RPM, and ran in to this.

rusefillc:
gotcha. let's give @andreika-git 48 hours?

mck1117:
works for me!

andreika-git:
> 
> 
> @andreika-git do you know anything about this check?

Yes, I remember it. The early instant RPM implementation was far less reliable than "normal RPM":
image
Here's a spike of 7279 RPM right at the start (2018-02-07, winter cold start at -12 degrees C). It was unpredictable without that check. 
The question is whether we've fixed it or not.

rusefillc:
We're recently added second trigger gap into many triggers to better handle start and stop noises

rusefillc:
This winter we need cold start experiments :)

mck1117:
> This winter we need cold start experiments :)

you can recreate cold by putting in SAE 60 oil and letting your battery run down a bit 🙂

Issue #785, opened 2019-05-04
Missing ADC Pins for ADC3, PF port


960:
Pins for 144 pin mcu, ADC3 missing in both code and TS

PF3 to PF10

rusefi:
multiple issues here actually
1) ADC channels are hard-coded
```
// is there a reason to have this configurable?
#ifndef ADC_SLOW_DEVICE
#define ADC_SLOW_DEVICE ADCD1
#endif /* ADC_SLOW_DEVICE */

// is there a reason to have this configurable?
#ifndef ADC_FAST_DEVICE
#define ADC_FAST_DEVICE ADCD2
#endif /* ADC_FAST_DEVICE */
```
2) even for hard-coded ADC channels we sometimes have two options for ADC input pin but rusEfi has only one option hard coded

First step of the change would probably be to switch from using "input channels" in configuration to using "brain pins" for input configuration

rusefi:
need custom compile with
-DSTM32_ADC_USE_ADC3=TRUE -DADC_SLOW_DEVICE=ADCD3 -DEFI_PIN_ADC9=GPIOF_3

maybe this would make channel9 PF3 work for CLT and IAT? if that would work we can go from there

See https://www.st.com/resource/en/datasheet/stm32f769ni.pdf

mck1117:
duplicate of #2238

rusefillc:
See also https://github.com/rusefi/rusefi/issues/3350

PR #3101, opened 2021-08-02
dead weird define

mck1117:
Not defined anywhere else.


rusefillc:
@andreika-git is that a prototype for ADC h7 something? do we need?

mck1117:
Iirc this was a thing for ola

rusefillc:
#785

Issue #3081, opened 2021-07-30
unit test for vehicle_speed.cpp


rusefillc:
looks like we have no coverage at all

https://github.com/rusefi/rusefi/blob/master/firmware/hw_layer/vehicle_speed.cpp
https://rusefi.com/docs/unit_tests_coverage/firmware/hw_layer/vehicle_speed.cpp.gcov.html

step 1: enable EFI_VEHICLE_SPEED in https://github.com/rusefi/rusefi/blob/master/unit_tests/efifeatures.h and get unit tests to even compile with EFI_VEHICLE_SPEED

step 2: fire 6 events into vehicle_speed.cpp and assert resulting vehicle speed sensor

PR #3103, opened 2021-08-03
Unit tests for vehicle speed.cpp #3081

alxrMironov:
I has found, what wrapping 
`
	CONFIG(vehicleSpeedSensorInputPin) = GPIOA_0;
	brain_pin_markUsed(GPIOA_0, "VSS" PASS_ENGINE_PARAMETER_SUFFIX);
	engineConfiguration->vehicleSpeedCoef = speedCoef;
`
into function causes big mess and tons of parameters, so i just copy-paste it.

Idk why `brain_pin_markUsed()` dont works, and vehicleSpeed still has `NOT_ASSIGNED`, so i used direct 'CONFIG(vehicleSpeedSensorInputPin) = GPIOA_0;' too.

### DIFF
--- a/unit_tests/efifeatures.h
+++ b/unit_tests/efifeatures.h
@@ -76,5 +76,3 @@
 #define EFI_FUEL_PUMP TRUE
 
 #define EFI_LUA TRUE
-
-#define EFI_VEHICLE_SPEED TRUE
--- a/unit_tests/efifeatures.h
+++ b/unit_tests/efifeatures.h
@@ -76,3 +76,5 @@
 #define EFI_FUEL_PUMP TRUE
 
 #define EFI_LUA TRUE
+
+#define EFI_VEHICLE_SPEED TRUE
--- a/unit_tests/tests/test_vehicle_speed.cpp
+++ b/unit_tests/tests/test_vehicle_speed.cpp
@@ -42,7 +42,6 @@ TEST(VehicleSpeedSensor, testValidSpeedDetection) {
 
 	// Init global variables
 	CONFIG(vehicleSpeedSensorInputPin) = anyPin;
-	brain_pin_markUsed(anyPin, vehicleSpeedSensorMessage PASS_ENGINE_PARAMETER_SUFFIX);
 	engineConfiguration->vehicleSpeedCoef = 0.5f;
 
 	// Valid speed 15kmh should be returned
@@ -58,12 +57,13 @@ TEST(VehicleSpeedSensor, testInvalidSpeed) {
 
 	// Init global variables
 	CONFIG(vehicleSpeedSensorInputPin) = anyPin;
-	brain_pin_markUsed(anyPin, vehicleSpeedSensorMessage PASS_ENGINE_PARAMETER_SUFFIX);
 	engineConfiguration->vehicleSpeedCoef = 0.5f;
 
 	// Invalid (slow) interval, should return 0 speed
 	simulatePeriodicSignalForCallback(eth, 0.5f, vsCallback PASS_ENGINE_PARAMETER_SUFFIX);
 	float measuredSpeed = getVehicleSpeed(PASS_ENGINE_PARAMETER_SIGNATURE);
+
+	// Direct comparasion as invalid speed shoud return true zero
 	EXPECT_EQ(0.0f, measuredSpeed);
 }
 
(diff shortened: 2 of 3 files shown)


unit_tests/efifeatures.h
```diff
@@ -76,3 +76,5 @@
 #define EFI_FUEL_PUMP TRUE
 
 #define EFI_LUA TRUE
+
+#define EFI_VEHICLE_SPEED TRUE
```
rusefillc, line 80: we already have it in https://github.com/rusefi/rusefi/blob/master/unit_tests/efifeatures.h#L16

unit_tests/tests/test_vehicle_speed.cpp
```diff
@@ -0,0 +1,74 @@
...
+	constexpr float speedCoef = 0.5f;
+
+	// Init global variables
+	CONFIG(vehicleSpeedSensorInputPin) = anyPin;
+	brain_pin_markUsed(anyPin, vehicleSpeedSensorMessage PASS_ENGINE_PARAMETER_SUFFIX);
+	engineConfiguration->vehicleSpeedCoef = speedCoef;
+
+	simulatePeriodicSignalForCallback(eth, freq, vsCallback PASS_ENGINE_PARAMETER_SUFFIX);
+	float measuredSpeed = getVehicleSpeed(PASS_ENGINE_PARAMETER_SIGNATURE);
+	EXPECT_EQ(returnedOnInvalidFreq, measuredSpeed);
```
rusefillc, line 72: we cannot use EQ for "float" type, float type is rarely equal. We use ASSERT_NEAR see other usages of ASSERT_NEAR

that's due to nature of float and double

unit_tests/tests/test_vehicle_speed.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+TEST(VehicleSpeedSensor, testValidSpeedDetection) {
+	WITH_ENGINE_TEST_HELPER(TEST_ENGINE);
+
+	constexpr float expectedSpeed = 15.0f;
+	constexpr float speedCoef = 0.5f;
+
+	// Init global variables
+	CONFIG(vehicleSpeedSensorInputPin) = anyPin;
+	brain_pin_markUsed(anyPin, vehicleSpeedSensorMessage PASS_ENGINE_PARAMETER_SUFFIX);
```
mck1117, line 48: you probably don't need `brain_pin_markUsed`

unit_tests/tests/test_vehicle_speed.cpp
```diff
@@ -0,0 +1,74 @@
...
+	EXPECT_NEAR(expectedSpeed, measuredSpeed, 0.01);
+
+}
+
+TEST(VehicleSpeedSensor, testInvalidSpeed) {
+	WITH_ENGINE_TEST_HELPER(TEST_ENGINE);
+
+	constexpr float freq = 0.5f; // Invalid period > 1sec
+	constexpr float returnedOnInvalidFreq = 0.0f;
+	constexpr float speedCoef = 0.5f;
```
mck1117, line 63: why so many extra constant variables that are used only once?

unit_tests/tests/test_vehicle_speed.cpp
mck1117: you can however use `EXPECT_FLOAT_EQ`: it checks that the floats are equal within +-4 LSB

see https://github.com/google/googletest/blob/master/docs/reference/assertions.md#floating-point-comparison-floating-point

unit_tests/tests/test_vehicle_speed.cpp
mck1117: These make is much harder to read.  For example, the last two lines of this test would be much clearer if written as

```
// Invalid (slow) interval, should return 0 speed
EXPECT_EQ(0, getVehicleSpeed(PASS_ENGINE_PARAMETER_SIGNATURE));
```

unit_tests/tests/test_vehicle_speed.cpp
```diff
@@ -0,0 +1,74 @@
...
+ * 	to simulate periodic signal on input pin with passed callback.
+ */
+static void simulatePeriodicSignalForCallback(
+	EngineTestHelper& eth,
+	float freqHz,
+	vss_callback_fp cb
+	DECLARE_ENGINE_PARAMETER_SUFFIX)
+{
+	constexpr auto periods = 50;
+	auto period = (1 / freqHz);
```
mck1117, line 29: Good use of `auto`!

PR #3124, opened 2021-08-09
Feature 3014-1 Migrating gcc_map_reader to gradle

Come1LLF00:
I completed `build.gradle` for `gcc_map_reader`. It works for gradle 6.1.1 from android and generates output like ant except for minor exceptions like:
+ 2 lines about versions of java and Ant in jar's manifest
+ creating extra folders `generated` and `tmp` ( 'cause it gradle )

Also, I "fixed" the run task in `build.xml` of `gcc_map_reader`.


rusefillc:
1) if we have gradle config to build a jar let's already remove ant script which does the same

2) looks like this currently depends on gradle in classpath? that's not ideal since we do not have control of what version would be installed into path etc. my preference would be for this new gradle script to be linked into existing android script so that I could execute via gradle wrapper which we have in android already. we will later refactor this area maybe but for now we get all java code a bit more connected by cross connecting all gradle scripts, nothing bad about it in my opinion

java_tools/gcc_map_reader/.gitignore
```diff
@@ -0,0 +1 @@
+.gradle
```
rusefillc, line 1: Is this needed now with android reference and invocation for m android folder?

java_tools/gcc_map_reader/build.gradle
```diff
@@ -0,0 +1,29 @@
...
+
+defaultTasks 'jar'
+
+dependencies {
+    compile fileTree( dir: 'lib', includes: [ 'junit.jar', 'annotations.jar' ] )
+}
+
+sourceSets {
+    main {
+        java.outputDir = file( 'build/classes' )
```
rusefillc, line 14: Why do we need non standard output folder? Please add comment into Gradle file if legit reason to make it obvious

java_tools/gcc_map_reader/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+defaultTasks 'jar'
+
+dependencies {
+    compileClasspath fileTree( dir: 'lib', includes: [ 'junit.jar', 'annotations.jar' ] )
+}
+
+/*
+  The deprecated now building by Ant script put the
+  generated classes into folder 'build/classes/<package-name>'.
+  But Gradle by default put the generated classes to
```
rusefillc, line 15: Right, but who cases about those classes location, why do we need legacy class location why not change?

PR #3125, opened 2021-08-09
Hotfix startScripts failure

Come1LLF00:
Just added mainClassName property to the whole build.gradle script


java_tools/gcc_map_reader/build.gradle
```diff
@@ -3,6 +3,8 @@ plugins {
     id 'application'
 }
 
+mainClassName = 'com.rusefi.Launcher'
```
rusefillc, line 6: I do not get this, how would this change what?

Come1LLF00: Well, it makes startScripts task work and doesn't break the run task which originally launches the rusefi_console.jar

rusefillc:
does not look like we need this

PR #3131, opened 2021-08-10
everybody is pulldown

### LINKED ISSUE #3113: #define EFI_DR_DEFAULT inconsistency between F4 and H7
F4 and probably F7 inherits 
#define EFI_DR_DEFAULT PIN_PUPDR_PULLUP

H7 says
#define EFI_DR_DEFAULT PIN_PUPDR_FLOATING

shall this be more unified or documented better?

See also #2525

mck1117:
fix #3113


rusefillc:
looks like this change does two things - H7 is now in sync is F4/F7 but also change F4/F7 default?!

@andreika-git  do you have an opinion on all this?

andreika-git:
pulldown is good.

Issue #3108, opened 2021-08-04
Switch vehicle speed consumers to read from sensor model


mck1117:
depends on #3107 

Switch callers from `getVehicleSpeed()` to `Sensor::get(SensorType::VehicleSpeed)`

Issue #3107, opened 2021-08-04
Add vehicle speed to sensor model


mck1117:
- Add to SensorType enum
- Add string in sensor.cpp
- create init_vehicle_speed.cpp, see https://github.com/rusefi/rusefi/pull/1736/files for a similar change

Issue #3106, opened 2021-08-04
Refactor vss towards flex fuel sensor


rusefillc:
Vss sensor should get much closer to flex fuel sensor 

A part of the change in s migration to Sensor framework see flex as example

mck1117:
As an intermediate change we can use functionPointerSensor to get it in the sensor model without full implementation, see init_lambda.cpp

mck1117:
depends on #3107

mck1117:
This is done!

PR #3132, opened 2021-08-10
Begin "vehicleSpeed" refactoring. Module switched to "FunctionalPointerSensor" class. #3106 #3107 #3108.

alxrMironov:


firmware/console/status_loop.cpp
```diff
@@ -689,13 +689,13 @@ void updateTunerStudioState(TunerStudioOutputChannels *tsOutputChannels DECLARE_
...
 #if EFI_VEHICLE_SPEED
-	float vehicleSpeed = getVehicleSpeed();
+	float vehicleSpeed = Sensor::get(SensorType::VehicleSpeed).value_or(0);
 	tsOutputChannels->vehicleSpeedKph = vehicleSpeed;
 	tsOutputChannels->speedToRpmRatio = vehicleSpeed / rpm;
 
 #endif /* EFI_VEHICLE_SPEED */
 #endif /* EFI_PROD_CODE */
-
+
```
mck1117, line 698: stray tab

firmware/controllers/algo/launch_control.cpp
```diff
@@ -55,7 +55,7 @@ bool LaunchControlBase::isInsideSwitchCondition() const {
  * then we have to return true, and trust that we would disable by other condition!
  */ 
 bool LaunchControlBase::isInsideSpeedCondition() const {
-	int speed = getVehicleSpeed(PASS_ENGINE_PARAMETER_SIGNATURE);
+	int speed = static_cast<int>(Sensor::get(SensorType::VehicleSpeed).value_or(0));
```
mck1117, line 58: no explicit cast required, perfectly safe to just assign a float to an int

mck1117: on second thought, why is `speed` an int at all? Just make it a float.

firmware/init/sensor/init_vehicle_speed_sensor.cpp
```diff
@@ -0,0 +1,26 @@
...
+};
+
+static GetVehicleSpeedWrapper vehicleSpeedWrapper;
+
+static FunctionPointerSensor vehicleSpeedSensor(SensorType::VehicleSpeed,
+[]() {
+	return vehicleSpeedWrapper.getVehicleSpeed();
+});
+
+void initVehicleSpeedSensor(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
```
mck1117, line 22: I don't see a call to `initVehicleSpeedSensor`, so the sensor never gets registered!

mck1117:
looks like this:
- fully does #3107 
- fully does #3108

firmware/controllers/algo/launch_control.cpp
alxrMironov: im will remove cast and make speed type changing with another PR and commit. Ok?

firmware/init/sensor/init_vehicle_speed_sensor.cpp
rusefillc: @alxrMironov ^^^ this one is probably the most important one to get this PR less red more green?

mck1117 (approved):

PR #3138, opened 2021-08-11
Hellen128 default

shadowm60:
Updated battery sense and SD card configuration to comply with schematic.


mck1117 (approved):
Looks ok to me!

Issue #3145, opened 2021-08-12
prevent invalid sensors for the moment after burn


mck1117:
1. sensors update
2. flash erase/burn happens (blocking CPU)
3. sensors are briefly invalid
4. sensors update, things are happy

let's avoid #3 happening, it means sometimes the fan relay will click on briefly after a burn

rusefillc:
What hard evidence do we have to support this theory?

mck1117:
The fan on my car clicks on after burn. I've seen exactly this happen.

Issue #3140, opened 2021-08-12
Nissan Altima: something clicks on "Burn"


rusefillc:
Nissan Almera with hellen

something clicks on Burn maybe FSIO ETB relay? #3139

mck1117:
It's probably the fan relay, if you have one.  During the flash, sensors go invalid since the ADC thread gets blocked, so when flash completes, there's a narrow window when things like relays might update without the having updated yet, so they briefly see invalid coolant temperature or something (in which case, fans turn on!)

mck1117:
#3145

rusefillc:
I think fan relay is not configured nor soldered

Xterra has CAN fan no idea what Almera does, this ticket is about Almera from yesterday

rusefillc:
we have ``burnconfig``

rusefillc:
weird dead ticket

PR #3146, opened 2021-08-12
inhibit sensor timeout during flash

mck1117:
#3145 and maybe #3140


rusefillc:
Nice stuff

rusefillc:
@mck1117 please send help

rusefillc:
@mck1117 how do we undraft?

PR #3160, opened 2021-08-15
reducing FSIO in order to enlarge Lua

rusefillc:


mck1117:
? We don't need to shrink this part, we need to shrink the pool(s) of LE

rusefillc:
do not follow

worst case why not both?

at the moment lua script size in config is only 256 is not that too little?

mck1117:
worst case we only need
`#define SYS_ELEMENT_POOL_SIZE 24`
instead of current 128

PR #3163, opened 2021-08-16
issue 3014 part 4 bin2header migration

Come1LLF00:
Just make `build.gradle` for `bin2header` and deleted its `build.xml`.


java_tools/bin2header/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id 'java-library'
+}
+
+repositories {
+    mavenCentral()
```
rusefillc, line 6: i wonder if we really need mavenCentral from each file, would not having it in one place be enough?

PR #3168, opened 2021-08-17
Fix cranking->idling taper phase for useSeparate*ForIdle tables

andreika-git:
Currently useSeparateAdvanceForIdle and useSeparateVeForIdle are broken because of https://github.com/rusefi/rusefi/pull/2900

We fix it by changing the isIdling() method's behavior. This method is used only by useSeparateAdvanceForIdle & useSeparateVeForIdle. And it doesn't know about "CrankToRunTaper" phase. The solution is that CrankToRunTaper should be actually CrankToIdleTaper, because Running phase has higher priority by TPS threshold, and it overrides the taper according to determinePhase(). So there is no transition from cranking to running. There is cranking to idling transition. And therefore isIdling() should return true for CrankToIdleTaper as well.

### DIFF
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -188,7 +188,7 @@ IIdleController::Phase IdleController::determinePhase(int rpm, int targetRpm, Se
 
 	// If still in the cranking taper, disable closed loop idle
 	if (crankingTaperFraction < 1) {
-		return Phase::CrankToRunTaper;
+		return Phase::CrankToIdleTaper;
 	}
 
 	// No other conditions met, we are idling!
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -507,8 +507,8 @@ float getIdleTimingAdjustment(int rpm) {
 	return idleControllerInstance.getIdleTimingAdjustment(rpm);
 }
 
-bool isIdling() {
-	return idleControllerInstance.isIdling();
+bool isIdlingOrTaper() {
+	return idleControllerInstance.isIdlingOrTaper();
 }
 
 static void applyPidSettings(DECLARE_ENGINE_PARAMETER_SIGNATURE) {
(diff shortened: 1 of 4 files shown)


andreika-git:
@mck1117 FYI

mck1117:
> Currently useSeparateAdvanceForIdle and useSeparateVeForIdle are broken because of #2900

This presumes that those should actually be applied during the taper.  On my engine, the VE is 5-10% different with the throttle open during the taper, since the taper is *by definition* not the idle state.

andreika-git:
> This presumes that those should actually be applied during the taper. On my engine, the VE is 5-10% different with the throttle open during the taper, since the taper is _by definition_ not the idle state.

Taper is by definition not the _running_ state either, because _runnin_ is by definition is when the throttle is open. So if a user enables useSeparateAdvanceForIdle/useSeparateVeForIdle, then the firmware CANNOT use running_VE when the throttle is closed. From the VE perspective, there is only running table or idling table. And the taper phase is more idling than running. Is the taper is disabled, that's the idling table would be used, not running. And the difference between them can be tremendous: for example, on my car, the the idle VE is 20, and the running VE is 36 (on the same RPM/load for idle region). Both are AFR-tuned. But if the running VE is used when idling (no throttle), i get AFR=8.5 (!)
image

I don't understand, what do you disagree with? The fact that useSeparate* just don't work for the first afterCrankingIACtaperDuration revolutions? Or with the fact that if useSeparate* is enabled, a user expects those tables to be used when the throttle is closed?

rusefillc:
We have new state mentioned in tests but no values have changed? Do we still need to add test coverage for this mode?

andreika-git:
> We have new state mentioned in tests but no values have changed? Do we still need to add test coverage for this mode?

That's not a new state, that's an old renamed state.

mck1117:
> And the taper phase is more idling than running. 

consider the electronic throttle case - it's very much running, if you have an ETB.  It's indistinguishable from pushing the throttle pedal.  The fact that our engines behave differently during this time is evidence enough that this new behavior needs a setting to enable it (default on is fine).

andreika-git:
Something like this?

rusefillc:
@andreika-git poke what's the plan to make it less draft?

@andreika-git poke unit test coverage :)

rusefillc:
@andreika-git says that he needs help with high-level strategy for unit test covering of this and/or similar change

#3322 created as a follow up

PR #3173, opened 2021-08-18
issue 3014 part 5 ts_plugin_launcher migration

Come1LLF00:
Almost, identically migrated from ant to gradle except for some additional files like `overview.html` that I couldn't exclude from the resulting jar 'cause I didn't find a way to filter dependencies' jar content. Also, I moved `rusefi_online_color_300.png` to `src/main/resources` as gradle requires.
NB!: use `shadowJar` to build jar 'cause standard `jar` task doesn't include binary files from dependencies.


rusefillc:
@Come1LLF00 it looks like this PR includes ChibisOS submodule change by mistake

please remove ChibisOS submodule change from this PR

PR #3188, opened 2021-08-23
Bunch of small fixes related to EG33 board

dron0gus:


firmware/controllers/flash_main.cpp
```diff
@@ -101,7 +101,7 @@ void setNeedToWriteConfiguration(void) {
 	needToWriteConfiguration = true;
 
 #if EFI_FLASH_WRITE_THREAD
-	if (allowFlashWhileRunning()) {
+	if (allowFlashWhileRunning() || (EFI_STORAGE_EXT_SNOR == TRUE)) {
```
mck1117, line 104: why not just return `true` from `allowFlashWhileRunning` if we have EFI_STORAGE_EXT_SNOR?

dron0gus: `allowFlashWhileRunning` is related to internal flash, AFAIR. And defined for each hw_port target...

PR #3193, opened 2021-08-25
rpm bins -> scaled

mck1117:
All RPM bins now use 50x scaling, and are stored in a uint8_t.

Saves 520 bytes.

bumps flash version 10003 -> 10010


racer-coder:
This should be a lot easier now using autoscale and scaled_channels with divisors.  I didn't see any code to handle the bins being off by 50x but that code won't be necessary any more.

mck1117:
Yeah, I shelved this for a while because it was a mess. It should be quite a bit easier now.

rusefillc:
@alxrMironovHome can you help?

Issue #2935, opened 2021-07-09
turbospeed input


rusefillc:
120krpm
 12compressor blades, sensor divide by factor of 8

if you theoretically have 8blade compressor you will have 1pulse per revolution 120k pulses per minute ie 2000hz

rusefillc:
andrey  2 hours ago
120krpm max, what is min?
New

Simon - Orchardperformance  21 minutes ago
Pretty much zero

Simon - Orchardperformance  21 minutes ago
Idling speed of the turbo can be sub 10k

Martin Matousek  16 minutes ago
aalso jacques sensor output is low (no pulses) below 5krpm or something, it probably allso have internal pullup to 5v (edited) 

Martin Matousek  15 minutes ago
and yes during idle turbospeed sensor can show 0krpm for sure and for smaller turbos 190 maybe even 200,210,220krpm is possible (edited) 

Simon - Orchardperformance  < 1 minute ago
Ours show a flicker from 1.5k to 2k.
Its pretty much flatline/ signal noise

Simon - Orchardperformance  < 1 minute ago
250k is right up at the top end of anything small available right now

rusefillc:
Related or not to https://github.com/rusefi/hellen76nissan-issues/issues/3?

rusefillc:
https://www.full-race.com/docs/Speed-Sensor-Installation-Instructions.pdf

mck1117:
250k rpm * 14 blades / 8 = 7.3khz

https://www.wolframalpha.com/input/?i=250000%2Fmin+*+14+%2F+8

mck1117:
an additional divide-by-8 would give us sub-1khz worst case

mck1117:
divide by 2 is easy with a D flip flop - chain more to divide by 2^n:

![image](https://user-images.githubusercontent.com/568254/127218349-66a12587-c79d-4ee9-95e3-f24eb767638a.png)

rusefillc:
Configuration fields added, next step is deciding if we copy/paste vehicle_speed.cpp or reuse vehicle_speed or else

See also - kind of depends on #3081

PR #3209, opened 2021-08-27
Turbo speed sensor implementation #2935

alxrMironov:


firmware/init/sensor/init_turbocharger_speed_sensor.cpp
```diff
@@ -0,0 +1,28 @@
+#include "pch.h"
+
+#include "init.h"
+#include "frequency_sensor.h"
+#include "turbocharger_speed_converter.h"
+
+static FrequencySensor turbochargerSpeedSensor(SensorType::VehicleSpeed, MS2NT(500));
```
mck1117, line 7: remember to set new sensor type!

rusefillc:
@alxrMironov poke

@mck1117 would you be fine with me merging as-is since progress anyway?

mck1117:
let's just wait, it's not hurting anything sitting.

PR #3213, opened 2021-09-01
[DRAFT] IAC H-Bridge Microstepping

andreika-git:
image

* A new setting "stepperDcInvertedPins" needed by RZ7899 H-bridge driver on Hellen boards. It's required because the outputs set to "low" only if both inputs set to "high":
  
image

Without this option the PWM signal won't work on one channel if the second one is set to "low" (because "low+low" means "disabled" for that chip).

* Microstepping support for Half-Step, 1/4-Step and 1/8-Step.

* Current limiting by setting min. and max. duty cycles for the H-bridge outputs. We don't measure the actual current, so we rely on this user-set values to save both the motor and the board from overheating.

* For example, this NEMA17 motor is rated 1.6A max, but direct 12V control draws >3A current, which is unacceptable:
Tested on a bench: https://www.youtube.com/watch?v=hIMrgqxTMaU


andreika-git:
This is a draft (TODO: unit-tests)

firmware/hw_layer/stepper_dual_hbridge.cpp
```diff
@@ -36,19 +50,57 @@ bool DualHBridgeStepper::step(bool positive) {
...
+			m_phase = (m_phase + numStepIncr) & tableSizeMask;
+			update(dutyMult);
+			// sleep 1/Nth of the pause time
+			pause(CONFIG(stepperNumMicroSteps));
+		}
+		return true;
+	}
+
     // step phase, wrapping
     if (positive) {
```
mck1117, line 68: why special case non-micro stepping? can't we just use the same phase table even for full stepping? I think all the math works out properly for 1x stepping.

andreika-git: Sure we can, but the simple and straightforward approach with just "numSteps=1" will give us a so-called "wave drive" method (aka "one phase on"):
image
And unfortunately it would give less torque than the current "full step max torque" method (aka "two phase on"):
image

Max torque drive on the left and wave drive on the right:
image

However it's ok to use the wave method for half-step (and microstepping): the torque loss is not that substantial because of corner smoothing:

image

image

mck1117: Ah hah, makes sense.  Mention that in a comment in-code!

firmware/hw_layer/stepper_dual_hbridge.cpp
```diff
@@ -22,12 +22,26 @@ static const int8_t phaseB[] =
...
 };
 
+static const int8_t microPhases[] = {
+    0,    20,  38,  56,  71,  83,  92,  98,
+    100,  98,  92,  83,  71,  56,  38,  20,
+    0,   -20, -38, -56, -71, -83, -92, -98,
+   -100, -98, -92, -83, -71, -56, -38, -20
+};
+
+static const int maxNumSteps = 8;
```
mck1117, line 32: these can be `constexpr` instead of `const` to force computation at compile time

firmware/hw_layer/stepper_dual_hbridge.cpp
```diff
@@ -36,19 +50,57 @@ bool DualHBridgeStepper::step(bool positive) {
...
+	}
+
     // step phase, wrapping
     if (positive) {
         m_phase = (m_phase + 1) & 0x03;
     } else {
         m_phase = (m_phase - 1) & 0x03;
     }
 
+    update(1.0f);
```
mck1117, line 74: should non-micro mode use `dutyMult`?

firmware/hw_layer/stepper.cpp
```diff
@@ -188,8 +188,12 @@ bool StepDirectionStepper::pulse() {
...
 }
 
-void StepperHw::pause() const {
-	chThdSleepMicroseconds((int)(MS2US(m_reactionTime)));
+void StepperHw::sleep(void) {
+	pause();
+}
+
+void StepperHw::pause(int divisor) const {
+	chThdSleepMicroseconds((int)(MS2US(m_reactionTime)) / divisor);
```
mck1117, line 196: We have to be a little careful here - since we use tick mode at 1khz, we can't sleep less than 1ms.  If you try to sleep less than 1ms, it will sleep for 0, essentially just a `yield()`, potentially starving lower priority threads.

mck1117: see #3214

mck1117: Maybe `sleepMs(max(1, reactionTime/divisor))` to ensure we always sleep

firmware/hw_layer/stepper_dual_hbridge.cpp
```diff
@@ -22,12 +22,26 @@ static const int8_t phaseB[] =
...
 {
     setReactionTime(reactionTime);
 
     m_motorPhaseA = motorPhaseA;
     m_motorPhaseB = motorPhaseB;
+
+    efiAssertVoid(CUSTOM_ERR_ASSERT, CONFIG(stepperNumMicroSteps) <= maxNumSteps, "stepperNumMicroSteps");
 }
 
 bool DualHBridgeStepper::step(bool positive) {
```
mck1117, line 47: This isn't actually micro stepping.... it's doing N micro steps per full step, but never less than one full step worth of movement.

Right now we still do full integer steps, not micro steps.  Should we instead represent as N * M steps, where N is the number of micro steps, and M is the number of actual steps from stop-to-stop on the stepper? Then setting positions 100 and 101 would be one full step apart, but in microstep mode, positions 100.25, 100.5, 100.75 would also exist between those positions.  The idle controller already emits a floating point position, so this wouldn't be too hard.

firmware/hw_layer/stepper_dual_hbridge.cpp
andreika-git: Maybe. 'dutyMult' is 0 by default (not 100) and will break the default behavior for existing users. Some measures have to be made...

firmware/hw_layer/stepper_dual_hbridge.cpp
andreika-git: Yes, exactly. Although this PR has already some noticeable advantages (less noise, motion smoothness, current chopping), it's also a preparation to utilize the fractional steps in the future. It should be added later in a separate PR.

rusefillc:
Andrey please replace useHbridges with useHbridgesToDriveIdleStepper in your branch

Issue #3159, opened 2021-08-15
Lua: Timer class


rusefillc:
our Lua API needs to be aware of time
I assume most of RE Lua is 32 bit? Let's have a special class just for Timer with 64 bit field or maybe reuse existing Timer class in Lua heap?

https://forums.tigsource.com/index.php?topic=22524.msg638134#msg638134

rusefillc:
```
int monster_create (lua_State * ls)
{
 // Monster is a C++ class defined somewhere...
 Monster * monster = new (lua_newuserdata(ls, sizeof(Monster)) Monster;
 luaL_getmetatable(ls, "MetaMonster");
    lua_setmetatable(ls, -2);
   
 return 1;
}

int monster_destroy (lua_State * ls)
{
 luaL_checktype(ls, 1, LUA_TUSERDATA);
 Monster * monster = reinterpret_cast<Monster*>(lua_touserdata(ls, 1));
 monster->~Monster();
 
 return 0;
}

int monster_rawr (lua_State * ls)
{
 luaL_checktype(ls, 1, LUA_TUSERDATA);
 Monster * monster = reinterpret_cast<Monster*>(lua_touserdata(ls, 1));
 monster->rawr(); // prints a menacing "rawr! I's gun eet u!" message
 
 return 0;
}

static const luaL_reg Monster_funcs[] =
{
    {"create", monster_create},
    {0, 0}
};

static const luaL_reg Monster_methods[] =
{
 {"__gc", monster_destroy},
 {"rawr", monster_rawr},
    {0, 0}
};

// Initialize Lua interface in some function...
luaL_newmetatable(ls, "MetaMonster");
lua_pushstring(ls, "__index");
lua_pushvalue(ls, -2);
lua_settable(ls, -3);

luaL_openlib(ls, 0, Monster_methods, 0);
luaL_openlib(ls, "Monster", Monster_funcs, 0);
```

rusefillc:
https://stackoverflow.com/questions/28723535/how-can-fix-it
https://stackoverflow.com/questions/19041215/lual-openlib-replacement-for-lua-5-2

PR #3216, opened 2021-09-03
lua timer is happy

mck1117:
depends on https://github.com/rusefi/luaaa/pull/1

#3159


mck1117:
now depends on https://github.com/rusefi/luaaa/pull/2

PR #3217, opened 2021-09-03
enable address sanitizer

mck1117:
Turns on the GCC address sanitizer.  It sniffs out things such as use-after-free, memory leaks, buffer overruns, etc.


mck1117:
ooh, it fails on mac!

Issue #3196, opened 2021-08-25
Stop engine button killed coil?


rusefillc:
Alex wonders if stop engine button killed his coil

rusefillc:
See #3180

rusefillc:
@mck1117 is this fixed now? can I CHANGELOG this?

rusefillc:
@mck1117 is this fixed by #3194?

rusefillc:
@mck1117 was this fixed by #3194?

rusefillc:
Related to #3071

rusefillc:
Actually we need #3071 which is currently https://github.com/rusefi/rusefi/pull/3220

mck1117:
This is probably fixed - it should now be impossible to overdwell a coil under any normal circumstance.

Let's go for changelog-worthy!

rusefillc:
Already changelog-ed https://github.com/rusefi/rusefi/commit/b2446a6ddb87e59e270a43283d19b59d32053626

PR #3220, opened 2021-09-05
overdwell protection

### LINKED ISSUE #3071: overdwell protection
If a sync problem happens between scheduling dwell start and coil firing, we can end up with stuck coils in the on state.

https://www.youtube.com/watch?v=0-bSRp9Yvw8

mck1117:
Prevents overdwell by always scheduling the coil fire, then canceling it later when we get more precise information.

This snip shows the engine sniffer with a hack installed to skip scheduling of every 10th spark firing schedule event.  The two dwells with arrows are 50% longer, since they were fired by overdwell protect, instead of by normal spark scheduling.

image

Without this change, those two coils would hang open, possibly melting the coil, to be eventually fired on the next engine cycle.

fix #3071 and probably #3196 too


rusefillc:
please unit test

mck1117:
> please unit test

image

rusefillc:
@mck1117 poke

rusefillc:
@mck1117 i believe that trailingSpark was pretty problematic and while I've made it better it is still a bit weird

more specifically I believe that
```
	/**
	 * todo: better method name since this method executes events in the FUTURE
	 * looks like such a method should be used only in some pretty narrow circumstances
	 * a healthy test should probably use executeActions instead?
	 */

void EngineTestHelper::clearQueue() {
	engine.executor.executeAll(99999999); // this is needed to clear 'isScheduled' flag
	ASSERT_EQ( 0,  engine.executor.size()) << "Failed to clearQueue";
}

```

I've replaced some usages and I while this branch is still failing that test I believe it's worth to rebase

PR #3226, opened 2021-09-05
Make can vss available again

shadowm60:


rusefillc:
ideal scenario would be to have test coverage

that could be kind of hard since we might not have support for dynamic re-initialization here? unless we do?

mck1117:
Why not just use a CAN sensor instead of bastardizing the VSS converter?

shadowm60:
@mck1117 i can create a new sensor for CAN, that dose not look so difficult, i wanted to keep all in one place, or make a new converter for CAN? any solution is good, i just want/need vss :)

rusefillc:
@shadowm60 please rebase and fix merge conflict,  I am ready to merge anything as long as we have green build

Issue #3174, opened 2021-08-18
second lua VM


mck1117:
Add a second lua VM that can run in higher priority thread for built-in system functions (fan, fuel pump, etc)

#2928

mck1117:
I no longer think we should have a second VM.

PR #3231, opened 2021-09-07
system lua: load a script!

mck1117:
#3174 

And do it in unit tests too!


mck1117:
looks like we're overflowing the main stack trying to load a script

rusefillc:
from 0000000020020ef8 to 0000000020020000

i think it wants 3K more?

mck1117:
oh I guess it does, oops

mck1117:
it fits!

PR #3292, opened 2021-10-01
put MAP in the sensor model

mck1117:
Required some new functionality to get map averaging to behave itself, but I think this solution is pretty clean!

Both fast and slow MAP use the same conversion function. Slow MAP uses traditional slow ADC.  Fast map uses a special path to update the fast MAP sensor.

#2667


firmware/init/sensor/init_map.cpp
```diff
@@ -32,13 +22,120 @@ static FunctionPointerSensor baroSensor(SensorType::BarometricPressure,
...
 
+// This converter is shared between both fast and slow: the only difference is
+// how the *voltage* is determined, not how its converted to a pressure.
+static LinearFunc mapConverter;
+static FunctionalSensor slowMapSensor(SensorType::MapSlow, MS2NT(50));
+
+// lowest reasonable idle is maybe 600 rpm
+// one sample per cycle (1 cylinder, or "sample one cyl" mode) gives a period of 100ms
+// add some margin -> 200ms timeout for fast MAP sampling
+static FunctionalSensor fastMapSensor(SensorType::MapFast, MS2NT(200));
```
andreika-git, line 33: Is it correct that the "fast" sensor has larger period than the "slow" one?
Usually frequency = 1 / period

mck1117: That 200ms period is the timeout - which in this case is how long we wait before falling back to the "slow" sensor.  The slow sensor's still updates at 200hz or whatever, but asynchronously, hence its much shorter timeout.

firmware/controllers/sensors/sensor_type.h
```diff
@@ -77,6 +77,9 @@ enum class SensorType : unsigned char {
 
 	TurbochargerSpeed = 35,
 
+	MapFast = 36,
```
rusefillc, line 80: code comment with a quick summary of what is fast MAP vs what is slow MAP

mck1117: I guess - they're just averaged (synchronous) map vs asynchronous map

rusefillc:
Drama at #3325

Issue #3302, opened 2021-10-03
USB Console thread runs out of stack and corrupts surrounding


dron0gus:
rusefi.map:
```
 .bss._ZL10usbChannel
                0x0000000020045e94      0x12c /tmp/rusefi.elf.QVDVPE.ltrans0.ltrans.o
 .bss._ZL10usbConsole
                0x0000000020045fc0      0x4d0 /tmp/rusefi.elf.QVDVPE.ltrans0.ltrans.o
```
gdb:
```
(gdb) p/x usbChannel
$23 = {<TsChannelBase> = {_vptr.TsChannelBase = 0x80ad878 <vtable for UsbChannel+8>, scratchBuffer = {0x0 <repeats 286 times>}, name = 0x80ad850}, m_channel = 0x200228e8}
(gdb) p &usbChannel.m_channel
$1 = (BaseChannel * const *) 0x20045fbc <usbChannel+296>
(gdb) watch *0x20045fbc
Hardware watchpoint 1: *0x20045fbc
... let it run and connect to TS

Thread 5 received signal SIGTRAP, Trace/breakpoint trap.
[Switching to Thread 537158712]
Sensor::get (type=SensorType::Rpm) at ./controllers/sensors/sensor.cpp:202
202		const auto entry = getEntryForType(type);
(gdb) bt
#0  Sensor::get (type=SensorType::Rpm) at ./controllers/sensors/sensor.cpp:202
#1  0x00264c20 in updateTunerStudioState (tsOutputChannels=0x20045cd8 <tsOutputChannels>) at ./console/status_loop.cpp:464
#2  0x00266d28 in prepareTunerStudioOutputs () at ./console/status_loop.cpp:878
#3  0x00264274 in TunerStudio::cmdOutputChannels (this=0x200219a0 <tsInstance>, tsChannel=0x20045e94 <usbChannel>, offset=0, count=256) at ./console/binary/tunerstudio_commands.cpp:24
#4  0x00263ea4 in TunerStudioBase::handleCrcCommand (this=0x200219a0 <tsInstance>, tsChannel=0x20045e94 <usbChannel>, data=0x20045e99 <usbChannel+5> "", incomingPacketSize=5)
    at ./console/binary/tunerstudio.cpp:704
#5  0x00263330 in tsProcessOne (tsChannel=0x20045e94 <usbChannel>) at ./console/binary/tunerstudio.cpp:528
[...]

PR #3308, opened 2021-10-03
reduce updateTunerStudioState stack usage

mck1117:
`updateTunerStudioState` uses a ton of stack on unoptimized builds, since it doesn't reuse stack space for multiple different locals/scratch work unless you enable optimization.  It will, however, reuse the same stack space if the work is being done by different functions!

When optimization is enabled, this change has almost no impact - all these new functions get inlined, and the stack space reused so very little is used in that case either way.

helps #3302

### DIFF
--- a/firmware/console/status_loop.cpp
+++ b/firmware/console/status_loop.cpp
@@ -459,6 +459,126 @@ extern HIP9011 instance;
 
 #if EFI_TUNER_STUDIO
 
+static void updateTempSensors() {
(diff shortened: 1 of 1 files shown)


mck1117:
Tested by setting the stack huge (no possibility of overflow), then observing that with an unoptimized build, the stack usage reduced on the USB stack by 232 bytes in the same conditions.

mck1117:
Even with this change, optimized builds use ~250 bytes of stack, and unoptimized uses ~750 bytes of stack :(

Issue #3325, opened 2021-10-06
Engine Stall due to fubar MAP sensor


BeerMoneyMotorsports:
NB1 Hellen
When pressing "burn" after a settings change, engine stalls and shuts down.  
Unsaved changes lights up briefly and goes out.  When this occurs, the blue led stays lit and then goes back to blinking.  
Test function= Cranking Base IAC position

Current Firmware date= 2021-10-06 12:12

https://rusefi.com/online/view.php?msq=668

log-
https://rusefi.com/online/view.php?log=360

mck1117:
Can you press the "no flash burn mode" button in the bottom right of the bench test page, then try it again?

BeerMoneyMotorsports:
@mck1117 i can try that.  

Reverted to a previous firmware v20210929 and issue is NOT present

rusefillc:
@BeerMoneyMotorsports at the moment you are burning not just any random setting but a very important for idle setting which directly affects idle

for us to know if this is about IAC specifically or not please attempt to change some much safer setting

changing compression ratio is best since that setting does absolutely nothing to the best of my knowledge

rusefillc:
smoking gun of "engine dies first, THEN it burns unsaved setting" is the fact that RPM goes down GRADUALLY. If we would have to start burning on a properly running engine we would have loss of communication without any RPM changes.

![image](https://user-images.githubusercontent.com/48498823/136269647-f6ee9134-7c4e-45a8-9d7d-8d8351e13d68.png)

rusefillc:
how did MAP sensor go from 35 to 5?

![image](https://user-images.githubusercontent.com/48498823/136272294-1bc3f5e9-a2ba-46e1-94c0-06247ab1f420.png)

PR #3328, opened 2021-10-06
fix map

mck1117:
maybe related to #3325


rusefillc:
huh? how is this change changing anything? is there some address magic somewhere?

mck1117:
those fields are referred to by name in `configureMapFunction`, but by position in `getMapCfg`.

For example:

```
case MT_MPX4250A:
    return {0.25, 20, 4.875, 250};
```

Those 4 numbers are volt1, kpa1, volt2, kpa2, but the fields were in the wrong order.

rusefillc:
So we need a unit test which would decode one known value for one known configuration?

PR #3330, opened 2021-10-06
basic knock response params

mck1117:


mck1117:
@OrchardPerformance, thoughts?

rusefillc:
@OrchardPerformance thoughts? good enough to move forward?

OrchardPerformance:
I agree with Dron that some more smarts are needed. 

Pulling some logic from the subaru and some general knowledge I suggest: 

- Knock retard aggression ( % value of how much timing is removed per knock )
- Table 8x8 RPM vs Load for maximum retard allowed 
- Single parameter for triggering fuel enrichment ( another %, the trigger is the cumulative value of the retard aggression )
- Single parameter for critical error ( again % like the enrichment trigger) 

- We already have a baseline noise table

Issue #3323, opened 2021-10-06
Get AndreiKA on more recent rusEFI


rusefillc:
At the moment @andreika-git uses some August fork of rusEFI 

in my opinion it's important to invest into getting @andreika-git back on master

https://github.com/andreika-git/rusefi/tree/hellen81_config

PR #3334, opened 2021-10-07
faster uniform adc cleanup

mck1117:
progress to #3323 

depends #3327


andreika-git:
> 1. F4/F7 uses separate fast ADC, which calls `onFastAdcComplete`
> 2. H7 uses fast "slow" ADC, which calls `onFastAdcComplete`

The problem is **when**  onFastAdcComplete is called.
We still have
`#define ADC_BUF_DEPTH_FAST      4`
which means that the interrupt will be called only after 4 measurements in a row, not one. So onFastAdcComplete may not be called at 10kHz frequency. Is it true?

andreika-git:
The whole point of EFI_FASTER_UNIFORM_ADC is to get ADC measurements at 40kHz, and consume them at 10kHz with averaging.

mck1117:
> No, they use getAvgAdcValue().

Check again, nothing in `onFastAdcComplete` uses `getAvgAdcValue`.

> And I'm not sure that hw oversampling at 40kHz is a good replacement for that, because it's more high frequency LPF.
>  get ADC measurements at 40kHz, and consume them at 10kHz with averaging.

Oversampling is not an alternative for a proper FIR or IIR filter.  Oversampling exists in our space essentially only to reduce quantization noise (and other uncorrelated noise, like thermal noise), not to prevent aliasing.

Analog inputs *require* a proper hardware antialiasing filter, then a sample rate sufficiently above the Nyquist rate to push aliases down below the noise floor.  This is exactly what the slow ADC does - it samples at some frequency, then applies a digital filter to further restrict the available bandwidth.

The special case of the H7 is that the sample buffer is updated at a full 10khz, but it's only observed by the `SlowAdcController` at 1khz or something, reducing the effective sample rate down from 10khz to 1khz.

> The whole point of EFI_FASTER_UNIFORM_ADC is to get ADC measurements at 40kHz, and consume them at 10kHz with averaging.

In fact taking N samples at N*10khz is worse than taking N samples "in an instant" at 10khz (as H7 hardware oversample does).  Multiple samples taken "at the same instant" will have the same effect of suppressing uncorrelated noise, but doesn't include the associated group delay of sampling over several intervals.

andreika-git:
> In fact taking N samples at N*10khz is worse than taking N samples "in an instant" at 10khz (as H7 hardware oversample does). Multiple samples taken "at the same instant" will have the same effect of suppressing uncorrelated noise, but doesn't include the associated group delay of sampling over several intervals.

EFI_FASTER_UNIFORM_ADC is not about noise, it's for trigger_ADC and detection of fast signal transients with a resolution finer than 1/10kHz. I understand that you maybe don't need it, but please don't kill this possibility because of that.

andreika-git:
> > resolution finer than 1/10kHz
> 
> so why not just run the ADC faster?

1. Don't we hit the limit of ADC sampling speed, considering the 4x oversampling?
2. We need to make only triggerSampleIndex faster, but not all 16 channels. That's why we have a different codepath for "if (++adcCallbackCounter >= ADC_BUF_NUM_AVG)" and calling onFastAdcCompleteInternal() only every 4th interrupt.

mck1117:
> detection of fast signal transients with a resolution finer than 1/10kHz

if only there were some way to do it autonomously in hardware, without an ADC....

andreika-git:
> > detection of fast signal transients with a resolution finer than 1/10kHz
> 
> if only there were some way to do it autonomously in hardware, without an ADC....

H7 has a hardware comparator, and Hellen mcu module is ready to use it in the future (I have some plans for it).

mck1117:
> it works even with 10kHz fast sampling, but 20kHz is more ample and safe

As you know, I'm still not sold on that method, as the digital input's threshold isn't very predictable or accurate.  Anyway, an option to set OS and rate per-board is reasonable.

PR #3339, opened 2021-10-09
per-MCU lua size

mck1117:


rusefillc:
@mck1117 simulator not happy

Issue #3359, opened 2021-10-17
Simulator crashes


rusefillc:
small problem: simulator crashes

bigger problem: we do not have CI for simulator

PR #3360, opened 2021-10-17
Fix simulator

mck1117:
No more crash on launch.  TS connects.

Adds a CI step to launch the simulator, then closes itself (cleanly) after 10 seconds.

#3359


rusefillc:
thank you @mck1117

PR #3386, opened 2021-10-21
Use SSH for Doxygen

chuckwagoncomputing:
Required secrets:
RUSEFI_SSH_SERVER
RUSEFI_SSH_USER
RUSEFI_SSH_PASS

Don't remove the FTP secrets yet, as they're still used by other workflows.

FTP runtime: 3h 32m 36s
SSH runtime: 6m 28s

### DIFF
--- a/.github/workflows/gen-docs.yaml
+++ b/.github/workflows/gen-docs.yaml
@@ -24,8 +24,8 @@ jobs:
       run: |
         if [ "${{github.ref}}" = "refs/heads/master" ]; then
           echo "::set-env name=RUSEFI_SSH_SERVER::${{secrets.RUSEFI_SSH_SERVER}}";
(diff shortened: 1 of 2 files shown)


rusefillc:
I think there is a defect with paths here and unintentional change:

we used to have multiple users with different relative home folders. With suggested universal username we will start uploading to wrong folder would not we?

incremental change would be to have not just RUSEFI_SSH_USER but specifically RUSEFI_DOXYGEN_SSH_USER

one day we can consider merging all usernames and adding relative path logic or variable but I would prefer not to think about this now :( can we please implement minimal viable change which means multiple SSH usernames?

rusefillc:
@chuckwagoncomputing I've just added the following secrets:

RUSEFI_SSH_SERVER
RUSEFI_DOXYGEN_SSH_USER
RUSEFI_DOXYGEN_SSH_PASS

PR #3389, opened 2021-10-21
lua can set sensors

mck1117:


mck1117:
Once upon a time, maybe Lua can receive CAN frames and then set sensors on the ECU...

PR #3420, opened 2021-10-29
is that the way or shall we else?

rusefillc:
zero dt gives us division by zero. do we want to fail fast?


rusefillc:
https://github.com/rusefi/rusefi/commit/747bbd9c6ab466b0c58ab5f5235127559d9bdf0f

PR #3426, opened 2021-10-29
better pch usage and gmock constructors

mck1117:
saves 5-8 seconds of unit test build time


rusefillc:
MacOS says "please fix me"

Issue #3435, opened 2021-10-31
remove background from "program" icon or replace icon


rusefillc:
at the moment ugly white background on icon

![image](https://user-images.githubusercontent.com/48498823/139594469-c28635c5-6f57-4e6a-877b-6c40e8845199.png)

need a better icon at least without ugly background

rusefillc:
and kind of broader question of all easily visible icons

rusefillc:
Maybe use $$$ toward design improvements

rusefillc:
@ksmola what shall we do about 

![image](https://user-images.githubusercontent.com/48498823/139608817-278c0829-1745-4ec6-b36e-8db9b50e0673.png)

background and all?

ksmola:
see latest PR

rusefillc:
Thank you @ksmola

That's one amazing leap total total domination!

PR #3437, opened 2021-11-01
Vector logos

ksmola:
see  #3435
console code needs to be updated to use upload48.png instead of upload48.jpg


rusefillc:
did I merge before or after latest change? what was the latest change?

PR #3444, opened 2021-11-02
autoscale fields in generated structs

mck1117:
Can now add `autoscale<ratio>` to scalar and array fields.  Emits a `scaled_channel<T>` instead of `T`, so the scaling is now encoded in to the type system, instead of manually scaling values in the firmware.  Two example fields are changed to the new autoscale: stoich ratio primary and secondary.


rusefillc:
I would like to do a bad thing and hold this change hostage of https://github.com/rusefi/proteus-Honda-OBD2A-adapter/issues/4#issuecomment-955029325

PR #3450, opened 2021-11-03
lock and delay gpio init

mck1117:
~prevent other threads from editing our pin while we aren't looking~ (we're already under lock there), and also wait 1ms after pin init to check startup state in case things take a second to happen


rusefillc:
is this related to https://rusefi.com/forum/viewtopic.php?p=42987#p42987?

https://rusefi.com/forum/viewtopic.php?p=42987#p42987 says "while driving" while this PR seems to be focused on init phase.

mck1117:
yes, that is related.

After further thought, it's probably a genuine hardware failure that we detected, and this won't help.

PR #3453, opened 2021-11-03
Basic framework for high pressure fuel pump control

racer-coder:
DO NOT PULL THIS AS IS.  It is completely untested, not with the simulator, not with real hardware.  I was hoping it could be a spring board to discuss whether the approach is ok before I start writing actual unit tests.  Obviously there's a bit of chicken and egg here before testing on an actual running motor.

There are a lot of TODO that hopefully you can provide feedback on.

The compiles in the simulator.

The idea is to calculate the volume of fuel that needs to be added per lobe.  Then apply a simple PI controller on top of that to manage the pressure.

From the volume we calculate the # of degrees of pump action we need.

Then it's a matter of using VVT and other data to trigger the pump at the right time.

### DIFF
--- a/firmware/controllers/algo/fuel_math.cpp
+++ b/firmware/controllers/algo/fuel_math.cpp
@@ -92,7 +92,7 @@ float getCrankingFuel3(
 	return crankingFuel;
 }
 
-floatms_t getRunningFuel(floatms_t baseFuel DECLARE_ENGINE_PARAMETER_SUFFIX) {
+float getRunningFuel(float baseFuel DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	ScopePerf perf(PE::GetRunningFuel);
 
 	ENGINE(engineState.running.baseFuel) = baseFuel;
@@ -109,7 +109,7 @@ floatms_t getRunningFuel(floatms_t baseFuel DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	efiAssert(CUSTOM_ERR_ASSERT, !cisnan(cltCorrection), "NaN cltCorrection", 0);
 	efiAssert(CUSTOM_ERR_ASSERT, !cisnan(postCrankingFuelCorrection), "NaN postCrankingFuelCorrection", 0);
 
-	floatms_t runningFuel = baseFuel * baroCorrection * iatCorrection * cltCorrection * postCrankingFuelCorrection;
+	float runningFuel = baseFuel * baroCorrection * iatCorrection * cltCorrection * postCrankingFuelCorrection;
 	efiAssert(CUSTOM_ERR_ASSERT, !cisnan(runningFuel), "NaN runningFuel", 0);
 
 	ENGINE(engineState.running.fuel) = runningFuel * 1000;
@@ -258,10 +258,10 @@ static float getCycleFuelMass(bool isCranking, float baseFuelMass DECLARE_ENGINE
 }
 
 /**
- * @returns	Length of each individual fuel injection, in milliseconds
+ * @returns	Mass of each individual fuel injection, in "thingy-grams"??? TODO FIX
  *     in case of single point injection mode the amount of fuel into all cylinders, otherwise the amount for one cylinder
  */
-floatms_t getInjectionMass(int rpm DECLARE_ENGINE_PARAMETER_SUFFIX) {
+float getInjectionMass(int rpm DECLARE_ENGINE_PARAMETER_SUFFIX) {
 	ScopePerf perf(PE::GetInjectionDuration);
 
 #if EFI_SHAFT_POSITION_INPUT
--- a/firmware/controllers/engine_cycle/high_pressure_fuel_pump.cpp
+++ b/firmware/controllers/engine_cycle/high_pressure_fuel_pump.cpp
@@ -13,56 +13,163 @@
 #include "high_pressure_fuel_pump.h"
 #include "spark_logic.h"
 
-#if EFI_HPFP
-
-#define LOBE_COUNT 3
+//#if EFI_HPFP
+#if 1
 
 class HpfpActor {
 public:
-	angle_t extra;
-	int phaseIndex;
+    uint8_t m_lobe_index;
+    float m_I_sum;
+    float m_pressureTarget;
 
-	AngleBasedEvent open;
-	AngleBasedEvent close;
+    AngleBasedEvent m_open;
+    AngleBasedEvent m_close;
+
 	DECLARE_ENGINE_PTR;
+
+    void start();
+
+private:
+    void scheduleNextCycle();
+    angle_t findNextLobe();
+    float calcFuelPercent(int rpm);
+    float calcPI(int rpm, float calc_fuel_percent);
+
+    static void hpfpPinTurnOn(HpfpActor *) {
+        enginePins.hpfpValve.setHigh();
+    }
+
+    static void hpfpPinTurnOff(HpfpActor *current) {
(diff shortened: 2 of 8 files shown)


firmware/controllers/algo/fuel_math.cpp
```diff
@@ -92,7 +92,7 @@ float getCrankingFuel3(
 	return crankingFuel;
 }
 
-floatms_t getRunningFuel(floatms_t baseFuel DECLARE_ENGINE_PARAMETER_SUFFIX) {
+float getRunningFuel(float baseFuel DECLARE_ENGINE_PARAMETER_SUFFIX) {
```
mck1117, line 95: let's avoid unrelated changes in a PR?

many small PRs is greatly preferred over one big PR

firmware/controllers/engine_cycle/high_pressure_fuel_pump.cpp
```diff
@@ -13,70 +13,172 @@
...
+    static void hpfpPinTurnOff(HpfpActor *current) {
+        RegisteredNamedOutputPin *output = &enginePins.hpfpValve;
+        output->setLow();
+
+        current->scheduleNextCycle();
+    }
 };
 
-static HpfpActor actors[LOBE_COUNT];
+static HpfpActor actor; // TODO: Instantiate one per bank? Need more sensors....
```
mck1117, line 50: let's start with single pump control - 99% of the engines we care about for now are 4cyl, single pump, but on a variable cam

firmware/controllers/engine_cycle/high_pressure_fuel_pump.cpp
```diff
@@ -13,70 +13,172 @@
...
+
+template<int TColNum, int TRowNum, typename vType, typename kType,
+         typename TValueMultiplier = efi::ratio<1>>
+auto makeMap3D(vType table[TRowNum][TColNum], kType rowBins[TRowNum], kType columnBins[TColNum]) {
```
mck1117, line 54: this will be unnecessary after https://github.com/rusefi/rusefi/pull/3452

firmware/integration/rusefi_config.txt
```diff
@@ -1486,7 +1485,40 @@ tChargeMode_e tChargeMode;
 int8_t[MAX_CYLINDER_COUNT iterate] fuelTrim;;"Percent", @@PERCENT_TRIM_BYTE_PACKING_DIV@@, 0, -25, 25, 2
 
 	float[GAP_TRACKING_LENGTH iterate] triggerGapOverrideTo;;"to", 1, 0, 0, 20, 2
-	int[326] mainUnusedEnd;;"units", 1, 0, -20, 100, 0
+
+! Someday there will be a 6th option for BMW S55 that uses a separate shaft just for HPFP
+#define hpfp_cam_e_enum "NONE", "Intake 1", "Exhaust 1", "Intake 2", "Exhaust 2"
```
mck1117, line 1490: separate shaft should just be an "unphased" option, right? There are other engines out there with variable cams, with the pump non-variable (for example, the C8 Z06's LT6 engine :) )

mck1117: or is that what `none` means?

firmware/integration/rusefi_config.txt
```diff
@@ -1486,7 +1485,40 @@ tChargeMode_e tChargeMode;
...
+        uint8_t[HPFP_LOBE_PROFILE_SIZE] hpfpLobeProfileQuantityBins;;"%", 0.4, 0, 0, 100, 1
+        uint8_t[HPFP_LOBE_PROFILE_SIZE] hpfpLobeProfileAngle;;"deg", 0.5, 0, 0, 125, 1
+        uint8_t[HPFP_DEADTIME_SIZE] hpfpDeadtimeVoltsBins;;"volts", 1, 0, 0, 255, 0
+        uint16_t[HPFP_DEADTIME_SIZE] hpfpDeadtimeMS;;"ms", 0.001, 0, 0, 65, 0
+        hpfp_target_t hpfpTarget;
+        uint16_t[HPFP_TARGET_SIZE] hpfpTargetLoadBins;;"load", 0.1, 0, 0, 6500, 1
+        uint16_t[HPFP_TARGET_SIZE] hpfpTargetRpmBins;;"RPM", 1, 0, 0, 65000, 0
+        hpfp_compensation_t hpfpCompensation;
+        uint16_t[HPFP_COMPENSATION_SIZE] hpfpCompensationLoadBins;;"cc/lobe", 0.001, 0, 0, 65, 3
+        uint16_t[HPFP_COMPENSATION_SIZE] hpfpCompensationRpmBins;;"RPM", 1, 0, 0, 65000, 0
```
mck1117, line 1519: I'd love to discuss different hpfp control strategies before we settle on one.  This looks mostly right, but I'd like to pick brains before we commit to it. Maybe post on forum and/or slack?

rusefillc:
Exciting stuff!

Any chance to have some test coverage in https://github.com/rusefi/rusefi/tree/master/unit_tests as part of this PR?

rusefillc:
maybe separate PR just to add new configuration fields so that we get code generation working and reduce drama with merge conflict contention?

new config fields even if dead in the beginning are valuable progress as long as we have 80% overall alignment

mck1117:
> maybe separate PR just to add new configuration fields so that we get code generation working and reduce drama with merge conflict contention?
> 
> new config fields even if dead in the beginning are valuable progress as long as we have 80% overall alignment

yes - agree. Smaller PRs good 🙂 

Also, that gives us a place to figure out the control strategy without tripping over implementation details at the same time.

firmware/controllers/algo/fuel_math.cpp
racer-coder: Yeah I can rip that stuff out and submit it separately.

firmware/integration/rusefi_config.txt
racer-coder: BMW N55/S55 engines actually have a separate shaft with a separate cam sensor.  Maybe it's so they don't have to time the pump shaft TDC on engine rebuilds?

https://www.realoem.com/bmw/enUS/showparts?id=3A93-USA-05-2015-F30-BMW-335i&diagId=11_4779
That shows where the HPFP mounts, along with part 6 - cam sensor.

NONE means no VVT.  Not sure those engines exist, but hey, why not....

racer-coder: Was looking through Motec documentation, and they have a setting for "reduction" meaning the pump shaft isn't 1:1 with the crank.  I bet that's what the extra cam sensor is for.  Since the pump valves themselves can float if driven too fast, then having a reduction gear on the pump shaft lets you rev the engine higher.  It lets you effectively get 2.8 etc lobes per cam instead of having to choose between 2 and 3.

firmware/controllers/engine_cycle/high_pressure_fuel_pump.cpp
racer-coder: How so?  It isn't immediately obvious to me.  FYI the point of this change is to just create the Map3D structure on the fly, rather than precreating it.  Precreating has no performance advantage, it takes 4 words of RAM, and it actually slower due to the vtable indirection (though that could be changed by marking some of the overridden methods as final).

mck1117:
One more thing: this is probably an opportunity to `ClosedLoopController`, which would make writing unit tests much easier, as it forces a clear distinction between inputs, target, open loop, closed loop, and output.

PR #3459, opened 2021-11-04
Create an interpolate3d function to match the interpolate2d, but for 3d tables.

racer-coder:
Remove some implicit C array to pointer conversions in Map3D by adding
(), *, and & where appropriate.  This allows us to remove getBinPtr.

Call interpolate3d from Map3d so there's no code duplication.

### DIFF
--- a/firmware/util/containers/table_helper.h
+++ b/firmware/util/containers/table_helper.h
@@ -29,7 +29,8 @@ template<int TColNum, int TRowNum, typename TValue, typename TColumn, typename T
 class Map3D : public ValueProvider3D {
 public:
 	template <typename TValueInit, typename TRowInit, typename TColumnInit>
-	void init(TValueInit table[TRowNum][TColNum], const TRowInit rowBins[TRowNum], const TColumnInit columnBins[TColNum]) {
+	void init(TValueInit (&table)[TRowNum][TColNum],
+                  const TRowInit (&rowBins)[TRowNum], const TColumnInit (&columnBins)[TColNum]) {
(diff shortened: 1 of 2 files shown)


firmware/util/containers/table_helper.h
```diff
@@ -29,81 +29,69 @@ template<int TColNum, int TRowNum, typename TValue, typename TColumn, typename T
...
 		static_assert(TValueMultiplier::den == TMult);
 		static_assert(TValueMultiplier::num == 1);
 
-		m_values = reinterpret_cast<TValue*>(&table[0][0]);
+		m_values = reinterpret_cast<TValue (*)[TRowNum][TColNum]>(&table);
 	}
 
-	void initValues(TValue table[TRowNum][TColNum]) {
-		m_values = &table[0][0];
+    void initValues(TValue (&table)[TRowNum][TColNum]) {
```
mck1117, line 72: looks like some tabs ended up as spaces?

mck1117 (approved):

firmware/util/containers/table_helper.h
racer-coder: What is the preferred style?  Tabs vs spaces, if tabs then how much does a tab represent, etc?
Ok to remove trailing space from blank lines?

I see: https://github.com/rusefi/rusefi/wiki/Code-Style
but I don't know Eclipse conventions.  Maybe just set emacs style to k&r and hope for the best?

mck1117: the important two in the preferred style are:
- tabs
- curlies on same line, not next line

K&R style is probably close enough.

we're not particularly strict about much else

PR #3460, opened 2021-11-04
Remove correct file in create_ini_image.sh

racer-coder:
I was getting disk full errors because I had two inis in a single zip.  Oops.

### DIFF
--- a/firmware/hw_layer/mass_storage/create_ini_image.sh
+++ b/firmware/hw_layer/mass_storage/create_ini_image.sh
@@ -12,20 +12,23 @@ FS_SIZE=$3
 SHORT_BOARDNAME=$4
 BOARD_SPECIFIC_URL=$5
 
+IMAGE=ramdisk.image
+ZIP=rusefi.ini.zip
+
 echo "create_ini_image: ini $FULL_INI to $H_OUTPUT size $FS_SIZE for $SHORT_BOARDNAME [$BOARD_SPECIFIC_URL]"
 
-rm -f rusefi.zip ramdisk_image.h
+rm -f $ZIP $IMAGE
 
 # copy *count*KB of zeroes
-dd if=/dev/zero of=ramdisk.image bs=1024 count=$FS_SIZE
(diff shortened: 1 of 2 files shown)


rusefillc:
root cause of the issue is copy-paste of magic file name, right?

while you are fixing the typo shall we go further and introduce a variable for file name and use the variable instead of magic file name twice?

mck1117:
why does nobody else see this?

rusefillc:
everybody else relies on GHA which does not keep state between builds?

racer-coder:
If the script succeeds, it correctly deletes the zip file. It's only if it
fails, then it will fail again.

I need to separately investigate why it failed mid run for me.

firmware/hw_layer/mass_storage/create_ini_image.sh
```diff
@@ -14,7 +14,7 @@ BOARD_SPECIFIC_URL=$5
 
 echo "create_ini_image: ini $FULL_INI to $H_OUTPUT size $FS_SIZE for $SHORT_BOARDNAME [$BOARD_SPECIFIC_URL]"
 
-rm -f rusefi.zip ramdisk_image.h
+rm -f rusefi.ini.zip ramdisk_image.h
```
rusefillc, line 17: root cause of the issue is copy-paste of magic file name, right?

while you are fixing the typo shall we go further and introduce a variable for file name and use the variable instead of magic file name twice?

rusefillc (changes requested):

firmware/hw_layer/mass_storage/create_ini_image.sh
racer-coder: Ok done.

FYI my root cause was the HPFP change.  Apparently we're just tripping over the 112kB limit for the FAT filesystem even though the ZIP file is only 88kB.  I'm not sure where the extra overhead is coming from, but I reduced the size of the root to work around this for now.  Can we bump the 112kB for the image size?  If not we'll probably run into this again.

PR #3464, opened 2021-11-05
Add 3d table support to the java config file parser.

racer-coder:
This removes the need to use custom for 3d tables.  Syntax (to avoid changing the ANTLR parser) is
to use 'x' instead of separate [], i.e.:
        uint16_t[HPFP_TARGET_SIZE x HPFP_TARGET_SIZE] hpfpTarget;;"kPa", 1, 0, 0, 65000, 0


racer-coder:
Am I supposed to commit the JAR file or does the build system commit that separately?

rusefillc:
please commit .jar

please commit a "poke" new line change into rusefi.input to cause cache miss code generation

rusefillc (approved):

racer-coder:
In testing I found the only change after calling gen_config.sh was to the mass storage gzip file.  Do you still want a dummy change to force the buildsystem to generate new headers?

rusefillc:
i want new line in rusefi.input to cause full rebuild of all .ini to be more sure that new jar works as intended

mck1117:
Thanks, this will make using #3444 much easier!

PR #3466, opened 2021-11-05
nissan MR18DE VVT cam pattern

mck1117:
cam pattern for nissan versa/tiida

It's not perfect, but it's probably close enough for the engine to run and collect a better pattern.


firmware/controllers/algo/engine_types.h
```diff
@@ -517,14 +517,16 @@ typedef enum {
...
+
 	// do not forget to edit "#define trigger_type_e_enum" line in integration/rusefi_config.txt file to propogate new value to rusefi.ini TS project
 	// do not forget to invoke "gen_config.bat" once you make changes to integration/rusefi_config.txt
 	// todo: one day a hero would integrate some of these things into Makefile in order to reduce manual magic
 	//
 	// Another point: once you add a new trigger, run get_trigger_images.bat which would run rusefi_test.exe from unit_tests
 	//
 	// todo: before increasing from 69 to 70 we shall use the half dozen of unused IDs
-	TT_UNUSED = 69, // this is used if we want to iterate over all trigger types
+	TT_UNUSED = 70, // this is used if we want to iterate over all trigger types
```
rusefillc, line 529:
```
// todo: before increasing from 69 to 70 we shall use the half dozen of unused IDs
```

rusefillc (changes requested):

firmware/controllers/algo/engine_types.h
mck1117: oh, good spot

mck1117: There are only actually one or two free.  Note that a few are marked `INVALID` in the TS list so that you can't pick them as they're VVT cam patterns.

rusefillc:
grrrrr https://github.com/mck1117/wideband/pull/20

mck1117:
> 0 of 36 checks passed 

brave man!

PR #3468, opened 2021-11-05
Allow scaled_channel to have both multiplier and divisor

racer-coder:
Disallow scaled_channel for float.  I can't think of a reason to allow it, and it gets in the way
of rounding.

Add separate template param to setTable; let the compiler sort out whether assignment can happen
between TElement and VElement without forcing them to be the same at function call time.

### DIFF
--- a/firmware/controllers/algo/rusefi_types.h
+++ b/firmware/controllers/algo/rusefi_types.h
@@ -96,7 +96,7 @@ typedef brain_pin_e egt_cs_array_t[EGT_CHANNEL_COUNT];
 
 #if __cplusplus
 #include "scaled_channel.h"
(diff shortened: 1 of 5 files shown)


racer-coder:
Ooops apparently just running unit tests isn't enough.  Compiling the simulator fails (which would have let me find issues before the build system)...

rusefillc:
why are we touching this now? :)

racer-coder:
So I can make HPFP RPM bins use uint8_t with a *50 factor.  Previously scaled_channel only handles /X not *X.

My plan is to add this, then fix autoscale to actually autoscale (read the factor that already exists in the config line and determine mul and div factors), then fix the tables to use this.

mck1117 (approved):

PR #3469, opened 2021-11-05
configurable etb maximum

### LINKED ISSUE #3457: Limit ETB to < 100%
@mck1117 and I through testing on my car found problems with ETB when value is 100% or greater. 

We need to limit the ETB to less than 100%. @mck1117 feel free to comment

mck1117:
fix #3457


rusefillc:
unit test on the way to "not draft"?

mck1117:
> 
> 
> unit test on the way to "not draft"?

yup

PR #3470, opened 2021-11-05
Determine the correct mul/div for autoscale from the TS config line instead of specifying it twice.

racer-coder:


java_tools/configuration_definition/RusefiConfigGrammar.g4
```diff
@@ -94,7 +94,7 @@ fieldOptionsList
 
 arrayLengthSpec: numexpr (ArrayDimensionSeparator numexpr)?;
 
-autoscale: Autoscale '<' numexpr '>';
+autoscale: Autoscale;
```
mck1117, line 97: we probably don't need a separate rule here, we can just put `Autoscale?` down below instead of `autoscale?`

java_tools/configuration_definition/src/main/java/com/rusefi/ConfigField.java
```diff
@@ -261,7 +259,39 @@ public boolean isFsioVisible() {
         return fsioVisible;
     }
 
-    public String autoscaleSpec() { return this.autoscaleSpec; }
+    public String autoscaleSpec() {
```
mck1117, line 262: looks like indentation is wonky here

racer-coder: Yeah I just can't win.  The C++ is consistently indented while the Java code is a mixture of tabs (8 space?) and 4 spaces (no tabs?).  Not sure which way to go with Java...

mck1117: our java should be all four-space

mck1117: .....because that's the intellij default

racer-coder: Ok updated my emacs rule for java-mode, fixed this function.

rusefillc:
.txt is the Model
.input/.ini is in the direction of View

if we want single source of truth it should be .txt not .input?

mck1117:
> .txt is the Model .input/.ini is in the direction of View
> 
> if we want single source of truth it should be .txt not .input?

Huh? This file doesn't touch the input at all?

racer-coder:
This doesn't change that - rusefi_config.txt is still the source of truth.  But rather than specifying the scale twice in the same line (once in autoscale<> and one in the field after the units), it only has to be specified once (the field after the units).

I do think we should have a less TS centric file format for specifying fields, and then generate TS ini from that... of course as you know I want a new GUI :-P  but I'm trying to pick my battles carefully.  I'm guessing replacing rusefi_config.txt with something else is probably not going to fly right now.

mck1117:
> I do think we should have a less TS centric file format for specifying fields, and then generate TS ini from that... of course as you know I want a new GUI :-P but I'm trying to pick my battles carefully. I'm guessing replacing rusefi_config.txt with something else is probably not going to fly right now.

you may notice that the new antlr-based parser does fully parse the file, then re-emits TS format, so we're actually not tightly bound to TS.

racer-coder:
So we parse the config file, reemit it, then parse it again (ConfigField, and friends?)  I must admit I was a bit confused when I found a proper parser sitting next to a regex parser and found I had to edit both to make it work.

mck1117:
No, the two parsers are airgapped.  The new one *can* emit the same headers/ts ini/etc that the currently used one does emit, but it isn't finished enough for that yet.  For now the proper parser is just used for stricter syntax checking (there was a lot of possible input that didn't cause a failure, but could emit garbage, at least now the new parser checks that the syntax is valid).

mck1117:
ps: https://rusefi.com/forum/viewtopic.php?t=1198

rusefillc:
my bad "TS config line" was the confusing part

rusefillc:
i am good with this PR as is

@mck1117 do you approve?

mck1117 (approved):

PR #3474, opened 2021-11-06
Move fuel density to a header file so it can be accessed by other components

racer-coder:


racer-coder:
Not sure where best to put this?  Seems a bit hacky.  On some ECU fuel density is user configurable... could go that route, but most people probably won't immediately know what to put here or that they need to put something there.

PR #3475, opened 2021-11-06
Configuration for HPFP control algorithm.

racer-coder:
Remove the old algorithm so the diff that includes the new algorithm doesn't intermix unrelated code.

The algorithm is the same one as before, but the configuration is separated out as requested.


rusefillc:
Seems be to causing #3481

PR #3479, opened 2021-11-06
TS parser improvements, compatibility

mck1117:
- handle autoscale the same as in #3470
- match tooltip behavior (don't require + prefix)
- handle nested definitions properly (signature in ini file)
- register signature hash with new parser


rusefillc:
@mck1117 please rebase and fresh .jar probably?

PR #3480, opened 2021-11-06
simplify signature.h generation

mck1117:
Remove unused fields from the generated signature header.  Just generate a single string, the actual signature.  Simplifies parsing for the new parser implementation and reduces churn a little.


firmware/gen_signature.sh
```diff
@@ -10,12 +10,6 @@ echo "Generating signature for ${SHORT_BOARDNAME}"
 
 # generate a unique signature
 date=`TZ=Europe/London date +"%Y.%m.%d"`
-echo "// Generated by gen_signature.sh" > ${SIGNATURE_FILE_NAME}
-echo "#define SIGNATURE_DATE $date" >> ${SIGNATURE_FILE_NAME}
-echo "#define SIGNATURE_BOARD ${SHORT_BOARDNAME}" >> ${SIGNATURE_FILE_NAME}
-
-echo "// SIGNATURE_HASH is a built-in variable generated by ConfigDefinition.jar" >> ${SIGNATURE_FILE_NAME}
```
rusefillc, line 17: I am attached to this line to be honest

my problem is I do not remember all the gears in our monster of mechanism. I need things self-documenting

What's the big picture here, what is driving the change?

I am happy to throw away _date and _board if unused

rusefillc (changes requested):

firmware/gen_signature.sh
mck1117: Neither `_date` nor `_board` exist anywhere in the firmware other than in this script, and in the generated headers.

rusefillc:
So step one https://github.com/rusefi/rusefi/commit/03107424d7378289b1bcb1d10d9141430d596ec8

mck1117:
I did also test this locally that the final string emitted in to the tunerstudio ini and `signature_xxx.h` header match what they used to before the change.

rusefillc:
step two https://github.com/rusefi/rusefi/commit/e9e88ff763a276422b1fe754cb2eab3b6791a32f

this way we leave all the comments as-is while killing complexity. Would that be acceptable?

mck1117:
comments are ok, they just have to match the format in rusefi_config.txt

PR #3482, opened 2021-11-06
Fix capitalization of subMenu for TS

### LINKED ISSUE #3481: TS project is not happy
image

@racer-coder to be honest I do not know what is wrong exactly but it's next to hpfp changes from yday

racer-coder:
This fixes #3481


rusefillc:
I've poked TS people to check if they would be interested to relax the syntax

PR #3495, opened 2021-11-07
Don't try to fire any timers shortly after a trigger edge.

racer-coder:
Instead fire them based on the previous edge.


racer-coder:
This won't pass unit tests.  No I didn't test this on real hardware and I have no numbers to back up my idea.

The idea is that we're busy right after receiving a tooth edge, and maybe events that need to be fired right on a tooth edge won't make it exactly on time.  So, instead handle those events relative to the previous tooth.

Maybe this doesn't matter much?  4usec at 6000 RPM is only 0.14 degrees.  And it only affects items scheduled within 0.14 degrees of a tooth.... but it's one of those insidious errors that you won't notice most of the time unless you look hard for it.

Yes I realize there are bigger fish to fry for jitter.  This just jumped out at me though as one that's easy to address.

mck1117:
+1 for bigger fish to fry

Due to our interrupt priorities, this probably doesn't actually make a difference.

rusefillc:
https://github.com/rusefi/rusefi/commit/656b8c3429c960ba5793b27cf25431c5920beb74

we have a bug here in case of two stroke do we?

Issue #2123, opened 2020-12-23
Wrap PAL driver into universal 'startDigitalCapture'


rusefillc:
Need a proper implementation for

```
#if HAL_VSS_USE_PAL
	ioline_t pal_line = PAL_LINE(getHwPort("vss", CONFIG(vehicleSpeedSensorInputPin)), getHwPin("vss", CONFIG(vehicleSpeedSensorInputPin)));
	efiExtiEnablePin("VSS", CONFIG(vehicleSpeedSensorInputPin), PAL_EVENT_MODE_BOTH_EDGES, vsExtiCallback, (void *)pal_line);
#elif HAL_USE_ICU
	digital_input_s* vehicleSpeedInput = startDigitalCapture("VSS", CONFIG(vehicleSpeedSensorInputPin));

	vehicleSpeedInput->widthListeners.registerCallback((VoidInt) vsAnaWidthCallback, NULL);
#else
	#error "HAL_USE_ICU or HAL_VSS_USE_PAL should be enabled to use EFI_VEHICLE_SPEED"
#endif /* HAL_VSS_USE_PAL, HAL_USE_ICU */
```

mck1117:
this is already done for trigger - it just needs to get pulled up a level

rusefillc:
digital_input_s should be abstract helper class regardless of chosen stm32 PAL=exti OR ICU timer hardware

the idea is the we have two copies of startDigitalCapture() depending on chosen hardware/compilation options, with higher level business logic calling same method and compiler picking chosen hardware implementation based on compilation settings

i.e. business logic does not have conditional compilation but actual startDigitalCapture implementation is compiled conditionally

mck1117:
Or startDigitalCapture goes down the list of possible capture options and sees if a pin can do each. Missing a type of capture, doesn't check it.

rusefillc:
@racer-coder I've mentioned this one on the phone

rusefillc:
Things have changed

PR #3497, opened 2021-11-07
fast exti interrupt handoff

mck1117:
- fast interrupt grabs timestamp
- hands off to slow interrupt
- slow interrupt calls trigger decoder

Less jitter! More faster! https://rusefi.com/forum/viewtopic.php?p=38501#p38501


mck1117:
Here's an example of the improvement:

An external STM32 is generating a 60-2 trigger pattern at ~5991 rpm (yes it's a strange number, no don't worry about it).

The DUT is connected with USB, logging to TS at 50hz.

Top is current master.  Bottom is with this PR.  X and Y axis scales are the same.

image

There is some aliasing visible in both (but masked somewhat by the noise in the first case!) with a period of ~7 seconds in both cases.  Not sure exactly what that is.

mck1117:
....and that was on a debug build.

dron0gus:
Is there any benefits of executing callbacks in IRQ context?
Why not ChibiOS mailboxes/events and bottom half in thread context?

mck1117:
> Is there any benefits of executing callbacks in IRQ context? Why not ChibiOS mailboxes/events and bottom half in thread context?

Because these interrupts are higher priority than the kernel itself, you can't call any OS functions.  `chSysLock()` doesn't disable these interrupts like it does normal ones!

dron0gus:
> > Is there any benefits of executing callbacks in IRQ context? Why not ChibiOS mailboxes/events and bottom half in thread context?
> 
> Because these interrupts are higher priority than the kernel itself, you can't call any OS functions. `chSysLock()` doesn't disable these interrupts like it does normal ones!

Question was about calling callback from `CH_IRQ_HANDLER(STM32_I2C1_EVENT_HANDLER)`. As it has IRQ OS prologue and epilogue I assume its priority is lower than OS's.
BTW, PENDSV can be used to delegate some actions to lower priority IRQ. Instead of occupying i2c interrupt.

mck1117:
> Question was about calling callback from `CH_IRQ_HANDLER(STM32_I2C1_EVENT_HANDLER)`. As it has IRQ OS prologue and epilogue I assume its priority is lower than OS's.

correct - `CH_IRQ_HANDLER(STM32_I2C1_EVENT_HANDLER)` in this PR has the same priority and behavior as the trigger input interrupts do today without this change.

firmware/controllers/engine_controller_misc.cpp
```diff
@@ -67,7 +67,8 @@ volatile uint32_t lastLowerNt = 0;
 volatile uint32_t upperTimeNt = 0;
 
 efitick_t getTimeNowNt() {
-	chibios_rt::CriticalSectionLocker csl;
+	// TODO: lock free so we don't have to disable interrupts!
+	__disable_irq();
```
racer-coder, line 71: May want to make this safe in case interrupts are already disabled when this function is called.  Or maybe I should go commit my lockfree version.

firmware/hw_layer/digital_input/digital_input_exti.cpp
```diff
@@ -103,9 +116,88 @@ digital_input_s* startDigitalCapture(const char *msg, brain_pin_e brainPin) {
...
+
+CH_IRQ_HANDLER(STM32_I2C1_EVENT_HANDLER) {
+	OSAL_IRQ_PROLOGUE();
+
+	for (size_t i = 0; i < 16; i++) {
+		auto& channel = channels[i];
+
+		if (channel.Timestamp != 0 && channel.Callback) {
+			channel.Callback(channel.CallbackData, channel.Timestamp);
+			channel.Timestamp = 0;
```
racer-coder, line 132: not safe if an event on the same channel comes in at the same time.

Now you're making me want to make timestamps 32-bit, though then they couldn't represent 0 anyways.
I think it's safer if you have a separate byte array (or byte field) representing which channels are active.  It's safer than this code, without being as complicated as using a uint16_t bit field and ldrex/strex to manipulate it.

firmware/hw_layer/digital_input/trigger/trigger_input_exti.cpp
```diff
@@ -24,9 +24,8 @@
 static ioline_t shaftLines[TRIGGER_SUPPORTED_CHANNELS];
 static ioline_t camLines[CAM_INPUTS_COUNT];
 
-static void shaft_callback(void *arg) {
+static void shaft_callback(void *arg, efitick_t stamp) {
```
racer-coder, line 27: Why does this file exists at all?  Seems like the API for using EXTI or ICU should be the same; why should there be trigger specific code for exti vs others?

firmware/controllers/engine_controller_misc.cpp
mck1117: +1 vote for lock free version :heart:

firmware/hw_layer/digital_input/trigger/trigger_input_exti.cpp
mck1117: if you want something to do: https://github.com/rusefi/rusefi/issues/2123

mck1117:
needs https://github.com/rusefi/ChibiOS/pull/18

firmware/hw_layer/digital_input/digital_input_exti.cpp
mck1117: Well, it's at least safe now, if not particularly pretty.  We're locked for only 4 instructions.

```
cpsid i
ldrd r2, r3, [r4, #16]      // load timestamp in to r2, r3
strd r6, r7, [r4, #16]      // zero timestamp field
cpsie i

```

rusefillc:
What's the plan for this amazing PR?

mck1117:
bench testing looked good to me, I think we're probably set

rusefillc:
@mck1117 merge this BEFORE or AFTER release?

mck1117:
this should wait until post-release

rusefillc:
shall I merge?

mck1117:
pull the lever!

PR #3505, opened 2021-11-08
Compile each project's PCH to a separate file to reduce developer friction

racer-coder:
Having to remove the precompiled header when switching from firmware to simulator to unit_tests and
back again is annoying.  Use a directory to store the precompiled header output.  Turn off
precompiled header warnings as GCC still complains even if it finds a valid one eventually.

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -18,6 +18,7 @@ PROJECT_DIR = .
 # Configure precompiled header
 PCH_DIR = $(PROJECT_DIR)/pch
 PCHSRC = $(PCH_DIR)/pch.h
+PCHSUB = firmware
 
 # Imported source files and paths
 CHIBIOS_CONTRIB = ChibiOS-Contrib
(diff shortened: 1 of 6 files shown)


racer-coder:
Sneak in a fix for windows srand vs srandom.

firmware/Makefile
```diff
@@ -61,7 +62,7 @@ endif
 
 # C++ specific options here (added to USE_OPT).
 ifeq ($(USE_CPPOPT),)
-  USE_CPPOPT = -std=c++17 -Wno-register -fno-rtti -fno-threadsafe-statics -fno-exceptions -fno-use-cxa-atexit -Winvalid-pch
```
mck1117, line 64: we should still include this option, right?

racer-coder: No, gcc will just spam you even if it eventually finds a valid pch file.  Stupid design IMO.... so I guess we just will have to rely on developers complaining if their builds get slow...

rusefillc:
great news windows CI is happy now!

in other news I am unable to compile locally now
I've removed pch file
and I am getting
```
Compiling stm32f4xx_hal_flash_ex.c
Compiling main_hardfault.c
make: *** No rule to make target 'pch.h', needed by 'pch/pch.h.gch/firmware'.  Stop.
make: *** Waiting for unfinished jobs....
"make -j12 all" terminated with exit code 2. Build might be incomplete.
```

I've already tried ``make clean`` and ``git clean -f -X``

I assume that we will have a couple of more contributors affected by this so let's post instructions at https://rusefi.com/forum/viewtopic.php?f=5&t=373

rusefillc:
my bad I've removed too much locally, looks good now!

PR #3506, opened 2021-11-08
Save ~2800 bytes of RAM by not caching angle->trigger mapping

racer-coder:
Memory seems more valuable than CPU; use the O(lg n) lookup by angle.  A side effect is that it
seems to fix trigger lookup for the second phase of TRIGGERTYPE 53 544 TT_TRI _TACH 0.00


mck1117:
sweeeeet

firmware/controllers/trigger/decoders/trigger_structure.cpp
```diff
@@ -62,7 +62,6 @@ TriggerWaveform::TriggerWaveform() :
 }
 
 TriggerFormDetails::TriggerFormDetails() {
```
mck1117, line 64: explicit constructor no longer needed if empty

racer-coder: fixed

racer-coder:
sorry this only saves 1440 bytes.  I thought it was double because of the way I was finding the large users.

rusefillc:
I am hesitant because we are flying blind without an explicit performance test and explicit performance requirement

Can I pull the "what about" and create a couple of tickets for the skeletons I am aware of?

rusefillc:
#3508 would be the first one

mck1117:
> I am hesitant because we are flying blind without an explicit performance test

This netted less change than we can measure with perf tracing. Seriously.

Testing with an externally generated 60-2 at ~6k RPM.

Whole interrupt takes ~22us.  `updateInstantRpm` claims to take ~3us in either case, which is low enough that the perf trace overhead becomes significant.

TIME FOR THE LOGIC ANALYZER AND BIT TWIDDLING

I set/reset an output pin when we enter/leave `updateInstantRpm`.

Before this PR: 1.46 us
After this PR: 2.66 us

Adds ~1.2us, about 200 CPU cycles on an F4 at 168MHz.

And this was on a 60-2, just about our worst case common scenario for trigger size.  The cost grows with O(log(n)) of the trigger size, and in cases like this, O(log(n)) rounds to O(1).

mck1117 (approved):

racer-coder:
Each iteration of the bisection was 11 instructions.  I'm not sure what the cycle count on them is.  I couldn't get it much lower without making the code very strange.  There are some weird things in the function, like calculating the # of buckets to check in floating point with a bunch of conditionals, only to cast it to an int later.  We could always cache that.

Also changing the lookup to ints would shave an instruction - floating point comparisons have one more instruction than integer comparisons.  Again, I'm not sure the cycle impact of that.

And we can hide the 1.2uS with #3495 :-)

mck1117:
Not sure we need to cut down the cycle count any more.  11/iteration is plenty good, especially with the function still legible.  1us per trigger tooth is a tiny fish to fry at this point.  Especially since it saves ~1.5k of memory.

PR #3514, opened 2021-11-09
Remove the EFI_NO_CONFIG_WORKING_COPY option - it is now the default.

racer-coder:
EDIT:  New goal:
    Any change entered via TunerStudio will now be applied immediately (modulo activeConfiguration
    stuff - TBD).  This makes for a more consistent user experience.  It also save 20,000 bytes of BSS.

### DIFF
--- a/firmware/console/binary/tunerstudio.cpp
+++ b/firmware/console/binary/tunerstudio.cpp
@@ -96,15 +96,6 @@
 
 extern persistent_config_container_s persistentState;
 
-#if !defined(EFI_NO_CONFIG_WORKING_COPY)
-/**
- * this is a local copy of the configuration. Any changes to this copy
- * have no effect until this copy is explicitly propagated to the main working copy
- */
-persistent_config_s configWorkingCopy;
-
-#endif /* EFI_NO_CONFIG_WORKING_COPY */
-
 static efitimems_t previousWriteReportMs = 0;
 
 static void resetTs(void) {
@@ -167,12 +158,8 @@ void tunerStudioDebug(TsChannelBase* tsChannel, const char *msg) {
 #endif /* EFI_TUNER_STUDIO_VERBOSE */
 }
 
-char *getWorkingPageAddr() {
-#ifndef EFI_NO_CONFIG_WORKING_COPY
-	return (char*) &configWorkingCopy.engineConfiguration;
-#else
-	return (char*) engineConfiguration;
-#endif /* EFI_NO_CONFIG_WORKING_COPY */
+static uint8_t *getWorkingPageAddr() {
+	return (uint8_t *) engineConfiguration;
 }
 
 static constexpr size_t getTunerStudioPageSize() {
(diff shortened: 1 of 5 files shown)


rusefillc:
amazing stuff!

firmware/console/binary/tunerstudio.cpp
```diff
@@ -101,7 +101,7 @@ extern persistent_config_container_s persistentState;
  * this is a local copy of the configuration. Any changes to this copy
  * have no effect until this copy is explicitly propagated to the main working copy
  */
-persistent_config_s configWorkingCopy;
+engine_configuration_s engineConfigWorkingCopy;
```
rusefillc, line 104: Can we move PR comment into codebase to a larger extent?

overall my point is that PR comments are often misplaced code comments. taken to extreme in general PR description could be tiny while whatever we have to say is best place in the code

rusefillc (changes requested):

firmware/console/binary/tunerstudio.cpp
racer-coder: PR comment describes the delta, which is usually only/most interesting at commit time.

rusefillc: i believe following could be added into code comments:

"scatter/gather overhead so reads and writes can span both portions"
"this XXX holds a copy while YYY does ZZZ"

mck1117: @rusefillc PR has changed a bit after some discussion on Slack

rusefillc: @mck1117 need your help - no brain available today for this PR. Relying on your approval.

mck1117: I want to do some testing in actual hardware, will do tonight.

racer-coder:
Updated version cleans up onlineApply... function.  I would like to leverage that to get rid of the notion of Burn.  Down the road we can repurpose "Burn" to mean actually flashing rather than "making some of the config live even though other parts of the config are already live"

firmware/console/binary/tunerstudio.h
```diff
@@ -42,8 +42,6 @@ bool handlePlainCommand(TsChannelBase* tsChannel, uint8_t command);
  */
 void handleQueryCommand(TsChannelBase* tsChannel, ts_response_format_e mode);
 
-char *getWorkingPageAddr();
```
rusefillc, line 45: if it's not in header shall we make it static to emphasize?

rusefillc:
Hold on, this is taking an unexpected turn, I wonder if applying settings on the fly is pretty much a no-no. We fatal on invalid settings, would not this change cause many settings to be invalid if done atomically one by one?

For instance you want to flip pins. If you apply settings right away you get a pin conflict on first step?

mck1117:
> 
> 
> Hold on, this is taking an unexpected turn, I wonder if applying settings on the fly is pretty much a no-no. We fatal on invalid settings, would not this change cause many settings to be invalid if done atomically one by one?
> 
> For instance you want to flip pins. If you apply settings right away you get a pin conflict on first step?

this doesn't change the behavior for `activeConfiguration` vs `engineConfiguration` - just how they're written in to `engineConfiguration` in the first place.  We still wait for the burn button to perform pin change callbacks and things.

rusefillc:
PS: so the proper way forward would be to add integration/unit test of the scenario I am trying to articulate as a prerequisite PR and then rebase this PR on master once it has additional coverage.

rusefillc:
So how do we handle "oh let's me disconnect TS without burning"? And TS re-connect after the fact?

Would TS read uncommited state as truth and save it into PC config?

racer-coder:
> Hold on, this is taking an unexpected turn, I wonder if applying settings on the fly is pretty much a no-no. We fatal on invalid settings, would not this change cause many settings to be invalid if done atomically one by one?
> 
> For instance you want to flip pins. If you apply settings right away you get a pin conflict on first step?

Maybe the code shouldn't fatal then, just put the engine in a "no run" state and wait for a valid config.

Normal engine management systems apply changes right away, unless it's a reboot-only change.  Burn shouldn't mean "use this," it should mean "save to flash."

racer-coder:
> > Hold on, this is taking an unexpected turn, I wonder if applying settings on the fly is pretty much a no-no. We fatal on invalid settings, would not this change cause many settings to be invalid if done atomically one by one?
> > For instance you want to flip pins. If you apply settings right away you get a pin conflict on first step?
> 
> this doesn't change the behavior for `activeConfiguration` vs `engineConfiguration` - just how they're written in to `engineConfiguration` in the first place. We still wait for the burn button to perform pin change callbacks and things.

I'm changing that next!  Another 6000 bytes.  I'm making a delta structure that listeners can use to read the old values (using template and #define magic).

rusefillc:
This is stressing me out. This PR is too big unexpectedly.

rusefillc:
We have no automated coverage for all this. I am not available to test this all manually. This PR can only go forward as a second PR after test coverage already added into master. See coverage report.

firmware/console/binary/tunerstudio.h
racer-coder: fixed

mck1117:
> This is stressing me out. This PR is too big unexpectedly.

calm down a little bit, as discussed, I'm going to try and test this exhaustively tonight.

There are three categories of config:

1. tables, etc that exist outside of engineConfiguration
2. mostly scalar parameters, a few tables that are inside engineConfiguration
3. stuff that uses activeConfiguration, pinouts, etc

For each of the above, this PR has the following effect:

1. none
2. These parameters apply when you type them in, not when you press burn (afaik, to test on real hw tonight)
3. none

racer-coder:
> Hold on, this is taking an unexpected turn, I wonder if applying settings on the fly is pretty much a no-no. We fatal on invalid settings, would not this change cause many settings to be invalid if done atomically one by one?
> 
> For instance you want to flip pins. If you apply settings right away you get a pin conflict on first step?

I finally have real hardware to try out.  Today, if I flip pins illegally and press Burn, all I get is a firmwareError.  It doesn't crash the box at all.  You can correct the scenario by changing the pin back, but the error doesn't clear until you reset the box.  This is a separate issue IMO - you should be able to clear errors without resetting the MCU.

So I don't see what the worry is at all - either we can handle invalid settings without crashing or we can't - if we can't that needs to be fixed whether settings happen right away or not.

rusefillc:
let's keep this open as source of inspiration

mck1117:
#3630

PR #3518, opened 2021-11-10
Track ram4 sections too.

racer-coder:
I don't know why I'm having trouble compiling this, but I think gradle itself is crashing.


rusefillc:
@racer-coder I've pushed fresh jar

PR #3519, opened 2021-11-10
Drop SingleChannelStateSequence in favor of MultiChannelStateSequenceWithData

racer-coder:
Most all the users were Multi* anyways, so just improve that:
1. Allow up to 8 waveforms to share one byte per timestamp.  It could be better but this is simple
   and gets most of the benefit.
2. Use a wrapper structure to handle reserving space for the arrays.  Makes the interface simpler
   and more rigid.  Also saves 4 bytes per Multi*.  Downside is access is now via -> and *,
   not . and (nothing).

Saves 224 bytes of BSS, 1832 bytes of RAM4/CCM, 824 bytes of TEXT (yay? but unexpected...), and
costs 26 bytes of RODATA (more yay?)

### DIFF
--- a/firmware/controllers/core/state_sequence.cpp
+++ b/firmware/controllers/core/state_sequence.cpp
@@ -9,56 +9,24 @@
 #include "state_sequence.h"
 #include "trigger_structure.h"
 
-MultiChannelStateSequence::MultiChannelStateSequence(unsigned maxWaveCount)
-	: wavePtr(reinterpret_cast<uint8_t *>(&switchTimes[maxWaveCount])) {
-	reset();
-}
-
-void MultiChannelStateSequence::reset(void) {
-	waveCount = 0;
-}
-
-float MultiChannelStateSequence::getSwitchTime(const int index) const {
-	return switchTimes[index];
-}
-
 void MultiChannelStateSequence::checkSwitchTimes(const float scale) const {
-	if (switchTimes[phaseCount - 1] != 1) {
-		firmwareError(CUSTOM_ERR_WAVE_1, "last switch time has to be 1/%f not %.2f/%f", scale,
-				switchTimes[phaseCount - 1], scale * switchTimes[phaseCount - 1]);
+	if (getSwitchTime(phaseCount - 1) != 1) {
+		firmwareError(CUSTOM_ERR_WAVE_1, "last switch time has to be 1/%f not %.2f/%f",
+			      scale, getSwitchTime(phaseCount - 1),
+			      scale * getSwitchTime(phaseCount - 1));
 		return;
 	}
 	for (int i = 0; i < phaseCount - 1; i++) {
-		if (switchTimes[i] >= switchTimes[i + 1]) {
-			firmwareError(CUSTOM_ERR_WAVE_2, "invalid switchTimes @%d: %.2f/%.2f", i, switchTimes[i], switchTimes[i + 1]);
+		if (getSwitchTime(i) >= getSwitchTime(i + 1)) {
+			firmwareError(CUSTOM_ERR_WAVE_2, "invalid switchTimes @%d: %.2f/%.2f",
+				      i, getSwitchTime(i), getSwitchTime(i + 1));
 		}
 	}
 }
 
-pin_state_t MultiChannelStateSequence::getChannelState(const int channelIndex, const int phaseIndex) const {
-	if (channelIndex >= waveCount) {
-		// todo: would be nice to get this asserting working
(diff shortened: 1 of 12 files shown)


racer-coder:
draft as I want to clean up the api to StaticAllocWrapper

mck1117:
any way we can break this down in to smaller atomic chunks so it's more digestible?

firmware/controllers/trigger/trigger_emulator_algo.cpp
```diff
@@ -106,7 +106,7 @@ static void updateTriggerWaveformIfNeeded(PwmConfig *state DECLARE_ENGINE_PARAME
 
 
 		TriggerWaveform *s = &engine->triggerCentral.triggerShape;
-		copyPwmParameters(state, s->getSize(), &s->wave);
+		copyPwmParameters(state, &*s->wave);
```
mck1117, line 109: address-of-dereference-of?

racer-coder: s->wave is a MultiChannelStateSequenceWithData, which is really just a uint8_t[some_size];
It does not derive from MultiChannelStateSequence.  So in order to access the Multi*Sequence, there are four overloaded operators: * and ->, both const and non-const versions.

So really the * is saying "give a reference the object you're managing" and & is saying "give me a pointer to it", since there is no operator that says give me the pointer.

Though maybe I could add
operator MultiChannelStateSequenceWithData*()
?

racer-coder:
> any way we can break this down in to smaller atomic chunks so it's more digestible?

Well I'm not sure it's much shorter, but really the bulk of the changes are in state_sequence.*.  The rest is mostly . to -> or extra * or &* to make it compile.  Plus a couple call sites that now use the new structure, but even that's only a couple lines.

racer-coder:
I'm dropping this review.  I'm going all in - gonna make triggers live in ROM.  No need for user-space SingleChannelStateSequence except for SimplePWM.  I'll move switchTimes into SingleChannel and take advantage of, as Matt pointed out, that it should just go back and forth.  So only store times, not states.  Then MultiChannel becomes an iterator wrapper over multiple SingleChannel - or maybe just make the user make separate outputters per pin - why even group them together?

rusefillc:
@racer-coder poke :)

racer-coder:
> 
> 
> @racer-coder poke :)
> 
> What's the plan for this PR?

It doesn't work on real hardware and I have no idea why.  My Nucleo H753 shows up on Wednesday :-)

rusefillc:
So hope to revisit once real hardware in hand?

racer-coder:
> So hope to revisit once real hardware in hand?

Yeah definitely.

racer-coder:
ready for review

firmware/controllers/core/state_sequence.cpp
```diff
@@ -9,42 +9,11 @@
...
-}
-
-void SingleChannelStateSequence::setState(int switchIndex, pin_state_t state) {
-	efiAssertVoid(OBD_PCM_Processor_Fault, state == 0 || state == 1, "wave state set");
-	pinStates[switchIndex] = state;
-}
-
-MultiChannelStateSequence::MultiChannelStateSequence() {
+MultiChannelStateSequence::MultiChannelStateSequence(unsigned maxWaveCount)
+	: wavePtr(reinterpret_cast<uint8_t *>(&switchTimes[maxWaveCount])) {
```
mck1117, line 13: possible to avoid `reinterpret_cast`?

racer-coder: I could make a union between switchTimes and waves, but in the end you need to know to skip maxWaveCount floats, not bytes, so you're going to have some dodgy pointer math somewhere.  I thought reinterpret_cast would be the lesser of the two evils.

firmware/controllers/core/state_sequence.h
```diff
@@ -35,42 +37,65 @@
...
- * Each element represents either a HIGH or LOW state - while at the moment this
- * is not implemented using a bit array, it could absolutely be a bit array
- *
- * This sequence does not know anything about signal lengths - only signal state at a given index
- * This sequence can have consecutive zeros and ones since these sequences work as a group within MultiChannelStateSequence
- *
- * @brief   PWM configuration for the specific output pin
- */
-class SingleChannelStateSequence {
+template<typename base_t, unsigned n_elem, typename... tail_t>
```
rusefillc, line 40: please note how you remove 6 lines of code comments while adding how many?

we need some basic code comments.

rusefillc (changes requested):

PR #3524, opened 2021-11-11
Change users of MultiChannelStateSequence to use the API

racer-coder:
Better than reaching into members that should be private.  I didn't feel like actually making them private though, as one user validates pinStates isn't NULL.


racer-coder:
I am going insane.  Can someone help me find out why this fails on real hardware?

PR #3527, opened 2021-11-11
Print BSS and RAM4 totals at the end of make.

racer-coder:
The compiler was outsmarting us: UNUSED_CCM disappeared entirely (probably because of the *0),
while UNUSED_RAM was helpfully moved to read-only flash.  It's possible making these not static
would have helped, but link time optimization might have gotten rid of them too.

Instead, print BSS and RAM4 totals at the end using the java script.  If java is unable to run,
don't error out the Makefile, just complain a little and move on.  Until the gcc_map_reader
actually works with RAM4, we print it manually.

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -409,3 +409,7 @@ openblt_clean:
 
 # Enable precompiled header
 include rusefi_pch.mk
+
+POST_MAKE_ALL_RULE_HOOK: $(BUILDDIR)/$(PROJECT).elf
+	@java -jar ../java_tools/gcc_map_reader.jar $(BUILDDIR)/$(PROJECT).map | grep Total || echo Unable to run gcc_map_reader
+	@$(TRGT)objdump -h $(BUILDDIR)/$(PROJECT).elf | grep -w ram4


rusefillc:
I am not convinced at all

We know from experience that https://github.com/rusefi/rusefi/blob/master/firmware/config/boards/microrusefi/compile_mre_f4_hardware_QC_special_build.sh fails very violently when UNUSED_RAM is violated

rusefillc:
let's try https://github.com/rusefi/rusefi/commit/1564f8aa025aa80c9c063016362d43a2cd5816c9

mck1117:
part of the problem is that there are parts of chibios itself that get removed by the link gc option (aka throw away things we never touch), which only happens on optimized builds

racer-coder:
~/src/rusefi/firmware$ make PROJECT_BOARD=microrusefi PROJECT_CPU=ARCH_STM32F4 -j4 > /dev/null
~/src/rusefi/firmware$ grep UNUSED build/rusefi.map
 .rodata._ZL15UNUSED_RAM_SIZE.lto_priv.0
                0x0000000008074694                _ZL15UNUSED_RAM_SIZE.lto_priv.0
_ZL15UNUSED_RAM_SIZE.lto_priv.0                   /tmp/rusefi.elf.iyHuwX.ltrans8.ltrans.o


rodata = read only data
maybe I'm using a different version of gcc? I just grabbed the latest.
gcc version 10.3.1 20210824 (release) (GNU Arm Embedded Toolchain 10.3-2021.10)

rusefillc:
image

racer-coder:
New commit comment:

If java is unable to run, don't error out the Makefile, just complain a little and move on.  Until
the gcc_map_reader actually works with RAM4, we print it manually.

PR #3529, opened 2021-11-11
load lua libs correctly

### LINKED ISSUE #3521: how do we use lua built in library functions
72a2bdba66a0cdfa3b3bea6d4f08f90b01e9c0d0 doesn't work, why?

mck1117:
fix #3521


rusefillc:
Thank you @mck1117

PR #3434, opened 2021-10-31
fresh lua

rusefillc:
```
git config --file=.gitmodules submodule.firmware/controllers/lua/luaaa.branch lua-2021-10-28
git submodule sync
git submodule update --init --remote
```

PR #3540, opened 2021-11-13
update luaaa

mck1117:


rusefillc:
note how https://github.com/rusefi/rusefi/pull/3434 changes .gitmodules while this PR does not

mck1117:
Had to turn off system lua since we don't use it and it makes luaaa saaaaad.

Issue #3537, opened 2021-11-12
Improve Vehicle Speed Sensor Tool Tips


droneperson42:
Under Sensors > Speed Sensor, the tool tips for "revolution to speed mult(coef)" and "Filter Reciprocal (Hz)" aren't that useful in helping users understand what they do and what values should be put in to make it work. I'm not sure exactly what they do but could be along the lines of:

"revolution to speed mult(coef)": Multiplier for converting wheel speed (Hz/rotations per second) to Km/h. Calculated by measuring wheel radius and substituting into this formula .... (Hz > Km/h: [2*pi*wheel_radius*3.6], or of it/s Km/h>Hz: 1/[2*pi*wheel_radius*3.6])

"Filter Reciprocal (Hz)": Number of teeth on (ABS ring/gearbox speed sensor, I'm not sure). Calculates wheel cycles/s or Hz based on how many counts it should expect to read per revolution.

thirstyone:
not sure if that's at all possible, but it would be nice to click a button in UI, move a vehicle 100 meters forward, click a button once again and have those coefficients calculated.

BeerMoneyMotorsports:
@mck1117 you have another issue for this?

mck1117:
this = ?

BeerMoneyMotorsports:
Fixing the VSS configuration parameters to be simpler and easier to comprehend.  Not the tool tips specifically.  There was Slack chatter about it which you were involved with.  You were changing to something like entering wheel diameter, gear ratio, etc and having math do it in the background?

rusefillc:
@BeerMoneyMotorsports yes

![image](https://user-images.githubusercontent.com/48498823/141871297-ff2021b9-8eac-4643-8843-8411bd1ac586.png)

PR #3542, opened 2021-11-13
better VSS configuration

mck1117:
Instead of a mystery ratio that users have to calculate, input things they can actually measure or count: wheel turns per km, axle ratio, and vss tooth count.

first part of #3537


rusefillc:
@mck1117 poke :)

rusefillc:
HW CI red?

mck1117:
this damn near killed me, the offset for the VSS sensor was WRONG!

mck1117:
oh my god it passed we're green

rusefillc:
Code generator on the way

PR #3548, opened 2021-11-15
Create a base class EngineModule for that contains various useful callbacks.

racer-coder:
This cleans up the API by not requiring the notifiers to know about who wants callbacks.  The only
place you need to update to add a module is in one place.


firmware/controllers/actuators/idle_thread.h
```diff
@@ -58,6 +58,9 @@ class IdleController : public IIdleController {
 	// CLOSED LOOP CORRECTION
 	float getClosedLoop(IIdleController::Phase phase, float tpsPos, int rpm, int targetRpm) override;
 
+	void onConfigurationChange(engine_configuration_s const * previousConfig) final;
```
mck1117, line 61: `override`?

mck1117:
It seems this misses a case that we use those callbacks for: they were pointers so you can inject a mock at unit test time, then check that the values from the mock are used correctly by other systems (for example, does idle controller timing adjustment get used in timing math?).

firmware/util/containers/type_list.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
+
+template<typename base_t, typename... tail_t>
+struct type_list {
```
mck1117, line 4: smells like `std::variant` 🙂

firmware/util/containers/type_list.h
```diff
@@ -0,0 +1,52 @@
...
+		      "Each type can only be listed once.");
+
+	template<typename func_t>
+	void apply_all(func_t const & f) {
+		mine.apply_all(f);
+		others.apply_all(f);
+	}
+
+	template<typename get_t>
+	auto get() -> typename std::enable_if<std::is_same<get_t, base_t>::value, get_t &>::type {
```
mck1117, line 18: we have c++14, we can use `std::enable_if_t` and `std::is_same_v`!

firmware/util/containers/type_list.h
racer-coder: If variant is a union, type_list is a struct.  Items are non overlapping.

mck1117: yep - this is `tuple` but with a built-in version of `std::apply`

firmware/controllers/actuators/idle_thread.h
racer-coder: eh, whatever.  Importantly, it needs to have one or the other to detect changes in the underlying API.  That's the only reason EngineModule needs virtual functions, otherwise we could rely on duck typing.

Because of the use of generic lambdas, it doesn't actually affect performance - the compiler knows the exact type and can call the functions directly withing using the vtable.

mck1117: Yes, I know that it will never actually hit the vtable.

Why does it need the base `EngineModule` at all? Why not rely on duck typing to enforce presence of any functions we try to call?

firmware/util/containers/type_list.h
```diff
@@ -0,0 +1,52 @@
...
+	template<typename has_t>
+	static constexpr bool has() {
+		return decltype(mine)::template has<has_t>() ||
+			decltype(others)::template has<has_t>();
+	}
+};
+
+template<typename base_t>
+struct type_list<base_t> {
+	base_t me;
```
mck1117, line 36: perhaps we could:
1. keep `base_t me;`
2. add `base_t* mePtr = &me;`
3. `apply_all` and `get` act on `mePtr`
4. add function `replace(base_t*)` that sets `mePtr` to the injected (mocked, perhaps) replacement

racer-coder: Yeah I think this can be made mockable.  Only for unit tests.  get<>() can hide the magic for unit tests.  Just need mock/replace method as you mention plus each replaceable class needs an interface_t defined inside it.

firmware/controllers/actuators/idle_thread.h
racer-coder: It's convenient to not have to declare dummies for each callback you don't care about.

firmware/util/containers/type_list.h
racer-coder: well, kinda - tuple allows duplicates, and apply calls your callback once with all the values as parameters.  type_list::apply_all calls your callback once for each parameter, so you don't have to detangle the parameters yourself.  Although maybe ... would handle that for you?

racer-coder: ok nevermind std::get does ensure type uniqueness if you get by type instead of index.  So I will remove a bunch of this and just use std::tuple; controllers can be wrapped in a Mockable<> which will have operator-> and a set method, and then have a custom apply method that makes the API nicer (unless I can make std::apply work well using ... somehow)

racer-coder: arg, with some things Mockable<> and some not, it becomes a total mess.  you need to do get<mockable<foo>> and duplicate detection doesn't see through the mockable.

So, back to a custom implementation.  Maybe I can lean more on tuple/get though and make simpler wrappers.

mck1117: this is simple enough that I don't think we need std tuple

rusefillc:
Master has changed beyong recognition. Not my fault!

rusefillc:
@racer-coder we have a code conflict

can you please rebase and I will merge right away?

mck1117:
@racer-coder can you include a case that calls `type_list<Mockable<base_t>>::set`?

racer-coder:
> @racer-coder can you include a case that calls `type_list<Mockable<base_t>>::set`?

Converted InjectorModel.  Saved 120 bytes of text, moved 12 bytes from data to Ram4 (used to be static InjectorModel, now embedded in Engine), also saved 4 bytes since we no longer have the IInjectorModel* anymore.  Did grow 12 bytes of rodata due to the added vtable entries from EngineModule.

PR #3551, opened 2021-11-15
Start the process of using a global engine ptr for unit tests.

racer-coder:
Passing around the engine, config, and persistent state wasn't buying anything.  Instead, use
EngineTestHelper(Base) to correctly set and clear the global variables.  Add a dummy check in
case some test tries to set them manually.


racer-coder:
Draft for obvious reasons (touchy subject?).  I could follow this up with a *ton* of cleanup but I don't want to pollute the diff too much.

rusefillc:
Why are we touching this?

What problem are we solving today?

I am not convinced that this is worth touching today.

mck1117:
> What problem are we solving today?

No more `PASS_ENGINE_PARAMETER_SIGNATURE` and friends would be nice.

unit_tests/tests/ignition_injection/test_fuel_map.cpp
```diff
@@ -152,32 +152,32 @@ TEST(misc, testAngleResolver) {
...
 	ASSERT_FLOAT_EQ(69.24, injectionStart.angleOffsetFromTriggerEvent);
 
 	printf("*************************************************** testAngleResolver 1\r\n");
-	findTriggerPosition(&ENGINE(triggerCentral.triggerShape), &ENGINE(triggerCentral.triggerFormDetails),&injectionStart, 0, engineConfiguration->globalTriggerAngleOffset);
+	findTriggerPosition(&ENGINE(triggerCentral.triggerShape), &ENGINE(triggerCentral.triggerFormDetails),&injectionStart, 0);
 	ASSERT_EQ(2, injectionStart.triggerEventIndex);
 	ASSERT_FLOAT_EQ(122.24, injectionStart.angleOffsetFromTriggerEvent);
 
 	printf("*************************************************** testAngleResolver 2\r\n");
-	findTriggerPosition(&ENGINE(triggerCentral.triggerShape), &ENGINE(triggerCentral.triggerFormDetails),&injectionStart, 56, engineConfiguration->globalTriggerAngleOffset);
```
rusefillc, line 180: is that "remove dead parameter" refactoring? is that a separate change?

racer-coder:
> Why are we touching this?
> 
> What problem are we solving today?

Working on the code has a lot of rough edges, this is one of them.  It's ugly, it's confusing until you get the hang of it, it's unnecessary and provides no benefit.  It falls under technical debt, IMO.

unit_tests/tests/ignition_injection/test_fuel_map.cpp
racer-coder: Apparently allow some configuration variables to be passed in as parameters rather than going through the global configuration.  Probably to avoid plumbing an entire configuration to findTriggerPosition?

It's related to the CONFIG_PARAM macro change - it used to mean "grab from config if production, otherwise grab from current scope (which probably means function parameter)."

rusefillc:
@racer-coder please rebase I believe master is less broken now

rusefillc:
@racer-coder merge conflicts it says

mck1117 (approved):

PR #3560, opened 2021-11-16
Programmatically remove EnginePtr/inject, EXPAND_Engine, and WITH_ENGINE_TEST_HELPER*

racer-coder:
for i in ": public EnginePtr " ", public EnginePtr" "EXPAND_Engine;" "EXPAND_Engine"; do
    git grep -l "$i" | xargs sed -i "s/$i//g"
done

git grep -l "inject" | xargs sed -i "/inject[(][)]/d"

for i in WITH_ENGINE_TEST_HELPER_SENS WITH_ENGINE_TEST_HELPER_BOARD_CALLBACK WITH_ENGINE_TEST_HELPER; do
    git grep -l "$i" | xargs sed -i "s/$i/EngineTestHelper eth/g"
done

git checkout firmware/controllers/core/engine_ptr.h
git checkout unit_tests/global.h

### DIFF
--- a/firmware/controllers/algo/engine.cpp
+++ b/firmware/controllers/algo/engine.cpp
@@ -462,14 +462,10 @@ void Engine::OnTriggerSyncronization(bool wasSynchronized) {
 #endif
 
 void Engine::injectEngineReferences() {
-	for (int camIndex = 0;camIndex < CAMS_PER_BANK;camIndex++) {
-	}
-
 	primaryTriggerConfiguration.update();
 	for (int camIndex = 0;camIndex < CAMS_PER_BANK;camIndex++) {
 		vvtTriggerConfiguration[camIndex].update();
 	}
-	triggerCentral.init();
 }
 
 void Engine::setConfig() {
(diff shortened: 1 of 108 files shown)


firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -52,10 +52,8 @@ TriggerCentral::TriggerCentral() : trigger_central_s(),
 }
 
 void TriggerCentral::init() {
```
mck1117, line 54: empty function

firmware/controllers/algo/engine.cpp
```diff
@@ -463,13 +462,8 @@ void Engine::OnTriggerSyncronization(bool wasSynchronized) {
 #endif
 
 void Engine::injectEngineReferences() {
-	triggerCentral.inject();
-	primaryTriggerConfiguration.inject();
 	for (int camIndex = 0;camIndex < CAMS_PER_BANK;camIndex++) {
```
mck1117, line 465: empty loop

racer-coder:
This was fine until I removed some dead code :-P

mck1117:
That HW CI failure is a known fluke, you're ok!

Issue #123, opened 2014-11-16
setHardwareUsTimer too large


rusefi:
With version 5333

See http://tapatalk.imageshack.com/v2/14/11/16/d7367a5c9ded7e7a38ef809459b5430b.jpg
See http://rusefi.com/forum/viewtopic.php?f=9&t=728&p=12376#p12374

Reported by: rusefi

mck1117:
I highly suspect this is gone, as it was last poked 5 years ago

PR #3564, opened 2021-11-17
Make engine* and friends be const pointers in production.

racer-coder:
This allows the compiler to see through the pointer and make accesses faster.
It saves 1336 bytes of text.

### DIFF
--- a/firmware/controllers/algo/engine.h
+++ b/firmware/controllers/algo/engine.h
@@ -399,5 +399,7 @@ void doScheduleStopEngine();
 // These externs aren't needed for unit tests - everything is injected instead
 #if !EFI_UNIT_TEST
 extern Engine ___engine;
-#endif // EFI_UNIT_TEST
+static Engine * const engine = &___engine;
+#else // EFI_UNIT_TEST
 extern Engine *engine;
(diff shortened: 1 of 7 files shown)


mck1117:
Where does all the text saving come from? The few instructions per usage of not dereferencing the pointer?

racer-coder:
It seems to be, though it's hard to sift through.  Before we'd have something like
reg = &engine; // ldr rX, [pc, #xyz] // get address of engine variable
reg = *reg; // ldr rX, [rX, #0] // load contents of engine variable
reg += 0x1234; // add offset to item within engine that we care about

after:
reg = &___engine.thing_we_care_about; // ldr rX, [pc, #123]
i.e. it's able to combine all three instructions into one.

In another case it rearranged the code so much I couldn't tell, but I still saw 6 less instructions (in a sea of 100-ish).
It's hard to diff the asm and make much sense of it, unfortunately.

mck1117:
Yeah, this definitely seems like one of those cases that COMPLETELY rearranges the machine code output, instead of making the same small change everywhere.  Still, that's a big net savings for a small change!

PR #3567, opened 2021-11-17
Final cleanup of ENGINE macros

racer-coder:
Fix some uses of ENGINE() not detected by regex due to parenthesis.
Remove now empty engine_ptr.h
Don't worry about EFI_* not being defined, the compiler defaults to 0 if they aren't.


rusefillc:
@racer-coder merge conflict

What EFI_* not to worry about? :)

racer-coder:
> What EFI_* not to worry about? :)

Before, engine_ptr.h had:
-#ifndef EFI_UNIT_TEST
-#error EFI_UNIT_TEST must be defined!
-#endif
-
-#ifndef EFI_SIMULATOR
-#error EFI_SIMULATOR must be defined!
-#endif
-
-#ifndef EFI_PROD_CODE
-#error EFI_PROD_CODE must be defined!
-#endif

I don't think that's necessary, the compiler seems to work fine without them.  For example, this compiles fine:
diff --git a/firmware/controllers/algo/engine.h b/firmware/controllers/algo/engine.h
index d75c26b72..31313ab83 100644
--- a/firmware/controllers/algo/engine.h
+++ b/firmware/controllers/algo/engine.h
@@ -39,9 +39,13 @@
 #include "global_execution_queue.h"
 #endif /* EFI_UNIT_TEST */
 
+#if !EFI_DOES_NOT_EXIST
+
 #define FAST_CALLBACK_PERIOD_MS 5
 #define SLOW_CALLBACK_PERIOD_MS 50
 
+#endif
+
 class RpmCalculator;
 struct AirmassModelBase;

mck1117:
I'd like to leave these in somewhere - they're there to catch misconfigured makefiles and such

```
#ifndef EFI_UNIT_TEST
#error EFI_UNIT_TEST must be defined!
#endif

#ifndef EFI_SIMULATOR
#error EFI_SIMULATOR must be defined!
#endif

#ifndef EFI_PROD_CODE
#error EFI_PROD_CODE must be defined!
#endif
```

racer-coder:
> I'd like to leave these in somewhere - they're there to catch misconfigured makefiles and such

ok they're back in.  It seems like you'd figure out pretty quickly one way or another, but it's small and easy to tuck away.

rusefillc:
I second @mck1117  - I prefer explicit FALSE for all since it's much easier to find FALSE than to prove that no value is defined

PR #3571, opened 2021-11-17
Make the module API more concise.

racer-coder:
engineModules.get<FOO>(). becomes modules<FOO>()->

I believe the new API is more conducive to supporting arrays, by doing std::array<T, N> or
std::array<Mockable<T>, N>, with the support of a helper class.


racer-coder:
@mck1117 slightly better?

I think if we have a
```
template<typename T, unsigned N>
struct type_list<std::array<T, N>> : std::array<type_list<T>, N>
with a helper apply_all(), has(), and get() methods, then this will work.

PR #3572, opened 2021-11-17
Add a lock for scheduleOrQueue events.

racer-coder:
Ignition callers are all tooth based, but aux valves calls it from both the microsecond timer and
the main initialization thread.  HPFP will also soon call it from both those places too.

### DIFF
--- a/firmware/controllers/engine_cycle/spark_logic.cpp
+++ b/firmware/controllers/engine_cycle/spark_logic.cpp
@@ -320,14 +320,24 @@ bool scheduleOrQueue(AngleBasedEvent *event,
 		/**
 		 * Spark should be scheduled in relation to some future trigger event, this way we get better firing precision
 		 */
-		bool isPending = assertNotInIgnitionList(engine->angleBasedEventsHead, event);
-		if (isPending) {
-#if SPARK_EXTREME_LOGGING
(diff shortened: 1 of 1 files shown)


racer-coder:
FW check failure was unrelated - happened during `apt-get update`.

firmware/controllers/engine_cycle/spark_logic.cpp
```diff
@@ -466,13 +476,29 @@ static void prepareIgnitionSchedule() {
...
 			// time to fire a spark which was scheduled previously
-			LL_DELETE2(engine->angleBasedEventsHead, current, nextToothEvent);
+
+			// Yes this looks like O(n^2), but that's only over the entire engine
+			// cycle.  It's really O(mn + nn) where m = # of teeth and n = # events
+			// fired per cycle.  The number of teeth outweigh the number of events, at
+			// least for 60-2....  So odds are we're only firing an event or two per
+			// tooth, which means the outer loop is really only O(n).  And if we are
+			// firing many events per teeth, then it's likely the events before this
+			// one also fired and thus the call to LL_DELETE2 is closer to O(1).
```
mck1117, line 500: in addition, the sum M + N is roughly proportional to the number of cylinders: engines with fewer trigger teeth will have more events per tooth, and engines with many teeth will only have an event or two per tooth. I'm fine that this looks O(N^2), since it's really close to O(1)

PR #3575, opened 2021-11-19
Extend LogField to support scaled_channel divisors.

### LINKED ISSUE #3574: scaled_channel: LogField to support mult and div
we have
```
class LogField {
public:
	template <typename TValue, int TMult = 1>
	constexpr LogField(const scaled_channel<TValue, TMult>& toRead, const char* name, const char* units, int8_t digits)
```

which does not seem to support new mult and div option. LogField needs to catch up for the improved LogField with mult and div

racer-coder:
Also move the type field for better packing.  Saves about 500 bytes of flash.
This fixes #3574.

### DIFF
--- a/firmware/console/binary_log/log_field.cpp
+++ b/firmware/console/binary_log/log_field.cpp
@@ -34,7 +34,7 @@ void LogField::writeHeader(Writer& outBuffer) const {
 	buffer[45] = 0;
 
 	// Offset 46, length 4 = Scale
-	copyFloat(buffer + 46, 1.0f / m_multiplier);
+	copyFloat(buffer + 46, m_multiplier);
 
 	// Offset 50, length 4 = shift before scaling (always 0)
 	copyFloat(buffer + 50, 0);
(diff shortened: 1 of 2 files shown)


racer-coder:
hold on I can shrink the size field too

firmware/console/binary_log/log_field.cpp
```diff
@@ -34,7 +34,7 @@ void LogField::writeHeader(Writer& outBuffer) const {
 	buffer[45] = 0;
 
 	// Offset 46, length 4 = Scale
-	copyFloat(buffer + 46, 1.0f / m_multiplier);
```
rusefillc, line 37: hold on why do we change this? is this because what we've called mult was actually a divider?

rusefillc:
@racer-coder i suspect master was broken at the time when you've created your branch
please rebase?

firmware/console/binary_log/log_field.cpp
racer-coder: I'm just reusing the word.  If you note in the constructor I'm now assigning `divisor/multiplier` to m_multiplier.  I suppose I could have changed the name to m_divisor?

PR #3576, opened 2021-11-19
fuel pump doesn't use fsio

mck1117:
....but it does use the fancy live data console stuff.

#2928


firmware/controllers/actuators/fuel_pump.txt
```diff
@@ -0,0 +1,5 @@
+struct_no_prefix fuel_pump_control_s
+	bit isPrime
+	bit engineTurnedRecently
+	bit isPumpOn
+end_struct
```
racer-coder, line 5: Does this get us anything, like the ability to view through the console or something?  I didn't see any other files or any indication other parts of the system can find this file.

One downside to bit fields is you can't default them to 0, so Valgrind gets unhappy that you're using uninitialized memory in unit tests, since we instantiate Engine on the stack.  For the real product, it isn't a problem as ChibiOS inits all RAM areas for us.

rusefillc: it's supposed to at least eventually

PR #3578, opened 2021-11-19
Expose console over serial port for H7 proc

racer-coder:
This way you can use the console/TunerStudio with the ST-Link


mck1117:
why not just use a second usb cable?

firmware/config/stm32h7ems/efifeatures.h
```diff
@@ -49,3 +49,16 @@
...
 #define LUA_SYSTEM_HEAP 100000
+
+// UART driver not implemented on H7, but the serial driver is...
+#undef TS_NO_PRIMARY
+#undef TS_PRIMARY_UART
+#define TS_PRIMARY_SERIAL SD3
+
+// todo: start using consoleSerialTxPin? Not sure
+#undef EFI_CONSOLE_TX_BRAIN_PIN
+#define EFI_CONSOLE_TX_BRAIN_PIN GPIOD_8
```
mck1117, line 60: pins should probably go in the `board.mk` for the respective nucleo board

mck1117: since right now these will get used for /every/ H7 board, these could cause pin conflicts with other boards that use PD8/PD9

racer-coder: This somewhat mimics what happens on F4, which has
```
#ifndef TS_PRIMARY_UART                                                                             
#define TS_PRIMARY_UART UARTD3                                                                      
#endif                                                                                              
```
and friends.  For some reason the F7/H7 more forcefully prevent this default.  I can move it but there's a lot of goofiness about TS_PRIMARY_UART vs TS_PRIMARY_SERIAL so getting the default to work correctly when one is defined and the other might be defined later is tricky.  The whole thing is just insane, the HAL layer should export available serial ports whether UART or Serial or whatever, then we'd just need one TS_PRIMARY_FOO variable to define which one to use.... or have some sentinel value like OFF to disable.

racer-coder:
> why not just use a second usb cable?

There's a serial port on the ST-Link, we should use it for something.

The doc has some dire warning about hooking up a USB cable to the main proc if the board is unpowered; I guess I can assume that both cables would get power at the same time and it'd be fine, but it's so much cleaner to just have the one.

mck1117:
> has some dire warning about hooking up a USB cable to the main proc if the board is unpowered

meh, I've never blown up an ST dev board doing this

firmware/config/stm32h7ems/efifeatures.h
mck1117: If you think this is insane, you should've seen what it replaced!

mck1117: And I don't mean moving the UARTD3 part - I mean moving the GPIOD_8 part.  Right now this will cause a pin conflict on an H7 proteus - those pins are used for electronic throttle control.

firmware/console/binary/tunerstudio_io_serial_ports.cpp
```diff
@@ -11,16 +11,23 @@
 #include "tunerstudio_io.h"
 #include "connector_uart_dma.h"
 
+// These may not be defined due to the HAL, but they're necessary for the compiler to do it's magic
+class UARTDriver;
```
mck1117, line 15: some of these are `struct`

firmware/config/stm32h7ems/efifeatures.h
racer-coder: changed how ports are defined; TS_NO_PRIMARY is now a 0-or-1 define, not a is-or-is-not-defined.  TS_PRIMARY_UART and TS_PRIMARY_SERIAL are folded into TS_PRIMARY_PORT so there's no confusion over ordering.

Moved defines to nucleo board.mk, though it seems most boards already define TS_NO_PRIMARY etc so perhaps the default of "on" was ok.  But putting it in board dir does make more sense, it's a board configuration.

firmware/console/binary/tunerstudio_io_serial_ports.cpp
racer-coder: I swear I've seen that matter before, but it did compile, so ... ?  My google-fu is failing me on this one (trying to figure out when it's ok and when it's not).

mck1117: I think it chucks a warning for it (or maybe only clang does?) but it still works fine.

firmware/console/binary/tunerstudio_io_serial_ports.cpp
```diff
@@ -11,16 +11,23 @@
...
-			UartDmaTsChannel primaryChannel(TS_PRIMARY_UART);
-		#else
-			UartTsChannel primaryChannel(TS_PRIMARY_UART);
-		#endif
-	#elif defined(TS_PRIMARY_SERIAL)
-		SerialTsChannel primaryChannel(TS_PRIMARY_SERIAL);
-	#endif
+#ifdef TS_PRIMARY_PORT
+	std::conditional_t<
+		std::is_same_v<decltype(TS_PRIMARY_PORT), UARTDriver>,
```
rusefillc, line 23: please add a one line comment explaining this C++ magic to less experienced C developers.

PR #3582, opened 2021-11-20
detect more overflows with asan

mck1117:
Detect array overflows with ASAN.  Detects the existing bug in TRI_TACH causing an overflow!

Also changes engine modules to be allocated via `unique_ptr` so that ASAN can detect overflows from those too.

Turns on ASAN in the simulator, and fixes a lua heap setup bug detected by that.


racer-coder:
I'm on the fence here; I think -fsanitize=bounds will solve the majority of this for us.  Do we still want to go down this path anyways?

firmware/util/overflow_detector.h
```diff
@@ -0,0 +1,28 @@
...
+	overflow_detect() {
+		__asan_poison_memory_region(&protector, sizeof(protector));
+	}
+
+	~overflow_detect() {
+		__asan_unpoison_memory_region(&protector, sizeof(protector));
+	}
+};
+
+#endif
```
racer-coder, line 28: // EFI_PROD_CODE

mck1117:
Well, I tried it, and `-fsanitize=bounds -fsanitize=undefined` doesn't actually catch that particular case!

racer-coder:
with bounds-strict:
```
Exporting TT_TRI_TACH
../firmware/controllers/trigger/trigger_decoder.cpp:207:36: runtime error: index 280 out of bounds for type 'float [280]'
tests/trigger/test_all_triggers.cpp:80:38: runtime error: index 280 out of bounds for type 'float [280]'
[       OK ] Triggers/AllTriggersFixture.TestTrigger/52 (2 ms)
```

No other errors popped up.  Up to you whether you really want to pursue this, as it'd require annotating a bunch of sites.

mck1117:
oh boy that was a lot of compiler settings to change and merge with no CI check mark :joy:

PR #3583, opened 2021-11-20
Fix potential buffer overrun in cyclic_buffer.

racer-coder:
This exposed a buffer overrun, so double the size of the buffer (to account for 720 degree engine
cycle vs 360 degree crank events).

Also use proper numeric limits when computing min/max.
Finally, add a lock around the call to cyclic_buffer that actually caused the contention.

### DIFF
--- a/firmware/controllers/algo/engine2.cpp
+++ b/firmware/controllers/algo/engine2.cpp
@@ -9,6 +9,7 @@
 
 #include "pch.h"
 
+#include "os_access.h"
 #include "speed_density.h"
 #include "fuel_math.h"
 #include "advance_map.h"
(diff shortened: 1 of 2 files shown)


firmware/controllers/trigger/decoders/trigger_structure.h
```diff
@@ -310,7 +310,7 @@ class TriggerFormDetails {
 	 * These angles are in event coordinates - with synchronization point located at angle zero.
 	 * These values are pre-calculated for performance reasons.
 	 */
-	angle_t eventAngles[PWM_PHASE_MAX_COUNT];
+	angle_t eventAngles[PWM_PHASE_MAX_COUNT * 2];
```
mck1117, line 313: I fixed this in my PR.

racer-coder: then I'll just have to complain more about your PR so I beat you to master

PR #3584, opened 2021-11-20
don't do things that don't work in clang

mck1117:


unit_tests/unit_test_rules.mk
```diff
@@ -63,7 +70,11 @@ endif
 
 # Enable address sanitizer for C++ files, but not on Windows since x86_64-w64-mingw32-g++ doesn't support it.
 ifeq ($(SANITIZE),yes)
-	USE_CPPOPT += -fsanitize=address -fsanitize=bounds-strict -fno-sanitize-recover=all
+	ifeq ($(IS_MAC),yes)
+		USE_CPPOPT += -fsanitize=address
+	else
+		USE_CPPOPT += -fsanitize=address -fsanitize=bounds-strict -fno-sanitize-recover=all
```
racer-coder, line 76: why only C++ files?  Should be all, right?

mck1117: only c++ because lua does some things asan doesn't like, but don't actually cause overruns.

rusefillc: @mck1117 that would have been an amazing code comment ^^^

PR #3586, opened 2021-11-20
correctly compute engine revolution number

mck1117:
This change helps the engine potentially start one cycle sooner, especially when using a crank-primary trigger.

Note how all the tests had to be adjusted to sync/measure RPM one cycle quicker than before!

tl;dr by starting the revolution counter at 0, `triggerIndexForListeners` is guaranteed to be zero at the very first sync point, so we can start everything the first time, instead of some number of sync points later.


mck1117:
This was also the cause of spurious print-outs of `engine stopped` to the console during cranking.

rusefillc:
Please changelog

PR #3587, opened 2021-11-21
Move LuaThread to CCM, and shuffle a couple items out.

racer-coder:
Lua's stack probably doesn't need DMA.  Allows us to group more of our unused RAM together
in one place.


mck1117:
Just an FYI, we have `NO_CACHE` for things that need to be DMA-able, since on Cortex-M7 parts, just because it's not in CCM doesn't mean DMA will behave well.

PR #3595, opened 2021-11-22
oh my god we've had no asserts

mck1117:


rusefillc:
why draft? is this about real vehicle testing?

mck1117:
no, it's because I wasn't sure we'd have an assert fail on HW CI or not (I assumed we wouldn't, and we didn't)

PR #3606, opened 2021-11-24
fix map avg logging

mck1117:


mck1117:
always log even if not actively sampling

Issue #3607, opened 2021-11-25
Revisit Sensor Sniffer


rusefillc:
With 1KHz data rate we do not need Sensor Sniffer in it's current form

Sensor Sniffer would be useful only if we go with higher data rate and larger buffer maybe probably once we have DRAM?

rusefillc:
![image](https://user-images.githubusercontent.com/48498823/143334707-3afcf5a3-d250-4c22-9296-dee635c7d271.png)

rusefillc:
Short-term we can reduce chart size but keep the code?

mck1117:
With engine phase logged, we can probably nix a few of the modes, like MAP sensing. I see no need for that if we can produce an awesome log like the one above.

mck1117:
This is pretty awesome:

![image](https://user-images.githubusercontent.com/568254/143365234-4ffa75f9-89b6-4d38-ab9e-cb39d6a14652.png)

rusefillc:
Sensor Sniffer would be useful for full speed ADC but that would mean a larger buffer?

mck1117:
We only need a buffer at all if we're interested in >1khz sampling. If not, TunerStudio is the buffer for us, and that has functionally infinite storage.

rusefillc:
yes, exactly. that's exactly what I am saying.

as of today we are not ready for 10KHz buffer but we probably want to keep the code.

PR #3609, opened 2021-11-25
goodbye MAP sensor chart

mck1117:
With fast TS, we don't need MAP sensor chart.

#3607


rusefillc:
we will need this code once we go 10KHz sensor sniffer

mck1117:
Huh? why would we? Proteus/Hellen/(I think MRE too) hardware inputs only have ~150hz of bandwidth, so it is impossible to need more than ~300hz sampling to get the full picture.

rusefillc:
oh you are saying this specific sensor has no meaning above 300Hz

mck1117:
Correct. Some sensor might, but not MAP.

PR #3622, opened 2021-11-26
run firmware builds in self hosted pool

mck1117:
Speed up firmware runners by running them in our own pool.  This avoids a full checkout (only a `git clean` then `git fetch` runs), and means no dependencies have to be installed for every single build that runs.

Typical runs are 1-2 minutes per firmware build. Currently there are 8 such runners running on my server, and eventually @ZHoob2004 plans to run some more on his with similar hardware.

image


rusefillc (changes requested):
We cannot just kill GHA runners and rely only on unreliable private runners

At the moment recovering back to slow but reliable GHA would mean complex history digging. We need a smarter way where switch back and forth is much simpler

mck1117:
It is true that this could pose a reliability issue for now, but the eventual goal is to have runners in the same pool running on multiple machines in multiple houses, etc.

rusefillc:
I really do not want to choose between speed and fall back effort. This PR cannot go forward as is.

.github/workflows/build-firmware-runners.yaml
```diff
@@ -0,0 +1,335 @@
...
+      matrix:
+        # What boards should we build for? In the 'include' section below,
+        # set up what each of these boards needs to build.
+        #
+        # see also gen_config where we have a similar list for all boards!
+        # see also build-primary-bundle where default/primary bundle is build separately
+        #
+        build-target: [
+            hellen72,
+            hellen81,
```
rusefillc, line 21: that would be the third place for same or similar list, a major copy-pasta :(

.github/workflows/build-firmware-runners.yaml
```diff
@@ -0,0 +1,335 @@
...
+            folder: hellen/hellen154hyundai
+            ini-file: rusefi_hellen154hyundai.ini
+
+          - build-target: cypress
+            folder: hellen
+            ini-file: rusefi_hellen_cypress.ini
+
+          - build-target: frankenso_na6
+            folder: frankenso
+            ini-file: rusefi_frankenso_na6.ini
```
rusefillc, line 93: We need to at least move ini file into a separate dictionary not to duplicate it?

PR #3627, opened 2021-11-27
alternator not industrial pid

mck1117:
We don't need industrial PID nonsense for alternator.  It also seems to break alternator debug outputs, which is relevant to https://rusefi.com/forum/viewtopic.php?f=19&t=2226


rusefillc:
technically a proper fix would be to add debug into industrial PID

PR #3630, opened 2021-11-28
Remove the EFI_NO_CONFIG_WORKING_COPY option

mck1117:
There are three categories of config:

1. tables, etc that exist outside of engineConfiguration
2. mostly scalar parameters, a few tables that are inside engineConfiguration
3. stuff that uses activeConfiguration, pinouts, etc

For each of the above, this PR has the following effect:

1. none
2. These parameters now apply when you type them in, not when you press burn
3. none

These have been tested on real hardware, in addition to the required testing change.

Look at all that memory it freed up! 20k!

inspiration acquired from #3514, most of credit to @racer-coder


firmware/console/binary/tunerstudio.h
```diff
@@ -71,7 +70,7 @@ post_packed {
 	short int count;
 } TunerStudioWriteChunkRequest;
 
-class TunerstudioThread : public ThreadController<CONNECTIVITY_THREAD_STACK> {
+class TunerstudioThread : public ThreadController<2 * CONNECTIVITY_THREAD_STACK> {
```
andreika-git, line 73: Do we need an explanatory comment here?

mck1117: whoops, that was a local change that snuck in

PR #3639, opened 2021-11-30
remove cache zip, rely on LazyFile instead

### LINKED ISSUE #3433: cache.zip in repo is that a problem or not
we like to re-generate cache.zip somes a few times a day

cache.zip being a binary artifact is probably not good in terms of git practices and repo size

technically all we need is "cache" folder with files in it, not a zip achieve

mck1117:
Remove the zip caching logic.  This had a few problems:

1. every time we changed config, it commits a new zip file to the repo.  This file is binary, so it causes significant repo bloat.
2. It fails to catch changes to the generation tool that cause a change to the output. edit: there exist a few invalid txt files already in the repo that weren't detected!
3. It doesn't save enough time to be worth it. `time ./gen_config_default.sh` takes 1.9 seconds for me.

fixes #3433


firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -38,15 +38,13 @@ WaveChart waveChart;
 static scheduling_s debugToggleScheduling;
 #define DEBUG_PIN_DELAY US2NT(60)
 
-trigger_central_s::trigger_central_s() : hwEventCounters() {
-}
-
-TriggerCentral::TriggerCentral() : trigger_central_s(),
+TriggerCentral::TriggerCentral() :
```
rusefillc, line 41: hold on why is this here? looks like a separate change?

mck1117: This was an unknown bug in the config generator.  It can't handle `define_constructor` properly. I can break that change out, though.

mck1117: #3640 ready pending green check marks

rusefillc:
@mck1117 does it say conflicts?

mck1117:
should be no conflict now

Issue #3645, opened 2021-12-02
Engine phase wrap around in log


NMSTEC:
Whichever trigger offset I add, will be subtracted from 720* and will show anything in between there as 0. Example 191* offset, 720-191 = 509* 509 is the maximum value log will show, and drop to 0 after that until 720* fully rotate. Example below.

![image](https://user-images.githubusercontent.com/45252464/144504972-a49743b6-88c4-46fa-bfad-24284e44db42.png)

NMSTEC:
TDC offset at 0 works properly: 
![image](https://user-images.githubusercontent.com/45252464/144506001-480f5fed-f72f-423f-befc-de51ee0d40c8.png)

NMSTEC:
Tested working. Thank you

PR #3646, opened 2021-12-03
use tooth angle instead of timed angle

mck1117:
improve engine phase logging by using actual tooth angle, instead of inferring angle from last sync point

this improves accuracy when cranking due to engine speed variation (think: harley cranking with 900cc cylinders)

relevant to #3645


rusefillc:
We do not account for time at all?

mck1117:
Nope, no time, this is only ever called just after tooth decode, so we know exactly where the engine is.

Issue #3660, opened 2021-12-05
Dumb CAN wake-up


NMSTEC:
1. rusEFI gets message "please go to sleep"
1. rusEFI turns off can TX
1. wait XXX seconds, where XXX is time for everyone to go to sleep
1. rusEFI goes in to low power wake-on-can mode

Just to clarify Harley graceful power off/power on is done via CAN not by controlling main relay

For much smarter implementation see #1731

rusefillc:
Oh shit :( Worst case we would have to implement a dedicated microcontroller just for CAN and wake up.

mck1117:
Can you enumerate the full list of what happens when you go from walking up to a parked bike, thru running engine?

rusefillc:
probably relevant https://www.st.com/resource/en/application_note/dm00096220-using-stm32f4-mcu-power-modes-with-best-dynamic-efficiency-stmicroelectronics.pdf

mck1117:
That's why I wanted clarity on what we need.  If we need to wake up for ANY packet, that's pretty easy.  If we need to only wake up for a particular packet, that's harder.

rusefillc:
step one: bike is parked for a week, things are whatever
step two: rider sits on the bike with RFID in his pocket, rider does not touch any buttons
step three: ass sensing sensor sends a CAN message to start waking up modules
if ECU does not wake up and announce itself within unknown time, dash would persist an error code with error light. 3 minute of manual fuckery to wipe out code


there are only two wires with +12 ever:
pin 67 has constant +12
pin 5 has +12 when physical switch is in IGNITION position

BCM expects CAN from ECU while no power on pin 5. Conclusion: MCU in ECU is supposed to feed itself on constant power pin 67.

@NMSTEC please correct me if any of this is not right
@NMSTEC please elaborate of shutdown procedure in terms of what CAN activity happens when

NMSTEC:
We need any packet whatsoever to wake up. As the rest of the vehicle does the same. 

Scenario 1. Walking to up to bike, nothing occurs. Turn IGN to on, and BCM wakes everyone up as well as turn "ign 12v" on. Wakes all up just by simply transmitting any packet. In this case 0x502 - 0x01(or 10? I cant recall, not at work)

[...]

PR #3666, opened 2021-12-06
testing stm32 stop/standby

mck1117:
testing for #3660 progress


rusefillc:
looks like only good for F4?

rusefillc:
@mck1117 can you please forum some info on the big picture?

PR #3674, opened 2021-12-08
smarter priming logic

mck1117:
Depends on #3662 

- Table for priming fuel mass vs. coolant temperature
- Configurable priming delay to build fuel pressure
- Uses full injector model (including injector pressure compensation!) to accurately prime even it fuel pump is still building pressure


mck1117:
Ripping the values out of the stock GM tune, I get really, really good starts: https://www.youtube.com/watch?v=lGjPbiyv49c

firmware/integration/rusefi_config.txt
```diff
@@ -1159,7 +1159,7 @@ int16_t tps2Max;Full throttle#2. tpsMax value as 10 bit ADC value. Not Voltage!\
 	
 	float targetVBatt;+This is the target battery voltage the alternator PID control will attempt to maintain;"Volts", 1, 0, 0, 30, 1
 	float alternatorOffAboveTps;+Turns off alternator output above specified TPS, enabling this reduced parasitic drag on the engine at full load.;"%", 1, 0, 0, 200, 2
-	float startOfCrankingPrimingPulse;+Prime pulse for cold engine, duration in ms\nLinear interpolation between -40F/-40C and fallout temperature\n\nSee also isFasterEngineSpinUpEnabled\nset cranking_priming_pulse X;"ms", 1, 0, 0, 200, 1
```
rusefillc, line 1162: in light of properties being removed and added we absolutely need some detailed release note for this change. Let's do this right in CHANGELOG? so far CHANGELOG entries were one liners, I think it's time to start creating longer entries

rusefillc (changes requested):

firmware/integration/rusefi_config.txt
mck1117: Changelog added.  I'm not as worried about removing old config here, since I've never seen these values changed from the defaults in someone's tune, because priming as it existed was pretty much useless (there was no possible way to actually deliver enough fuel).

PR #3677, opened 2021-12-08
CAN ISO-TP progress (+unit-tests fix)

andreika-git:


firmware/hw_layer/drivers/can/can_msg_tx.cpp
```diff
@@ -87,20 +90,6 @@ void CanTxMessage::setBit(size_t byteIdx, size_t bitIdx) {
 	m_frame.data8[byteIdx] |= 1 << bitIdx;
 }
 
-#else
```
mck1117, line 90: why this change?

rusefillc: if we can remove code and tests are still passing it must be unneeded code?

mck1117: but the code didn't get removed, it just got #ifdef'd in every single function instead of all at once

rusefillc: @mck1117 please forgive me I was excited to try if this would have fixed PCAN. CAN defines are a mess at the moment, this is technical debt right now while I am excited about getting PCAN ready.

PR #3678, opened 2021-12-08
CAN ISO-TP fix?

andreika-git:


mck1117:
merged red?

rusefillc:
@mck1117 i believe it was red due to master red

Issue #3687, opened 2021-12-10
Second CAN


rusefillc:
proteus HW has two CAN busses already
Lua has "bus index" parameter already

we have can2RxPin and can2TxPin configuration fields

We need to add all the wiring to have second CAN bus work in parallel to the first one which uses canTxPin canRxPin

rusefillc:
```
txPayload = {}
txPayload[1] = 123;

function onTick()
 txCan(1, 0x611, 1, txPayload)
 txCan(2, 0x612, 1, txPayload)
end
```

PR #3688, opened 2021-12-10
multi CAN plumbing

mck1117:
#3687


firmware/hw_layer/drivers/can/can_hw.cpp
```diff
@@ -294,4 +300,15 @@ bool getIsCanEnabled(void) {
 	return isCanEnabled;
 }
 
+CANDriver* detectCanDevice(size_t logicalIndex) {
+	switch (logicalIndex) {
+	case 0:
+		return detectCanDeviceImpl(engineConfiguration->canRxPin, engineConfiguration->canTxPin);
+	case 2:
```
rusefillc, line 307: 0 and 2? Please add code comment

rusefillc (changes requested):

firmware/hw_layer/drivers/can/can_hw.cpp
mck1117: typooooooooo

Issue #3544, opened 2021-11-14
MAP phase sensing


rusefillc:
Harley does not have a physical CAM sensor.

Harley is asking for MAP sensor to be used for engine phase detection

rusefillc:
Let's see if we can grab a nice MAP chart using TS

1) use TunerStudio BETA for better performance
2) set data rate to MAXIMUM
3) change logging configuration to only log MAP ton increase data flow
4) set widest MAP sampling windows so that all angles are sampled

TODO do we have un-averaged fast 'raw' decoded MAP or MAP voltage @mck1117 ?

NMSTEC:
We've attempted to take a high speed MAP debug log with TS Beta but MAP values do not refresh often, Andrey wonders if code changes would be needed

@mck1117 can you please look at 
[2021-11-24_11.01.17.mlg.zip](https://github.com/rusefi/rusefi/files/7598143/2021-11-24_11.01.17.mlg.zip)

NMSTEC:
![image](https://user-images.githubusercontent.com/45252464/143319162-3489faee-e9f8-4116-8b6c-d0354adcc7c9.png)

NMSTEC:
[DataLogs.zip](https://github.com/rusefi/rusefi/files/7598949/DataLogs.zip)

NMSTEC:
![image](https://user-images.githubusercontent.com/45252464/143319491-64ab504f-7b6e-4423-94bf-e9d951906cf8.png)

NMSTEC:
[map w enginephase.zip](https://github.com/rusefi/rusefi/files/7598964/map.w.enginephase.zip)

rusefillc:
@mck1117 I do not trust these flat regions those are too constant. Does this look suspicious to you?

NMSTEC:
![image](https://user-images.githubusercontent.com/45252464/143327360-7f883eac-ca83-4dcf-9eeb-dad0ac503a8d.png)
[2021-11-24_16.03.29.mlg.zip](https://github.com/rusefi/rusefi/files/7599272/2021-11-24_16.03.29.mlg.zip)

Seems fixed.

rusefillc:
@stefanst I bet that's not the shape you were expecting :)

rusefillc:
![image](https://user-images.githubusercontent.com/48498823/143360739-bb83d5ff-fb18-4b34-9d15-545161c404fa.png)

315-405 ratio between spikes

mck1117:
[...]

PR #3689, opened 2021-12-10
maybe more working MAP sampling

mck1117:
we need to check instant MAP, not averaged MAP

#3544


mck1117:
Oooooooh it looks really good!

image

Issue #3045, opened 2021-07-24
2 byte table axis


rusefillc:
at the moment we are flexible with TS table cell size - it could be 1 byte or 4 byte

but each axis cell still always take 4 bytes. for 8x8x1 table we have 8x2x4 bytes for axis would be nice to have an option of axis taking less memory than data in the table

step 1 extra flexibility for 3D interpolation template?
step 2 something else?

rusefillc:
should it actually be 1 byte for RPM?

how do we want to pack load axis into two bytes?

mck1117:
> how do we want to pack load axis into two bytes?

we already do it in a few places, scaled by (iirc) 50 or 100

rusefillc:
@alxrMironovHome can you help?

rusefillc:
Relates to #3707

rusefillc:
@mck1117 shall we close?

PR #3704, opened 2021-12-20
ui & config for fuel/ign trim tables

mck1117:
Adds a 4x4 trim to adjust fuel % and ignition degrees +/-, for every cylinder.   Since they share axes (all fuel trims share, all ign trims share), it only uses 416 bytes of config space.

Since this is already an incompatible change requiring a change to `FLASH_DATA_VERSION`, this PR also:

- converts table axes to uint16_t (rpm, load where safe to do so) instead of using float #3045 
- removes some unused config spacers, leaving 400 bytes available at the end of `engine_confguration_s` for added parameters.

Between these three changes, nets 1356 bytes of config (and therefore RAM) savings.


mck1117:
depends on #3692 and  #3705

mck1117:
image

PR #3706, opened 2021-12-20
Do compiler check first

dron0gus:


rusefillc:
I did something similar in master already does it not work?

dron0gus:
> I did something similar in master already does it not work?

Ups. Let me update.

dron0gus:
cpp files are compiled after c. So gcc version is checked after all c files get compiled. That may take some time.

rusefillc:
Can we put cpp before C?

mck1117:
> Can we put cpp before C?

Not without changing the inner makefile bits of chibios, no.  Just use a C file.

dron0gus:
Check is moved to first compiled c-file.

PR #3708, opened 2021-12-21
tables support non-integer-scaled scaled_channel

### LINKED ISSUE #3707: Map3D needs some changes to allow autoscale RPM bin
See https://github.com/rusefi/rusefi/commit/76088f686fa9df26ac69da103275a211c15de506

looks like Map3D is not ready for one of axis being autoscaled? what about just table being autoscaled and not axis?

to clarify - I am less excited about one-byte RPM at the moment and mostly focused on having the actual table autoscaled, either way it looks like Map3D is not fully compatible with mult+div autoscale?

mck1117:
fix #3707

### DIFF
--- a/firmware/util/containers/table_helper.h
+++ b/firmware/util/containers/table_helper.h
@@ -73,10 +73,10 @@ private:
         m_values = &table;
 	}
 
-	template <int TRowMult>
-	void initRows(const scaled_channel<TRow, TRowMult> (&rowBins)[TRowNum]) {
+	template <int TRowMult, int TRowDiv>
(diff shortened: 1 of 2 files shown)


firmware/util/containers/table_helper.h
```diff
@@ -69,28 +69,28 @@ class Map3D : public ValueProvider3D {
 		m_values = reinterpret_cast<TValue (*)[TRowNum][TColNum]>(&table);
```
racer-coder, line 69: Do you want to update this function?  Or doesn't matter since it only affects values, not keys?

Issue #3702, opened 2021-12-20
V-Twin - Per bank Timing table


NMSTEC:
V-Twin engine is not symmetrical with front and rear engines having very different cooling. Because of that we would need front and rear banks to run completely independent timing maps.

We can keep the same axis points.

I want additional Timing table for second bank in addition to current timing table. 

Blah blah blah:
Vtwins. Rear cyl runs way hotter than front. Needs separate tables for Front and Rear cyls. This will also to apply to V4 bikes (Aprillia).

rusefillc:
it looks like @NMSTEC prefers completely independent separate timing maps but at the moment @mck1117 suggest that we go with 6x6 per-bank or per-cylinder trimming/adjustment map

NMSTEC:
After much deliberation we have agreed trim would be enough for our specific case.

mck1117:
this is fixed with cylinder ignition trim

PR #3709, opened 2021-12-21
cylinder ignition trim

mck1117:
Read the table, do some math, set per-cylinder ignition advance!

#3702


rusefillc:
@mck1117 draft or merge? is this about missing unit test?

Issue #3703, opened 2021-12-20
V-Twin - Additional VE map 


NMSTEC:
V-Twin engine is not symmetrical with front and rear engines having very different cooling. Because of that we would need front and rear banks to run completely independent fuel maps.

We can use the same Axis Values between both maps

Blah blah: As mentioned on #3702 rear cylinders run on different fuel level/VE from the front due to temp differences.

mck1117:
like #3702 we don't need per-cylinder maps, we can use a cylinder trim table per cylinder

mck1117:
possible dupe #3284

mck1117:
this is fixed with per-cylinder trim tables, as in #3284

Issue #3284, opened 2021-09-30
individual cylinder fuel trims


rusefillc:
@blundar need more details on this feature

would it be in % extra, positive or negative, with "0" meaning no trim? or anything else?

rusefillc:
@mck1117 says something like "give #7 5% more gas"

blundar:
I don't think it really matters as long as it is clear.

I think I personally prefer 1.0 = no change, 0.9 = 10% reduction, 1.1 = 10%
increase
0 = no change +10 = 10% increase, -10 = 10% decrease would be ok

rusefillc:
fancy boys have 8x12 tables for each cylinder

rusefillc:
![image](https://user-images.githubusercontent.com/48498823/135382552-7c1d48f6-c91e-46c5-9b64-10c1b8e0eadf.png)

rusefillc:
Status: configuration maybe is ready, TS view maybe ready, no code.

mck1117:
@blundar says

> if you can set your axes points, 4x4 is plenty
> 4x4 lets you exclude idle, exclude cruise/off boost
> that's 110% functional.

PR #3715, opened 2021-12-22
hook up fuel trims

mck1117:
#3703 
#3284


rusefillc:
@mck1117 draft or merge? is this about missing unit test?

Issue #3305, opened 2021-10-03
rusefi_Console: Too long re-connection time


andreika-git:
TunerStudio is able to restore the connection in a few seconds after the USB cable is connected again.
But the Console has a too long reconnection timeout (>10 seconds?). It would be perfect if we could match the TS timings.

mck1117:
>  But the Console has a too long reconnection timeout (>10 seconds?). It would be perfect if we could match the TS timings.

That is, if it works at all. Usually the console just crashes for me upon reconnect attempt.

rusefillc:
Q: what is this about?
A: @andreika-git often need to reset ECU while bench testing stuff like stepper etc so we want a lighter ECU reset workflow

rusefillc:
@alxrMironovHome can you help?

rusefillc:
@andreika-git how does it look with the change?

rusefillc:
@alxrMironovHome discovered that drama was about poking driver with port number which was not yet valid at that point. that was hanging-ish the driver.

now with the fix we wait for port to become officially available again and things seem to have improved!

mck1117:
Fixed by #3936

PR #3716, opened 2021-12-22
Java Console reconnection fixing (by issue #3305)

alxrMironovHome:
With this changes JC is reconnecting after the 5s timeout (as planned???)

### DIFF
--- a/java_console/io/src/main/java/com/rusefi/io/LinkManager.java
+++ b/java_console/io/src/main/java/com/rusefi/io/LinkManager.java
@@ -46,6 +46,9 @@ public class LinkManager implements Closeable {
     public static final String LOG_VIEWER = "log viewer";
     private final CommandQueue commandQueue;
 
+
+    private String lastTriedPort;
+
     private LinkConnector connector = LinkConnector.VOID;
     private boolean isStarted;
     private boolean compositeLogicEnabled = true;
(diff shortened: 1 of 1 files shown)


java_console/io/src/main/java/com/rusefi/io/LinkManager.java
```diff
@@ -46,6 +46,9 @@ public String unpack(String packedLine) {
     public static final String LOG_VIEWER = "log viewer";
     private final CommandQueue commandQueue;
 
+
+    private String lastTriedPort;
```
mck1117, line 50: please explicitly initialize (perhaps to null)

java_console/io/src/main/java/com/rusefi/io/LinkManager.java
```diff
@@ -46,8 +46,10 @@ public String unpack(String packedLine) {
     public static final String LOG_VIEWER = "log viewer";
     private final CommandQueue commandQueue;
 
+    private String lastTriedPort = null;
+
     private LinkConnector connector = LinkConnector.VOID;
-    private boolean isStarted;
+    private boolean isStarted = false;
```
rusefillc, line 52: no reason to change that line

java_console/io/src/main/java/com/rusefi/io/LinkManager.java
```diff
@@ -272,13 +275,22 @@ public void send(String command, boolean fireEvent) throws InterruptedException
...
-        connector.restart();
+        close(); // Explicitly kill the connection (call connectors destructor??????)
+
+        String[] ports = getCommPorts();
+        boolean isPortAvaliableAgain = Arrays.stream(ports).anyMatch(lastTriedPort::equals);
+        if(isPortAvaliableAgain) {
+            connect(lastTriedPort);
+        }
+
+       // connector.restart();
```
rusefillc, line 286: we do not need dead code, please remove if if you are removing it

rusefillc (changes requested):
this removes LinkConnector#restart usages so let's remove all LinkConnector#restart implementations and code

overall my problem is that I do not understand this change. what is fundamentally different between new code and old code?

PR #3723, opened 2021-12-28
adjustment to #3722

mck1117:
Using getShaftSynchronized doesn't work for non-decoder VVT modes, like "first half rising", "vtwin map", etc. Perhaps the best thing is to wait for the thing we KNOW we're waiting for, an adjustment of crank phase from the VVT system, so that's what this PR waits for.


firmware/controllers/engine_cycle/main_trigger_callback.cpp
```diff
@@ -352,7 +352,7 @@ void mainTriggerCallback(uint32_t trgEventIndex, efitick_t edgeTimestamp) {
 	ScopePerf perf(PE::MainTriggerCallback);
 
 	if (noFiringUntilVvtSync(engineConfiguration->vvtMode[0]) 
-		&& !engine->triggerCentral.vvtState[0][0].getShaftSynchronized()) {
```
rusefillc, line 355: now that we are fixing recent changes It sounds like we need more test coverage

rusefillc (changes requested):

firmware/controllers/engine_cycle/main_trigger_callback.cpp
mck1117: Done, looks like the map cam test was sort of broken anyway :)

mck1117:
OK, finally green!

Issue #3721, opened 2021-12-27
NB1/NB2 triggers review


mck1117:
Breaking off NB1/2 as its own issue, as it's a different decoder to NA.  #3717 

@BeerMoneyMotorsports says nb1 recording. 4 starts. 3rd was no start while 4th fired right up. they were sequential attempts. no time between.

https://rusefi.com/online/view.php?log=517

mck1117:
In that log:

- Crank looks great.  We sync it no problem.
- VVT makes a bunch of garbage right at the start of cranking, what's up with that?

![image](https://user-images.githubusercontent.com/568254/147441990-25cca91b-bfb4-48a3-9213-3fd870ceccec.png)

That mess of a blob was a total of 33 VVT edges in 0.32 second.

mck1117:
Third attempt in that log is an even worse disaster:

![image](https://user-images.githubusercontent.com/568254/147444392-a46d0e8b-fcb7-4d22-94ef-a6e46be68a2d.png)

mck1117:
And the 4th attempt is perfect!?

![image](https://user-images.githubusercontent.com/568254/147444432-37939e4f-b21e-4573-a481-57764fc1a0a6.png)

BeerMoneyMotorsports:
log of second NB1.  had some oddities of rpm spike.

https://rusefi.com/online/view.php?log=518

rusefillc:
so you need three points to connect logic analyzer

GND - we usually have it on proto area
for CAM/CRANK the simplest option could be to solder little wire stubs at the back side of the PCB right to relevant ECU connector pins

0.5M is good enough rate for our purposes while saving logic analyzer trace. Please IM if I can help with logic analyzer UI.

BeerMoneyMotorsports:
Second NB1 updated firmware.  
![image](https://user-images.githubusercontent.com/82395131/147510995-6a7973e9-a4f7-4788-9033-e1cb5a7d21bc.png)

https://rusefi.com/online/view.php?log=519

BeerMoneyMotorsports:
> so you need three points to connect logic analyzer
> 
> GND - we usually have it on proto area for CAM/CRANK the simplest option could be to solder little wire stubs at the back side of the PCB right to relevant ECU connector pins
> 
> 0.5M is good enough rate for our purposes while saving logic analyzer trace. Please IM if I can help with logic analyzer UI.

[...]

PR #3726, opened 2021-12-30
extract isSyncPoint function

mck1117:
Extract isSyncPoint as its own function.  Prep for some custom decode logic to solve problems with NB2 trigger sync during cranking.

progress on #3721

Slightly improves performance as we can bail out (`return false;`) as soon as we find a gap that doesn't match.


unit_tests/tests/trigger/resources/nb2-cranking-good.csv
```diff
@@ -481,10 +481,6 @@ Time [s],crank,cam
 6.940991417,0,1
 6.941894250,1,1
 6.959307333,0,1
-6.959354333,1,1
```
mck1117, line 484: this is a blip of noise picked up by the logic analyzer from the ignition system

unit_tests/tests/trigger/resources/nb2-cranking-good.csv
```diff
@@ -980,28 +976,3 @@ Time [s],crank,cam
 9.816113833,1,1
 9.829231167,0,1
 9.829947667,1,1
-9.850056500,0,1
```
mck1117, line 983: this is a mess of noise that happens at key-off that makes the test harder to deal with (ie, cut the log off while the engine is still running happily)

PR #3731, opened 2022-01-02
96-00 miata connector site pic

BeerMoneyMotorsports:


rusefillc:
@BeerMoneyMotorsports please elaborate what is wrong with the current naming convention why do you suggest a change? Do you plan to rename all other files in https://github.com/rusefi/rusefi/tree/master/misc/www/images/site folder according to the new naming convention?

I am also worried about spaces in file names. Would you be able to take care of HTML part of the change with regards to spaces?

rusefillc (changes requested):
Please make file name more uniform.

rusefillc:
@ksmola do we still care for image sizes? 3000px wide image that is formatted to be 480px on the main page? good bad do not care? does this add up or is this no longer a concern in 2022?

mck1117:
It's probably worth a shrink if only for load time

PR #3736, opened 2022-01-03
GDI/HPFP: Set HPFP current/timing params on MC33816

racer-coder:
obviously completely untested.  Please test!  I would use a spare injector or pump that you don't care about and test with an oscilloscope.  Yes I think you can use an injector since all you really need is something to sink 12V at various current levels.  Start by testing on a spare high impedance port injector since you know it'll never sink too much current while you make sure "on/off" works.

I would start with something like peak=5A, hold=3A, hold_off=10us, and hold_tot=10ms.


rusefillc:
@racer-coder does https://github.com/rusefi/rusefi/commit/e9498f64db408ac85c3322a75cb92a7e19f3eb29 look right to you?

racer-coder:
yes, switching to autoscale is fine (I couldn't decide whether to do that), and the other changes match correctly IMO.  Yeah those are the defaults I'd use for those 4 params.

Issue #3739, opened 2022-01-04
GHA execution throttling


rusefillc:
In order to cut the pipeline let's execute some jobs less often than always

rusefillc:
@chuckwagoncomputing I've failed to implement this :( Any chance you would be interest in this approach to GHA performance?

chuckwagoncomputing:
This issue isn't actionable as-is. Please clarify how it should be determined whether a job should be run. Coin toss? Phase of the moon? Only on Tuesdays? The color of the Queen's hat?

rusefillc:
@chuckwagoncomputing my bad

see https://github.com/rusefi/rusefi/commit/0c22c9ffb3aee4494a9221537b984bc3419f320e which is part of master now. it shows skip-rate values and execution-throttle.sh tool which is currently unused in master.

rusefillc:
@chuckwagoncomputing does it look like we currently skip at all?

chuckwagoncomputing:
Looks like it's working to me. Here is an example: https://github.com/rusefi/rusefi/runs/5998787781?check_suite_focus=true

rusefillc:
![image](https://user-images.githubusercontent.com/48498823/163583609-0560adf0-fbcd-414a-8801-6b4148fc30e1.png)

PR #3741, opened 2022-01-04
fix firmware yaml

mck1117:


rusefillc:
yes #3739 does not work as indented

as of right now I am not sure why :(

eg33 i would keep without throttling since an example of custom board with custom enums. maybe 50%

PR #3743, opened 2022-01-05
improve maf output channels and logging

mck1117:
- Separate log output channels for estimated and measured airmass
- MAF has a raw gauge just like everybody else


rusefillc:
@mck1117 poke

Issue #3720, opened 2021-12-26
stm32_stop just reboots stm32f4discovery


rusefillc:
stm32f407 discovery see also https://github.com/rusefi/rusefi/wiki/Dev-Quick-Start

need to fix/implement MCU stop with wake-up on pin PA0 or any other pin of your choice. Method will be
invoked via existing USB serial port connectivity, frontend application to send commands already exists.

current non working "stm32_stop" method at https://github.com/rusefi/rusefi/blob/master/firmware/hw_layer/ports/stm32/stm32f4/mpu_util.cpp

work would need to be summitted as github pull request

bin/send_command "stm32_stop"

stm32f4discovery with fresh firmware
PA0 jumper wire to GND, no other wires
mini USB, micro USB

send stm32_stop via console

expected: some sort of stop
actual result: firmware reboots

rusefillc:
@NMSTEC do you have an stm32f407 discovery board?

NMSTEC:
I do not. Will order from the link you sent, and will fuck with proteus upon my return.

rusefillc:
@NMSTEC report same or similar on Proteus F7 with some hw mod

NMSTEC:
STM32F7 on proteus. Cut PA0 trace, soldered on jumper to RX pin of CAN. Can is not active at this time, and running on bench from USB only. stm32_stop forces reboot. Cut jumper, leaving PA0 hanging in the air, problem remains.

rusefillc:
maybe relevant:
```
The important part is to always clear the wakeup flag (via PWR_CR_CWUF_Msk) in the PWR->CR register. Otherwise the device will immediately wake up after attempting to enter the stop mode.
```

NMSTEC:
Resolved: https://github.com/rusefi/rusefi/pull/3783

mck1117:
dddddone!

PR #3747, opened 2022-01-06
Stop & wakeup working on rising edge

NMSTEC:
stm32_stop works. used NVIC to trigger wakeup. Will need revisiting for changing to Falling edge, or will require hardware changes for signal inversion.

### DIFF
--- a/firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
+++ b/firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
@@ -178,21 +178,12 @@ void sys_dual_bank(void) {
 void stm32_stop() {
 	SysTick->CTRL = 0;
 	SCB->SCR |= SCB_SCR_SLEEPDEEP_Msk;
-	//RCC->AHB1RSTR = RCC_AHB1RSTR_GPIOERST;
 	enginePins.errorLedPin.setValue(0);
 	enginePins.runningLedPin.setValue(0);
 	enginePins.communicationLedPin.setValue(0);
 	enginePins.warningLedPin.setValue(0);
 
-	/* THIS IS ALTERNATIVE HAND FUCKERY WAY OF DOING THIS. HAL way is below.
-	// configure mode bits
-    EXTI->IMR |= EXTI_IMR_IM0;              // interrupt request from line 0 not masked
-    EXTI->RTSR |= EXTI_RTSR_TR0;           // rising trigger enabled for input line 0
 
-    // Enable interrupt in the NVIC
-    NVIC_SetPriority(EXTI0_IRQn, 0);
-  	NVIC_EnableIRQ(EXTI0_IRQn);
-*/
 	palEnableLineEvent(PAL_LINE(GPIOA, 0), PAL_EVENT_MODE_RISING_EDGE);
 
 	PWR->CSR1 |= PWR_CSR1_WUIF;
--- a/firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
+++ b/firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
@@ -210,19 +210,11 @@ void stm32_stop() {
 
 void stm32_standby() {
 	SysTick->CTRL = 0;
-	__disable_irq();
-	RCC->AHB1RSTR = RCC_AHB1RSTR_GPIOERST;
-
-	// configure mode bits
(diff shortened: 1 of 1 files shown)


firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
```diff
@@ -209,7 +221,8 @@ void stm32_standby() {
 
 	// Wait for event - this should never return as it kills the chip until a reset
 	__WFE();
+	__WFE();
```
rusefillc, line 224: do we now need __WFE _twice_? your PR adds second time?

firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
```diff
@@ -209,7 +221,8 @@ void stm32_standby() {
...
 
 	// Lastly, reboot
-	NVIC_SystemReset();
+	//NVIC_SystemReset();
```
rusefillc, line 227: if we do not need something simply remote it

firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
```diff
@@ -177,20 +177,32 @@ void sys_dual_bank(void) {
...
-
+	SCB->SCR |= SCB_SCR_SLEEPDEEP_Msk;
+	//RCC->AHB1RSTR = RCC_AHB1RSTR_GPIOERST;
+	enginePins.errorLedPin.setValue(0);
+	enginePins.runningLedPin.setValue(0);
+	enginePins.communicationLedPin.setValue(0);
+	enginePins.warningLedPin.setValue(0);
+	
 	// configure mode bits
+    EXTI->IMR |= EXTI_IMR_IM0;              // interrupt request from line 0 not masked
```
mck1117, line 188: can you try `palEnableLineEvent(line, mode);` instead?

mck1117: or even `efiExtiEnablePin`

mck1117: see digital_input_exti.cpp line 71-ish

firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
mck1117: this looks like unrelated change that snuck in, it's in the next function

firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
NMSTEC: Funny enough, yes and no. 
So if we are using WFE then yes, needs twice. Its an oddity: https://devzone.nordicsemi.com/f/nordic-q-a/10932/can-someone-explain-to-me-how-nordic-nrf51822-sleep-mode-works

However, with WFI we do not need it. Standby is not used right now at all, and may be removed altogether. I am working on testing it now. Hence why I didn't much care for the standby function.

firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
rusefillc: can we please remove standby already? do we know enough to know that we do not need it?

firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
rusefillc: looks like this code comment with magic link has to be in the code then

NMSTEC: Yeah, but since we are waking on interrupt instead of Event, might as well forget about it. especially since I couldnt get Event to work anyway, (at least for wakeup). WFI is more direct, and tested to be working.

firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
NMSTEC: Not yet. I want to finish testing current consumption in both modes, and see what would work. I do believe we can probably dump standby, but please wait for confirmation.

NMSTEC:
#3720

NMSTEC:
Apologies for the double commit. Didn't realize I did correctly set up GIT on my end.

mck1117 (approved):

rusefillc:
great stuff thank you!

PR #3748, opened 2022-01-06
TLE8888: do not allow direct gpio drive for Push Pull outputs

dron0gus:
Some register settings also needs to be changed. Change of direct
drive input is not enought.


dron0gus:
Just rebase and forced push to restart tests.

PR #3750, opened 2022-01-07
Move out of CMM

dron0gus:
This needs additional investigation. But allocating these structs
in CMM cause HardFault

### DIFF
--- a/firmware/controllers/actuators/idle_hardware.cpp
+++ b/firmware/controllers/actuators/idle_hardware.cpp
@@ -17,8 +17,10 @@
 #include "dc_motors.h"
 #if ! EFI_UNIT_TEST
 #include "stepper.h"
-static StepDirectionStepper iacStepperHw CCM_OPTIONAL;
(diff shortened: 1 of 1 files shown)


firmware/controllers/actuators/idle_hardware.cpp
```diff
@@ -17,8 +17,8 @@
 #include "dc_motors.h"
 #if ! EFI_UNIT_TEST
 #include "stepper.h"
-static StepDirectionStepper iacStepperHw CCM_OPTIONAL;
-static DualHBridgeStepper iacHbridgeHw CCM_OPTIONAL;
+static StepDirectionStepper iacStepperHw;// CCM_OPTIONAL;
```
rusefillc, line 20: plz kill dead code or add a comment, but not leave uncommented dead code

rusefillc (changes requested):

firmware/controllers/actuators/idle_hardware.cpp
dron0gus: done!

PR #3752, opened 2022-01-07
per-board pre-low-power handlers

mck1117:
Some boards might need to do special things before going to low-power mode.

### DIFF
--- a/firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
+++ b/firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
@@ -178,13 +178,15 @@ void sys_dual_bank(void) {
 void stm32_stop() {
 	SysTick->CTRL = 0;
 	SCB->SCR |= SCB_SCR_SLEEPDEEP_Msk;
+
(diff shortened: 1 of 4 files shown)


firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp
```diff
@@ -178,13 +178,15 @@ void sys_dual_bank(void) {
...
 	SCB->SCR |= SCB_SCR_SLEEPDEEP_Msk;
+
+	// Turn off the LEDs, those use some power
 	enginePins.errorLedPin.setValue(0);
 	enginePins.runningLedPin.setValue(0);
 	enginePins.communicationLedPin.setValue(0);
 	enginePins.warningLedPin.setValue(0);
 
-
-	palEnableLineEvent(PAL_LINE(GPIOA, 0), PAL_EVENT_MODE_RISING_EDGE);
```
rusefillc, line 187: let's keep PA0 implementation as stm32f4 discovery implementation?

rusefillc (changes requested):

PR #3758, opened 2022-01-08
Enable ENABLE_AUTO_DETECT_HSE only if STM32_HSECLK is not defined

dron0gus:
Some board (like stm32f429-disco) with SDRAM need clock to be
enabled in __early_init (before .bss and .data initialization).
So if STM32_HSECLK defined - skip clock detection.


rusefillc:
@dron0gus please help h81 compile again

rusefillc:
@dron0gus please help h81 compile again

Issue #3496, opened 2021-11-07
XY idle ve table


BeerMoneyMotorsports:
Idle ve is currently rpm only. It should be rpm  vs load with minimum of 6x6. 

Issue- load changes enough to cause a 3 point afr change at same rpm.  Lights, fan, window switches, etc.

BeerMoneyMotorsports:
@mck1117 do you have time soon to assist me with this?  Hopefully i can educate myself some.

mck1117:
this is done

PR #3781, opened 2022-01-12
XY Idle VE Table

mck1117:
complete #3496 

Miraculously a 4x4 3D table fits in the same hole the old float 2D 8-wide table did!


rusefillc:
@mck1117 poke

mck1117:
release is out, let's go for it

PR #3788, opened 2022-01-13
Randomly skip firmware build jobs

chuckwagoncomputing:
This should satisfy #3739

@rusefillc Note that I didn't use the bash script you wrote. This allows us to do the "dice roll" before checking out the repo.


rusefillc:
https://github.com/rusefi/rusefi/issues/3817 follow-up created

@chuckwagoncomputing just FYI really not the biggest priority

Issue #3777, opened 2022-01-12
flex fuel cranking fuel multipliers


mck1117:
Separate fuel cranking quantity table for E100, interpolated between E0 and E100 tables.

mck1117:
@blundar should it be linear interpolate between E0 and E100 for cranking fuel qty?

mck1117:
@OrchardPerformance and a random 2015 Ford tune downloaded from HPTuners says yes, two tables with linear interpolate is good enough.

PR #3791, opened 2022-01-15
flex cranking math

mck1117:
Separate flex fuel cranking multiplier table, interpolate between the existing table and the new table based on E content.

#3777


rusefillc:
@mck1117 poke?

PR #3795, opened 2022-01-16
does this help?

mck1117:


rusefillc:
nope. not helping so far.

mck1117:
image

how the heck are we 100k bytes overflowed

PR #3797, opened 2022-01-16
Openblt proteus

dron0gus:
Switch to HSI
Enable for F4 and F7 non-special builds.


rusefillc:
@mck1117 would HW CI help here once it gets revived?

PR #3814, opened 2022-01-19
SD card read prefetcher

mck1117:
Part of our performance trouble with the SD card is that we rx a request over USB, request the card, wait for a response (!), read our data, then send it over USB.  That's a lot of waiting and copying while the PC is waiting for us to read a block.

This change prefetches the next block in sequence while we read out the current one.  Since most operations done by the PC are long sequential reads (copying out log files!), this improves performance by having the next block already ready (or at least in progress) when the PC requests it.

Nearly doubles sequential read speed on an F4 from 545kB/s to 925kB/s in my tests.
H7 sees a similar gain, from 645kB/s to 1.1MB/s!


rusefillc:
@mck1117 poke?

mck1117:
@rusefillc I think this broke something, can you revert?

rusefillc:
@mck1117 reverted

Issue #3815, opened 2022-01-19
what's the difference between setRpmValue and assignRpmValue


mck1117:
Why do we have two very similar sounding functions?

PR #3816, opened 2022-01-19
fix always instant rpm option

mck1117:
With this option on, we get stuck in cranking mode.

#3815


rusefillc:
so it really means it's asking for a unit test right?

Issue #3817, opened 2022-01-20
Well, for release builds we want to build everything?


rusefillc:
My perfectionism is asking for release builds to build everything :(

See #3788

PR #3818, opened 2022-01-20
Don't skip if branch contains "release"

chuckwagoncomputing:
#3817

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -194,7 +194,7 @@ jobs:
 
     steps:
     - name: Execution throttle early exit
-      if: ${{ matrix.skip-rate && contains(github.ref, "release") }}
+      if: ${{ matrix.skip-rate && contains(github.ref, 'release') }}
       # todo: how do we produce well-defined releases with skip-rate approach? For I can think of hacking the line below before a release
       run: if (($(($RANDOM % 100)) < ${{ matrix.skip-rate }})); then echo "skip=not_true" >> $GITHUB_ENV; fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-firmware.yaml
```diff
@@ -194,9 +194,9 @@ jobs:
 
     steps:
     - name: Execution throttle early exit
-      if: ${{ matrix.skip-rate }}
+      if: ${{ matrix.skip-rate && !contains(github.ref, 'release') }}
```
rusefillc, line 197: Please add code comments explaining what it does and how? i.e. both user documentation and implementation details?

rusefillc (changes requested):

rusefillc:
@chuckwagoncomputing I am sorry I do not get it :(

My understanding is that all master builds are the same. At the moment my manual release process is to download files from https://rusefi.com/build_server/ and to manually attach those to new GH release

I am wonder if/what is the smarter way to release but at the moment I am not aware of a better process. Is your fix assuming some smarter-than-what-I-do release process?

chuckwagoncomputing:
Oh I'm the one stupid here. I remembered seeing branches in the branch list with "release" in the name, e.g. 2017.02.20_release_1.0.1

rusefillc:
@chuckwagoncomputing yes while creating a release I think GHA demands a tag or branch name so I manually type the name of new branch to create.

Which is horrible since technically it's not atomic - branch is created while creating release and the artifacts are manually uploaded separately.

In my opinion with just monthly releases it's NOT worth to automate it unless someone knows a REALY cheap way to automate this.

For the specific task at hand can we have a manually invoked action which does same firmware like master but with a special environment variable which we would use to remove the randomness?

PR #3835, opened 2022-01-25
ask the user what they want to do

mck1117:
Instead of doing weird things like closing the console after 5 seconds (?), ask the user what they want to do instead.  There are legitimate cases where pressing YES on this dialog is reasonable.

fix #3848

### DIFF
--- a/java_console/io/src/main/java/com/rusefi/io/DfuHelper.java
+++ b/java_console/io/src/main/java/com/rusefi/io/DfuHelper.java
@@ -10,6 +10,7 @@ import com.rusefi.ui.StatusConsumer;
 
 import javax.swing.*;
 import java.io.IOException;
+import java.util.concurrent.CountDownLatch;
 
 import static com.rusefi.Timeouts.SECOND;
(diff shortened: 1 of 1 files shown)


rusefillc:
Please elaborate on what would be a legitimate case here.

My priority is to reduce the chance of bricked unit. I do not trust the users.

NMSTEC:
> Please elaborate on what would be a legitimate case here.
> 
> My priority is to reduce the chance of bricked unit. I do not trust the users.

Incorrectly loaded FW already trying to fix it, testing SW you build yourself.
Or add "advanced user console/option".

I've hit this myself multiple times.

This may be a difference of opinion, but I agree keep users far from destructive things, but there needs to be a limit to that. A dialog saying "YOU MAYBE DO STUPID. DONT DO STUPID. WANNA DO STUPID?" should be plenty enough.

rusefillc:
Incorrectly loaded how? You broke it funny your problem to fix

Considering the cost to unfuck just a warning is not enough. Prevention mechanism has to be much more complex, maybe text entry field like github does in danger zone

PR #3849, opened 2022-01-27
complex wall wetting model

mck1117:
Instead of fixed values, compensate tau/beta for CLT and MAP like the OEMs do

depends on #3848


mck1117:
image

rusefillc:
@mck1117 not green?

Issue #3744, opened 2022-01-06
alphax-2chan


rusefillc:
@andreika-git any help would be appreciated! we have the folder with C code and yaml but nothing is real so far

PR #3856, opened 2022-01-30
alphax 2ch crank hall/vr options

mck1117:
@chuckwagoncomputing is this the right thing to do for pins that have multiple internal functions? These two pins can either be a VR interface, OR two hall inputs. Those functions each use a different STM32 pin.

#3744


rusefillc:
@mck1117 I am pretty sure that the answer is YES

PR #3859, opened 2022-01-31
mitigate F7 flash error by making the image smaller

mck1117:
F7 flash problem doesn't happen if we're smaller than 512K total image size.

Turn off INI drive for now (TS can auto load the ini from the internet...) to save enough flash to get the F7 firmware down under 512K, hiding the erase issue.


rusefillc:
@mck1117 poke?

mck1117:
Looks good to me.

PR #3861, opened 2022-02-01
safer file name

rusefillc:


firmware/util/math/efi_pid.h
```diff
@@ -1,5 +1,5 @@
 /**
- * @file pid.h
```
mck1117, line 2: Comment about why the name?

PR #3863, opened 2022-02-01
we have other places to set these MCU values

rusefillc:


mck1117:
looks like kinetis/cypress `hw_ports.mk` will need `MCU = cortex-m4`

image

rusefillc:
what a mess :( looks like not worth touching at all

mck1117:
what? you were two lines away from it working

rusefillc:
i am not sure that new working state was better than old working state

PR #3864, opened 2022-02-01
remove system lua VM

mck1117:
Looks like we've chosen to do this instead by hard coding this logic, supplementing with user Lua as necessary, exposing state via live data.

#3174


rusefillc:
Any reason why we need this now or just general clean-up?

mck1117:
Flash and RAM savings since a few things have overflowed recently.

rusefillc:
I do not follow. Master is green. Please be specific with the problem statement.

I find system VM potentially useful thus my preference to keep.

PR #3865, opened 2022-02-01
Remove FSIO, byeeeeeeeee

mck1117:
#2928


rusefillc:
please split this PR into a few much smaller ones 🧠 too small

for example it looks like there is a part migrating some logics from FSIO to hard-coded let's PR that separately

mck1117:
small enough for :brain: now?

PR #3867, opened 2022-02-01
dead pwm tester code

mck1117:


rusefillc:
what are the arguments for removal? looks like a useful toolset to assert scheduler precision?

rusefillc:
@mck1117 ^^^^?

mck1117:
oh hi

This file seems to have been disabled and had no changes since at least January 2015, 7 years ago.

It's not actually representative of what we care about for running an engine, For that - it's best to actually just feed a trigger signal and inspect the resulting injection/ignition timing with a scope.

rusefillc:
my preference is to keep this dead code

PR #3872, opened 2022-02-01
enum: add pins for L9779WD-SPI and L9779WD

dron0gus:
BTW, we almost reached brain_pin_e limit as byte.


rusefillc:
we kind of need to address a few bugs in the code generator and we should be able to increase enum size in one click. but only once/if we address bugs in code generator :(

rusefillc:
See #3874

PR #3883, opened 2022-02-02
dead StartupFuelPumping

mck1117:
Whatever this was supposed to do we clearly don't need since startup is really good these days.


rusefillc:
@andreika-git do you remember/care about that feature?

andreika-git (approved):
If primingFuelPulse works correctly, then we don't really need manual pedal fuel pumping.

andreika-git:
btw, we also have "primingSquirtDurationMs" which is not used anymore

Issue #2641, opened 2021-05-07
memory: remove status_loop LOGGING_BUFFER


mck1117:
consuming 1800 bytes, do we need this?

rusefillc:
@mck1117 what do you think, do we need it?

PR #3884, opened 2022-02-02
less dependence on that logger

mck1117:
- refactoring around `printOverallStatus` and friends
- name functions better and add comments to explain what is going on there
- remove dead settings buffer config parameter 

#2641


rusefillc:
@mck1117 poke

Issue #3672, opened 2021-12-08
remaining mock ADC


rusefillc:
We still have some usages of MockAdcState setMockVoltage shall we refactor away to unify sensor mocking?

PR #3888, opened 2022-02-03
remove mock adc

mck1117:
#3672


rusefillc:
same unit tests fails extremely consistently there might be a signal there

any chance to split PR into smaller ones?

mck1117:
I really don't understand how changing two lines in a hardware CI test makes a java unit test fail

rusefillc:
I have little faith in that unit test I am pretty sure that one of these real-tcp-ip tests fails sometimes. Maybe very bad luck strike.

But I am not open to the risk of merging this, maybe smaller PRs are the way to go.

rusefillc:
Amazing it just makes no sense.

mck1117:
I had accidentially checked in `Fields.java` before, which is the only explanation I can come up with

PR #3894, opened 2022-02-05
fix f4 standby

mck1117:
fix https://github.com/rusefi/alphax-2chan/issues/22

Tested with both TJA1051 and TJA1442.


mck1117:
tl;dr you have to clear the wake flag after configuring it if the pin was high when you configured it.

image

mck1117:
update: it still immediately wakes if you're transmitting CAN frames

PR #3900, opened 2022-02-07
support lps22

mck1117:
LPS22 is nearly the same as LPS25, easy to support both by identifying at runtime.

### DIFF
--- a/firmware/hw_layer/sensors/lps25.cpp
+++ b/firmware/hw_layer/sensors/lps25.cpp
@@ -42,8 +42,10 @@ bool Lps25::init(brain_pin_e scl, brain_pin_e sda) {
 	{
 	case expectedWhoAmILps22:
 		m_type = Type::Lps22;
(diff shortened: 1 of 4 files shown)


opelpanfan:
Tested both LPS25 and LPS22 and both work just fine with this change, same 103kPa reading and reacts to changes

firmware/hw_layer/sensors/lps25.cpp
```diff
@@ -9,7 +9,8 @@
 #include "lps25.h"
 
 static constexpr uint8_t addr = 0x5C;
-static constexpr uint8_t expectedWhoAmI = 0xBD;
+static constexpr uint8_t expectedWhoAmILps22 = 0xB1;
```
rusefillc, line 12: please add CHANGELOG entry

rusefillc (changes requested):

PR #3905, opened 2022-02-09
Engine sniffer refinements

mck1117:
- Change layout slightly based on height (don't render too much)
- Slightly improve performance
- Simplify rendering logic for UpDownImage
- Remove gross mouse-over big red text
- Fix line stroke spillage out from dashed lines in to borders of sniffer items
- Render 1 decimal place instead of 2 for angles (who cares about 0.01 degree?)
- Don't overlap row labels (left side) with rendered events


mck1117:
Big:
image
Medium:
image
Small:
image

Issue #3912, opened 2022-02-11
rusEFI console does not stay alive due to CRC error with in outputs


rusefillc:
H7 proteus
master
console would disconnect after a few seconds

rusefillc:
![image](https://user-images.githubusercontent.com/48498823/153529765-777c9faa-05c3-4ab8-8b07-f7007a9a2dcc.png)

rusefillc:
```
Fri Feb 11 10:05:39 EST 2022: onDataArrival alive for 272 second(s) rusEFI 2022.02.11.proteus_f4.1528776780 
Fri Feb 11 10:05:39 EST 2022: onDataArrival alive for 273 second(s) rusEFI 2022.02.11.proteus_f4.1528776780 
Fri Feb 11 10:05:39 EST 2022: onDataArrival alive for 273 second(s) rusEFI 2022.02.11.proteus_f4.1528776780 
Fri Feb 11 10:05:39 EST 2022: onDataArrival alive for 273 second(s) rusEFI 2022.02.11.proteus_f4.1528776780 
```

rusefillc:
Seems separate from #3923

PR #3913, opened 2022-02-11
rusEFI console does not stay alive #3912

rusefillc:
While that's not solving all problems for me this must be fixing out channel concurrency for purposes of CRC calculation


rusefillc:
this change is making h7 connection unreliable as @mck1117 predicted

```
This might cause usb problems on some chips, don't merge yet
``

PR #3915, opened 2022-02-11
executeCommand knows about opcodes

mck1117:
tangential to #3912

### DIFF
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -2004,4 +2004,3 @@ end_struct
 
 #define CAN_ECU_SERIAL_RX_ID 0x100
 #define CAN_ECU_SERIAL_TX_ID 0x102
-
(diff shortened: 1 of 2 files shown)


java_console/io/src/main/java/com/rusefi/binaryprotocol/BinaryProtocol.java
```diff
@@ -590,17 +604,14 @@ private boolean sendTextCommand(String text) {
 
     public static byte[] getTextCommandBytes(String text) {
         byte[] asBytes = text.getBytes();
-        byte[] command = new byte[asBytes.length + 1];
-        command[0] = Fields.TS_EXECUTE;
-        System.arraycopy(asBytes, 0, command, 1, asBytes.length);
-        return command;
+        return asBytes;
```
rusefillc, line 607: looks like cause of #3928

PR #3916, opened 2022-02-11
Chunk outputchannels

mck1117:
Read output channels in multiple chunks, avoid CRC errors from mutable output channels. #3912


mck1117:
Looks good except for core8 build also broken in master.

PR #3917, opened 2022-02-11
move allowLongResponse option to ECU side

mck1117:
We should check this on the ECU, not on the console.


rusefillc:
allowLongResponse on the client side serves as yet another protection against corrupted bytes on physical channels where corruption is possible. Is your argument that CRC covers that already?

mck1117:
CRC ensures we do not parse corrupt packets on the client, correct.

The fact that *sometimes* large packets aren't safe to send is an implementation detail _on the ECU side_, so that's where we should check if we're doing something valid.

PR #3920, opened 2022-02-11
Added some additional flags to show/hide menu items in TS

opelpanfan:
Some additional flags.

#define ts_show_tle8888 true
#define ts_show_rotary true
#define ts_show_gdi true
#define ts_show_experimental true
#define ts_show_fancy_hardware true
#define ts_show_transmission true
#define ts_show_rolling_launch false
#define ts_show_antilag false
#define ts_show_launchcontrol true

### DIFF
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -1998,6 +1998,16 @@ end_struct
 #define ts_show_hardware_simulator true
 #define ts_show_sd_pins true
 
+#define ts_show_tle8888 true
+#define ts_show_rotary true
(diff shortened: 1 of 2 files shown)


rusefillc:
I think we are mixing too much here

8888 is clearly a board attribute

but show_rolling_launch? why do we need fine control over rolling_launch?

opelpanfan:
Not all boards have TLE8888 so thought would be useful to hide that menu, and regards show_rolling_launch that option was hidel anyway.

rusefillc:
too much stuff in same PR. I am unable to review so much things artificially coupled together

PR #3922, opened 2022-02-11
AIM dash TPS, CLT,IAT, and Batt V fix

opelpanfan:
Based on AIM can protocol
https://www.aimtechnologies.com/support/racingecu/AiM_CAN_101_eng.pdf

FIxed TPS, CLT, IAT, Batt V values

Tested on real device and updated 4 values shows correct data.


mck1117:
We shouldn't need to do a multiply while writing a scaled channel. See if you can do it by changing the mult/div parameters instead.

rusefillc:
Add url of pdf as code comment?

PR #3934, opened 2022-02-13
OponBLT: drop cmake required version from 3.15 to 3.10

dron0gus:


rusefillc:
https://github.com/feaser/openblt/pull/10

PR #3960, opened 2022-02-23
Test bench: implement injector test for dead time calculation

dron0gus:
This is preparation to implement TS dialog where user can define on time, off time and iteration count for injector test.

### DIFF
--- a/firmware/console/binary/output_channels.txt
+++ b/firmware/console/binary/output_channels.txt
@@ -287,7 +287,8 @@ uint16_t rpmAcceleration;dRPM;"RPM/s",1, 0, 0, 0, 0
 
     int luaInvocationCounter;;"count",1, 0, 0, 0, 0
     int luaLastCycleDuration;;"nt",1, 0, 0, 0, 0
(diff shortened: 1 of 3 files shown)


mck1117:
this is just an alternative to using rusEFI console command for the same job?

dron0gus:
> this is just an alternative to using rusEFI console command for the same job?

Yes. Just to do the same job using GUI.

mck1117:
I think it might be possible to have a pc variable that sets part of the bytes sent for bench test command, might be worth investigating

dron0gus:
> I think it might be possible to have a pc variable that sets part of the bytes sent for bench test command, might be worth investigating

Need to dig into other's ECU ini files.
But documentation says nothing about this http://tunerstudio.com/index.php/support/manuals/tsdevmanuals/80-command-buttons

mck1117:
> But documentation says nothing about this

yeaaaah, the documentation doesn't say a lot of things....

mck1117:
Had a look at an MS3 ini, and nope, it just sets normal config fields on the ECU then sends a plain (constant) "do the thing" bench test command.

rusefillc:
@dron0gus shall we merge this or do you plan changed or else?

PR #3961, opened 2022-02-24
Bench test clean

dron0gus:
Do not pass strings as arguments where possible
Do not use strings as temp variables to store floats

### DIFF
--- a/firmware/controllers/bench_test.cpp
+++ b/firmware/controllers/bench_test.cpp
@@ -53,9 +53,6 @@
 #endif // BOARD_TLE8888_COUNT
 
 static bool isRunningBench = false;
-// yes that's a bit lazy/crazy to have those as strings
-static char offTimeBuffer[8];
-static char counterBuffer[8];
 
 bool isRunningBenchTest(void) {
 	return isRunningBench;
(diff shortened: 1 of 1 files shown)


firmware/controllers/bench_test.cpp
```diff
@@ -129,153 +126,153 @@ static int count;
...
 
-static void pinbench(const char *delayStr, const char *onTimeStr, const char *offTimeStr, const char *countStr,
-		OutputPin* pinParam, brain_pin_e brainPinParam) {
-	delayMs = atoff(delayStr);
-	onTime = atoff(onTimeStr);
-	offTime = atoff(offTimeStr);
-	count = atoi(countStr);
-
-	brainPin = brainPinParam;
+static void pinbench(float dly, float on, float off, int cnt,
```
mck1117, line 129: no contractions of short words please

dron0gus: All self-explaining names are used by static globals

mck1117:
```
struct state
{
    int count;
    float on;
    .......
};
state benchState;



{
    benchState.count = count;
    benchState.on = on;
     ...........
}
```

dron0gus: `state` can be merged to `BenchController`. But lets do it in next iteration.

PR #3962, opened 2022-02-26
core 8 overrides and stepper-as-brain-pin

mck1117:
Stepper config should be brain pins, as those are internal to ECU, and are only ever STM32 pin names.

Move/add some config to core8 config overrides.


rusefillc:
@mck1117 it seems legit red

mck1117:
and now it's legit green!

PR #3969, opened 2022-02-27
embiggen lua space

mck1117:
We have memory available, let's blow it on LUAAAAAAA!


mck1117:
Because this kind fellow on the forum ran out of memory: https://rusefi.com/forum/viewtopic.php?p=44608#p44608

PR #3975, opened 2022-03-02
better flash logic and warning message

mck1117:
- Use `enum class` instead of oldschool scope-spilling enum
- Don't fire a warning message in case of CRC mismatch when we have blank flash (this is an expected case, why would we warn?)
- minor cleanups


rusefillc:
@mck1117 poke

PR #3976, opened 2022-03-02
Tried to enable USART1 on pins PA9/PA10 but compiler gives error

opelpanfan:
image (1)


mck1117:
Try doing it in `board.mk` instead: https://github.com/rusefi/rusefi/blob/9dc87be35dad83eb1ddf444b713ab7ea444d5a55/firmware/config/boards/proteus/board.mk#L34

opelpanfan:
Same Error 
image

rusefillc:
```
/home/runner/work/rusefi/rusefi/gcc-arm-none-eabi-9-2020-q2-update/bin/../lib/gcc/arm-none-eabi/9.3.1/../../../../arm-none-eabi/bin/ld: build/obj/hal_uart_lld.o (symbol from plugin): in function `VectorD4':
(.text+0x0): multiple definition of `VectorD4'; build/obj/hal_serial_lld.o (symbol from plugin):(.text+0x0): first defined here
collect2: error: ld returned 1 exit status
```

UART hardware could be used by two drivers - serial or uart, but not by both. if you are looking to enable one option you would also need to disable the conflicting other option as well

which i think is STM32_SERIAL_USE_USART1

let me push a useful change first

mck1117:
> let me push a useful change first

see my PR: https://github.com/rusefi/rusefi/pull/3978

rusefillc:
I hope that https://github.com/rusefi/rusefi/commit/4f15fc61797e52f06f1f2f7505a6fca88395c897 would be helpful

you now need to rebase or merge and with that change you should be able to disable serial driver if you want to use UART and not SERIAL

open question if that was even the intention. if it was it's totally the right one - UART is more reliable than SERIAL, at the cost of extra DMA usage which sometimes conflicts with other hardware

mck1117:
@opelpanfan why close? don't you want this merged so that official firmware builds have serial?

opelpanfan:
Sorry, I'm not good with git yet. Done separate PR #3979 as it was for me easier to understand.

mck1117:
> Sorry, I'm not good with git yet. Done separate PR #3979 as it was for me easier to understand.

Ahh no problem, didn't see the new one :)

Issue #3980, opened 2022-03-05
ECU prints unknown command for all commands


mck1117:
ECU prints unknown command even when it's a known command

![image](https://user-images.githubusercontent.com/568254/156882122-bc2d3396-f075-454f-a153-cf6f151ffde5.png)

rusefillc:
@dron0gus could this be related to some of your recent changes?

dron0gus:
Yes. This is my fault. Nothing bad happens, only return code is incorrectly interpreted and error message emitted.
Will fix it on weekend.

mck1117:
Yep, that's why I assigned to you :)

No big deal, just created an issue so we didn't forget about it.

mck1117:
also let's not do this:

![image](https://user-images.githubusercontent.com/568254/156906916-4d40f135-a4a6-42c5-95a6-50968868f1f9.png)

dron0gus:
https://github.com/rusefi/rusefi/pull/3987

PR #3987, opened 2022-03-08
Fix unknown command #3980

dron0gus:


firmware/util/cli_registry.cpp
```diff
@@ -552,10 +552,7 @@ void handleConsoleLine(char *line) {
 
 	int ret = handleConsoleLineInternal(line, lineLength);
 
-	if (ret == 0) {
-		efiPrintf("unknown command [%s]", line);
-		return;
-	} else if (ret < 0) {
+	if (ret < 0) {
```
mck1117, line 555: do we still want to check >0 vs <0 for unknown vs. failed command?

rusefillc:
@dron0gus what do you think about ^^^?

firmware/util/cli_registry.cpp
dron0gus: Unknown command error reported from `handleConsoleLineInternal` and -1 is returned.
`handleConsoleLine` reports toplevel error if command is failed to handle due to any reason. Exact reason is emitted by `handleConsoleLineInternal` or `handleActionWithParameter`

PR #3988, opened 2022-03-10
Launch fuel and spark cut issue

opelpanfan:
Fix to Launch Spark and Fuel Cut.

Two statements contradicts each other 
!limitedSpark && !limitedFuel
and 
limitedSpark and limitedFuel will never be true to match &=


rusefillc:
@opelpanfan please clarify 'draft' state

mck1117 (approved):

mck1117:
I would love a test that fails before this PR, and passes with it.

rusefillc:
Oh I would totally love to see the test as well

@opelpanfan any chance to have a separate PR with a test which proves wrong behavoir?

mck1117:
One option is to move this logic in to `limp_manager.cpp` instead, where we have already validated that it works. We would just have to then write a very simple test to check that limp_manager.cpp requests cut in case of launch.

opelpanfan:
Sorry I'll take a look at how to write tests. Tested on hardware before pushing PR.

opelpanfan:
Done code move to limp_manager.cpp

Sorry dont know how to write tests, but tested on HW.

LimpState LimpManager::allowInjection() const {
	if (!m_allowInjection) {
		return {false, m_allowInjection.clearReason};
	}
	if (!m_transientAllowInjection) {
		return {false, m_transientAllowInjection.clearReason};
	}
	if (engine->launchController.isLaunchFuelRpmRetardCondition()) {
		return {false, ClearReason::LaunchFuel};
	}
	return {true, ClearReason::None};
}

LimpState LimpManager::allowIgnition() const {
	if (!m_allowIgnition) {
		return {false, m_allowIgnition.clearReason};
	}
	if (!m_transientAllowIgnition) {
		return {false, m_transientAllowIgnition.clearReason};
	}	
	if (engine->launchController.isLaunchSparkRpmRetardCondition()) {
		return {false, ClearReason::LaunchSpark};
	}
	return {true, ClearReason::None};
}


Added two new ClearReason's 
enum class ClearReason : uint8_t {
	None, // 0
	Fatal,
	Settings,
	HardLimit, // 3
	FaultRevLimit,
	BoostCut, // 5
	OilPressure,
	StopRequested, // 7
	EtbProblem, // 8
	LaunchSpark, // 9
	LaunchFuel // 10
};

mck1117:
> LaunchSpark, // 9
> LaunchFuel // 10

Don't need two, we can use the same code for both. We separately log fuel/ign cut reason.

mck1117:
> LimpState LimpManager::allowIgnition() const {

please do this in `updateState`

opelpanfan:
ok doing now and update PR

firmware/controllers/limp_manager.cpp
```diff
@@ -75,6 +75,17 @@ todo AndreiKA this change breaks 22 unit tests?
 		allowSpark.clear();
 */
 	}
+	
+	// Fuel cut if launch control engaged
+	if (engine->launchController.isLaunchFuelRpmRetardCondition()) {
```
mck1117, line 80: do we need to check `isLaunchCondition` too?

mck1117: nope, looks like these functions check that internally. Sweet!

opelpanfan: it is already checked as part of bool LaunchControlBase::isLaunchRpmRetardCondition()

firmware/config/boards/microrusefi/connectors/generated_ts_name_by_pin.cpp
```diff
@@ -25,10 +25,10 @@ const char * getBoardSpecificPinName(brain_pin_e brainPin) {
 		case GPIOD_1: return "12 - Ignition 4";
 		case GPIOB_1: return "40 - AN volt 9, Aux Reuse";
 		case GPIOB_0: return "36 - AN volt 8, Aux Reuse";
-		case GPIOB_9: return "AUX J2 PB9";
 		case GPIOD_7: return "14 - GP Out 5";
-		case GPIOB_8: return "AUX J2 PB8";
+		case GPIOB_9: return "AUX J2 PB9";
```
rusefillc, line 29: GRRRRRR

I am very concerned when people do not care what they include in PRs :(

firmware/config/boards/microrusefi/connectors/generated_ts_name_by_pin.cpp
```diff
@@ -25,10 +25,10 @@ const char * getBoardSpecificPinName(brain_pin_e brainPin) {
 		case GPIOD_1: return "12 - Ignition 4";
 		case GPIOB_1: return "40 - AN volt 9, Aux Reuse";
 		case GPIOB_0: return "36 - AN volt 8, Aux Reuse";
-		case GPIOB_9: return "AUX J2 PB9";
```
rusefillc, line 28: this change does not belong in this PR

rusefillc (changes requested):

firmware/config/boards/microrusefi/connectors/generated_ts_name_by_pin.cpp
opelpanfan: I don't think it was me.

opelpanfan: I haven't pushed this on.

[Auto-generated configs and docs](https://github.com/rusefi/rusefi/pull/3988/commits/0ce9438397c00070dcee53641612e8d447f94404)

mck1117: you did push it, because you mis-merged it from master

rusefillc: You are totally right I am just being defensive and aggressive by default please forgive me, some personal drama today

https://github.com/rusefi/rusefi/issues/3989

opelpanfan: how do I fix this?

opelpanfan: change back to         case GPIOB_8: return "AUX J2 PB8";

rusefillc: considering the value of this fix here and identifying https://github.com/rusefi/rusefi/issues/3989 we can just ignore the imperfection it and merge as is

firmware/controllers/limp_manager.cpp
```diff
@@ -140,6 +151,6 @@ LimpState LimpManager::allowIgnition() const {
 	}
 	if (!m_transientAllowIgnition) {
 		return {false, m_transientAllowIgnition.clearReason};
-	}
+	}
```
rusefillc, line 154: @opelpanfan mostly because I am a grumpy old man. Why did we need to change this line?

Please consider reviewing every line you push/PR.

firmware/controllers/limp_manager.h
```diff
@@ -13,7 +13,8 @@ enum class ClearReason : uint8_t {
 	BoostCut, // 5
 	OilPressure,
 	StopRequested, // 7
-	EtbProblem,
+	EtbProblem, // 8
+	LaunchCut // 9
```
rusefillc, line 17: please add a comma so that next person changing here is not changing your line

PR #3997, opened 2022-03-14
TPS AE engages during crank fix #3996

### LINKED ISSUE #3996: TPS AE engages during crank
Expected result: TPS vs TPS to not be active until engine is running.
Actual result: During cranking TPS will move to crank position, which may trip AE.

Conditions: ETB is disabled while engine not spinning.

I am unsure if cranking RPM is overriden by TPS acceleration or it stays static until out of crank condition however due to my engine type, we have long enrichment times, so it may roll over into normal VE table.

image
https://rusefi.com/online/view.php?log=694

rusefillc:


mck1117:
Check if not spinning. We might want AE during cranking to help the engine start or whatever

rusefillc:
@mck1117 @NMSTEC I am confused. I believe ask is NOT to enrich while cranking which is same as spinning?

mck1117:
We have
- not spinning
- cranking 
- Running

We want AE in all but the first one.

PR #3740, opened 2022-01-04
support better wbo format

mck1117:
Support both AEM X-Series, older rusEFI wideband firwmare, and newer rusEFI wideband firmware with better data format (more debug info!)

progress on #2953 and #3233

Issue #3233, opened 2021-09-11
warning code for WBO#0 and WBO#1 error codes


rusefillc:
Do we have a warning for lack of communication with WBO#0 and/or WBO#1?

Do we have a warning for error communicated by WBO#0 and/or WBO#1?

mck1117:
progress on this: https://github.com/mck1117/wideband/commit/4bb9ac6162afb7b7c0b4b8b4cbb8aa8adba9de83

Issue #2953, opened 2021-07-12
CEL to indicate WBO error


BeerMoneyMotorsports:
We need to implement The CEL to relay WBO errors. We have no gauge to assist in this, so CEL implementation is critical. It would be great to tie it to the blue error led so the error will be shown in CEL blinking and can be identified without opening the case. At minimum, flashing CEL for any wbo error is required.

BeerMoneyMotorsports:
This should include a toggle in TS to shut it off for those NOT running internal WBO.

mck1117:
We can do it for some external WBO as well - both AEM xseries and us indicate invalid (not some random AFR reading) when the sensor is warming up, overheated, or missing.  As for our WBO controller, internal is exactly the same as external.

BeerMoneyMotorsports:
@mck1117 poke poke.

BeerMoneyMotorsports:
is this on the docket at all?  @rusefillc @mck1117

PR #4000, opened 2022-03-14
support better wbo format

mck1117:
try again on #3740 

Support both AEM X-Series, older rusEFI wideband firwmare, and newer rusEFI wideband firmware with better data format (more debug info!)

progress on https://github.com/rusefi/rusefi/issues/2953 and https://github.com/rusefi/rusefi/issues/3233


rusefillc:
@mck1117 poke

mck1117:
let's do it

PR #4005, opened 2022-03-17
aux linear sensors

mck1117:
Aux linear sensors, like how we have aux temp sensors. I have a coolant pressure sensor, Jimmy has an extra pressure sensor post-supercharger, etc. Sometimes you need some extra linear sensors to get logged.


mck1117:
Android build is broken because there was a unilateral decision to unnecessarily couple all components to one another as tightly as possible.

rusefillc:
I am cherry picking into master to help this specific PR

and #4016 to address root cause of the drama

PR #4007, opened 2022-03-18
Fresh GCC 10

rusefillc:


rusefillc:
@mck1117 note how only Proteus F7 fails?!

rusefillc:
@mck1117 any interest?

PR #4023, opened 2022-03-24
output channel tweaks

mck1117:
- better boost controller logging
- scaled channels for PID status (saves space)
- remove dead MAP cam fields
- fix signed/unsigned and limits for a few channels
- add boost PID state
- pass pid_status_s as a ref instead of ptr
- add units/limits to some channels


firmware/console/binary/output_channels.txt
```diff
@@ -260,7 +260,7 @@ uint16_t rpmAcceleration;dRPM;"RPM/s",1, 0, 0, 0, 0
 	int16_t autoscale accPedalSplit;;"%",{1/@@PACK_MULT_PERCENT@@}, 0, 0, 0, 0
 
     int16_t autoscale m_requested_pump;"DI: Pump Angle";"deg",{1/@@PACK_MULT_ANGLE@@}, 0, 0, 0, 0
-    int16_t autoscale boostControlTarget;"Pump Angle";"deg",{1/@@PACK_MULT_PRESSURE@@}, 0, 0, 0, 0
```
mck1117, line 263: this typo was a real WTF

mck1117:
The failures are duplicates of good builds, will self-heal in `master`. Looks green to me.

PR #4024, opened 2022-03-28
shall we?

rusefillc:
@mck1117 opinions?


mck1117:
this is undefined behavior - the pointer you pass points to the stack but is persisted later. We don't need the header though, we can just move the class definition to the cpp file instead.

mck1117:
Actually, I'd prefer no change at all. It becomes impossible to write granular unit tests if we make the declaration for the converter inaccessible.

PR #4029, opened 2022-03-31
prototype overcurrent protected driver

mck1117:
Implement it as a gpio chip, since we need to inject extra logic between "set pin to on" and actually flipping the output, and the gpiochip api is what we have to do that :)

We could also in theory use this for ECU outputs if we wanted to roll our own output protection (or have some interestingly shaped trip curve).


rusefillc:
We now have 2 byte pin IDs let's not PI hack? https://github.com/rusefi/rusefi/commit/a90f8060736af926c93a17a3be84ac759b1c0981

firmware/controllers/algo/rusefi_hw_enums.h
```diff
@@ -279,33 +279,28 @@ enum class Gpio : uint16_t {
 	CAN_PIN_6 = 256,
 	CAN_PIN_7 = 257,
 
-	PROTECTED_PIN_0 = 290,
```
mck1117, line 282: I don't think good things happen if we put a gap here.

PR #4043, opened 2022-04-04
Flex fuel cranking improvements

### LINKED ISSUE #4041: why CUSTOM_ERR_ZERO_CRANKING_FUEL?
image

no start at the beginning of this log:
[no idle.zip](https://github.com/rusefi/rusefi/files/8405553/no.idle.zip)

using this tune:
https://rusefi.com/online/view.php?msq=935

mck1117:
Fix #4041

- put a default in the E85 table
- Add an enable bit for flex cranking logic instead of just enabling by default
- Interpolate to E85 instead of E100, since E85-100 doesn't really exist anyway.
- Warn for zeroes in those tables.
- Correctly interpolate based on flex sensor result


mck1117:
❌ will self heal after merge

rusefillc:
Why do we need the new flag? I forgot if that's to follow some pattern or if that makes good sense.

mck1117:
New flag lets you add a flex sensor without having to worry about it changing things. Second table isn't necessary if you're running JUST E10 or JUST E85, only necessary if you're actually flex fueling.

PR #4048, opened 2022-04-06
Add interface for selecting GearController and TransmissionController classes

chuckwagoncomputing:
I'm PRing this before continuing work because I've never built something with selectable classes before, and I'm not sure if this is the proper way to architect it.


chuckwagoncomputing:
I forgot to mention, this also adds a GM 4L6X transmission controller which for now is just a duplicate of SimpleTransmissionController.

PR #4064, opened 2022-04-11
overhaul some docs surrounding rusefi_config.txt

chuckwagoncomputing:
Review would be appreciated, this reflects my workflow, but it might not reflect everyone's.

### DIFF
--- a/firmware/rusefi.cpp
+++ b/firmware/rusefi.cpp
@@ -91,6 +91,7 @@
  * Tunerstudio output channels are handled completely manually, as opposed to settings where offsets are managed automatically by the gen_config scripts.
  * Adding channels to tunerstudio_outputs.h induces offsets which are kept track of in comments next to each channel.
  * In the OutputChannels section of rusefi.input, these values are 3rd token in the configuration string.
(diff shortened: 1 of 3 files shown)


firmware/rusefi.cpp
```diff
@@ -84,17 +84,28 @@
  * 2.777ms, so we schedule spark firing at '2.777ms after the 690 position sensor event', thus combining trigger events
  * with time-based offset.
  *
+ * @section tunerstudio Getting Data To and From Tunerstudio
+ *
+ * Contains the enum with values to be output to Tunerstudio.
+ * console/binary/tunerstudio_outputs.h
+ * Tunerstudio output channels are handled completely manually, as opposed to settings where offsets are managed automatically by the gen_config scripts.
```
mck1117, line 91: this is no longer true, have a look at `output_channels.txt`!

chuckwagoncomputing: And offsets aren't handled manually anymore?

mck1117: make a change in `output_channels.txt`, run `gen_output_channels.sh` and see the magic!

chuckwagoncomputing:
@rusefillc I guess this is good to go if it looks good to you.

PR #4068, opened 2022-04-15
allow setting injection phase by start or center of injection

mck1117:
Currently we schedule by end-of-injection, as is the default. It's useful to be able to tune GDI by start-of-injection, so allow that (and center, while we're at it) too.

Also, uh, diesel needs SOI, but we don't talk about that.


rusefillc:
maybe a test?

OrchardPerformance:
Well to make things complicated, there is an argument for DI that scheduling EOI and not SOI gives better results. You can start injection as early as you like as long as the exhaust is shut
Tuning EOI gives you control over the amount hitting the piston at the last stage of injection and the evaporation time available for the final fuel injected

by scheduling EOI you can set a base table of EOI that is pretty flat due to only mechanical concerns and the spark timing as the boundary and then let the ECU do what the hell it wants in terms of picking an SOI. 
Combine with a duty cycle counter to tell you when you need to up the pressure and a chosen arbitrary limit near TDC for the start boundary and it simplifies the tuning of DI a bit.

Issue #4073, opened 2022-04-16
Users are confused about FATAL error and reboot


rusefillc:
rusEFI FATAL is only reset on full reboot

We clearly have users confused about need to reboot, we clearly have users not realizing that TS reconnect or USB reconnect sometimes does not cause reboot

rusefillc:
![image](https://user-images.githubusercontent.com/48498823/163688968-46bc44bd-bbc2-42c7-be2c-7745d2ff7962.png)

rusefillc:
idea1: show uptime in fatal message, get fatal message to tick

idea2: auto-reboot every 120 seconds after fatal? Very not sure about that idea2

PR #4074, opened 2022-04-16
Users are confused about FATAL error and reboot #4073

rusefillc:
looks like a low hanging usability improvement but does not work :(


rusefillc:
image

rusefillc:
i got this to _sometimes_ work but sometimes I get no error message at all and that does not look right

also value is not ticking, could be that TS just keeps displaying the first error it has received?

image

PR #4075, opened 2022-04-16
More fragment refactoring: encode by type, not enum index

mck1117:
One overload of `getLiveDataAddr` per live data object, then all we have to do is call them in a big list to get pointers. No ugly cast. Plus, this opens the door to the list being `constexpr`, which can save some memory/cycles (any use of `reinterpret_cast` generally excludes you from being `constexpr`).

The generated file just looks like this now:
image

Then we can remove the big `switch(...)` based `getStructAddr`.


rusefillc:
reminder that we will soon have two ETBs and two of something else

mck1117:
> reminder that we will soon have two ETBs and two of something else

Fans. We have two fans as well.

No reason we couldn't encode somehow in the generation side "we have N of this thing", then it generates
```
getLiveDataAddr<Thing>(0),
getLiveDataAddr<Thing>(1),
```

rusefillc:
that's a bit too fancy refactoring too quickly, please let me finish the feature using my legacy ways

PR #4076, opened 2022-04-17
abandoned new parser

rusefillc:
this code was never hooked into unit tests and PROD workflow, is it now time to drop?


mck1117:
That code still runs - it's still doing the stricter antlr-based syntax checking on the config file.

I would love to let it replace the existing parser.

rusefillc:
does not run at the moment I've commented it out to make my life easier, seemed liked the right way considering how it does not produce anything.

Let's keep for rest of 2022 and delete if still unused in 2021

PR #4077, opened 2022-04-17
knock threshold table fix

mck1117:
- fix knock threshold table (why were there two, we only read one?)
- improve knock gauges
- make `KnockController` an engine module


mck1117:
>  java.lang.IllegalStateException: todo: implement default tsInfo for long form

this smells like a failure not due to a change here, so I think this is actually green

rusefillc:
master is green @mck1117 please rebase

PR #4078, opened 2022-04-18
set SHORT_BOARD_NAME from mk, not shell script

mck1117:
We have makefiles for these things, let's not recreate make's functionality in shell scripts.

This makes the change for non-Hellen boards.


rusefillc:
it's legit red @mck1117

PR #4081, opened 2022-04-18
pass parameters instead of `export` in build scripts

mck1117:


rusefillc:
is this a legit red?

mck1117:
look it's self healing

PR #4093, opened 2022-04-23
More untested transmission features

chuckwagoncomputing:
Use ISS for gear detector if it is present
Measure shift time
Use separate pressure control curves for shifts
Apply 3-2 shift solenoid during the 3-2 downshift
Wrapped TCU stuff in `#if EFI_TCU` statements
Added gauge for torque converter ratio, but haven't added measurement for that yet


rusefillc:
Output_channels are now a legacy way to gauge

Pretty much the 'views'/models/livedata.yaml idea that we split the pile of state into smaller domain specific states

PR #4104, opened 2022-04-25
allow enum_class in rusefi_hw_enums.h

mck1117:
With `enum class X` we can force a particular size for an enum, instead of relying on the (insane) method of dropping some huge value in there.


rusefillc:
Wow 82 files. Thank you for progress and unification!

PR #4108, opened 2022-04-26
Brain pin is enum class

mck1117:
Make `brain_pin_e` an `enum class`, instead of c-style `enum`. This lets us explicitly size it without any magic - it's now forced to be a uint8_t.

This is 90% a regex find-replace of `GPIO([A-Z])_([0-9]+)` -> `Gpio::$1$2`, with a few real edits to the config generator and a few functions that touch this enum.


rusefillc:
scary changes to scripts

image

mck1117:
> scary changes to scripts

Works fine.

image

rusefillc:
ohh it's pretty green?

rusefillc:
we have a conflict with 4chan fix sorry

PR #4118, opened 2022-04-28
h128 i2c?

rusefillc:


shadowm60 (approved):
It looks ok for now but if we would have baro lps25 we would need deinit after a read, or check in i2c if the same pin and function is requested.

PR #4120, opened 2022-04-30
Next steps

rusefillc:


firmware/controllers/algo/rusefi_enums.h
```diff
@@ -392,10 +391,9 @@ typedef enum {
 
 	ES_AEM = 6,
 
-	Force_4_bytes_size_ego_sensor = ENUM_32_BITS,
 } ego_sensor_e;
 
-typedef enum {
+typedef enum __attribute__ ((__packed__)) {
```
mck1117, line 396: enum class?

rusefillc: yes but trying to get 80% results at 20% of the effort

mck1117: except that what you've done may have undefined behavior

PR #4136, opened 2022-05-04
set MRE overrides for spi2/3 pins

mck1117:


rusefillc:
Does this hard-wire AUX GPIO extension port to be SPI3? Because hard-wiring extention port to be always SPI does not seem right?

We can do a lot with _defaults_ but let's be careful with _override_ as it's un-reversible?

mck1117:
We can hard-wire the pins, but you're right, we shouldn't actually enable SPI unless the user needs it.

firmware/config/boards/microrusefi/board_configuration.cpp
```diff
@@ -208,16 +212,14 @@ void setBoardDefaultConfiguration() {
 	setupDefaultSensorInputs();
 
 
-	// SPI for SD card
```
rusefillc, line 211: Shall we keep spi2=true and improve the comment to "SPI for official 0.6.0 SD card"?

since that's now part of official HW i think we want spi2 enabled by default

mck1117: it's enabled in the override

firmware/config/boards/microrusefi/board_configuration.cpp
```diff
@@ -163,16 +162,21 @@ void setBoardConfigOverrides() {
 	engineConfiguration->canTxPin = Gpio::B6;
 	engineConfiguration->canRxPin = Gpio::B12;
 
+	// SPI2 for onboard SD card
+	engineConfiguration->is_enabled_spi_2 = true;
```
mck1117, line 166: @rusefillc spi2 enabled here

PR #4140, opened 2022-05-06
Trigger gm 3 vs 5

mck1117:
Somehow my engine has 5/10 degree gaps, while Mark's engine has 3/12 degree gaps. I don't know why.


rusefillc:
See also

https://www.mediafire.com/?40mfgeoe4ctti
http://www.ls1gto.com/forums/archive/index.php/t-190549.htm
http://www.ls2.com/forums/showthread.php/834483-LS-Timing-Reluctor-Wheels-Explained

PR #4152, opened 2022-05-09
cam/crank sync improvements

mck1117:
- ~~Clarify `Engine::getOperationMode` and `doesTriggerImplyOperationMode`~~ now #4153
- Don't switch from wasted -> sequential until we get VVT phase information
- ~~Set a maximum RPM above which we don't try to use the cam to help sync the crank~~ now #4156 
- ~~Encapsulate `vvtSyncCounter`~~ now #4154

related #4099

PR #4162, opened 2022-05-10
move logic in to trigger decoder

mck1117:
A bunch of logic was in `engine.cpp` that really should be part of the trigger decoder itself.

This PR moves it in.

Unblocks #4152 being able to reset the "full sync" flag when we get a trigger error.


mck1117:
Oops red

PR #4164, opened 2022-05-11
don't rely on triggerStateListener to detect trigger errors

mck1117:
Check that various sync problems cause `getShaftSynchronized()` to return false.


rusefillc:
once/if we are happy with this change can we merge test-only or mostly-test part first and then merge production code as separate merge?

mck1117:
> once/if we are happy with this change can we merge test-only or mostly-test part first and then merge production code as separate merge?

yeah, I already had to break part off in to #4165

rusefillc:
This PR has a functional change at the moment? Split one green PR into two green PRs?

unit_tests/tests/trigger/test_cam_vvt_input.cpp
```diff
@@ -141,8 +140,8 @@ TEST(trigger, testNB2CamInput) {
 
 	int totalRevolutionCountBeforeVvtSync = 5;
 	// need to be out of VVT sync to see VVT sync in action
-	eth.fireRise(25);
-	eth.fireRise(25);
+	eth.fireRise(25 * 70 / 180);
```
mck1117, line 143: Hah! Mistake in the test caught by this change.

PR #4182, opened 2022-05-18
properly print extended CAN ids

mck1117:


mck1117:
the one red is a fluke, looks good to me

PR #4190, opened 2022-05-19
pwm uses correct channels

mck1117:
ugh, made a mistake in #4188


firmware/hw_layer/ports/stm32/stm32_common.cpp
```diff
@@ -287,12 +287,12 @@ static expected<stm32_pwm_config> getConfigForPin(brain_pin_e pin) {
 	}
 };
 
-stm32_hardware_pwm pwms[5];
+static stm32_hardware_pwm hardPwms[5];
```
mck1117, line 290: this is just for good measure since there was maybe another global symbol called `pwms`

PR #4194, opened 2022-05-24
honda trigger sync ratio

mck1117:
improved with data from real car


rusefillc:
Add code comments clarifying usage on honda obd2? Move to honda-specific file? Just ideas

mck1117:
these are appropriate changes for everybody with a 12t crank - not just Honda. There isn't anything Honda-special about this change.

Issue #4195, opened 2022-05-24
Support individually-wired wasted spark on odd cylinder engines (especially 3cyl)


ElDominio:
Add support for individually-wired wasted spark on odd cylinder engines, current example case:

Yamaha FZ09 engine
3cyl
24-2 crank trigger
3 ignition coils

Configured as 3 cylinders fires strangely as expected, had to use as a V6 to get it to work

rusefillc:
@ElDominio can you draw a picture of explain how does wasted spark odd cylinders should work?

ElDominio:
Think V6 but lop off one bank

rusefillc:
![IMG_20220526_152913](https://user-images.githubusercontent.com/48498823/170563061-7fc9afe3-6a99-4525-a3d0-0f1461139622.jpg)

mck1117:
Every output fires every 360 degrees. For a 1-2-3 firing order engine, output 1 fires at angle 0, output 2 at 120, and output 3 at 240. Output 1 fires again at 360, then 2 at 480, then 3 at 600.

ElDominio:
Does this worl? Mitsubicha runs fine on seuqnetial but have not tested individually wired wasted

ElDominio:
@rusefillc does this need testing or something?

rusefillc:
related to #6868

rusefillc:
problem: original implementation was focusing on engines with even number of cylinders and half the number of coil, like your miata NA/NB. Wasted spark with individual coils was always a hack build on top of that original notion of half-the-cylinders-number-of-coils

in reality individual coils can be used in wasted spark regardless of number of cylinders and even on odd fire two cylinder engines, again as long as induvial coils

one of key parts in the code
```
	// In spin-up cranking mode we don't have full phase sync info yet, so wasted spark mode is better
	// However, only do this on even cylinder count engines: odd cyl count doesn't fire at all
	if (ignitionMode == IM_INDIVIDUAL_COILS && (engineConfiguration->cylindersCount % 2 == 0)) {
		bool missingPhaseInfoForSequential =
			!engine->triggerCentral.triggerState.hasSynchronizedPhase();

		if (!engineConfiguration->oddFireEngine && (engine->rpmCalculator.isSpinningUp() || missingPhaseInfoForSequential)) {
			ignitionMode = IM_WASTED_SPARK;
		}
	}
```

rusefillc:
[...]

Issue #1594, opened 2020-07-11
Cranking -> Running transition is a mess


mck1117:
Cranking and running modes are both sequential, and this mess happens during the transition from cranking to running:

![image](https://user-images.githubusercontent.com/568254/87221391-91c1fa00-c320-11ea-955f-0c84f29d3b07.png)

mck1117:
Note that the duration of these two intervals is approximately the same:
![image](https://user-images.githubusercontent.com/568254/87221413-bc13b780-c320-11ea-8816-dd59bfd0b13e.png)

What's happening is that the first small injection is scheduled, then the injection phase is updated, then at the next tooth its scheduled AGAIN.  Since we target the end-of-injection timing, the first injection got scheduled expecting the duration of the longer cranking pulse, not the shorter running pulse.

rusefi:
Poke @mck1117

rusefillc:
@mck1117 what is the state of this issue?

mck1117:
there are more doubled injections here: https://github.com/rusefi/rusefi/issues/3275#issuecomment-927444156

rusefillc:
@mck1117 poke :)

rusefillc:
@mck1117 I assume you were busy for the last five months?

Closing since no one cares to use many words, not actionable as is.

mck1117:
> no one cares to use many words

huh?

mck1117:
This is still definitely a bug.

mck1117:
It happens roughly like this:

1. Long injection scheduled, so we start the injection very early to hit the requested end-of-injection time.
2. Injection duration becomes very short, so it starts when the long injection would've, but then completes very early.
3. New start time based on the short injection hasn't happened yet, so we start it again, getting a doubled injection.

rusefillc:
Related to #4507

PR #4208, opened 2022-05-28
Injector scheduled by angle, not tooth index

mck1117:
Instead of pre-computing which tooth should schedule injector opening, just check it using the current engine phase angle.  Fully decouples injection logic from any concept of the trigger, only coupling it to "you are at phase angle X, next call will be at phase angle Y".

builds on #4206 and #4207

tiny incremental progress towards making #4195 easier to implement

potentially related to #1594


rusefillc:
@mck1117 poke

mck1117:
This has made the bug in our scheduling more obvious, but not made it any worse. I think it's OK to merge for now.

PR #4213, opened 2022-05-30
remove trigger duty cycle calculation

mck1117:
We calculated trigger duty cycle, and then ignored it. Not even sure what trigger duty cycle was supposed to be useful for, and it hadn't been touched in 7 (!) years.


rusefillc:
i guess the dream was to use this for trigger validation until VR has destroyed that dream

PR #4218, opened 2022-05-30
actually run the new parser

mck1117:
what if we used it instead of deleting it #4214 

Write a bunch of unit tests for the parser part of it, layout/output coming soon.


rusefillc:
I see zero usages this PR just revives dead code, what's the point?

Please integrate the dead code into existing unit test coverage

rusefillc:
ps: also merge conflict it says

mck1117:
> I see zero usages this PR just revives dead code, what's the point?

That's not quite true - just parsing the code with the antlr-generated parser has side effects. That side effect being that if the input file doesn't match the grammar, it will throw an error and end the program.

rusefillc:
Well, yes. We have the benefit of running parser, but that's just a piece of a useful implementation. Can we please push more than that?

mck1117:
So far just tests for the parser, not the layout or output logic yet.

image

PR #4219, opened 2022-05-30
Write output channels with new parser

mck1117:
depends #4218 

Tested with unit tests, and that it outputs a bit-identical output_channels.ini except that it doesn't output any padding bits (like `unusedBit_230_8`, for example).


rusefillc:
why do we kill percent_t usability and replace it with copy-pasta?

mck1117:
because currently it generates like this, as an enum (wtf?):

image

PR #4220, opened 2022-05-31
test new parser TS layout

mck1117:
Mostly test the layout logic by looking at the TS output.

### DIFF
--- a/java_tools/configuration_definition/src/main/java/com/rusefi/newparse/layout/ArrayLayout.java
+++ b/java_tools/configuration_definition/src/main/java/com/rusefi/newparse/layout/ArrayLayout.java
@@ -58,7 +58,7 @@ public class ArrayLayout extends Layout {
 
     @Override
     public String toString() {
-        return "Array of " + this.prototypeLayout.toString() + " length " + this.length[0] + " " + super.toString();
+        return "Array of " + this.prototypeLayout + " length " + this.length[0] + " " + super.toString();
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


java_tools/configuration_definition/src/main/java/com/rusefi/newparse/layout/ArrayLayout.java
```diff
@@ -58,7 +58,7 @@ public void setOffsetWithinStruct(int offset) {
 
     @Override
     public String toString() {
-        return "Array of " + this.prototypeLayout.toString() + " length " + this.length[0] + " " + super.toString();
+        return "Array of " + this.prototypeLayout + " length " + this.length[0] + " " + super.toString();
```
rusefillc, line 61: why would you "this." here?

java_tools/configuration_definition/src/main/java/com/rusefi/newparse/layout/StructLayout.java
```diff
@@ -4,12 +4,11 @@
 import com.rusefi.newparse.parsing.*;
 
 import java.io.PrintStream;
-import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.List;
 
 public class StructLayout extends Layout {
-    /*private*/public List<Layout> children = new ArrayList<>();
+    /*private*/public final List<Layout> children = new ArrayList<>();
```
rusefillc, line 11: ``/*private*/public`` is that some weird MS style?

mck1117: I think those were intended to be private, but I made them public for some reason...

PR #4226, opened 2022-06-01
Remove third trigger channel

mck1117:
All real cases that would have needed the third channel are covered by the cam/vvt inputs instead.

Also remove the four different modes that are all different ways of interpreting the same Honda 24+4+1 distributor, but are inferior to the new "12 tooth crank" + "single tooth cam" method of doing it.


rusefillc:
Great stuff! Please CHANGLELOG trigher remival

mck1117:
I think it's happy this time...

mck1117:
Looks like the two red will self-heal.

Issue #4200, opened 2022-05-26
USB TS update rate


rusefillc:
At the moment my MRE via mini USB shows 249 which is too close to 250 and is not 999 as I recall on some record setting picture. Do we expect more than 249?

![image](https://user-images.githubusercontent.com/48498823/170396164-7a93ecae-ef78-4f42-8107-b3fe4b926619.png)

![image](https://user-images.githubusercontent.com/48498823/170396095-e30c8a1c-6ea1-4afb-8e16-2a37f7101504.png)

rusefillc:
@mck1117 says

```
output channels are ~4x blocking factor
some limitation (not us, I don't think, maybe windows, maybe usb?) limits to ~1000 USB transactions/second
```

rusefillc:
We clearly saw 999 which was clearly not random

![image](https://user-images.githubusercontent.com/48498823/171502801-4ec52a26-a8bd-45c8-ad55-ac4550195fc0.png)

rusefillc:
@mck1117 i believe there was an update recently? can we link?

mck1117:
yes, blocking factor was increased which got us back to ~500hz over USB #4703

PR #4227, opened 2022-06-01
500 byte blocking factor

mck1117:
#4200, this gets us 500hz. I tried even bigger and it didn't get faster, so I'm not sure where the perf went. It doesn't seem to be the ECU since it's pinned at exactly 500.0/499.9.


mck1117:
tested on F4 and H7, seems OK to me

PR #4111, opened 2022-04-26
new autoscale constraint

rusefillc:
```
$ ./gen_config.sh | grep RRRR
mkdir: cannot create directory ‘build’: File exists
RRRRRRRRRRRRRRRR Unexpected scale of 0.1 without autoscale on ConfigField{name='timeConstant', type='uint16_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.1 without autoscale on ConfigField{name='deadband', type='uint8_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.1 without autoscale on ConfigField{name='minAfr', type='uint8_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.1 without autoscale on ConfigField{name='maxAfr', type='uint8_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.1 without autoscale on ConfigField{name='knockRetardAggression', type='uint8_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.1 without autoscale on ConfigField{name='knockRetardReapplyRate', type='uint8_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.1 without autoscale on ConfigField{name='dwellVoltageCorrVoltBins', type='uint8_t', arraySizes=[8]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.02 without autoscale on ConfigField{name='dwellVoltageCorrValues', type='uint8_t', arraySizes=[8]}
RRRRRRRRRRRRRRRR Unexpected scale of {1/1000} without autoscale on ConfigField{name='applyNonlinearBelowPulse', type='uint16_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.001 without autoscale on ConfigField{name='multisparkSparkDuration', type='uint16_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.001 without autoscale on ConfigField{name='multisparkDwell', type='uint16_t', arraySizes=[]}
RRRRRRRRRRRRRRRR Unexpected scale of 5 without autoscale on ConfigField{name='tchargeBins', type='uint8_t', arraySizes=[16]}
RRRRRRRRRRRRRRRR Unexpected scale of 0.01 without autoscale on ConfigField{name='tchargeValues', type='uint8_t', arraySizes=[16]}
[...]

PR #4233, opened 2022-06-03
use yaml like you're supposed to

mck1117:
Let's use yaml like it was intended, instead of parsing an array (?)


rusefillc:
c'mon can we please address https://github.com/rusefi/rusefi/pull/4111 before this? :)

Issue #4259, opened 2022-06-18
verbose trigger sync should reuse engineSnifferRpmThreshold


rusefillc:
we cannot afford verbose trigger at high RPM same as we cannot afford engine sniffer at high RPM

we shall apply same engineSnifferRpmThreshold logic and setting into verbose trigger sync

mck1117:
Do we even need verbose trigger print out? We log the same sync data to TS log at ~500hz already...

rusefillc:
We definitely do not log _all_ the info from console verbose lines. For instance #4172

PR #4260, opened 2022-06-18
no verbose trigger sync print out

mck1117:
We have better data in TS already, why print it out to text log that hurts performance?

#4259


rusefillc:
I trust your intentions but you are factually incorrect here.

PR #4261, opened 2022-06-18
new parser TS output fixes, config file fixes

mck1117:
- Use double instead of float to avoid loss of precision
- Fix some config file inconsistencies that cause output differences
- Yaml parser supports new parser
- Parse enum bit count correctly


rusefillc:
please rebase to get happy green

Issue #455, opened 2017-06-29
RPM hard limit depending on CLT


rusefi:
Variable RPM hard limit depending on engine temperate would be a nice feature right from Lexus LFA

See https://www.youtube.com/watch?v=PopzpO-Iu94

AlexHLinS:
I've discovered the idea and current map's set. 
Can we use the Warmup fuel multiplier for these? The model will be: [RPM Limit] = [RPM hard limit] / [CLT Multipler] (e.g. engine has a red line at 7000 rpm, CLT multiplier at 0 Celsius equal 1.28 => rpm limit set to 7k / 1.28 = 5,5k rpm) 
I  offer to add to the "LimpManager" method named as setCLTRevLimit(bool CLRRpmTag). This one will check if current m_faultRevLimit, and set it to calculated only if calculated is less than existing.
If you applying my idea, I'll go be going to do this.

rusefillc:
i would not mix Warmup and hard limit, those are separate things to be configured separately

we need a new curve here, could be a small 4 element curve

AlexHLinS:
Ok, got it.

mck1117:
with #4267, this is ddddone!

PR #4267, opened 2022-06-22
clt based rpm limit

mck1117:
implement #455

image


mck1117:
Everything failed will self heal.

PR #4269, opened 2022-06-22
logic analyzer uses exti

mck1117:
ICU must die!

relevant to #3032


firmware/development/logic_analyzer.cpp
```diff
@@ -88,44 +82,42 @@ void WaveReader::onFallEvent() {
...
-		reader->hw = nullptr;
+		reader->line = 0;
 		return;
 	}
-		
 
 	reader->name = name;
 
-	reader->hw = startDigitalCapture("wave input", brainPin);
+	reader->line = PAL_LINE(getHwPort("logic", brainPin), getHwPin("logic", brainPin));
```
rusefillc, line 114: c'mon

the smart way would have been to use the suggested universal input caputure API

mck1117: ....an api that doesn't exist.

PR #4272, opened 2022-06-23
clean up sd overrides

mck1117:


firmware/config/boards/hellen/hellen128/board_configuration.cpp
```diff
@@ -148,7 +148,14 @@ static void setHellen128ETBConfig() {
 void setBoardConfigOverrides() {
 	setHellen176LedPins();
 	setupVbatt();
-	setSdCardConfigurationOverrides();
+
+	engineConfiguration->sdCardSpiDevice = SPI_DEVICE_2;
+	engineConfiguration->spi2mosiPin = H_SPI2_MOSI;
```
rusefillc, line 153: we shall extract a method, at least for hellen boards? one for spi2 another for spi3

mck1117: yep, I noticed that too.

PR #4280, opened 2022-06-24
fix hellen na6 sd card config

mck1117:


mck1117:
red will be fixed by #4279, this one is green

PR #4287, opened 2022-06-26
new parser writes datalogs

### LINKED ISSUE #4451: invalid ini
image

mck1117:
#4441 let's move forward
fixes #4451


rusefillc:
should this or something be deleting legacy code already?

mck1117:
> should this or something be deleting legacy code already?

We could do that.

firmware/console/binary/output_channels.txt
```diff
@@ -68,7 +68,7 @@ uint16_t rpmAcceleration;dRPM;"RPM/s",1, 0, 0, 5, 0
 	int16_t autoscale vvtPositionB1I;@@GAUGE_NAME_VVT_B1I@@;"deg",{1/@@PACK_MULT_ANGLE@@}, 0, 0, 0, 0
 
 	uint16_t autoscale chargeAirMass;@@GAUGE_NAME_AIR_MASS@@;"g",{1/1000}, 0, 0, 0, 0
-	uint16_t autoscale crankingFuelMs;crankingFuelMs\nairmass in mg, 0-65 grams;"ms",{1/@@PACK_MULT_FUEL_MASS@@}, 0, 0, 0, 0
+	uint16_t autoscale crankingFuelMs;;"ms",{1/@@PACK_MULT_FUEL_MASS@@}, 0, 0, 0, 0
```
rusefillc, line 71: separate RP?

rusefillc:
My focus right now is on https://github.com/rusefi/rusefi/commit/ae8eb3073a4ecf0c4f056344e10286c8e0f09c05

rusefillc (changes requested):
Please extract output_channels.txt change into separate PR

rusefillc (changes requested):
I believe that this PR should be removing TsOutput

mck1117:
> My focus right now is on [ae8eb30](https://github.com/rusefi/rusefi/commit/ae8eb3073a4ecf0c4f056344e10286c8e0f09c05)

Yes, that should be pretty easy shortly.

mck1117:
> I believe that this PR should be removing TsOutput

that's still in use by the normal config generation, but not by live data

firmware/console/binary/output_channels.txt
mck1117: removed for now, will clean things like this later

rusefillc:
I am failing to see the value of this change without TsOutput removal - as it it's introducing descepancy between two extremely similar tasks. I believe that TsOutput removal is the bare minimal of newparse if you choose to go with live data as first newparse usage.

mck1117:
I don't understand your question. The live data tool as modified in this PR makes no call to TsOutput, but `gen_config.sh` does, so it isn't removed yet.

`DataLogConsumer`, `OutputsSectionConsumer`, and the constructor arguments to `TsOutput` are now removed since those were now dead code. There is nothing more to remove.

rusefillc:
my understaidnig of TsOutput constructor usages in master is "OutputsSectionConsumer and TSProjectConsumer are extremely similar"

my common sense is "if we replace OutputsSectionConsumer same PR should also be replacing TSProjectConsumer "

my common sense is "if we remove OutputsSectionConsumer same PR should also be removeing TSProjectConsumer "

Q: does this PR treat OutputsSectionConsumer and TSProjectConsumer the same way by migrating those together?

rusefillc:
do we want to change random things, or seek some stability in the world?

rusefillc:
@mi-hol here we have a great idea which is I. The works for a couple of years and in my opinion the transition period of extra effort to maintain old and new is not justified

@mck1117 I still believe that we should either completely migrate to newparse or officially give up. What do you think @mck1117?

mi-hol:
> @mi-hol here we have a great idea which is I. The works for a couple of years and in my opinion the transition period of extra effort to maintain old and new is not justified

@rusefillc apologize but I fail to understand what you try to tell me. What I see is a harsh comment discouraging further progress on this matter. It's your choice how this will end up. From my view a balance between requirements of developer and owner are required to progress as a team.

rusefillc:
Abandoned

PR #4299, opened 2022-06-28
quick fix #4298

### LINKED ISSUE #4298: rpm hard limit limited at 12500
rpmHardLimit and many other fields which use uint8_t and RPM_1_BYTE_PACKING_MULT of 50 are currently limited at 12500

moted people need higher RPM limit

rusefillc:


mck1117:
woah you can't just change the scaling factor like that

there's no rush here - let's do the right thing

rusefillc:
@mck1117 please elaborate. is your concern with configuration migration or with loss of precision or else?

mck1117:
let's #4300 instead

PR #4300, opened 2022-06-28
rescale RPM values

### LINKED ISSUE #4298: rpm hard limit limited at 12500

mck1117:
- Each RPM parameter gets its own scale and limits, as appropriate for that parameter.
- Switch a few other tables from float to scaled ints.
- Remove RPM_1_BYTE_PACKING_MULT, since we only ever interact with these channels via `autoscale`, so there's no need for an accessible define.

Fix #4298 properly.

Also saves 84 bytes of config, so that's cool.


rusefillc:
looks like this might be actually green

mck1117:
I'm going to wait for green master

Issue #4084, opened 2022-04-19
Mark reports that detectHellenBoardType hangs up HD and Nucleo


rusefillc:
detectHellenBoardType hangs up HD hellen and Nucleo

rusefillc:
```
2022-06-19_11_28_31_739: EngineState: applyNonPersistentConfiguration()
2022-06-19_11_28_31_739: EngineState: initializeTriggerWaveform(TT_TOOTHED_WHEEL_60_2/8)
2022-06-19_11_28_31_739: EngineState: Read valid configuration from flash!
2022-06-19_11_28_31_740: EngineState: Hellen board pin states = 0 1
2022-06-19_11_28_31_740: EngineState: * Hellen 144-pin mcu detected!
2022-06-19_11_28_31_740: EngineState: Starting Hellen Board ID detection...
2022-06-19_11_28_31_741: EngineState: pin_markUsed: boardId on PF1
2022-06-19_11_28_31_741: EngineState: pin_markUnused: PF1
2022-06-19_11_28_31_741: EngineState: pin_markUsed: boardId on PF1
2022-06-19_11_28_31_742: EngineState: pin_markUnused: PF1
2022-06-19_11_28_31_742: EngineState: pin_markUsed: boardId on PF1
2022-06-19_11_28_31_742: EngineState: pin_markUnused: PF1
2022-06-19_11_28_31_742: EngineState: pin_markUsed: boardId on PF0
2022-06-19_11_28_31_743: EngineState: pin_markUnused: PF0
2022-06-19_11_28_31_743: EngineState: pin_markUsed: boardId on PF0
2022-06-19_11_28_31_743: EngineState: pin_markUnused: PF0
2022-06-19_11_28_31_744: EngineState: pin_markUsed: boardId on PF0
2022-06-19_11_28_31_744: EngineState: pin_markUnused: PF0
2022-06-19_11_28_31_744: EngineState: pin_markUnused: PF0
2022-06-19_11_28_31_745: EngineState: pin_markUnused: PF1
2022-06-19_11_28_31_745: EngineState: * RESULT: BoardId = 2, R1 = 510.000000000, R2 = 750.000000000 (Elapsed time: 306 ms)
2022-06-19_11_28_31_745: EngineState: pin_markUsed: Primary UART RX on PD5
2022-06-19_11_28_31_745: EngineState: pin_markUsed: Primary UART TX on PD6
[...]

PR #4307, opened 2022-06-30
hellen board ID detect improvements

mck1117:
- Return -1 in case of failed detection (missing hardware, etc), and return the real ID  (instead of +1) in case of detection.
- Successfully detects cases of two missing resistors, one missing resistor, or missing capacitor (aka, any possible hardware failure) without hanging or taking too long
- Remove dependency on GPT, instead using the precision timer we already have. This actually improved the jitter/timing accuracy, since we don't have to do the goofy `Tadj` step. Confirmed with a scope that timing is now actually more accurate.
- Clarify timestamp math (naming, mostly)
- Reduce the maximum number of allowed Newton–Raphson iterations.  All the testing I did never saw convergence take longer than ~5 iterations, so 20 is certainly plenty.
- Compare resistors by ratio, instead of by subtraction. Since the resistor series is roughly a geometric series, the "half way point" between two is the geometric mean, not the arithmetic mean. Therefore, we should compare by ratio error instead of absolute error.
- Move the call to `detectBoardType()` to a more logical place, the earliest possible place it could be called.

improves upon the solution in #4084
likely helps #4149


rusefillc:
Pending a collection of "before" state real hw results just to be safe. EtA unclear.

mck1117:
This provided an identical appearing trace on an oscilloscope for a half dozen different resistors I tried on the breadboard.

rusefillc:
before
4chanA
2022-07-02_05_16_25_763: EngineState: * RESULT: BoardId = 5, R1 = 510.000000000, R2 = 1200.000000000 (Elapsed time: 309 ms)


4chanB
2022-07-02_05_17_32_598: EngineState: * RESULT: BoardId = 12, R1 = 510.000000000, R2 = 4700.000000000 (Elapsed time: 327 ms)


4chanC
2022-07-02_05_18_15_676: EngineState: * RESULT: BoardId = 101, R1 = 620.000000000, R2 = 510.000000000 (Elapsed time: 306 ms)

PR #4312, opened 2022-07-02
implement CAN driver for simulator

mck1117:
Kirby had a really good idea - what if the simulator connects to a real CAN device on your computer so you can prototype lua scripts with your computer instead of an ECU?

This PR will connect rusEFI's can channel 1 to your linux machine's `can0` interface.

Tested in loopback with a real CAN interface (tx a packet, rusEFI can rx the same packet back!), Kirby will test soon while actually connected to a bus.


rusefillc:
is chibios repo a better repo for the new hal_can* files?

mck1117:
Maybe, maybe not. We already have other custom chibios-flavor drivers (which this is) checked in to this repo.

PR #4321, opened 2022-07-05
fix lua CAN RX of 11b frames

mck1117:
Previously the ID could get passed with some garbage in the high bits when processing an 11b frame.

### DIFF
--- a/firmware/CHANGELOG.md
+++ b/firmware/CHANGELOG.md
@@ -23,6 +23,11 @@ Release template (copy/paste this for new release):
 
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


rusefillc:
Is there anything we can test here? Unit or hal in simulator?

mck1117:
Not really, no. The problem was in fact discovered when the simulator _didn't_ have this bug in the first place, since it wasn't susceptible to putting garbage in the high ID bits for standard frames. Same script was moved to a real stm32 and it didn't work, since that driver _does_ sometimes result in random high bits if you read `CAN_EID` for a standard frame.

Issue #4326, opened 2022-07-07
remove thread_controller.h


mck1117:
Looks like we copied something that ChibiOS already has - `BaseStaticThread`.

Probably don't need `ThreadController` at all.

mck1117:
Maybe not delete but simplify, it's still useful to store priority at construction time.

PR #4327, opened 2022-07-07
Simpler thread controller

mck1117:
#4326


firmware/controllers/engine_controller.cpp
```diff
@@ -154,14 +153,12 @@ class EngineStateBlinkingTask : public PeriodicTimerController {
 
 		if (is_running) {
 			// blink in running mode
-			enginePins.runningLedPin.setValue(counter % 2);
```
rusefillc, line 157: more than one atomic PR here? looks like /s/Start/start/ occupies a big percent of "files chagnged"?

it would be great to extract trivial huge rename into separate PR

Issue #4308, opened 2022-06-30
Uniform reset pattern for all Lua adjustments


rusefillc:
We have
```
	IgnitionState() {
		luaTimingMult = 1;
	}
```
which has nothing to do with
https://github.com/rusefi/rusefi/blob/7f8a26f7798b5e00478840b80e2a92c565954859/firmware/controllers/lua/lua.cpp#L312

See also https://github.com/rusefi/rusefi/pull/4306#issuecomment-1171485649

PR #4331, opened 2022-07-08
Add a timeout for Lua ETB adjustment

mck1117:
If your lua script hangs/blows up/has a bug, we don't want the throttle to get stuck open. Require the script to write a throttle position at least every 0.2 second.

sort of #4308


mck1117:
todo: check behavior with unit test

PR #4340, opened 2022-07-14
pt2001 updates

mck1117:
- Remove unused RAM constants
- Compile actual generated files from PT2001 IDE to avoid copy/paste


firmware/hw_layer/mc33816/rusefi/sample_code/PT2001_LoadData.c
```diff
@@ -1,21 +1,54 @@
...
- * mc33816_data.c
- *
- * For historic reasons rusEFI source code refers to all this as mc33816 while the microcode
- * is signed with MC33PT2001 key. to use real mc33816 you would have to replace binary data with true mc33816 version
- *
- * this is manual copy-paste of sample_code/PT2001_LoadData.c
- * see mc33816/rusefi/readme.md
- */
+/*******************************************************************************
+* Example Code
```
rusefillc, line 2: is this "example code"?

mck1117: It's not, but this is the exact file that the tool generates. There doesn't seem to be any way to control the template it uses to generate this file.

PR #4343, opened 2022-07-14
GDI: add min/max boost time config

mck1117:
fault detection!

Also adjust the default RAM parameters to the correct ones for a Bosch HDEV5 injector instead of whatever random defaults we had.


mck1117:
red X will self heal

Issue #4336, opened 2022-07-08
idle clarify part 2


rusefillc:
new ticket to continue #4335 since amazing problem statement

Maybe we should get educating me out of the ticket realm. There are still a few things I don't quite understand and that make it harder for me to tune idle:
Is there a list explaining what the names of the parameters in the log/ TS gauges mean? I searched the Wiki using obvious seeming terms, but was unable to find anything
Is the idle-status error the difference between target rpm and actual rpm? If so, it seems to get calculated only occasionally.
The I-term in the log I posted (https://rusefi.com/online/view.php?log=924) varies between 0.0 and 0.3. It appears to be maxed out at 0.3, even though there is still an error. It has authority to change the value of the output between -5 and +10. What is the relation between I-term and output?
There appears to be no log field for the P-term
What is the difference between the open-loop idle slider and the offset value when in CL mode?

Also: One thing that tripped me up was a wrong AC input. The ECU though the AC was turned on and jacked up the idle accordingly. There was no way I could see in the log that this was happening, looking at the idle values. I was able to diagnose it, because I realized the AC fan was running constantly.
TLDR: In order to tune PID idle, I would like to see how my total output value is made up from it contributing elements.

rusefillc:
Yes, error is difference between target and actual. 


RE: A/C portion diag

your log has "targetRpmAcBump" which was since renamed to "idle: A/C bump", is that part of the ask @stefanst ? or does zero value of targetRpmAcBump in your logs not matching expectations?

your log also says "AC enabled"

stefanst:
Maybe I need to formulate better:
In order for me to troubleshoot idle with an idle valve, I would like to see the following in the log:
- current IAC setting
- all factors contributing to that setting in a manner that when I add them up a+b+c+.... = IAC
- current target idle rpm and contributing factors

Also an explanation how "Min", "Offset" and "Open Loop Idle Value" work. Especially the difference between the OL value and Offset seems not intuitive to me.  

I assume that Min/Max is that total PID output Min/Max, whereas we have separate Min/Max values for I-term only. But I am not certain. It could also be that Min/Max could be applied to each P and D separately. 

Tool tips could be more verbose. I particularly dislike the tips for "derivative filter loss" and "anti windup frequency"- they just state that 0.1 is a good value. 
The tip for "CicPID" could also be enhanced. What is it? At least say "Normally this should be FALSE" or TRUE, whatever the case may be. 

[...]

PR #4350, opened 2022-07-15
remove idle pid cic, simplify idle UI

mck1117:
- We've seen very good results on many vehicles without PID CIC, so it's unnecessary to keep around and confuse users.
- ~~Force 0 PID offset and remove the UI for it since this function is fulfilled by the open loop tables instead.~~ #4355

#4336


rusefillc:
is this two unrelated PRs here - one for offset another or CIC? can we please split since the offset part I am happy to merge

mck1117:
> can we please split since the offset part I am happy to merge

#4355

rusefillc:
the whole feature removal side is disproposionatelly emotional and this specific feature removal would be especially troublesome. with everything going on unfortunatelly I have no capacity even to see this PR in the list of PRs.

PR #4355, opened 2022-07-17
remove idle offset

mck1117:
Force 0 PID offset and remove the UI for it since this function is fulfilled by the open loop tables instead.

#4336


rusefillc:
@mck1117 please changelog since some users could have used offset setting

PR #4363, opened 2022-07-21
GDI: detect too short/too long boost phase

mck1117:
Detect whether the boost phase is too short or too long, facilitated with some timers.


rusefillc:
@mck1117 are you available to start moving PT2001 code into libfirmware?

mck1117:
yep

PR #4364, opened 2022-07-21
Add "Reset to OpenBLT" button

dron0gus:
Similar to "Reset to DFU", but uses different seed


mck1117:
does openblt already read this seed...?

mck1117:
> does openblt already read this seed...?

@rusefillc 😢

dron0gus:
@mck1117 openblt will support this in next PRs. :)

Issue #4401, opened 2022-07-29
Additional CAN messages


BeerMoneyMotorsports:
I have noticed we are missing a few CAN messages on the Bus.  There are probably more that i have not noticed.  

1- Low Fuel Pressure
2- High Fuel Pressure
3- Flex %

rusefillc:
Do we need to update https://github.com/rusefi/rusefi/blob/master/firmware/controllers/can/rusEFI_CAN_verbose.dbc @mck1117 ?

10AE0739:
I'll add to this. I'd like to see target AFR CAN messages so that it can be displayed on CAN Gauges.

mck1117:
this isn't done, the dbc needs updating

PR #4402, opened 2022-07-29
more CAN verbose

mck1117:
- Send VVT positions and targets
- Send both lambdas 
- Send flex percent #4401
- Send both fuel pressures #4401

Also don't directly read the `Value` field of `expected`, since it may now contain garbage in case of invalid, instead of 0.


rusefillc:
looks cool and abandonet, let me give this a try

mck1117:
let me resurrect

mck1117:
in the process of fixing another PR

mck1117:
shouuuuld be happy now, let's see if green check marks agree

mck1117:
it's happy!

PR #4414, opened 2022-08-04
allow logging multiple warnings at once

mck1117:
Previously, we would only print one warning every N seconds, even if multiple were active.

Now, we print every warning every N seconds, so that you don't miss any when multiple are active.

For example, if you forgot to plug in your electronic throttle, it should fire three warnings: TPS 1 primary low, TPS 1 secondary low, TPS 1 inconsistent. But before this change, it would print one of those each time, basically at random, but usually only ever see one of those messages.


rusefillc:
@mck1117 this is green and useful? poke

mck1117:
green and useful yes, but something is wrong with cyclic_buffer

mck1117:
this now works OK

Issue #4424, opened 2022-08-09
VVT target to allow negative values


ElDominio:
4B11 VVT issues: positive duty makes Exhaust VVT go negative

As title says, that happens, but VVT angle cannot be negative. Need to target negative angle since this is expected operation

rusefillc:
duty? angle? I am very confused

tune URL is best

ElDominio:
https://rusefi.com/online/index.php?vehicleName=Lancer+NotEvo+fifrst&user_id=675&sorts[uploadedMsq]=-1&sorts[uploadedLog]=-1

Increasing exhaust VVT duty makes the VVT angle go into negative numbers (when idling at 0 degrees)

rusefillc:
@ElDominio what is "actual behaviour" and what is "expected behaviour"

I am starting to wonder if that's about VVT target table not allowing negative values

Are you asking for VVT target to start accepting negative values?

![image](https://user-images.githubusercontent.com/48498823/183550831-dd765a0a-f496-45e0-99e9-30f69851a9ce.png)

ElDominio:
Correct, that would be best

rusefillc:
just a reminer that you would need to lower PID min value to include your negative range

mck1117:
there's a other step, the control is reversed for his exhaust cam

rusefillc:
For reversed control do we simply use negative P?

mck1117:
> For reversed control do we simply use negative P?

It's much more user friendly to not do that.

mck1117:
> just a reminer that you would need to lower PID min value to include your negative range

PID min is the min output - not the min target

rusefillc:
https://github.com/rusefi/ts-issues/issues/52

ElDominio:
https://rusefi.com/online/view.php?log=967

VVT exhaust angle target is negative, PID max and min are 90 and -90 respectively. No change in VVT angle, VVT direction for exhaust is set to "retard"

https://rusefi.com/online/view.php?log=968

This log has VVT direction set to "advance", and once the target is changed to a positive value(at about 30s), actual VVT goes rapidly towards -40 (max VVT angle on exhaust), I assume because the PID is seeing it get farther and farther away from +10 (the target) .

ElDominio:
👀

PR #4425, opened 2022-08-09
add option to invert vvt control

mck1117:
#4424


rusefillc:
Do we really need 4 oprions? Can we see bank 1 anx bank two working differently?

rusefillc:
Lets start with only two bits

mck1117:
> Do we really need 4 oprions? Can we see bank 1 anx bank two working differently?

that's a fair point, I don't think any engines have fundamental control differences between banks

rusefillc:
please CHANGELOG

Issue #4442, opened 2022-08-14
Struct array for outputs


rusefillc:
We need an array of VVT PID statuses to help with #4424

PR #4444, opened 2022-08-15
A safe place where stuff is consistent

rusefillc:


rusefillc:
Between legacy implementation supports #4442 and newparse integration being so limited and local and lack of progress with newparse with more complex use-cases I choose to revert to consistent full legacy implementation instead of the spotty limited unfinished newparse

Issue #4460, opened 2022-08-19
use instant RPM more


mck1117:
Slow RPM is fine, but has some limitations, and instant RPM looks great. Why compute both?

rusefillc:
A bit vague?

mck1117:
Instant RPM is better than non-instant RPM. Let's switch to use instant RPM all the time.

PR #4461, opened 2022-08-19
enable alwaysInstantRpm in some tests

mck1117:
Progress to #4460.

This is all tweaks to when we get RPM (always instant gets us rpm earlier), and exact RPM numbers that change a little.


rusefillc:
Do we want some fancy "assert RPM" which would be taking both RPMs? Just asking

Issue #4459, opened 2022-08-19
350Z VQ35 tripple trigger issues


ElDominio:
https://rusefi.com/online/view.php?log=972

At 4.3s, RPM jumps to 900 even though the real engine did not do this, then at 6.35s goes from 300rpm to 150, which is exactly half of reality. Once it is somewhat stable the car starts and never has issues again

mck1117:
This is due to a limitation in how we measure RPM. We have an "alternative" way of doing it too called "instant rpm" that looks much better in your log. #4460 created.

rusefillc:
@ElDominio can you please elaborate? At the moment i do not understand your update

ElDominio:
In the log it shows the RPM jumping wildly on cranking, even though cranking is stable at 300 RPM.

Also in WOT pulls, RPM is jumping to 11k ish
https://rusefi.com/online/view.php?log=973

ElDominio:
https://rusefi.com/online/view.php?log=974

Lowered dwell by .2, problem is quite mitigated, pending further testing

mck1117:
Lots of weird things are going on.

Why is your MAP 120kPa? Turbo but only 3psi?

Why is your accelerator pedal dropping out during the pull?

Those RPM doublings are on trigger sync losses.

ElDominio:
Centrifugal supercharger
Acc pedal? no idea i might be scared mid pull, i'm also not revving out to max rpm since i usually do lower rpm limits until im comfortable with how the car behaves

ElDominio:
https://rusefi.com/online/view.php?log=976

VVT intake #1 can be seen jumping about 4 degrees (from 0 to 4) in idle, looks like a step function. Might be of interest?

ElDominio:
https://rusefi.com/online/view.php?log=977

This has a few more

rusefillc:
looks like crank decoding issue, everything else is a consequnce of crank decoding issue

![image](https://user-images.githubusercontent.com/48498823/185703703-c0efc558-e967-4466-beba-242dcd286263.png)

ElDominio:
Here is a max data rate log at WOT-ish
https://rusefi.com/online/view.php?log=979

at 1.96s the issue is slightly apparent, at 3.9s it is very apparent

ElDominio:
![IMG_20220819_164410__01.jpg](https://user-images.githubusercontent.com/5685108/185704319-bb8926ea-1750-4635-8099-b79bddb8be45.jpg)

[...]

PR #4473, opened 2022-08-23
log missed exti events

mck1117:
Might help narrow down trigger issues. #4459 etc


rusefillc:
@mck1117 is this specific to #4459 or shall we request fresh data from @NMSTEC https://github.com/rusefi/hellen-gm-e67/issues/54?

PR #4480, opened 2022-08-24
log the state of hasSynchronizedPhase() for future diagnostic help.

rusefillc:
this change started as lesser of two evils but it now collapses with tests using EngineTestHelperBase

### DIFF
--- a/firmware/controllers/trigger/trigger_decoder.cpp
+++ b/firmware/controllers/trigger/trigger_decoder.cpp
@@ -93,6 +93,21 @@ void TriggerDecoderBase::resetCurrentCycleState() {
 
 #if EFI_SHAFT_POSITION_INPUT
 
+void PrimaryTriggerDecoder::resetHasFullSync() {
+#if EFI_UNIT_TEST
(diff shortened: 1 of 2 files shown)


firmware/controllers/trigger/trigger_decoder.cpp
```diff
@@ -389,7 +404,7 @@ angle_t PrimaryTriggerDecoder::syncEnginePhase(int divider, int remainder, angle
 	}
 
 	// Allow injection/ignition to happen, we've now fully sync'd the crank based on new cam information
-	m_hasSynchronizedPhase = true;
+	engine->engineState.isSynchronizedPhase = true;
```
mck1117, line 407: wtf? please don't split that state outside of the trigger decoder

rusefillc: WTF yourself for "let me just manually copy this boolean from here to there"

https://github.com/rusefi/rusefi/commit/2baae2adc0a44510044cf3d145759877785b5f53

Issue #4292, opened 2022-06-27
Time delay before DFCO


ElDominio:
I remember some technical reason for this behavior but I don't remember it. The real  reason i would like this is so cars with pops can have, for example, 2-3 seconds of pops then go into DFCO. 

I think the technical reason is to control the amount of engine braking available but I'm not going to lie lol

ElDominio:
Any progress on this?

10AE0739:
I was also looking for this feature in settings.

rusefillc:
@mck1117 shall we close this ticket?

mck1117:
@ElDominio want to give it a try?

ElDominio:
Of course, I'll try to get it tested before sunday

10AE0739:
I'm running the 8/30 snapshot on my Hellen NB1. This works well. Thank you.

mck1117:
Reports of it working, looks like we're done here!

rusefillc:
We have this one and #4500 I wonder what else can use a time delay

PR #4482, opened 2022-08-25
dfco delay

mck1117:
#4292


firmware/controllers/algo/fuel/dfco.h
```diff
@@ -14,4 +14,5 @@ class DfcoController : public EngineModule {
 	bool m_isDfco = false;
```
rusefillc, line 14: (not directly related to this PR)
is it time for dfco_controller.txt to get some LiveData logging?

mck1117: some day, probably 🙂

Issue #3940, opened 2022-02-16
Extra Idle Air if Throttle is Pressed behavior


ElDominio:
I think this function should be slightly improved so it can better be used as a sort of dashpot.

Suggestions:
1) Make it engage when the ECU goes out of "idle" state, not when throttle is pressed
2)  Implement a check for RPM/s that when under a RPM/s threshold, the value is slowly (maybe the rate at which idle returns to normal can also be configured?) returned to normal idle values.

Example:
User pressed throttle, engine RPMs increase significantly
Once the RPMs start falling, check rate of change of RPM to make sure we aren't dropping like a stone and possibly shutting off (or still falling into steady idle)
Once RPM/s is relatively low, the additional idle air subtracts slowly, to be able to return to idle without an undershoot

![image](https://user-images.githubusercontent.com/5685108/154221785-4679a6fd-abf1-4c83-b88c-a4283f55c539.png)

rusefillc:
@andreika-git any interest?

mck1117:
@ElDominio do you have a log that demonstrates the behavior you want this change to avoid? That would be helpful at least for us to test against, and maybe to offer alternatives that are already possible today.

rusefillc:
and tune obviously

at the moment we are affected by https://github.com/rusefi/web_backend/issues/158 :(

ElDominio:
The behavior this would try to avoid, especially in Open Loop operation, is a dip of RPM as RPM comes down from not-idle. 

This post https://forums.linkecu.com/topic/13259-dash-pot-settings/?do=findComment&comment=85016 explains it quite well, and even some OEM's (basically all USDM) use this even in closed loop mode. You crack the idle valve open to a certain % as RPMs are dropping from above idle, wait for them to stabilize (for .5s-2s, depends on how hard said engine decelerates), then slowly return idle % to the base table. It allows for a much more mellow return to idle, and much more consistent in my experience.

mck1117:
Here's an example return to idle on rusEFI (5.3 LS, medium-ish cam, 228/230 duration):

![image](https://user-images.githubusercontent.com/568254/161491765-020d3f9d-c4c9-4208-90e5-44d3f0587044.png)

My idle timing control settings are probably more aggressive than they need to be, but you get the idea.

BTW, you can emulate this behavior by setting your open loop idle position to be a little bit (50-100 RPM) higher than target, then let the closed loop control pull it down once you're close.

[...]

PR #4486, opened 2022-08-25
[DRAFT] Air by rpm taper

andreika-git:
https://github.com/rusefi/rusefi/issues/3940

### DIFF
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -110,6 +110,14 @@ percent_t IdleController::getRunningOpenLoop(float rpm, float clt, SensorResult 
 
 	running += iacByTpsTaper;
 
+	float airTaperRpmUpperLimit = engineConfiguration->idlePidRpmUpperLimit + engineConfiguration->airTaperRpmRange;
+	iacByRpmTaper = interpolateClamped(
+		engineConfiguration->idlePidRpmUpperLimit, 0,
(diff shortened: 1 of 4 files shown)


mck1117:
oops, looks like some overlap with the PR I just made, but that's OK: https://github.com/rusefi/rusefi/pull/4487

firmware/controllers/actuators/idle_thread.cpp
```diff
@@ -110,10 +110,18 @@ percent_t IdleController::getRunningOpenLoop(float clt, SensorResult tps) {
 
 	running += iacByTpsTaper;
 
+	float airTaperRpmUpperLimit = engineConfiguration->idlePidRpmUpperLimit + engineConfiguration->airTaperRpmRange;
+	iacByRpmTaper = interpolateClamped(
```
mck1117, line 114: I don't understand this logic, why do we add more position at higher RPM?

This seems like it would encourage the engine to stall as it returns to idle.

image

mck1117: this is also not what #3940 asked for, that issue is asking for "predictive idle entry" where we look at current RPM, dRPM/dt, and some time window in the future about whether we're going to enter idle soon

rusefillc:
it's not ideal but my preference is simply to merge it so that @ElDominio gives it a try

So 30 days from now either @ElDominio says that he needs it or we revert it if he does not, at least that's my current thinking

PR #4491, opened 2022-08-27
tunerstudio: GPPWM gauges: unique names, use Note as label

dron0gus:
This fixes issue when only GPPWM 4 output Gauge was available.


rusefillc:
@dron0gus why was only number 4 displayed? looked like it was four unique names even before the change to use names?

Also do we have good default names? We need good default names since we probably use empty names for gauges now? :(

dron0gus:
@rusefillc couse all 4 have same name and overwrite each other.
Screenshot from 2022-08-27 22-09-33
TS construct name of gauge from two pieces: name and label. Now we use Note for Label. So another half is always exist.

PR #4493, opened 2022-08-28
Correction of some warnings. Fix for "Simulator" build in linux

Finist64:
Fix build "Simulator" for linux.
Fix of some warnings when building the project.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,7 +11,9 @@
 		"system_error": "cpp",
 		"vector": "cpp",
 		"xhash": "cpp",
-		"xtree": "cpp"
+		"xtree": "cpp",
+		"cstring": "cpp",
+		"cstdio": "cpp"
 	},
 	"editor.detectIndentation": false,
 	"editor.insertSpaces": false,
--- a/firmware/config/engines/mazda_miata_vvt.cpp
+++ b/firmware/config/engines/mazda_miata_vvt.cpp
@@ -362,10 +362,12 @@ static void setCommonMazdaNB() {
 	set4EC_AT();
 }
 
+#pragma GCC diagnostic ignored "-Wunused-function"
 static void setMazdaMiataEngineNB1Defaults() {
 	setCommonMazdaNB();
 	strcpy(engineConfiguration->engineCode, "NB1");
 }
+#pragma GCC diagnostic pop
 
 static void setMazdaMiataEngineNB2Defaults() {
 	strcpy(engineConfiguration->engineCode, "NB2");
--- a/firmware/config/engines/vw_b6.cpp
+++ b/firmware/config/engines/vw_b6.cpp
@@ -14,6 +14,7 @@
 #include "mre_meta.h"
 #include "proteus_meta.h"
 
+#pragma GCC diagnostic ignored "-Wunused-function"
 static void commonPassatB6() {
 	setCrankOperationMode();
 	engineConfiguration->trigger.type = TT_TOOTHED_WHEEL_60_2;
@@ -127,6 +128,7 @@ static void commonPassatB6() {
 	engineConfiguration->injectionMode = IM_SEQUENTIAL;
 	engineConfiguration->crankingInjectionMode = IM_SEQUENTIAL;
 }
+#pragma GCC diagnostic pop
 
 /**
  * set engine_type 39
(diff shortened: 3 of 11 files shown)


firmware/config/engines/mazda_miata_vvt.cpp
```diff
@@ -362,10 +362,12 @@ static void setCommonMazdaNB() {
 	set4EC_AT();
 }
 
+#pragma GCC diagnostic ignored "-Wunused-function"
```
rusefillc, line 365: that's a hack not real solution

real solution would be to move static method closer to real usages so that it is conditionally compiled together with usages

firmware/controllers/lua/can_filter.cpp
```diff
@@ -12,7 +12,7 @@ CanFilter* getFilterForId(size_t busIndex, int Id) {
 		auto& filter = filters[i];
 
 		if (filter.accept(Id)) {
-			if (filter.Bus == ANY_BUS || filter.Bus == busIndex) {
+			if (filter.Bus == ANY_BUS || filter.Bus == (int) busIndex) {
```
rusefillc, line 15: @mck1117 what do you think? is that a proper solution here?

firmware/config/engines/vw_b6.cpp
```diff
@@ -14,6 +14,7 @@
 #include "mre_meta.h"
 #include "proteus_meta.h"
 
+#pragma GCC diagnostic ignored "-Wunused-function"
```
rusefillc, line 17: @mck1117 what do you think?

firmware/controllers/algo/engine_configuration.cpp
```diff
@@ -164,7 +164,9 @@ __attribute__((weak)) void boardOnConfigurationChange(engine_configuration_s* /*
  * See preCalculate which is invoked BOTH on start and configuration change
  */
 void incrementGlobalConfigurationVersion() {
-	engine->globalConfigurationVersion++;
+	//https://www.modernescpp.com/index.php/volatile-and-other-small-improvements-in-c-20
```
mck1117, line 167: this field should not be volatile in the first place, it provides no value

firmware/controllers/lua/lua_hooks.cpp
```diff
@@ -27,7 +27,7 @@ using namespace luaaa;
 #endif
 
 static int lua_readpin(lua_State* l) {
-	auto msg = luaL_checkstring(l, 1);
+	auto msg __attribute__((unused)) = luaL_checkstring(l, 1);
```
mck1117, line 30: if this is unused in some scenario, why not only create it if in that scenario?

firmware/controllers/lua/lua_hooks.cpp
```diff
@@ -616,7 +616,7 @@ void configureRusefiLuaHooks(lua_State* l) {
 	lua_register(l, "findCurveIndex", [](lua_State* l) {
 		auto name = luaL_checklstring(l, 1, nullptr);
 		auto result = getCurveIndexByName(name);
-		if (result == EFI_ERROR_CODE) {
+		if (result == (int) EFI_ERROR_CODE) {
```
mck1117, line 619: EFI_ERROR_CODE is a hack at best, we should probably remove it all together and use `expected<T>` instead.

simulator/simulator/can/hal_can_lld.cpp
```diff
@@ -31,6 +31,7 @@
 #include <sys/ioctl.h>
 
 #include <cstring>
+#include <bits/stdc++.h>
```
mck1117, line 34: why do we need this?

.vscode/settings.json
```diff
@@ -11,7 +11,9 @@
 		"system_error": "cpp",
 		"vector": "cpp",
 		"xhash": "cpp",
-		"xtree": "cpp"
+		"xtree": "cpp",
```
mck1117, line 14: don't need this part of the change, please avoid `git add .`

firmware/config/engines/mazda_miata_vvt.cpp
mck1117: +1, we should never be turning on/off warnings unless there's a really good reason to, and this isn't one

rusefillc:
Thank you @Finist64 !!! Some stuff merged some stuff inspired https://github.com/rusefi/rusefi/pull/4496

PR #4499, opened 2022-08-29
idle always uses instant rpm

mck1117:
Instant RPM is well proven these days, let's always use it for idle instead of offering the option.

#4460


rusefillc:
@mck1117 merge conflict

mck1117:
> merge conflict

somehow I keep merge conflicting myself 😛

PR #4508, opened 2022-09-01
just provoking CI

rusefillc:


mck1117:
hardware qc build builds with `-o0`

hardware qc build is skipped 90% of the time

mck1117:
local build says we have about 20k free

PR #4516, opened 2022-09-02
I am sure that's not magic

rusefillc:
@mck1117 can you please help? why does this red?

### DIFF
--- a/firmware/controllers/algo/engine.h
+++ b/firmware/controllers/algo/engine.h
@@ -37,6 +37,7 @@
 #include "ignition_controller.h"
 #include "alternator_controller.h"
 #include "dfco.h"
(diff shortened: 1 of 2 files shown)


firmware/controllers/algo/engine.h
```diff
@@ -87,7 +88,7 @@ struct AirmassModelBase;
 #define CYCLE_ALTERNATION 2
```
mck1117, line 88: this define is dead 🙂

mck1117:
This is a weird one, I'm really not sure why it lets you touch those protected functions if forward declared. Those functions are `const`, so it's pretty safe to remove the `protected` from `FuelComputerBase` and `FuelComputer`.

PR #4492, opened 2022-08-27
tunerstudio: indicators: no blank indicators in off state

dron0gus:
I'd like to see something on indicator even it is off.
From this:
Screenshot from 2022-08-27 20-55-11
to this:
Screenshot from 2022-08-27 19-47-14

PR #4521, opened 2022-09-03
fix critical error readout

mck1117:
bug from #4492 @dron0gus FYI

this indicator must be EXACTLY like this

This is why this guy couldn't get the critical error: https://rusefi.com/forum/viewtopic.php?p=46203#p46203


abelom:
works fine with the fix

image

PR #4531, opened 2022-09-05
// todo: what is broken _exactly_?

rusefillc:


rusefillc:
Now I am confused @mck1117 

Was not something broken here a few days ago? See code comments, do you remember the discussion?

mck1117:
Was it? I didn't make that comment. In general, it's a bad idea to "launder" a value through the scaled_channel then back to a float. That's asking for precision to get lost in really unpredictable ways, not to mention that the float/int conversions cost something.

rusefillc:
I agree when it's scaled_channel<int> but here we have scaled_channel<float> which is a bit odd to begin with

scaled_channel code comments imply that scaled_channel is integer and somehow we ended up with float implementation. I am confused about why we have what in this area.

Do we agree that "LiveData" model should often? always? be kosher to just use without limitations?

Hold on, just got more confused:

```
	int vvtCamCounter = (int)0;
	/**
	 * offset 28
	 */
	scaled_channel<float, 1, 1> mapVvt_MAP_AT_SPECIAL_POINT = (float)0;

```

we generate PLAIN int and scaled float which does no scaling even? WAT?

mck1117:
> I agree when it's scaled_channel but here we have scaled_channel which is a bit odd to begin with

I don't know what the original problem was.  Why was the comment added in the first place? I don't know if it was actually broken or not, or whether you added the comment because you weren't sure whether it was OK or not (in 5941ad8fb70826fef6eed62b12fea480325d9461)

>  imply that scaled_channel is integer

`scaled_channel` is not an integer. `scaled_channel` lets you store some value, passed as a float (UNLESS mul=div=1), which is stored as the type you specify, with some scale factor. That storage type can be a float, int, uint64_t, whatever you want. It could even be some other custom type that defines enough math operators to work.

To/from a `scaled_channel` with mul=div=1 is literally free, not just approximately free. There are special conversion operators present for "exact read/write" with mul/div are both 1.

> we generate PLAIN int and scaled float which does no scaling even? WAT?

I don't think `MAP_AT_SPECIAL_POINT` needs `autoscale` keyword. Looks like copy-paste error to me. Basic smoke test of building firmware without it is OK. We should remove autoscale from everywhere it isn't needed.

Note: binary SD log only knows how to log scaled_channel, so it's needed even on unscaled fields if we want to log them to SD (for now).

rusefillc:
https://github.com/rusefi/libfirmware/blob/0b8f0d4949a8bc7bcd77d1ff0bc022e7a0fd351f/util/include/rusefi/scaled_channel.h#L16 calls integer explicity, it looks like at least the comment is outdated?

Issue #4477, opened 2022-08-23
ECU does not fire inj/ign correctly after swapping settings 3A92


ElDominio:
I use one 4chan ECU for:
1) Honda B20 (running a modified 24-1 modified trigger)
2) Mitsubishi 3A92 3 cylinder

When I moved the ECU from 3A92 to B20, injector 4 did not fire until i updated firmware (dk what this has to do with it)

Then, I swapped back to 3A92,  and now coil 3 does not fire (shows on RusEFI console that Coil3 is not being commanded to fire)

Logs and tune at
https://rusefi.com/online/view.php?log=996

mck1117:
snip of rusefi console?

please include all the information you can about an issue

ElDominio:
![image](https://user-images.githubusercontent.com/5685108/186282121-a6ef778b-7e8c-47d7-8775-c32415e5ea07.png)

mck1117:
I don't understand your trigger configuration

![image](https://user-images.githubusercontent.com/568254/186282610-d30670eb-8669-4f03-83a6-3666430fd0e2.png)

![image](https://user-images.githubusercontent.com/568254/186282640-2fe8ef58-91ef-43e4-a925-8befbbfc5db6.png)

Does this engine have vvt? Even if not, the "first" cam must always be in the spot labeled "intake cam". But you have no cam inputs configured at all?

mck1117:
There is a real bug here though, we're trying to do wasted spark on an odd-cylinder engine which doesn't really work properly right now.

mck1117:
For now you should set "Require cam/VVT sync for ignition" to `true`

ElDominio:
It does have VVT, and it is only on the exhaust cam, and only has that cam sensor. Intake has no sensor.

It did run correctly before with these settings, quite well in fact https://youtu.be/XcJ1Yui71R4 haha

I'll swap those settings and report back once I can (just stored the car)

mck1117:
Ok, well, the cam input set as secondary is never correct for this setup.

mck1117:
see also #4195

ElDominio:
Will try to test this on Thursday

ElDominio:
Not working, now it doesn't start since both coils 1 and 3 fire at the same time

log and tune: https://rusefi.com/online/view.php?log=1019

console snip: 
![image](https://user-images.githubusercontent.com/5685108/188017483-4e2f0dce-3783-439e-86ff-aed73b98f998.png)

[...]

PR #4537, opened 2022-09-05
Engine Sniffer into TS data logs fix #4535

### LINKED ISSUE #4535: Engine Sniffer into TS data logs
We have amazing data rate, let's dump injector and coil state right into data logs

rusefillc:
image

We can see logic state of coils and injectors in MLV

I belive that would help troubleshoot stuff like #4477

### DIFF
--- a/firmware/controllers/engine_cycle/main_trigger_callback.cpp
+++ b/firmware/controllers/engine_cycle/main_trigger_callback.cpp
@@ -114,6 +114,44 @@ void turnInjectionPinHigh(InjectionEvent *event) {
 	}
 }
 
+void InjectorOutputPin::setHigh() {
+    NamedOutputPin::setHigh();
+	// this is NASTY but what's the better option? bytes? At cost of 22 extra bytes in output status packet?
+	switch (injectorIndex) {
+	case 0:
+		engine->outputChannels.injectorState1 = true;
+		break;
+	case 1:
+		engine->outputChannels.injectorState2 = true;
+		break;
+	case 2:
+		engine->outputChannels.injectorState3 = true;
+		break;
+	case 3:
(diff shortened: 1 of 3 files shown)


rusefillc:
@mck1117 polymorphism or not?

@mck1117 horrible switch with bits or byte array?

mck1117:
wait, what? what's the goal here?

rusefillc:
@mck1117 description updated

mck1117:
what does `polymorphism or not` mean?

rusefillc:
@mck1117 are concurned about making setHigh/setLow virtual and ending up with runtime lookup tables?

mck1117:
> horrible switch with bits or byte array?

As I think I mentioned once before, this is _technically_ unsafe (BUT). Bit operations aren't atomic on any architecture I know of (you have to read, modify, write), while operations to different bytes (even adjacent bytes) are.  However, this should be OK since everything touching these bits (injection, ignition) are running at the same interrupt priority, plus it's not "critical" data but instead "nice to have" data.

I think byte array is overkill, and unfortunately C++ doesn't let you index arrays of bits like it does an array of bytes, so switch it is.

mck1117:
> are concurned about making setHigh/setLow virtual and ending up with runtime lookup tables?

They may not need to be virtual at all. Everyone that calls these is calling an `InjectorOutputPin` or `InjectorOutputPin`, not a `NamedOutputPin`. So we could add `setHighAndReport` or something for injection/ignition to call that have this extra reporting bolted on, with no virtual call required.

rusefillc:
@mck1117 obviously virtual could be avoided easily

question is if we WANT virtual for OOP or else, or if we want to AVOID virtual for CPU costs or else.

Issue #4529, opened 2022-09-05
All hard cuts (boost cut especially) should have hysterisis


ElDominio:
Client is riding the boost cut limit, making it activate/deactive very slightly on the top end. Example here:
https://rusefi.com/online/view.php?log=1028

ECU should disable injection for an amount (user configurable) of boost before enabling again, as to reduce confusion of misfire vs actual cut

mck1117:
yeah I hit the same thing the other day where I was getting very "buzzy" boost cut due to rapidly cycling on/off

mck1117:
TODO after #4541: do we care about configurable hysteresis, or is fixed fine?

ElDominio:
I always say configurable cause a 10kpa hysteresis on a car that make 150kpa total is not the same as 10kpa on a car that makes 300

ElDominio:
Sstill relevant

ElDominio:
STILL RELEVANT

ElDominio:
![image](https://user-images.githubusercontent.com/5685108/230741663-2c84793c-2a2f-475b-b3b3-2a0250ed548b.png)

in area
https://rusefi.com/online/view.php?msq=1438&dialog=limitsAndFallback

![image](https://user-images.githubusercontent.com/5685108/230741686-0546974b-6b4f-46be-bd81-f32a944e15c0.png)


A delay would be nice too but that's another separate ticket lol

rusefillc:
@ElDominio poke?

PR #4541, opened 2022-09-05
add hysteresis to limp rpm, boost, injector duty

mck1117:
Add hysteresis to force the driver to back off, and maybe save components.

#4529


rusefillc:
@mck1117 please changelog

PR #4547, opened 2022-09-07
only report vvt position if we have full sync

### LINKED ISSUE #4546: Non-first VVT inaccuracy if engine not synchronized yet
We need to not report VVT position until the primary trigger has full (ie, disambiguated) sync. Otherwise, we may get bogus cam position readings for other cams that do not contribute to engine sync.

mck1117:
fix #4546


rusefillc:
any test? maybe simplest one to implement would be one of real recordings processing tests?

mck1117:
Well - it's already kind of tested by some existing tests that ensure we didn't break vvt 1 intake.

But yes, it is worth testing other cams too.

mck1117:
OK - now it's tested. Nissan test now uses both cams as the sync cam, and in case of the second cam (the new case), the test fails without the new check added.

PR #4549, opened 2022-09-08
huh? dead code? how can we get NAN from eventAngles table?

rusefillc:


mck1117:
assertion failure is good enough for me, shouldn't have NAN there.

PR #4568, opened 2022-09-12
tle8888 dead code

dron0gus:


rusefillc:
oh nice that this hack is dead!

@dron0gus can you please PR some comments on somewhere in the code in terms of which place/thread/else does TLE8888 init loop if we first power stm32 with USB and only later power on 8888?

PR #4575, opened 2022-09-13
digital_input: check for valid callback not name!

dron0gus:


rusefillc:
lol :)

mck1117:
Look up in `efiExtiEnablePin`, you need to change the order that fields are assigned if you want to do this. Name was assigned last, so everything else was guaranteed valid if name was set.

PR #4592, opened 2022-09-19
add fuel/spark cut codes to TS indicators

mck1117:
Indicators now show the actual spark/fuel cut reason, not just that its cut. That also means we no longer need to actually send the bit of "cut" vs "not cut", since that logic can be in TS instead of copying bits around on the ECU.


mck1117:
No cut:
image
Cut:
image

PR #4597, opened 2022-09-20
terrible horrible typo

mck1117:
Horrible terrible typo that causes `someSortOfTriggerError` to get stuck `true` forever after an error.


rusefillc:
I cannot think of better names but it looks like something is not ideal with method naming

PR #4600, opened 2022-09-21
taking InjectorModel from list of EngineController

rusefillc:


mck1117:
? Why a separate type list?

rusefillc:
For now I am just trying to complehend enough of C++ to understand the relationship between Mockable and type_list and who provides what value how.

PR #4602, opened 2022-09-21
Added Fan1, Fan2, FuelPump and Main relay status to 0x5FE AIM dash

opelpanfan:
Update to broadcast FuelPump, Fan1, Fan2 and MainRelay statuses to 0x5FE AIM Dash Message.
That will be consumed by AIM Dash and also by ECUMASTER PMU16 to turn the fuel pump, fan, and main relay.

### DIFF
--- a/firmware/controllers/can/can_dash.cpp
+++ b/firmware/controllers/can/can_dash.cpp
@@ -1243,6 +1243,20 @@ static void populateFrame(Aim5f7& msg) {
 	msg.LambdaTarget1 = engine->fuelComputer->targetLambda;
 	msg.LambdaTarget2 = engine->fuelComputer->targetLambda;
 }
+// User Specific frames for EcuMaster PMU
+struct Aim5fe {
+	scaled_channel<uint16_t, 1> Fan1Status;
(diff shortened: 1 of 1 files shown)


firmware/controllers/can/can_dash.cpp
```diff
@@ -1243,6 +1243,20 @@ static void populateFrame(Aim5f7& msg) {
 	msg.LambdaTarget1 = engine->fuelComputer->targetLambda;
 	msg.LambdaTarget2 = engine->fuelComputer->targetLambda;
 }
+// User Specific frames for EcuMaster PMU
```
rusefillc, line 1246: I am overall OK with the change but please improve code comments:

please clarify what packets follow what spec what version. maybe even a URL of spec into the source code

EcuMaster reference is super confusing. Are those packets totally random user-defined format? if that's just random format Lua is the proper way to go. if that's totally random custom format why does it matter who is the consumer EcuMaster or not?

rusefillc (changes requested):

opelpanfan:
I'll try to use Lua for that.

PR #4615, opened 2022-09-24
insist on setBoardConfigOverrides fix #4614

### LINKED ISSUE #4614: insist on setBoardConfigOverrides
At the moment setBoardConfigOverrides is applied only on ECU boot at configuration reset

https://github.com/rusefi/hellen154hyundai/issues/65 smoke is asking for setBoardConfigOverrides to be applied on any config change

rusefillc:


mck1117 (approved):

firmware/CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ Release template (copy/paste this for new release):
  -   
 
 ### Fixed
- -   
+ - Insist on setBoardConfigOverrides #4614
```
mck1117, line 20: maybe human description, not opaque issue title?

"enforce board configuration overrides more strictly"

simulator/simulator/rusEfiFunctionalTest.cpp
```diff
@@ -181,3 +181,6 @@ CANDriver* detectCanDevice(brain_pin_e pinRx, brain_pin_e pinTx) {
 	return &CAND1;
 }
 #endif // HAL_USE_CAN
+
+void setBoardConfigOverrides() {
```
mck1117, line 185: does engine_configuration.cpp not get compiled for these targets? This function is defined there as `weak` so it can get replaced by boards that need it.

image

mck1117, line 185: does engine_configuration.cpp not get compiled for these targets? This function is defined there as weak so it can get replaced by boards that need it.

image

mck1117 (changes requested):

simulator/simulator/rusEfiFunctionalTest.cpp
rusefillc: Those changes were to make build green

PR #4621, opened 2022-09-25
infer useOnlyRisingEdge by trigger type

### LINKED ISSUE #4470: Damn what is "useOnlyRisingEdgeForTriggerTemp" from 2018?
```
	bool useOnlyRisingEdgeForTrigger = triggerConfig.UseOnlyRisingEdgeForTrigger;
	this->useOnlyRisingEdgeForTriggerTemp = useOnlyRisingEdgeForTrigger;
```

"temp" implies that something was supposed to get better _later_?

### LINKED ISSUE #4606: ignition scheduling problem when "use only rise" set to false
simulator, 12-1, "use only rise" set to false:

image

"use only rise" set to true:

image

mck1117:
We don't need to let the user pick this - we can infer it from the trigger type. We've had a few issues recently where users shot themselves in the foot with an unexpected selection on this option.

There are three categories of trigger with different "use only rise" behavior:

1. Old engines with obscure triggers
2. Modern engines with high tooth count (24-1, 60-2, etc)
3. VR sensors (or hall with one useless edge, ex Miata NB)

Why does this change work with each:

1. has a specific trigger implementation for that engine/sensor/pattern combination, so we can safely hard-code whether to useOnlyRise
2. doesn't need the added precision from falling edges, and would just add unnecessary CPU load, so we can force it off.
3. doesn't have useful falling edges anyway. This category fully overlaps with 1 and 2: 'weird' triggers we know the right answer, and skipped-tooth doesn't need the falling edge.

fix #4470
more completely fix #4606

### DIFF
--- a/firmware/config/boards/hellen/hellen121nissan/board_configuration.cpp
+++ b/firmware/config/boards/hellen/hellen121nissan/board_configuration.cpp
@@ -135,8 +135,6 @@ void setBoardDefaultConfiguration() {
 	// Some sensible defaults for other options
 	setCrankOperationMode();
 
-	engineConfiguration->vvtCamSensorUseRise = true;
-	engineConfiguration->useOnlyRisingEdgeForTrigger = true;
 //	setAlgorithm(LM_SPEED_DENSITY);
     // at least this starts
 	engineConfiguration->fuelAlgorithm = LM_ALPHA_N;
--- a/firmware/config/boards/hellen/hellen154hyundai/board_configuration.cpp
+++ b/firmware/config/boards/hellen/hellen154hyundai/board_configuration.cpp
@@ -190,7 +190,6 @@ void setBoardDefaultConfiguration() {
 	setCrankOperationMode();
 
 	engineConfiguration->vvtCamSensorUseRise = true;
-	engineConfiguration->useOnlyRisingEdgeForTrigger = true;
 	setAlgorithm(LM_SPEED_DENSITY);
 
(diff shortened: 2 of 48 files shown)


rusefillc (changes requested):
Can we extract test_issue_898 change into separate PR?

mck1117:
yes, there are some other bits to peel out of here before it's ready

rusefillc:
Shall we address https://github.com/rusefi/rusefi/issues/4668 before we merge this or any other PR?

mck1117:
> Shall we address https://github.com/rusefi/rusefi/issues/4668

that requires rolling around on the floor with a multimeter, which I don't have the time for right now

mck1117:
I think it's ready.

rusefillc:
How would a user know if his specific trigger is both edges or just one edge?

What do you think about adding a note right into each trigger image, i.e. a change to triggers.txt generator and TriggerImage.java?

rusefillc:
@mck1117 ^^^?

mck1117:
> How would a user know if his specific trigger is both edges or just one edge?

The goal is that the user doesn't have to know, care, or understand "one edge" vs "both edges" for a trigger.  See the cases enumerated in the description above. Ideally if "show logic levels" is set, it'll show all edges, and if not set, it'll show whatever edges we're actually decoding based on the trigger type. This works with VVT input after #4747, not sure if it does for primary.

PR #4622, opened 2022-09-25
plumbing parts of 4621

mck1117:
Pull out plumbing changes of #4621. No actual change to behavior yet.

This PR just makes the change to avoid reading engineConfiguration whenever possible, instead reading `useOnlyRisingEdges` on the trigger shape itself.

related  #4470


mck1117:
woohoo it's green!

Issue #4474, opened 2022-08-23
Set a complete fuel cut when BattV = 0v


ElDominio:
This would make cars with main relay control shut off the moment key is shut off, then turn off the main relay after 1sec as usual. Clients keep asking why their car stays on after they key off

rusefillc:
Let's unpack if multiple action items here:

1) is this happening with USB power or without or both
2) why 1 second for main relay what is it about

mck1117:
besides, we should already be cutting the main relay when the ignition voltage goes low

ElDominio:
1) happening on both
2) some cars get angry if it all shuts down at once? idk lol but if we want to keep MR alive for any reason, cutting fuel should still be done even if MR is on

shadowm60:
On motorcycles there is an issue sometimes that voltage regulator dies, this would make voltage to jump over 18v, maybe this issue could be broken down that main relay/injectors are fired if the voltage is in a preconfigured voltage range?

rusefillc:
So, overall: we have no idea what's going on and we have a lot of ideas :)

let's start the game of elimintation: @BeerMoneyMotorsports @stefanst did you experience anything relevant on Miata NB2?

BeerMoneyMotorsports:
No issues with my NB2 or any  reported by my customers..

rusefillc:
@ElDominio it sounds like this issue is only specific to specific vehicles.

Question1: are you aware of vehicles where we do NOT have the issue?
Question2: any model/year where we DO have the issue? Wiring diagrams etc?

ElDominio:
Hellen154hyundai

69 mustang with a 4chan controlling main relay

mck1117:
@ElDominio can you draw a rough diagram of:

- ignition wire from key to ECU (and anything else relevant)
- main relay wiring to/from ECU
- main relay wiring to other engine hardware

rusefillc:
@mck1117 see edit - looks like Hellen154hyundai  is the one to explore? https://github.com/rusefi/rusefi_documentation/tree/master/OEM-Docs/Hyundai

mck1117:
does the behavior still happen with USB connected? ie, can you log it?

rusefillc:
@mck1117 what shall we do in your opinion? close or ask a question?

ElDominio:
Remind me and I'll get a video of it happening

mck1117:
[...]

PR #4628, opened 2022-09-28
require ignition on for firing

mck1117:
#4474


rusefillc:
@mck1117 poke? close or merge or else?

PR #4629, opened 2022-09-28
fix instant rpm clearing bug

mck1117:
`checkIfSpinning` is too strict - we don't want to reset RPM state if we simply haven't seen a sync point yet, but only if we've stopped seeing trigger events.  With the old logic, we never got instant RPM on the first sync point, and didn't get it until the second sync point, slowing down engine sync by one or two whole crank revolutions!

Tested on a real car to actually improve starting, as we start firing inj/ign events one trigger cycle earlier.


rusefillc (changes requested):
@mck1117 please CHANGELOG

PR #4631, opened 2022-10-02
SENT protocol capture implementation

dron0gus:


firmware/check.c
```diff
@@ -14,11 +14,4 @@
 	#error "GCC compiler >= 9.3.1 required"
 #endif
 
-// Firmware builds but does not actually run on GCC10 :(
-// todo: one day someone should fix it, do YOU want to help?
-#if (GCC_VERSION > 1000)
```
rusefillc, line 19: why here? let's not?

dron0gus: I will cleanup this when it ready.

firmware/hw_layer/drivers/sent/sent.cpp
```diff
@@ -13,89 +13,112 @@
 
 #include "sent.h"
 
-struct sent_channel {
+typedef enum
```
mck1117, line 16: if you really want to like c++, go for the magic of `enum class`!

dron0gus: Step by step. I got too many errors when try to convert to enum class.

rusefillc:
We also have https://github.com/rusefi/rusefi/pull/4648

PR #4641, opened 2022-10-06
gcc 10

mck1117:
Current issue is that the code size is much bigger from GCC 10 for no apparent reason.

Seems the same at -Os, much bigger (~20%?) at -O2 than the same program from 9.3.1


mck1117:
tested locally, this does appear to make a binary that works

mck1117:
Looks good to me - don't worry about 429 failure, that's fixed by #4679. In fact, that PR was developed against GCC10.

PR #4643, opened 2022-10-06
gcc10 preparation -fno-inline-functions

rusefillc:
cherry picking -fno-inline-functions


mck1117:
let me confirm whether this actually changes anything on gcc9, I believe this defaults to off at -o2 in gcc9, and ON at -o2 in gcc10, which was causing most of the difference

mck1117:
confirmed that this makes no difference on 9.3.1 at -O2

mck1117 (approved):

Issue #4635, opened 2022-10-05
4B11 trigger error, double RPM


ElDominio:
https://rusefi.com/online/view.php?log=1065

at 7.8 seconds RPM jumps to double RPM

rusefillc:
Double RPM is usually cam noise but this tune already has 3000 cam sync limit

https://rusefi.com/online/view.php?msq=1249&dialog=triggerConfiguration_IO

![image](https://user-images.githubusercontent.com/48498823/194115942-126db8e4-17ef-4ba0-95e5-f63fbcd3997a.png)

rusefillc:
camResyncCounter did not go up so that's NOT cam resync

Maybe it's legit trigger noise? Ideally root cause of noise should be addressed. I can add resiliency gap ratios into decoder but that would only hide the problem. 

Reminder about https://github.com/rusefi/rusefi/issues/4507 :(

![image](https://user-images.githubusercontent.com/48498823/194117529-30c103e6-550b-4649-8544-bdb0a71b5d36.png)

ElDominio:
Any type of filtering would be nice, since the volume of these units will be with plug and play harness.

mck1117:
> I can add resiliency gap ratios into decoder but that would only hide the problem.

This won't help, the warning is that there were *too many* teeth. Also, look at instant RPM: it's fuzzy:

![image](https://user-images.githubusercontent.com/568254/194180596-23d57fe5-c702-4c3a-991f-6acbc6565272.png)

He's going to try inverting trigger, adjust offset, and see if it goes away (ie, some hall sensors are crap on one edge or the other)

ElDominio:
Fixed by setting cam to inverted.

mck1117:
@ElDominio, did the instant rpm get not-fuzzy?

mck1117:
That looks pretty fuzzy to me. Did you try inverting the primary trigger?

ElDominio:
Yes, that log is inverted.

ElDominio:
UPDATE:
Not fixed. I had set a max cam/crank sync RPM, which made it not lose sync. If Max RPM for sync = 0, it loses sync at about 5k rpm. Setting a max RPM for sync = 3000 lets it rev out without a problem, but VVT doesn't work (since it seems this value makes the ECU just completley forget about the existence of VVT)

Log with Max Sync RPM disabled (=0)
https://rusefi.com/online/view.php?log=1069

Log with max cam sync = 4000RPM
[...]

PR #4645, opened 2022-10-08
fix 36-2-1 trigger

mck1117:
36-2-1 has the wrong number of teeth on either side of the -1 gap

This change:
- Fixes the trigger pattern
- Adds test of real running 4b11 engine by checking that `triggerToothAngleError` is reasonable for every single tooth
- Add that same test to some other triggers for good measure

#4635


rusefillc:
Please changelog

rusefillc (changes requested):
changelog

Issue #4651, opened 2022-10-10
efiPrintf does not accept some special chars


dron0gus:
Message parser on PC side confused by '\r' or '\n' (or both). And console skips some following messages.
```
I 221010 115627.167 [communication executor1] MessagesCentral - postMessage EngineState: Sleeping...
I 221010 115628.174 [communication executor1] MessagesCentral - postMessage EngineState: Restarting at 115200
I 221010 115628.177 [communication executor1] MessagesCentral - postMessage EngineState: sending AT
I 221010 115628.177 [communication executor1] EngineState - EngineState.unknown: `msg`Received 5 +OK
I 221010 115628.178 [communication executor1] EngineState - Invalid key [] value [`msg]
I 221010 115628.178 [communication executor1] EngineState - EngineState.unknown: Received 5 +OK
I 221010 115628.178 [communication executor1] EngineState - EngineState.unknown: `msg`sending AT+VERSION
I 221010 115628.178 [communication executor1] EngineState - Invalid key [] value [`msg]
I 221010 115628.178 [communication executor1] EngineState - EngineState.unknown: sending AT+VERSION
I 221010 115628.179 [communication executor1] EngineState - EngineState.unknown: `msg`Received 5 +VERS`msg`command AT+VERSION
I 221010 115628.179 [communication executor1] EngineState - Invalid key [] value [`msg]
I 221010 115628.179 [communication executor1] EngineState - EngineState.unknown: Received 5 +VERS`msg`command AT+VERSION
I 221010 115628.179 [communication executor1] EngineState - Invalid key [Received 5 +VERS] value [`msg]
I 221010 115628.179 [communication executor1] EngineState - EngineState.unknown: command AT+VERSION
I 221010 115628.179 [communication executor1] EngineState - EngineState.unknown:  failed -1`msg`FAIL! Only 0 commands (of 4 total) were passed to the Bluetooth module!`msg`Finished with -4 return`msg`UART: TS: ERROR: no second byte`msg`TunerStudio size=756 / total=653 / errors=1 / H=3 / O=313 / P=0 / B=0`msg`TunerStudio W=0 / C=0 / P=0`
I 221010 115628.180 [communication executor1] EngineState - Invalid key [ failed -1] value [`msg]
[...]

PR #4652, opened 2022-10-10
Logging asserts

dron0gus:
Related to https://github.com/rusefi/rusefi/issues/4651
Wrong symbol can still come from one of arguments, like:
`efiPrintf("This is test %s", "this is wrong char\r\n");`
So check should be done for final buffer.
Also check for '\r' too.
Or just simply replace all not-allowed chars with space.

@rusefillc please cherry-pick at least first patch at this series.

### DIFF
--- a/firmware/util/loggingcentral.cpp
+++ b/firmware/util/loggingcentral.cpp
@@ -182,11 +182,6 @@ void efiPrintfInternal(const char *format, ...) {
 	}
 #endif
 #if (EFI_PROD_CODE || EFI_SIMULATOR) && EFI_TEXT_LOGGING
-	for (unsigned int i = 0; i < strlen(format); i++) {
-		// todo: open question which layer would not handle CR/LF properly?
-		efiAssertVoid(OBD_PCM_Processor_Fault, format[i] != '\n', "No CRLF please");
-	}
-
 	LogLineBuffer* lineBuffer;
 	msg_t msg;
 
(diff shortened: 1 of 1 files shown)


firmware/util/loggingcentral.cpp
```diff
@@ -210,6 +208,13 @@ void efiPrintfInternal(const char *format, ...) {
 	// Ensure that the string is comma-terminated in case it overflowed
 	lineBuffer->buffer[sizeof(lineBuffer->buffer) - 1] = LOG_DELIMITER[0];
 
+	for (unsigned int i = 0; i < strlen(lineBuffer->buffer); i++) {
```
mck1117, line 211: we know the length of the written buffer, right? why do we need `strlen` again?

dron0gus: Done!

firmware/util/loggingcentral.cpp
```diff
@@ -204,11 +202,20 @@ void efiPrintfInternal(const char *format, ...) {
...
 	va_end(ap);
 
 	// Ensure that the string is comma-terminated in case it overflowed
-	lineBuffer->buffer[sizeof(lineBuffer->buffer) - 1] = LOG_DELIMITER[0];
+	if (len > sizeof(lineBuffer->buffer) - 1)
+		len = sizeof(lineBuffer->buffer) - 1;
+	lineBuffer->buffer[len] = LOG_DELIMITER[0];
+
+	for (size_t i = 0; i < len; i++) {
+		/* just replace all non-printable chars with space
```
mck1117, line 214: `//` vs `/* */` comments?

mck1117 (approved):

PR #4653, opened 2022-10-10
Bluetooth jdy33

dron0gus:


dron0gus:
Tested on MRE with JDY-33 attached using extra connector, UART4.
Please ping me if this breaks something on other BT modules...

rusefillc:
Will test when I have next batch of any boards with JDY33

Issue #4650, opened 2022-10-09
use triggerToothAngleError to detect wrong trigger (and/or noise)


mck1117:
We compute the delta between the actual and expected timing of every trigger tooth (`triggerToothAngleError`), and expect this to be near zero. If it's far from zero, that means the trigger pattern is maybe wrong or mis-synced.

Issue #987, opened 2019-10-22
better handling of unexpected trigger shape


rusefi:
at the moment we have poor usability around trigger shape troubleshooting

1) engine sniffer shows what exactly? electrical levels or logic levels with "inverted" accounted for?

2) does engine sniffer match trigger images from trigger library https://rusefi.com/wiki/index.php?title=Manual:Software:Trigger

3) do we have links to trigger library in TS? in console?

4) shall we overlap expected shape on actual shape right in engine sniffer?

PR #4656, opened 2022-10-11
detect and ignore doubled trigger edges

mck1117:
Detect when we get a "doubled edge" trigger input.

Will enable the second check (very wrong tooth timing) later, as it causes problems at the moment.

#987 #2222 #4650


mck1117:
this looks also OK, proteus CI still sad 🙁

rusefillc:
What do we perceive to be root cause here? Is it expected to get double edges or it is some specific hardware issue?

Do we want to changelog this improvement?

firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -641,6 +641,62 @@ void TriggerCentral::decodeMapCam(efitick_t timestamp, float currentPhase) {
 	}
 }
 
+bool TriggerCentral::isToothExpectedNow(efitick_t timestamp) {
```
rusefillc, line 644: See open comments on ticket

Please rebase to get green build

rusefillc (changes requested):

rusefillc:
@ElDominio tell me about 4b11 is it Hall or VR?

rusefillc:
@mck1117 please blink if alive

mck1117:
I'm alive! Was racing this weekend and getting ready before that.

rusefillc:
What do we perceive to be root cause here? Is it expected to get double edges or it is some specific hardware issue?

Do we want to changelog this improvement?

rusefillc:
@mck1117 is this abandoned?

ElDominio:
My only question with this is how do some other ECU companies trigger properly on both edges (no crazy sync loss)

mck1117:
> My only question with this is how do some other ECU companies trigger properly on both edges (no crazy sync loss)

sort of unrelated discussed offline re: #4635

mck1117:
> What do we perceive to be root cause here? Is it expected to get double edges or it is some specific hardware issue?

It is a case we can detect cheaply, so why not detect it? Perhaps it will be a useful metric to detect issues we know about but haven't narrowed down to a root cause.

rusefillc:
> > What do we perceive to be root cause here? Is it expected to get double edges or it is some specific hardware issue?
> 
> It is a case we can detect cheaply, so why not detect it? Perhaps it will be a useful metric to detect issues we know about but haven't narrowed down to a root cause.

I like this justification! My preference would be to have this filtering enabled by default but have an option to turn it off, what do you think? I am fine adding the option later myself.

mck1117:
> I like this justification! My preference would be to have this filtering enabled by default but have an option to turn it off, what do you think? I am fine adding the option later myself.

Right now it only filters truly doubled edges (no trigger should have 1/2 degree tooth spacing), but sure

Issue #3985, opened 2022-03-07
SD card log fields list should be auto-generated


rusefillc:
At the moment we generate list of fields we log in TS and so we log 600 data points via Tuner Studio "start log" button

We shall auto-generate list of entries in SD card logs as well since at the moment SD card has only about 60 data points :(

rusefillc:
And now Drew has his this :( https://rusefi.com/forum/viewtopic.php?p=46394#p46394

PR #4657, opened 2022-10-11
allow SD logging non-scaled_channel

mck1117:
- Allow `LogField` to accept plain integer/float, instead of requiring `scaled_channel`.
- Remove `autoscale` in places where we have 1:1 scaling, using the new logic.

useful for #3985


rusefillc:
Thank you!

mck1117:
That hardware ci error seems unrelated, all PRs have the same failure. Plus, no SD card on HW CI so it shouldn't be failing.

PR #4661, opened 2022-10-13
build tweaks for size

mck1117:
- Turn off 16-byte function alignment on MCUs without a cache - saves 1-15 bytes added between functions for alignment
- use `chsnprintf` instead of `snprintf`
- Don't bother with trying to compute the locale-specific decimal point for lua (`lua_getlocaledecpoint`), just use `'.'`.
- Turn on section "sort by alignment", avoids useless fill gaps between linked objects.


mck1117:
Looks like just our favorite proteus fail.

Issue #3680, opened 2021-12-09
Throttle synchronization


rusefillc:
Zeiss says:

```
Throttle synchronization: this is now only the thought process of me, I do not know if it is 100% correct. The M70 has two independent throttles, each with its own TPS. Via the TPS Proteus knows how far the ETB are open, but only approximately. This means that if the values (angle to resistance) differ from each other, the ETB are opened differently. Correct? Via the ClosedLoop, the injection quantity is adapted to the air quantity, in each case per bank. Only it can be that one bank delivers more power than the other, because one bank gets more air (and as a consequence also gasoline) than the other. This is the reason why the two ETBs have to be synchronized against each other. It is not enough to calibrate them against the end stops, they must be balanced against each other. When they are properly balanced, not only is the idle much better and more stable, but so is throttle response and power.

The strategy here would be to measure the air mass (=MAF) or pressure (=MAP) and then open or close the ETB until the values match. BMW controls the ETB in the opposite direction during the adjustment. For this we need the evaluation for the second MAF or MAP sensor. I personally would prefer MAF, a normal analog one, not a digital one. After all, we already have the implementation for MAF/MAP, you just need to duplicate it and implement the algorithm for synchronization. It would make not only me happy, but everyone who has a V-motor with two ETBs.
```

billstewartx:
Jaguar V12 from 1995 also has 2 MAP sensors and 2 throttle bodies. So it would also benefit from this sync update.  I have a V12 setup  for a test mule.

HeinrichG-V12:
I've found so interesting infos how BMW syncronized the TBs:

1. get air mass from MAF (MAF because those engines use MAF sensors)
2. calculate injection amount (get this value from fuel map, "injection amount Y for air mass X")
3. calculate ti-time from injection amount (ti-time is injection duration or injector OPEN time)
4. use ti-time for TB sync

I think it absolutely doesn't matter if you have a MAP or MAF, the most important thing is to have one ti-signal per bank and compare these two signals to synchronize the TBs. In the end you don't synchronize the opening angles but the throughput.

mck1117:
> use ti-time for TB sync

this is a vestige of how the computer was built (easier to compute with injection time than fuel quantity), but we probably don't want to do that. We live in a world now where the fuel _pressure_ might be different per bank (dual bank GDI, anyone?), so the injection time might be different for each bank to deliver the same fuel.

HeinrichG-V12:
[...]

PR #4677, opened 2022-10-18
prep for multiple MAP sensors

mck1117:
Encapsulate MAP averaging in to its own sensor class so that we can add a second one more easily.

#3680


mck1117:
Ready! Soon we will have multi-MAP!

PR #4679, opened 2022-10-19
more f429 business

mck1117:
- missed a comma in firmware yaml
- Add vscode build config
- LED pins
- Init ethernet pins at the correct time
- Minor tweaks that let you build without LTO (easier debugging maybe?)

Ethernet tested working on default build for both F429 and H743 nucleo boards.


firmware/console/ethernet_console.cpp
```diff
@@ -81,23 +81,6 @@ struct EthernetThread : public TunerstudioThread {
 static EthernetThread ethernetConsole;
 
 void startEthernetConsole() {
-#ifndef STM32H7
-	// TODO: why does this break H7? I thought the pins were the same?
```
rusefillc, line 85: Does this remove init from nucleo nucleo-F7 config?

rusefillc (changes requested):

firmware/console/ethernet_console.cpp
mck1117: In fact, now they just use the same file.

mck1117:
current status:

image

firmware/console/ethernet_console.cpp
```diff
@@ -81,23 +81,6 @@ struct EthernetThread : public TunerstudioThread {
...
-	efiSetPadMode("ETH_RMII_REF_CLK",  Gpio::A1, PAL_MODE_ALTERNATE(0xb));
```
rusefillc, line 86: you've killed my fancy pin names

not that those are really needed

mck1117: those fancy names use extra flash memory 🙂

PR #4684, opened 2022-10-20
Adc unregister

### LINKED ISSUE #4683: Sensor: unregister on the fly
have CLT assigned to any ADC input
set input to NONE, click 'burn'
go to lua

```
cltSensor = Sensor.new("clt")
```

expected behaviour: happy script 

actual result:
```
2022-10-19_19_52_40_965: EngineState: LUA ERROR loading script: [string "-- start of technical boring details please i..."]:50: Tried to create a Lua sensor of type Clt, but one was already registered.

```

at the moment full power cycle is needed for real analog sensor to un-register

rusefillc:
Sensor: unregister on the fly fix #4683

### DIFF
--- a/firmware/hw_layer/adc/adc_subscription.cpp
+++ b/firmware/hw_layer/adc/adc_subscription.cpp
@@ -95,7 +95,7 @@ static AdcSubscriptionEntry* findEntry() {
 	efiSetPadUnused(getAdcChannelBrainPin("adc unsubscribe", entry->Channel));
 #endif // EFI_PROD_CODE
 
-	if (entry->Sensor != nullptr) {
+	if (entry->Sensor) {
 		entry->Sensor->unregister();
 	}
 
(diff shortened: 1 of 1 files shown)


firmware/hw_layer/adc/adc_subscription.cpp
```diff
@@ -95,6 +95,10 @@ static AdcSubscriptionEntry* findEntry() {
 	efiSetPadUnused(getAdcChannelBrainPin("adc unsubscribe", entry->Channel));
 #endif // EFI_PROD_CODE
 
+	if (entry->Sensor) {
```
mck1117, line 98: we don't need to use `entry->Sensor`, we can just use the `sensor` parameter (which should is `== entry->Sensor`)

mck1117: And then also don't need a branch.

firmware/hw_layer/adc/adc_subscription.cpp
```diff
@@ -95,6 +95,8 @@ static AdcSubscriptionEntry* findEntry() {
 	efiSetPadUnused(getAdcChannelBrainPin("adc unsubscribe", entry->Channel));
 #endif // EFI_PROD_CODE
 
+	entry->Sensor->unregister();
```
mck1117, line 98: no no, just do `sensor.unregister();`

rusefillc: Sorry, Netflix and past midnight on east coast

mck1117: > Netflix and past midnight

it happens!

PR #4691, opened 2022-10-22
progress: rpm register based on trigger input being none or not

rusefillc:


mck1117:
use the helper function instead of checking if gpio equal to unassigned

PR #4693, opened 2022-10-22
allow all pins for fast inputs on default builds

mck1117:
We use PAL now, so all pins work for "fast" inputs on Frankenso, etc.


mck1117:
poke, we have a real user asking for it with a very frankensteined frankenso:

image

PR #4694, opened 2022-10-23
Fixed links

Akash190104:
Fixed links to redirects to make the readme look less cluttered. Added the git clone line.


rusefillc:
thank you @Akash190104 !

PR #4705, opened 2022-10-27
 prep for multiple MAP sensors

mck1117:
try #4677 again


mck1117:
Looks like it fits on kinetis now!

PR #4723, opened 2022-10-31
TS UI suggestions by ELDominio

dron0gus:
Idle position from Extra 2 to Idle category
idle AC adders to AC dialog
Fan idle adders to Fan dialog


ElDominio:
General consensus in slack is to duplicate, what do you think?

PR #4725, opened 2022-10-31
TS: UI: rename Pin to input or output where possible

dron0gus:
Rename Pin to input or output where possible
Keep Pin term only for internal connections (SD, SPI...)


rusefillc:
Sweeeeeet:)

PR #4738, opened 2022-11-06
does gcc11 know something?

rusefillc:


rusefillc:
@nmschulte any chance you would be available to PR the fix? need smarter conditional compilation here

rusefillc:
@nmschulte no I mean fix legit compilation erro specifically with cypress configuration, see red checkmark on github compilation status

nmschulte:
I'll see what I can do.

nmschulte:
FWIW, I cannot get even `master` to build w/ `PROJECT_BOARD=cypress PROJECT_CPU=cypress`, complaining about `EFI_FILE_LOGGING` in `ChibiOS/os/various/fatfs_bindings/fatfs_diskio.c`:

```
ChibiOS/os/various/fatfs_bindings/fatfs_diskio.c:30:2: error: #error "MMC_SPI or SDC driver must be specified"
   30 | #error "MMC_SPI or SDC driver must be specified"
      |  ^~~~~
```

I'm very new to the project so missing a lot of context.  I'm not using the `setup_linux_environment.sh` or any other configuration-like mechanisms, just running `make` -- it seems like this should work but perhaps I'm missing something re: configuration.  I'll analyze the Jenkins/CI scripts, and also try the version of GCC referenced in `setup_linux_environment.sh`, or anything else you direct me to.

mck1117:
cypress and kinetis are "weird" - try running their compile scripts instead of just bare `make`

nmschulte:
> cypress and kinetis are "weird" - try running their compile scripts instead of just bare `make`

That resolved things (`export USE_FATFS="no", etc.), thanks; I'm able to reproduce the issue.  Is this "weird"ness intentional, then?  Seems like the board/cpu could set these "as normal."  Anyway, I'll dig more into the semantic issue later tongiht :+1:.

mck1117:
> Is this "weird"ness intentional, then?

Not really, no. IMO all those settings should get set in the `*.mk` files, not in shell scripts. The makefiles are far, far more portable than the wacky shell script way of doing it. That might be another good one to tackle.

PR #4747, opened 2022-11-07
fix vvt logging logic

mck1117:
Make VVT logging logic do the right thing (fix the todo in #4621). It shouldn't be looking at `engineConfiguration->vvtCamSensorUseRise` at all.

### DIFF
--- a/firmware/controllers/trigger/trigger_central.cpp
+++ b/firmware/controllers/trigger/trigger_central.cpp
@@ -210,7 +210,7 @@ static angle_t wrapVvt(angle_t vvtPosition, int period) {
 	return vvtPosition;
 }
 
-static void logFront(bool isImportantFront, efitick_t nowNt, int index) {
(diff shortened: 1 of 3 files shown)


firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -210,42 +210,37 @@ static angle_t wrapVvt(angle_t vvtPosition, int period) {
...
-
-			addEngineSnifferVvtEvent(index, FrontDirection::DOWN);
-			addEngineSnifferVvtEvent(index, FrontDirection::UP);
 		}
 	}
 }
 
 void hwHandleVvtCamSignal(TriggerValue front, efitick_t nowNt, int index) {
 	TriggerCentral *tc = getTriggerCentral();
-	if (tc->directSelfStimulation || !tc->hwTriggerInputEnabled) {
```
rusefillc, line 245: A bit surprised to see these lines removed under "logging logic"?

mck1117: as you should be, that was for testing 🙂

PR #4750, opened 2022-11-07
fix BACKUP_FLASH_ADDR assertion semantic

nmschulte:
for https://github.com/rusefi/rusefi/pull/4738#issuecomment-1306111813

### DIFF
--- a/firmware/hw_layer/ports/cypress/backup_ram.cpp
+++ b/firmware/hw_layer/ports/cypress/backup_ram.cpp
@@ -35,7 +35,7 @@ static void backupInit() {
 	// we cannot trust the saved data anymore, until it's saved in backupRamFlush()
 	// so we mark is as 'pending'
 	backupRam[backupStateOffset] = BACKUP_PENDING;
-	intFlashWrite(BACKUP_FLASH_ADDR + backupStateOffset, (char *)backupRam, sizeof(backupRam[backupStateOffset]));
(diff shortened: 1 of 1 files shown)


mck1117:
should we just be comparing against 0? Or is this actually a pointer? (ie, do we ever dereference `BACKUP_FLASH_ADDR` as a pointer?)

rusefillc:
@mck1117 this all started with gcc11 detecting suspicious/illegal memcpy operations probably around intFlashRead/intFlashWrite

mck1117:
> @mck1117 this all started with gcc11 detecting suspicious/illegal memcpy operations probably around intFlashRead/intFlashWrite

I'm aware of that. I'm asking what the _right_ thing to do is, instead of casting a pointer type to an integer for a comparison (which is a code smell at best)

nmschulte:
`BACKUP_FLASH_ADDR` is typically (except for a `...Write) cast as `flashaddr_t`, which is a `uintptr_t`, so I opted for this than to use `NULL`

mck1117:
I would be in favor of making `flashaddr_t` a pointer type, instead of an integer type. We're using it as a pointer (cast to char* right before the memcpy...)

mck1117:
`uintptr_t` is not a pointer - it's an integer type guaranteed to be at least as large as the pointers on your architecture. Used when you need to store a pointer as an integer instead of a pointer.

nmschulte:
Ah I see.  My C++ is rusty.  Is `uint32_t *` appropriate then?

PR #4754, opened 2022-11-08
Gcc 11 hw ci

mck1117:


nmschulte (approved):

mck1117:
Compiler at least runs on hw ci, we can merge this with red as I think it's the same issue as in the base branch.

PR #4756, opened 2022-11-08
setup_linux_environment.sh tar uses --auto-compress

nmschulte:
avoids e.g.:

```
bzip2: (stdin) is not a bzip2 file.
tar: Child returned status 2
tar: Error is not recoverable: exiting now
```


nmschulte:
already part of https://github.com/rusefi/rusefi/pull/4757

Issue #4759, opened 2022-11-08
We have two usages of wget and one usage of curl


rusefillc:
We have three different lines downloading same GCC compiler - two usages of wget and one usage of curl

We shall extract "download_and_uncompress_gcc.sh" script and reference that script from three places shall we?

rusefillc:
@nmschulte @mck1117 can you please ^^^?

nmschulte:
#4761 resolves this.

PR #4761, opened 2022-11-08
factor GCC provision script

nmschulte:
for #4759 -- uses some bash variable expansion magic which may prove problematic for certain systems
`provide_gcc.sh` downloads and extracts GCC toolchain and creates a colloquially named `gcc-arm-none-eabi` link for use; this could change to accept a variable for that name instead, but this colloquial name allows to encapsulate this term and avoid having to pass it back or require the use-sites to determine the name (which changes among releases/builds).


nmschulte:
I had a logic error, deleting the archive before determining the directory name ... fixed, but at the expense of two CI runs ... :facepalm:

PR #4762, opened 2022-11-08
use build_support ARM GCC 11.3 toolchain

nmschulte:
via https://github.com/rusefi/build_support/pull/5
for https://github.com/rusefi/rusefi/pull/4680#issuecomment-1307961836

### DIFF
--- a/firmware/provide_gcc.sh
+++ b/firmware/provide_gcc.sh
@@ -2,7 +2,7 @@
 
 # Download and extract GCC arm-none-eabi toolchain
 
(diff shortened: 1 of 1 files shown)


rusefillc:
image

@nmschulte please add return code check to fail task if download fails? even while it should not be failing now :)

nmschulte:
> @nmschulte please add return code check to fail task if download fails? even while it should not be failing now :)

somehow the `set -e` I had intended was lost; it's there now, should resolve this problem

PR #4765, opened 2022-11-10
random tidying

nmschulte:


firmware/controllers/settings.cpp
```diff
@@ -1076,14 +1076,14 @@ static void setValue(const char *paramStr, const char *valueStr) {
 #if EFI_PROD_CODE
 	} else if (strEqualCaseInsensitive(paramStr, CMD_VSS_PIN)) {
 		setVssPin(valueStr);
-#endif // EFI_PROD_CODE
+#endif /* EFI_PROD_CODE */
```
rusefillc, line 1079: we have no consistency here :(

mck1117: IMO we should have no single line `/* */` comments anywhere

PR #4770, opened 2022-11-12
unneeded logic in ETB init?

rusefillc:


mck1117:
Nope, we very much need that logic. It lets you configure either throttle without bothering to do the math on any that don't exist.

unit_tests/tests/actuators/test_etb.cpp
```diff
@@ -199,16 +200,25 @@ TEST(etb, initializationNoPrimarySensor) {
...
 	EtbController dut;
 
-	// Needs pedal for init
-	Sensor::setMockValue(SensorType::AcceleratorPedal, 0.0f, true);
+	EngineTestHelper eth(TEST_ENGINE, [](engine_configuration_s* engineConfiguration) {
+	});
 
-	// Redundant, but no primary configured
-	Sensor::setMockValue(SensorType::Tps1, 0.0f, true);
+	engineConfiguration->tps1_1AdcChannel = EFI_ADC_NONE;
```
mck1117, line 206: we should never be setting pins in a unit test like this, there are explicit TPS init tests, and this is not one of them

rusefillc:
it looks to me that hasSensor(functionToTpsSensorPrimary(function) should be incapsulated into "Sensor::isRedundant(m_positionSensor)". Actually it looks like Sensor::isRedundant(m_positionSensor) is always false when hasSensor(functionToTpsSensorPrimary(function) is false.

@mck1117 please clarify why do we need both checks and not only the second check

mck1117:
looking

please include info like that instead of this:

mck1117:
> why do we need both checks and not only the second check

If the sensor DOES NOT exist, then you probably didn't mean to operate this throttle in the first place.

If the sensor DOES exist but is not redundant, the throttle should not run, and we should throw an error because it means you intended to operate this throttle, but your TPS isn't safe.

The first case simply returns `false`, and the second case throws an error then returns `false`.

rusefillc:
perfect, we agree that both return "false" and it sounds like we agree that "no primary" is a very narrow debatable use-case?

i would prefer to add simplicity for my evil SENT reasons at the cost of not handling no primary without exception

if you do not want ETB and you have no TPS users would still be able to disable ETB function. my point "ETB function says ETB it's legit to fail if unhappy ETB setup"

mck1117:
> i would prefer to add simplicity for my evil SENT reasons at the cost of not handling no primary without exception

It looks like the TPS init logic actually skips init of the whole thing if not configured. We can maybe just make the change dron proposed in his PR instead of removing completely. That covers all cases - preserves existing logic, but also lets you supply your own redundant SENT sensor instead.

> my point "ETB function says ETB it's legit to fail if unhappy ETB setup"

If we want to do that, we should at least fire a separate fatal instead of a misleading one about something that isn't the real issue

rusefillc:
Looks like I will close in light of https://github.com/rusefi/rusefi/issues/4688#issuecomment-1312367020

PR #4797, opened 2022-11-18
Docker containerized build

nmschulte:
TL;DR, examples: --

build the images:
`UBUNTU_VERSION=20.04 sh -c 'docker-compose --file rusefi.docker-compose.yml build --build-arg UBUNTU_VERSION=${UBUNTU_VERSION}'`

build the firmware w/ "build-primary-bundle" flow, though this needs work to split or unify ... :
`UBUNTU_VERSION=20.04 docker-compose --file rusefi.docker-compose.yml run --rm firmware`

... w/ "build-firmware" flow (currently commented out, in favor of the "build-primary-bundle" currently):
`UBUNTU_VERSION=20.04 docker-compose --file rusefi.docker-compose.yml run --rm firmware kinetis PROJECT_CPU=kinetis USE_VERBOSE_COMPILE=true`

----

This is the rough beginnings of creating a portable/containerized build environment, currently for firmware and simulator projects.  It's functional but still needs a lot of work to reach its full potential, but the existing build scripts need factored/unified or other work to help support it and to avoid/remove duplication of concerns.

The names used aren't perfect yet either, but they're a good start and shouldn't collide or have portability concerns.  Some patterns I already think to change are:
- the firmware and simulator images are only used for build; they can be names as such
- the `rusefi.` prefix on `docker-compose.yml` can probably be changed to like `build.`, or dropped if no separate composition seems fruitful
  -  I envision one for running the simulator and apps and tools for example
- `firmware/provide_gcc.sh` doesn't belong in `firmware/` necessarily
- the flow(s) (i.e. `firmware/build_docker.sh`) should really belong directly in the `Dockerfile`s, to reap all of the benefits of containerization
- the containers run as/with root permissions, which causes permissions issues with the created files on the host
  - possibly a simple fix is to leverage `USER` (though it's unclear how to identify the/a valid uid, but maybe `1000` goes far; otherwise maybe an `ARG` or `ENV` can make this flexible)
  - Docker's user namespace remapping helps mitigate this in a better way, but it was immature/required system-level changes to enable last I know

The `UBUNTU_VERSION` build argument defines the tag of `ubuntu` image to build with, and can be set via the `docker-compose` build paradigm with the pattern shown above.  `.env` files can be used (and specified via CLI so as to provide a namespace and aid portability, a la `rusefi.docker-compose.yml`), but this pushes further toward providing an orchestration/script to issue these commands.


rusefillc:
@nmschulte please rebase since some commits were cherry picked

rusefillc:
@nmschulte this looks abandoned feel free to reopen in case of progress

nmschulte:
It's on the back back burner; I use this to build, if I improve it to the point of sharing and allowing further integration / use I will.  Thanks!

Issue #3014, opened 2021-07-21
Let's admit it it's time to move to gradle


rusefillc:
Historically all java code was build by ant

At the moment we already have android app building using gradle so much of the code already builds under gradle and we have both and and gradle which is not optimal

we shall completely migrate from ant to gradle for all java code

rusefillc:
Related to #2857

rusefillc:
@Come1LLF00 is this related to this change? two sets of generated files looks like one set uses wrong path?

![image](https://user-images.githubusercontent.com/48498823/127712221-3a100520-2a9b-4887-9741-592149990137.png)

rusefillc:
this could be a great task to offshore with $$$ once I have $ and time to manage

rusefillc:
overall status: we are pretty close!

very few outstanding issues:

open question java console artifact size if compiled with ant VS if compiled with shadow_jar gradle task

open question of java console shadow_jar has all .dll and other resources

rusefillc:
@nmschulte we already have java_console gradle it just PROBABLY does not produce amazing result

```
cd java_tools
gradlew :ui:shadowJar
```

nmschulte:
(OpenJDK 8) `ant jar`: 11249099 bytes - 100%

```
rusefi/java_console$ ls -l ../java_console_binary/rusefi_console.jar 
-rw-r--r-- 1 nmschulte nmschulte 11249099 Nov 20 21:28 ../java_console_binary/rusefi_console.jar
```

(OpenJDK 8) `./gradlew :ui:shadowJar`: 23177023 bytes - 206%
```
@desmas-l-hpdo:rusefi/rusefi/java_tools$ ls -l ../java_console/ui/build/libs/ui-all.jar 
-rw-r--r-- 1 nmschulte nmschulte 23177023 Nov 20 22:35 ../java_console/ui/build/libs/ui-all.jar
```

nmschulte:
Here's a repository w/ the contents of the `gradle` jar atop the `ant` jar; the last commit shows a lot of extra files, many of them `.../test/` files, though the `ant` jar has some of this as well.

https://github.com/nmschulte/rusefi-java_console-jar-ant-gradle-diff

rusefillc:
@nmschulte thank you this was extremely helpful I did not realize close we are!

without com.ibm.icu jar file is only 20% larger I can live with that. Next step understanding the differences around embedded .cpp source code resources and checking what files does ant version have which are not present in gradle version

[...]

PR #4800, opened 2022-11-19
javac builds use iso-8859-1 encoding

nmschulte:
```
[javac] /rusefi/rusefi/java_console/io/src/main/java/peak/can/basic/MutableTPCANHandle.java:4: error: unmappable character for encoding ASCII
[javac]  * @LastChange $Date: 2020-08-07 16:40:53 +0200 (ven., 07 août 2020) $
[javac]                                                             ^
```


rusefillc:
This PR seems to focus on Ant config files while we are moving to gradle https://github.com/rusefi/rusefi/issues/3014

nmschulte:
It seems `gradlew` still respects the `build.xml` then, as these changes resolve the problem (when using `gradlew`, via `misc/jenkins/functional_test_and_build_bundle/build_current_bundle.sh`, via e.g. `misc/jenkins/build_java_console.sh`).

Is there a different place I should do this?

rusefillc:
No, ant and gradle output into different folder thus clean different folders

nmschulte:
I pruned this to just the `java_console` build, which should eventually go away with #3014 (and the shell script ported to `gradlew` at the same time).

PR #4809, opened 2022-11-21
add /usr/sbin to PATH in create_ini_image*.sh

nmschulte:
e.g. `./gen_default_everything.sh` fails on Debian systems because e.g. `mkfs.fat` is not in regular user's `PATH`
```
hw_layer/mass_storage/create_ini_image.sh: line 26: mkfs.fat: command not found
```

### DIFF
--- a/firmware/hw_layer/mass_storage/create_ini_image.sh
+++ b/firmware/hw_layer/mass_storage/create_ini_image.sh
@@ -15,6 +15,9 @@ BOARD_SPECIFIC_URL=$5
 IMAGE=ramdisk.image
 ZIP=rusefi.ini.zip
 
+# mkfs.fat and fatlabel are privileged on some systems
+PATH="$PATH:/usr/sbin"
+
 echo "create_ini_image: ini $FULL_INI to $H_OUTPUT size $FS_SIZE for $SHORT_BOARDNAME [$BOARD_SPECIFIC_URL]"
 
 rm -f $ZIP $IMAGE
--- a/firmware/hw_layer/mass_storage/create_ini_image_compressed.sh
+++ b/firmware/hw_layer/mass_storage/create_ini_image_compressed.sh
@@ -14,6 +14,8 @@ BOARD_SPECIFIC_URL=$5
 
 IMAGE=ramdisk.image
 
+# mkfs.fat and fatlabel are privileged on some systems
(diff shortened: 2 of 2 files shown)


rusefillc:
@mck1117 what's your take?

Looks like an undocumented hack in the script.
Hack line needs a comment in the script I do not like "please go figure out based on history" approach to hack documentation.

But bigger picture is this a good change?

nmschulte:
Would it be clearer if I wrote explicitly in the comment that it's a Debian system that suffers?

rusefillc:
Yes it would be more clearer

@chuckwagoncomputing @ZHoob2004 what's your take on the overall hack?

nmschulte:
Can do.  It's common for Linux distros not to have `[/usr]/sbin` in normal users' `PATH`, so I expect this affects more than just Debian.  I assume Canonical/Ubuntu package `dosfstools` differently, because they're different and can't be bothered.

I also want to note that git commit history (a la `git blame`) are extremely useful with good commit messages: beats the piss out of littering the code with comments, but I digress ... no biggie: the comments are indeed convenient.

mck1117:
why is mkfs.fat not in your path? isn't that a broken package that didn't symlink itself properly?

mck1117:
> I also want to note that git commit history (a la `git blame`) are extremely useful with good commit messages: beats the piss out of littering the code with comments, but I digress ... no biggie: the comments are indeed convenient.

I've been trying for years to convince Andrey of this. It's 2022, we have git blame now.

rusefillc:
git blame is amazing only at depth 1

nmschulte:
> git blame is amazing only at depth 1

You just need better tooling; I can find you some?

nmschulte:
> why is mkfs.fat not in your path? isn't that a broken package that didn't symlink itself properly?

Historically, `mkfs.*` were only used to manipulate devices, which requires root permissions (hence `.../sbin` vs `.../bin`), and `.../sbin` is not in a normal (non-root) users' `PATH`.  It's still not on Debian, but apparently is in Ubuntu (presumably via symlink), and I'm guess it's hit/miss for other distros.

ref: https://forums.debian.net/viewtopic.php?t=65714
ref: https://www.linuxquestions.org/questions/linux-newbie-8/mkfs-not-working-in-red-hat-9-a-96945/

Issue #3614, opened 2021-11-25
logging of live data structs was: data points


mmatousek89:
@mmatousek89 reports

Why so many channels are hidden in debug?
Can we have all of them logged all time?

@rusefillc response
live data structures should be merged with existing outputs for TS, TS should see all live data structures as yet another output channel and should be able to log either of it

implementation detail
``
FragmentEntry
``

mck1117:
examples?

mmatousek89:
tps diff
wastegate dc from table
etb pri/sec diff
.... and 70others

rusefillc:
@mmatousek89 historically we were concerned about RAM and traffic

At this point I believe we can afford to have way more gauges not as debug mode.

instead of asking for "70 others" it would be best to ask specifically for the ones you need this week/next week to get results quicker in a more focused way

rusefillc:
Depends on #197

rusefillc:
@racer-coder can you help? that would mean finishing #197

racer-coder:
I agree that everything should be exposed.  I think it comes down to philosophical "how should the core of the software operate", which I'm kinda trying to tackle partly with rethinking how tables work.

Using Motec M600/800/etc software as an example,
1. First you configure all the inputs.  Inputs can be direct connected sensors or received over canbus/expansion blocks.
2. Engine management goes and computes a bunch of intermediaries.  Estimated air mass, VE after compensations, Fuel required, injector duration for fuel required, etc
3. ALL of these variables - inputs + computed variables - are available to be logged, transmitted over CAN, or to be used as inputs into other tables for more computation
4. Outputs (solenoids etc) are based on the variables + some logic

Since there are so many variables, you need a tree view to fit them so you aren't scrolling forever through some giant list.  Link actually lets you type a search phrase in the dialog box that selects variables.

So yes, there should be a centralized place to describe all variables - whether inputs or not, including units, precision, scale, # bytes, etc.  Then all the logic in the software should refer to this centralized variable list.

I think we kinda already have this for sensors, but it should be extended to everything else too.

As for logging them "all the time" or not, that comes down to how much storage we have, but I assume the MCU is powerful enough and I assume we use an external microSD card so we should have plenty of bandwidth and storage space to log everything.  Link ECU lets you select which channels you log and what refresh rate you choose for each of them.

rusefillc:
@racer-coder I think we are pretty much on the same page here

[...]

PR #4820, opened 2022-11-23
etbStatus_error does the job

rusefillc:
image

### DIFF
--- a/firmware/controllers/actuators/electronic_throttle.cpp
+++ b/firmware/controllers/actuators/electronic_throttle.cpp
@@ -511,14 +511,6 @@ expected<percent_t> EtbController::getClosedLoop(percent_t target, percent_t obs
 		m_shouldResetPid = false;
 	}
 
-	// Only report the 0th throttle
-	if (m_function == ETB_Throttle1) {
-#if EFI_TUNER_STUDIO
-		// Error is positive if the throttle needs to open further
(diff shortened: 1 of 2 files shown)


firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -511,14 +511,6 @@ expected<percent_t> EtbController::getClosedLoop(percent_t target, percent_t obs
 		m_shouldResetPid = false;
 	}
 
-	// Only report the 0th throttle
-	if (m_function == ETB_Throttle1) {
-#if EFI_TUNER_STUDIO
-		// Error is positive if the throttle needs to open further
-		engine->outputChannels.etb1Error = target - observation;
```
nmschulte, line 518: Help me understand, what's the goal here?  Is this property dead now, then?

rusefillc: @nmschulte manually copying values into "outputChannels" is part of the legacy pattern

these days we automatically log some "models" thus on the image you would see same value logged twice

this migration is still not 100% complete and there are probably corner cases. in this specific case it looks like the legacy entry is no longer needed

rusefillc: @nmschulte vaguely related to #3614

PR #4764, opened 2022-11-10
avoid ctime, use UTC ISO 8601 for CMD_DATE format

nmschulte:
for https://github.com/rusefi/rusefi/issues/4662

I have not tested these changes in any way beyond that the builds succeed.

~~Should this concern w/ backward compatibility at all?  E.g. introduce i.e. `CMD_DATE_ISO` and retain `CMD_DATE` for some time?~~ Seemingly no (see below) -- ideally the altered logic will account for this and not cause crashes about mismatches, but may.

PR #4824, opened 2022-11-23
avoid ctime in GPS UART

rusefillc:
@nmschulte looks like that's an atomic change separate?

### DIFF
--- a/firmware/hw_layer/sensors/gps_uart.cpp
+++ b/firmware/hw_layer/sensors/gps_uart.cpp
@@ -11,14 +11,12 @@
  * Kot_dnz 2014
  */
 
+#if EFI_UART_GPS
 #include "pch.h"
 
 #include <string.h>
-// todo: MISRA does not like time.h
-#include <time.h>
 
-#if EFI_UART_GPS
-
+#include "rusefi_types.h"
 #include "console_io.h"
 #include "eficonsole.h"
 #include "nmea.h"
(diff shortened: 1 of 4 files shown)


firmware/hw_layer/sensors/gps_uart.cpp
```diff
@@ -30,42 +28,43 @@ static THD_WORKING_AREA(gpsThreadStack, UTILITY_THREAD_STACK_SIZE);
...
-
-static void onGpsMessage(char *buffer) {
-
+static void onGpsMessage(const char * const buffer) {
 	gps_location(&GPSdata, buffer);
-	date_get_tm(&curTm);
 
-	if (GPSdata.quality == 4 && GPSdata.GPStm.tm_year > 0 && GPSdata.GPStm.tm_sec != curTm.tm_sec) {
-		// quality =4 (valis GxRMC), year > 0, and difference more then second
-		date_set_tm(&GPSdata.GPStm);					// set GPS time
```
nmschulte, line 65: @rusefillc this comes from `rtc_helper`, whose API changed, hence the bundling with the other changes in #4764

rusefillc: yes, that one is already in master but https://github.com/rusefi/rusefi/pull/4764/commits/a00ac0bf038809707a87dad2091595dd6d294981 was a bad example for making the point about atomic changes

PR #4829, opened 2022-11-26
without ValueProvider3D interface how do we mock?

rusefillc:
We can inline ValueProvider3D and get rid of constructor argument but would that mean wiping out test?


mck1117:
We already do elsewhere in the file like this:

```c++
EXPECT_CALL(dut, getTargetLambda(1000, FloatEq(0.8f)))
		.WillOnce(Return(5.0f));
```

I think the removed test is redundant.

PR #4830, opened 2022-11-26
refactoring: constexpr fuel computer

rusefillc:
separating refactoring from functional changes


mck1117:
once 4829 merges this will be the same as the original

rusefillc:
@mck1117 please note lack of yaml part of the change here

this PR makes a point to NOT make any functional changes

Issue #4816, opened 2022-11-21
Proteus and 4chan fail to auto-calibrate specific VAG ETB


rusefillc:
4chan rev D with JLC assembled TLE9201
start with default tune - ETB is oscillating trying to shut down or something? PCB lights are cycling, looks like some sort of shutdown of either 6A questionable wall power supply or TLE9201 protection? Same questionable wall power supply is happy to drive a much larger Ford SENT ETB - log of oscillation https://rusefi.com/online/view.php?log=1101

Now pressing "auto calibrate" getting un unhappy TPS calibration https://rusefi.com/online/view.php?msq=1310 log https://rusefi.com/online/view.php?log=1102

![image](https://user-images.githubusercontent.com/48498823/203099642-c0eb763f-5fc8-4146-8597-4268dfc3f7be.png)

![image](https://user-images.githubusercontent.com/48498823/203098337-d5af29cb-e2d8-4bed-b010-5f153d759f20.png)

rusefillc:
Oh and ETB does open completely during auto-calibration

also I have at least two units of same kind, here's another piece

![image](https://user-images.githubusercontent.com/48498823/203100800-e5ab8d39-7317-4174-8dfc-260bf3dc9d39.png)

mck1117:
I think video would be handy here

rusefillc:
@mck1117 https://youtu.be/Pwmy6Y5feR4

rusefillc:
Same exact behavior with Proteus

mck1117:
Do you have a TS log? We should be logging the raw voltage.

mck1117:
> weird not inverted TPS calibration here? with a 4 wire TPS it's not like I could wire things in a wring way?

“not inverted” is pretty normal, lots of throttles have both signals with the same sign but different slope magnitude and offset.

rusefillc:
> Do you have a TS log? We should be logging the raw voltage.

Two 4chan logs linked all the way on top, do you mean Proteus logs? I sure can grab those but behavior is 200% the same exact.

rusefillc:
This issue here is a combination of #4832 and shitty power supply where power supply was shutting down before TLE9201 would overload.
![ps-bad](https://user-images.githubusercontent.com/48498823/204117293-f563df9c-4ad8-4af0-bc4a-a471970653f4.jpeg)

[...]

PR #4838, opened 2022-11-27
ETB compensates for battery voltage

mck1117:
This ensures that your ETB will behave the same at any battery voltage, including during cranking/failed alternator/etc. Importantly this means that when you autotune PID gains with the engine off, the extra 2-3 volts don't make your throttle overtuned when you start the engine.

Sort of related to #4816 since a weak battery could cause autocal fail in theory too.


mck1117:
let's wait for release for this one

rusefillc:
@mck1117 ready to merge?

mck1117:
tested, works nicely

rusefillc:
plz CHANGELOG

Issue #4798, opened 2022-11-18
Remove/make injector duty limit configurable


ElDominio:
I have customers complaining about random cuts, after  seeing a VIDEO of injector duty, it goes over 100%, triggering the duty limit. Can we have an option that when duty is >96%, the injector outputs go static(stuck to ground without closing).

If the drivers can support this, I believe this is an easy way to let them lean out and blow their shit up, which is not our fault since the ECU is reporting that it is the current behavior; maybe choose between current behavior and this

rusefillc:
I wonder if those users are running a recent version with https://github.com/rusefi/rusefi/pull/4592 improvement

the https://github.com/rusefi/rusefi/pull/4592 with nice indicators was supposed to have helped?

ElDominio:
Indicators work, but people swear their fuel system is enough and the ECU is holding them back

ElDominio:
This is very problematic in cars where transient goes over for a split second and the limit stays engaged with the throttle at WOT (fuel is cut until car is no longer WOT). This is happening with a drift car when the user stabs the throttle close to rev limit and transient fueling makes duty go over 100% for a split second. Is there any progress on this? @mck1117 @rusefillc

rusefillc:
transient and short term 100% duty looks like a great argument to remove the limit

mck1117:
> transient and short term 100% duty looks like a great argument to remove the limit

there has been some extensive offline discussion about this - something like only engaging the limiter after some period of time is probably the right solution. Allow very brief over 100% moments, but still cut if you're far over 100% for a "long time".

It also needs some fixes to injection scheduling to happen first.

rusefillc:
> It also needs some fixes to injection scheduling to happen first.

absolutely, things would break if condition is just removed

ElDominio:
Again with my theorycrafts, couldn't the fueling algorithm just "lock" when injector duty reaches say 96%, stop calculating additional fuel and justkeep the duty cycle locked at 96%, while UI reports unreal duty? (tag: Norman says)

mck1117:
Even more evil is that we could display actual duty * 1.042 so that 96% duty displays as 100% 😈

But we shouldn't do that, there are real solutions here too.

ElDominio:
please jesus do something about this, every time i run into this at the dyno i waste valuable dyno time

rusefillc:
[...]

PR #4841, opened 2022-11-27
Allow scheduling_s to come from a pool

mck1117:
Allow getting scheduling events from a pool, instead of bringing your own along.

This will allow scheduling the same event multiple times, so that overlapping injection scheduling actually works. If you schedule over 100% duty cycle, things will be fine as you can have multiple opens/closes scheduled at the same time, and until the last close runs, the injector will stay pinned open. #4798 

Needs more testing before merge!


mck1117:
let's wait for release for this one

PR #4844, opened 2022-11-27
TriggerScheduler cleanup

mck1117:
remove unused trgEventIndex in new code path

### DIFF
--- a/firmware/controllers/system/timer/trigger_scheduler.cpp
+++ b/firmware/controllers/system/timer/trigger_scheduler.cpp
@@ -73,14 +73,13 @@ bool TriggerScheduler::scheduleOrQueue(AngleBasedEventOld *event,
 }
 
 bool TriggerScheduler::scheduleOrQueue(AngleBasedEventNew *event,
-		uint32_t trgEventIndex,
 		efitick_t edgeTimestamp,
 		angle_t angle,
 		action_s action,
 		float currentPhase, float nextPhase) {
 	event->enginePhase = angle;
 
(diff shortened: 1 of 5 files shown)


firmware/controllers/system/timer/trigger_scheduler.cpp
```diff
@@ -73,14 +73,13 @@ bool TriggerScheduler::scheduleOrQueue(AngleBasedEventOld *event,
...
 bool TriggerScheduler::scheduleOrQueue(AngleBasedEventNew *event,
-		uint32_t trgEventIndex,
 		efitick_t edgeTimestamp,
 		angle_t angle,
 		action_s action,
 		float currentPhase, float nextPhase) {
 	event->enginePhase = angle;
 
-	if (event->shouldSchedule(trgEventIndex, currentPhase, nextPhase)) {
+	if (event->shouldSchedule(TRIGGER_EVENT_UNDEFINED, currentPhase, nextPhase)) {
```
rusefillc, line 82: this line looks odd, shall we add a code comment?

rusefillc:
Wow I've missed the whole Old/New. Please add code comments elaborating on why do we have two and what is the plan.

```
struct AngleBasedEventOld : public AngleBasedEventBase {
};

struct AngleBasedEventNew : public AngleBasedEventBase {
};
```

firmware/controllers/system/timer/trigger_scheduler.cpp
mck1117: removed via better design

Issue #4832, opened 2022-11-27
ETB overheats due to constant isTpsError true/false/true/false jitter: this seems obvious enough to disable ETB at least on non running engine 


rusefillc:
Proteus
default settings
VAG ETB
https://rusefi.com/online/view.php?log=1111

evil combination over-heats TLE9201 on Proteus while consuming around 5A

![image](https://user-images.githubusercontent.com/48498823/204117222-52081531-e262-4669-a92d-0bbee3238273.png)

rusefillc:
@mck1117 @ElDominio we can definitely detect that condition of isTpsError jitter

Question: how do we handle it? Turn throttle off for 10 seconds? what if we are in the middle of a turn? take whatever TPS value and apply it for next 1 second? else?

ElDominio:
I downloaded the log, I can see isTpsError  is 1 for maybe 0.01 seconds at a time; is duty being set to 0 when this happens? On most ETB systems I have seen, errors are summed and when it goes over a certain amount of errors for a certain amount of time (or even, only if the error state is true for more than a second), THEN ETB is considered failed and ETB duty set to zero, probably for this reason

mck1117:
what if we just change the defaults so they're invalid for all throttles? Correct me if I'm wrong, but the loop we get here is

1. default throttle calibration is sometimes valid for this throttle
2. ECU tries to control throttle
3. driver chip overheats
4. user is unable to autocalibrate because chip may be overtemp'd
5. user stuck with bad defaults

mck1117:
>  evil combination over-heats TLE9201 on Proteus while consuming around 5A

How did you power Proteus in this scenario?

ElDominio:
But i can do calibration just fine? Or are you talking about PID calibration?

ElDominio:
Hello, this is to report that the problem was indeed PID settings, and i used vales that seem logical (P close to 1, i close to 10-15, D close to .1) and the TLS is cool to the touch. I suggest PID autocal uses old vales to start, so it is more of an aid than a "we can do this automatigically"

But in the end, the problem is no more. Thanks everyone

mck1117:
can you create a log of good/bad PIDs?

rusefillc:
@ElDominio here we have "ETB overheats due to constant isTpsError true/false/true/false jitter: this seems obvious enough to disable ETB at least on non running engine" which is NOT something you have experienced

PID drama would be https://github.com/rusefi/rusefi/issues/4833

[...]

PR #4854, opened 2022-11-30
ETB error counter logic

mck1117:
- Each ETB only checks its own TPS
- Increment counter on rising edges, not on both edges
- Reset counter if something "weird" happens (ignition power cycle, autotune)
- Reorder logic so counter reset works
- Update test to check multiple cycles of good/bad TPS
- Update test to check that `etbErrorCode` changes correctly
- "jitter" -> "intermittent"

#4832


rusefillc:
@mck1117 just shipped you x2 VAG pigtails and x2 Nissan pigtail

At the moment VAG ETB is the best way I know to reproduce the problem on the bench

rusefillc:
I believe that master detects intermittent TPS while this PR seems to kill that feature. Let me make some evil picking and add a new test into master.

mck1117:
> I believe that master detects intermittent TPS while this PR seems to kill that feature. Let me make some evil picking and add a new test into master.

wait a sec - this does absolutely count intermittent TPS

rusefillc:
@mck1117 please _rebase_ in light of https://github.com/rusefi/rusefi/commit/405b23d6823fa287dbe0ebe1952cf09479cd5c3b

PR #4880, opened 2022-12-07
Strictly validate stack use in hardware CI

### LINKED ISSUE #4881: stack overflow setting date
Looks like `sscanf` uses lots of stack to do its thing, overflowing the comms stack.

from #4764

mck1117:
There's a stack overflow with setting the date, this should catch it, right?

try to detect  #4881

fix #4881


mck1117:
hmm why did it succeed....

mck1117:
hah, there we go!

image

rusefillc:
Did you guys merge fix and detection into one PR? :(

Issue #4876, opened 2022-12-06
vr-discerete trigger oscilloscope


mck1117:
All the hardware is present to show the user an analog oscilloscope view of the trigger waveform.

Analog input connected to pins on ADC3, then digital hardware produces the real trigger output, but we can use the analog input for debugging.

mck1117:
sneak preview:

![image](https://user-images.githubusercontent.com/568254/206087937-870e5caf-6892-4e47-b12e-3aee2c8d4e27.png)

mck1117:
> I NEEED THIS for my SVX trigger issue debuging.

If you replace max992x with https://github.com/andreika-git/hellen-one/tree/master/modules/vr-discrete, you'll have this! As an added benefit, the discrete VR implementation supports software inversion (never have to swap wires like you do on MAX)

mck1117:
This is perfectly usable now, closing. We can make more issues for specific features.

PR #4885, opened 2022-12-11
Basic trigger scope implementation

mck1117:
Implement basic version of #4876 

Looks like this when I touch a floating pin with my finger:

image


rusefillc:
This is very very cool!!!

Also please CHANGELOG

PR #4873, opened 2022-12-05
Jammed ETB detection

mck1117:
Detect if the integrator is high/low for a long period of time, indicating very bad PID tuning or a mechanically stuck throttle (Did you leave a towel in the intake? Is the throttle broken?)

#1336

PR #4894, opened 2022-12-14
Jammed rebase

rusefillc:


mck1117 (changes requested):
let's #4873 and not complicate things, it needs some testing first

PR #4898, opened 2022-12-15
support Toyota ETCS-i tps/pps

mck1117:
Toyota ETCS-i has similar behavior to the "Ford TPS" mode we already have where one of the sensors maxes out at partial travel, and the other fully linear.

This PR makes the "ford TPS" mode available to the PPS as well, and adds a config field for the "secondary maximum" value for each TPS and PPS.

### DIFF
--- a/firmware/init/sensor/init_tps.cpp
+++ b/firmware/init/sensor/init_tps.cpp
@@ -101,7 +101,7 @@ public:
 	{
 	}
 
-	void init(bool isFordTps, RedundantFordTps* fordTps, const TpsConfig& primary, const TpsConfig& secondary) {
+	void init(bool isFordTps, RedundantFordTps* fordTps, float secondaryMaximum, const TpsConfig& primary, const TpsConfig& secondary) {
 		bool hasFirst = m_pri.init(primary);
 		if (!hasFirst) {
 			// no input if we have no first channel
@@ -125,7 +125,7 @@ public:
 
 		if (isFordTps && fordTps) {
 			// we have a secondary
-			fordTps->configure(5.0f, 52.6f);
+			fordTps->configure(5.0f, secondaryMaximum);
 			fordTps->Register();
 		} else {
 			// not ford TPS
@@ -161,6 +161,7 @@ static RedundantPair tps2(tps2p, tps2s, SensorType::Tps2);
 // Used only in case of weird Ford-style ETB TPS
 static RedundantFordTps fordTps1(SensorType::Tps1, SensorType::Tps1Primary, SensorType::Tps1Secondary);
 static RedundantFordTps fordTps2(SensorType::Tps2, SensorType::Tps2Primary, SensorType::Tps2Secondary);
+static RedundantFordTps fordPps(SensorType::AcceleratorPedal, SensorType::AcceleratorPedalPrimary, SensorType::AcceleratorPedalSecondary);
 
 // Pedal sensors and redundancy
 static FuncSensPair pedalPrimary(1, SensorType::AcceleratorPedalPrimary);
@@ -181,19 +182,32 @@ void initTps() {
 
 	if (!engineConfiguration->consumeObdSensors) {
 		bool isFordTps = engineConfiguration->useFordRedundantTps;
+		bool isFordPps = engineConfiguration->useFordRedundantPps;
 
-		tps1.init(isFordTps, &fordTps1,
+		float tpsSecondaryMaximum = engineConfiguration->tpsSecondaryMaximum;
+		if (tpsSecondaryMaximum < 10 || tpsSecondaryMaximum > 90) {
+			// most ford TPS seem to be ~52.6%, default to that if not set
+			tpsSecondaryMaximum = 52.6f;
+		}
+
+		tps1.init(isFordTps, &fordTps1, tpsSecondaryMaximum,
 			{ engineConfiguration->tps1_1AdcChannel, (float)engineConfiguration->tpsMin, (float)engineConfiguration->tpsMax, min, max },
 			{ engineConfiguration->tps1_2AdcChannel, (float)engineConfiguration->tps1SecondaryMin, (float)engineConfiguration->tps1SecondaryMax, min, max }
 		);
 
-		tps2.init(isFordTps, &fordTps2,
+		tps2.init(isFordTps, &fordTps2, tpsSecondaryMaximum,
 			{ engineConfiguration->tps2_1AdcChannel, (float)engineConfiguration->tps2Min, (float)engineConfiguration->tps2Max, min, max },
 			{ engineConfiguration->tps2_2AdcChannel, (float)engineConfiguration->tps2SecondaryMin, (float)engineConfiguration->tps2SecondaryMax, min, max }
 		);
 
+		float ppsSecondaryMaximum = engineConfiguration->ppsSecondaryMaximum;
+		if (ppsSecondaryMaximum < 10 || ppsSecondaryMaximum > 90) {
+			// most Toyota PPS seem to be ~65%, default to that if not set
+			ppsSecondaryMaximum = 65;
+		}
+
 		// Pedal sensors
-		pedal.init(false, nullptr,
+		pedal.init(isFordPps, &fordPps, ppsSecondaryMaximum,
 			{ engineConfiguration->throttlePedalPositionAdcChannel, engineConfiguration->throttlePedalUpVoltage, engineConfiguration->throttlePedalWOTVoltage, min, max },
 			{ engineConfiguration->throttlePedalPositionSecondAdcChannel, engineConfiguration->throttlePedalSecondaryUpVoltage, engineConfiguration->throttlePedalSecondaryWOTVoltage, min, max }
 		);
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -445,7 +445,7 @@ end_struct
 injector_s injector
 
 bit isForcedInduction;Does the vehicle have a turbo or supercharger?
-bit useFordRedundantTps;On Ford vehicles one of the sensors is not linear on the full range, i.e. in the specific range of the positions we effectively have only one sensor.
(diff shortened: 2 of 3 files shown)


nmschulte:
I'll test this evening; thanks for the quick turn support!

I wonder, is renaming from "ford" to something more generic like "shifted" or "biased" (still noodling on the concept) possible?  Is renaming a config bit/property a worrisome event?

nmschulte:
I didn't get to check in-vehicle as I ran into an issue with rusEFI console; it would fail to connect (and continually retry) w/ a message like "expecting to receive 2 bytes -- received 0 in 5000ms," and when this occurs I'd have to reset the board (or else e.g. TS would not connect either -- upon reset, TS would happily connect, but not otherwise).  I played with a bunch of other tuning/settings in TS (about Base Engine, Trigger, VVT, Ignition, etc.) so perhaps this caused issues; I will have to flash the older working build to diagnose.

mck1117:
> I will have to flash the older working build to diagnose.

I don't think we've touched anything interesting recently that would change comms behavior...

nmschulte:
Perhaps I just need to try a different console build: https://github.com/rusefi/rusefi/issues/4899

nmschulte:
This branch seems to enable Toyota ETCS-i to work.  Some observations:
- ETCS-i requires connecting motor in reverse polarity to that specified in JCE10/Lexus IS300 FSM
  - M- (pin 1) to h-bridge positive, M+ (pin 2) to h-bridge negative
- JCE10/Lexus IS300 FSM indicates stock ECU operates at 500 Hz (unverified by oscilloscope)
  - I'm currently operating at 2 kHz as the noise is preferred here, though all values I tested seem noisier than stock ECU
- `fordRedundantAps` seems necessary, even though a value of `100` seems appropriate
  - **disabling causes accel pedal error at ~15% until about 97.7%**
    - **_log: https://desmas.net/lexus_is300/toyota-etcs_i-no_fordRedundantPps-2022-12-18_11.05.20.mlg_**
  - toggling this requires an ECU reset, producing a critical error and this dialog to appear: image
- `fordRedundantTps` value of `64.5` seems to allow my ETB full-range without TPS error
  - **TPS/APS between ~57% and ~59% cause accel pedal error and ETB to drop out; dwelling in this range for too long causes ETB to stop function, regained by +12V power (or ECU reset) cycle**
    - **_log: https://desmas.net/lexus_is300/toyota-etcs_i-100_fordRedundantPps-2022-12-18_11.05.20.mlg_**
- `etbDutyAverageGauge`, `etbDutyRateOfChangeGauge`, `etbPpsErrorCounterGauge`, and `etbTpsErrorCounterGauge` always display 0, seem ineffective
- I'm driving the clutch motor at constant 50% duty via low-side driver GPPWM
  - I intend to drive this via ETB channel (after it's fixed), though it's unclear how I'll fix the duty cycle with this setup
  - meters (clamp and direct) measure nearly exactly 1 A DC draw (AC does not register w/ true RMS meter)
    - DC supply indicates 0.61 A delta at 14.4 V between low-side dis/connect (0.51 A at 12 V; 0.9 A via clamp DC)
- throttle motor resistance is about the same as clutch motor impedance: ~5 Ω
  - 10% duty (@ 2 kHz) draws ~0.1 A

firmware/init/sensor/init_tps.cpp
```diff
@@ -125,7 +125,7 @@ struct RedundantPair {
 
 		if (isFordTps && fordTps) {
 			// we have a secondary
-			fordTps->configure(5.0f, 52.6f);
+			fordTps->configure(5.0f, secondaryMaximum);
```
nmschulte, line 128:
```suggestion
			fordTps->configure(MAX_TPS_PPS_DISCREPANCY, secondaryMaximum);
```

firmware/init/sensor/init_tps.cpp
```diff
@@ -125,7 +125,7 @@ struct RedundantPair {
...
 			fordTps->Register();
```
nmschulte, line 129: Should `RedundantPair` and/or `FuncSensPair` need a `deinit` or to otherwise [about here] unregister `m_redund`/`fordTps` in the case the controlling setting is toggled during run, to avoid `Duplicate registration for sensor "..."`?

nmschulte:
> - `fordRedundantAps` seems necessary, even though a value of `100` seems appropriate
>   - **disabling causes accel pedal error at ~15% until about 97.7%**   
>     - **_log: https://desmas.net/lexus_is300/toyota-etcs_i-no_fordRedundantPps-2022-12-18_11.05.20.mlg_**
> ...
> - `fordRedundantTps` value of `64.5` seems to allow my ETB full-range without TPS error
>   - **TPS/APS between ~57% and ~59% cause accel pedal error and ETB to drop out; dwelling in this range for too long causes ETB to stop function, regained by +12V power (or ECU reset) cycle**
>     - **_log: https://desmas.net/lexus_is300/toyota-etcs_i-100_fordRedundantPps-2022-12-18_11.05.20.mlg_**

After some investigation, it seems this issue is with PPS.  However, viewing the log and the code I do not understand why; the scale of the graph makes interpreting the primary vs secondary values difficult.
image
_log: https://desmas.net/lexus_is300/2022-12-18_13.17.24.mlg_

nmschulte:
> - `fordRedundantAps` seems necessary, even though a value of `100` seems appropriate
>   - **disabling causes accel pedal error at ~15% until about 97.7%**

Okay I figured this out.  The code defaults to `65` if the value is greater than `90`, and the ETCS-i sensors blow the `MAX_TPS_PPS_DISCREPANCY = 5.0f` check, hence why "`100` seems appropriate;" `65` would effectively scale the sensor to never blow this `delta` check, though the inputs are linear.

So it seems ETCS-i demands ability to relax the delta discrepancy.  This could happen strictly in code, or could be exposed as a tune setting.

firmware/init/sensor/init_tps.cpp
nmschulte: see https://github.com/mck1117/rusefi/pull/4/files for fix

nmschulte:
> I still need to determine the actual max values when driven by the throttle cable vs turning to max by hand; that's likely what's at play here.

I was able to test the ETCS-i installed in the car (same unit/model as tested on the bench, above).

First, the FSM indicates this for monitoring VPA2 (vs VPA1): 
> # Typical Malfunction Thresholds
> ## Detection Criteria
> `|VPA - (VPA2 * 0.8)|`*
> \* _Corrected by learning value_

image

----

With 14.4 V (@ ~1 °C ambient) to Proteus 0.6, these are the values from Tuner Studio:

signal | accelerator closed | accelerator opened fully by hand | accelerator opened fully by pedal
--- | --- | --- | ---
VPA (PPS1) | ~0.525 V | ~3.770 V | ~3.755 V
VPA2 (PPS2) | ~2.150 V | ~4.965 V | ~4.965 V

_VPA2 (PPS2) reaches max at VPA (PPS1) of ~2.750 V._

----

With only 5 V USB:

signal | accelerator closed | accelerator opened fully by hand | accelerator opened fully by pedal
--- | --- | --- | ---
VPA (PPS1) | ~0.500 V | ~3.595 V | ~3.610 V
VPA2 (PPS2) | ~2.065 V | ~4.750 V | ~4.750 V

_VPA2 (PPS2) reaches max at VPA (PPS1) of ~2.625 V._

----

In all situations, VPA2 clips before VPA reaches full range, just as with the bench unit (operated through full range by hand), even though the FSM and other sources indicate this is outside of the usable range.  It's possible (and likely) that the throttle cable on the vehicle has been adjusted out of range, though I could not find anything in the FSM about adjusting this.  The only instructions are like "dis/connect accelerator cable," and in fact sometimes to operate the sensor fully by hand.  This is in contrast with most/all other specifications the FSM provides, including for the TPS and e.g. engine mechanical clearances.

firmware/init/sensor/init_tps.cpp
```diff
@@ -181,19 +182,32 @@ void initTps() {
...
 		);
 
-		tps2.init(isFordTps, &fordTps2,
+		tps2.init(isFordTps, &fordTps2, tpsSecondaryMaximum,
 			{ engineConfiguration->tps2_1AdcChannel, (float)engineConfiguration->tps2Min, (float)engineConfiguration->tps2Max, min, max },
 			{ engineConfiguration->tps2_2AdcChannel, (float)engineConfiguration->tps2SecondaryMin, (float)engineConfiguration->tps2SecondaryMax, min, max }
 		);
 
+		float ppsSecondaryMaximum = engineConfiguration->ppsSecondaryMaximum;
+		if (ppsSecondaryMaximum < 10 || ppsSecondaryMaximum > 90) {
```
nmschulte, line 204: This range-resetting (and limiting) behavior seems unfriendly (and unnecessarily restrictive).
1) it'd be better to clamp the value than reset it without notice
2) it'd be better to just accept the given value (with the possible exception of 0, and _maybe_ 100, defaulting to something "sane").

I'd argue the same about the TPS application, but not knowing the details of the original concern about Ford TPS I can't argue against retaining it (though the unnoticed reset is still unfriendly).

mck1117: The concern about being too lenient is that you can set it to 0%, which will defeat the redundancy on your throttle completely. I've changed it to just not allow small values, but allow 100 (since that's actually safe)

firmware/init/sensor/init_tps.cpp
```diff
@@ -181,19 +189,32 @@ void initTps() {
 
 	if (!engineConfiguration->consumeObdSensors) {
 		bool isFordTps = engineConfiguration->useFordRedundantTps;
```
nmschulte, line 191: Should these use `activeConfiguration` too?

mck1117:
nmschulte was still playing with getting this working right on his throttle, but ok

nmschulte:
I'll put up a separate PR for my efforts; they're "once removed" from what this enables, and this seemed to be working well enough for what was intended.  Thanks!

PR #4904, opened 2022-12-18
organize ETB H-Bridge No1/2 into panels

nmschulte:
image

Is selectively disabling the panels against practice?  Wondering because I guess bits/pins there-in will retain their value if not enabled, which makes unsetting them a hassle.  What's the rusEFI practice around this?


nmschulte:
> What's the rusEFI practice around this?

It seems TPS behaves the same way as this branch does about ETB:
image

PR #4933, opened 2023-01-03
S105: successed TS connection over CAN

dron0gus:


dron0gus:
Screenshot from 2023-01-04 01-26-12

PR #4936, opened 2023-01-04
S105: pinout fixes, defaults

dron0gus:


firmware/config/boards/s105/board_configuration.cpp
```diff
@@ -4,15 +4,83 @@
...
  * @todo    Add your board-specific code, if any.
  */
+
+static void setInjectorPins() {
+	engineConfiguration->injectionPins[0] = Gpio::D8;
+	engineConfiguration->injectionPins[1] = Gpio::D9;
+	engineConfiguration->injectionPins[2] = Gpio::D10;
+	engineConfiguration->injectionPins[3] = Gpio::D11;
+
+	// Disable remainder
```
mck1117, line 14: the remainder should be disabled by default, no need to do this part

rusefillc: @mck1117 thank you https://github.com/rusefi/rusefi/commit/6ef5fef5900e0a07d61eb8a34c4b5dea0bf6335a

PR #4938, opened 2023-01-05
Support non-uniform dividers on ADC inputs

dron0gus:
S105 board has different dividers on CLT and IAT vs MAP and TPS channels.
Add option and board callback `getAnalogInputDividerCoefficient()` to support getting ADC input dividers from board helper instead of `engineConfiguration->analogInputDividerCoefficient`
Should not affect boards with no `EFI_NON_UNIFORM_ANALOG_DIVIDER` option


rusefillc:
Another idea, no arguments for pros and cons:
__weak__ functions getCltDividerOverride getIatDividerOverride etc

firmware/hw_layer/sensors/hip9011_logic.h
```diff
@@ -16,7 +16,7 @@
 #endif
 
 #define HIP9011_BAND(bore)		(900 / (CONST_PI * (bore) / 2))
-#define HIP9011_DESIRED_OUTPUT_VALUE	5.0f
+#define HIP9011_ANALOG_OUTPUT_MAX		5.0f
```
rusefillc, line 19: huh how is this HIP change in this PR? should that be a separate PR?

rusefillc:
can we/shall we split into three PRs with two preparations (hip and map)?

dron0gus:
> Another idea, no arguments for pros and cons: **weak** functions getCltDividerOverride getIatDividerOverride etc

In this case ADC code needs to know what user (CLT, IAT) is trying to get value. This is not good in terms of modular programming.
Also (not in case of S105 with its fixed pin functions) ADC input channel divider is parameter of ADC input, not of a function that is connected to this input.

firmware/hw_layer/sensors/hip9011_logic.h
dron0gus: https://github.com/rusefi/rusefi/pull/4939
https://github.com/rusefi/rusefi/pull/4940

firmware/hw_layer/adc/adc_inputs.h
```diff
@@ -15,6 +15,13 @@
 #define SLOW_ADC_RATE 500
 #endif
 
+#if EFI_NON_UNIFORM_ANALOG_DIVIDER == FALSE
```
mck1117, line 18: I would prefer

```c++
__attribute__((weak)) float getAnalogInputDividerCoefficient(adc_channel_e) {
    return engineConfiguration->analogInputDividerCoefficient;
}
```

This will do the same thing, but requires no defines anywhere, just adding the function in the board's cpp file.

firmware/config/boards/s105/board.mk
```diff
@@ -31,6 +31,9 @@ DDEFS += -DSTM32_USB_USE_OTG2=FALSE
 DDEFS += -DEFI_FILE_LOGGING=FALSE
 USE_FATFS = no
 
+# This board has different dividers on adc inputs
+DDEFS += -DEFI_NON_UNIFORM_ANALOG_DIVIDER=TRUE
```
mck1117, line 35: If we use `__attribute__((weak))`, we don't need any defines to select whether to use it or not.

firmware/hw_layer/adc/adc_inputs.h
dron0gus: Personally I don't like weak definition. As it sometimes cause strange issues that is hard to debug.
But, np, I'll rewrite it to use weak attribute.

mck1117: Hm, I've never had an issue with it like that. And in this case, LTO will inline it anyway when optimization is enabled.

Issue #3227, opened 2021-09-06
Unable to change broken settings


rusefillc:
Nik has FATAL error due to VSS PAL on PA8 conflicting with joystick on something else Px8

problem is we get 

![image](https://user-images.githubusercontent.com/48498823/132144977-8318c6ce-e539-4ff5-9120-1b7e2b4649c0.png)

while trying to change settings to good ones

rusefillc:
```
9/5/21 8:00:07 PM :Info: CrC matched skipped controller read on page:1
9/5/21 8:00:07 PM :Debug: Read All Data
9/5/21 8:00:07 PM :Debug: isBlank Time: 0 ms.
9/5/21 8:00:07 PM :Debug: DiffTime: 4 ms.
9/5/21 8:00:07 PM :Info: Difference Report disabling shared table Models
9/5/21 8:00:07 PM :Debug: Received Instruction: Retrieve Config Error, Page: 0
9/5/21 8:00:14 PM :Info: Time to save msq: 81.3948
9/5/21 8:00:14 PM :Info: dirtyData.size(): 1
9/5/21 8:00:14 PM :Debug: Received Instruction: Write Chunk, Page: 1
9/5/21 8:00:14 PM :Info: dirtyData.size(): 0
9/5/21 8:00:14 PM :Debug: Flush lasted: 213ms.
9/5/21 8:00:14 PM :Debug: Burn Page anonymous: 1
9/5/21 8:00:14 PM :Debug: Received Instruction: , Page: 1
9/5/21 8:00:14 PM :Info: Queueing burn to page:1
9/5/21 8:00:14 PM :Info: Sending, 0 bytes.
9/5/21 8:00:14 PM :Info: Loaded Restore point C:\Users\Dell2019\Documents\TunerStudioProjects\nik sept 5\CurrentTune.msq
	Sending, 0 bytes.
9/5/21 8:00:14 PM :Info: dirtyData.size(): 0
9/5/21 8:00:14 PM :Info: dirtyData.size(): 0
9/5/21 8:00:14 PM :Debug: Flush lasted: 1ms.
9/5/21 8:00:14 PM :Debug: Burn Page anonymous: -1
9/5/21 8:00:14 PM :Debug: Received Instruction: , Page: 1
9/5/21 8:00:14 PM :Info: skip burn to page:1, lastWritePage = -2
9/5/21 8:00:17 PM :Debug: burned page FAILED! 1
[...]

PR #4945, opened 2023-01-06
remove joystick

mck1117:
As far as I can tell, nobody has used joystick in a very long time.

tangential #3227


rusefillc:
Can we please also remove "newparse"? I know it's just a bit of a stretch but technically there are some similarities between dead joystick and unborn newparse

mck1117:
it's a valid discussion but seems irrelevant to this ticket

rusefillc:
I am sorry I have no capacity to be discussing this via typing.

My preference is not to merge this PR at this PR, my preference is not to see it on the list, allow me to close.

PR #4948, opened 2023-01-06
improve iat ignition table

mck1117:
Switch the ignition IAT correction table to use IAT/load instead of IAT/RPM.

This is what the OEMs do - very little RPM dependence, mostly dependent upon load/iat.

image

Also includes more sensible default values in the table.

image


rusefillc:
@mck1117 please CHANGELOG

mck1117:
I was just on my way to do that :)

Issue #4896, opened 2022-12-15
Toyota ETCS-i APS (ETB APP) [and TPS]


nmschulte:
_First, according to the documentation and references, this [APS] situation isn't intended in normal/stock operation, so this issue serves at the moment to simply document the situation: manual actuation on the bench, beyond the designed/intended [APS] ranges.  **However, intended/typical TPS suffers similarly!**  Notes about ETCS-i magnetic clutch included for completion._

Toyota's ETCS-i (ETB) APS (APP) outputs behave similarly to some (e.g. certain Ford; https://github.com/rusefi/rusefi/blob/master/firmware/controllers/sensors/redundant_ford_tps.cpp) TPS outputs: one signal reaches max before the other when actuated across the entire range.  In this [bench] situation, the VPA2 (redundant/secondary APP) signal reaches max (~5V) before VPA (primary APP).  This causes rusEFI to signal an error (again, when using the full mapping/range, on the bench).

Additionally, the ETCS-i's TPS output (VTA1 and VTA2) behaves similarly, though this is a typical situation.  This would be ideal to support via rusEFI.

![image](https://user-images.githubusercontent.com/8540239/207802265-06c93316-4f59-44e2-962d-2d186f1285c7.png)
_(image via https://alflash.com.ua/ETCS.htm)_

I will check soon (<24hr hopefully) the APS output ranges in-vehicle, though the documentation notes the typical/allowable range.  Ideally rusEFI would like to support this situation as well.

----
_APS (from Lexus IS300 FSM, DI-297, DI-299)_
![image](https://user-images.githubusercontent.com/8540239/207802852-12c407a7-631f-497d-b85c-71148ee69bad.png)
![image](https://user-images.githubusercontent.com/8540239/207805255-e53e23ae-b589-46ef-876b-9a2a4acf5b15.png)

_TPS (from Lexus IS300 FSM, DI-96, DI-99)_
![image](https://user-images.githubusercontent.com/8540239/207803056-f681875e-e120-4c8e-8879-cc6f2d25f844.png)
![image](https://user-images.githubusercontent.com/8540239/207805009-4bb80a6e-1f55-4dfc-b1a4-0917c139256b.png)

----

Finally, relatedly, ETCS-i has also a magnetic clutch [dis]engaging the control motor, operated via the ECM.  The ECM pulses the circuit to avoid excess power consumption, affording ~0.9 A (from main/12V it seems).

[...]

PR #4964, opened 2023-01-12
TPS, "monitored" RedundantSensor [2/2]

nmschulte:
See first [1/2]: #4966.

- tidies/clarifies some `init_tps` logic/code
- adds Broken / Experimental `*SecondaryMonitors` config bits to configure the redundancy to purely monitor (second/secondary sensor checks the first/primary) and resolve directly the primary sensor

The "monitored" feature allows to avoid the step-wise transition at the partial redundancy switch-point, which is difficult/impossible to remove even in static/bench conditions (and likely to vary in dynamic/real-world conditions).  Ex. Toyota ETCS-i:

Averaging (`tpsSecondaryMaximum`: `64.0`% [60.8% threshold @ 4.81325 V]; stepwise highlighted):
image

Monitoring (`tpsSecondaryMaximum`: `64.0`% [60.8% threshold @ 4.81325 V]; stepwise eliminated):
image

The stepwise transition can be minimized by choosing a[n unachievable] "perfect" `tpsSecondaryMaximum`, but there will always be a small step due to the 0.5% resolution of this field and that the sensors can/will vary and drift.  More precision can be had by tuning the `tps#SecondaryMax` config in coordination, but this makes the auto-tune features useless, and isn't ideal.


mck1117:
+317 -251 is too big for one PR, can we break this down a bit?

nmschulte:
> +317 -251 is too big for one PR, can we break this down a bit?

I removed the melding of `FordRedundantSensor` and related to #4966.

firmware/controllers/sensors/redundant_sensor.cpp
```diff
@@ -9,33 +9,63 @@ RedundantSensor::RedundantSensor(SensorType outputType, SensorType first, Sensor
...
-		return UnexpectedCode::Inconsistent;
+	float sensor2Value = sensor2.Value;
+
+	// Partial redundancy; useful for some sensors: e.g. Ford and Toyota ETCS-i
+	if (m_secondMaximum != 100) {
+		// scale the second sensor value accordingly, proportioning to the first sensor
+		sensor2Value *= m_secondMaximum / 100;
+
+		// The partial redundancy threshold, slightly less than 100% to avoid issues near full-range
+		float threshold = m_secondMaximum * 0.95;
```
nmschulte, line 49: Using `m_maxDifference` to proportion this was dumb (as in first PR), hence the change back to a fixed percentage.  I don't know why -10% was chosen initially, but obviously I like -5% better...

firmware/controllers/sensors/redundant_sensor.cpp
```diff
@@ -9,33 +9,63 @@ RedundantSensor::RedundantSensor(SensorType outputType, SensorType first, Sensor
...
+
+		// Check second sensor is below partial redundancy switch-over threshold
+		if (sensor2Value > threshold) {
+			// Check first sensor is at or above partial redundancy switch-over threshold
+			if (sensor1.Value > threshold - m_maxDifference) {
+				return sensor1.Value;
```
nmschulte, line 55: This could instead `sensor2Value = sensor1.Value;` and fall-through to the averaging case if for some reason that was preferred...

firmware/controllers/sensors/sensor_info_printing.cpp
```diff
@@ -36,24 +35,28 @@ void CanSensorBase::showInfo(const char* sensorName) const {
 #endif // EFI_CAN_SUPPORT
 
 void RedundantSensor::showInfo(const char* sensorName) const {
-	efiPrintf("Sensor \"%s\" is redundant combining \"%s\" and \"%s\"", sensorName, getSensorName(m_first), getSensorName(m_second));
+	// TODO(nms): print about partial redundancy
```
nmschulte, line 38:
```suggestion
```

firmware/init/sensor/init_tps.cpp
```diff
@@ -13,7 +12,9 @@
 #define MAX_TPS_PPS_DISCREPANCY 5.0f
 #endif
 
-struct TpsConfig {
+#define TPS_ADC_UNITS_PER_VOLT 200
```
nmschulte, line 15: Maybe this should live in a higher scope, for other (existing?) use?

nmschulte:
I couldn't target my own repo/branch (the first one), but this PR is based upon the first, will clear up if that merges; otherwise you'll have to view the diff locally or so.

PR #4977, opened 2023-01-17
Update yq, fix bug in pinout generation

chuckwagoncomputing:


rusefillc:
@chuckwagoncomputing CI failure was legit, reverting now

PR #4982, opened 2023-01-20
fix quoting

mi-hol:


rusefillc:
@chuckwagoncomputing do we need such quotes?

@mi-hol are there any other yaml files with quotes?

How does https://rusefi.com/docs/pinouts/Benelli_Walbro/ look?

chuckwagoncomputing:
Quotes aren't required in YAML unless you have special characters in your string, which seems to be the case with at least some of these.

mi-hol:
> How does https://rusefi.com/docs/pinouts/Benelli_Walbro/ look?

@rusefillc 
Looking good, except for C3 & C4 (which I copied from AlphaX-4chan and adapted, seems to be a bug there but "covered" by individual pins https://rusefi.com/docs/pinouts/hellen/alphax-4chan/)
image

  - pin: C3C4
    function: 'Crank Sensor - Phonic Wheel (VR) NOTE!! is + and - side assignment required??'
    class: event_inputs
    type: vr
    color: white/black

PR #4983, opened 2023-01-20
fix combined pin C3C4

mi-hol:


rusefillc:
@mi-hol yes VR cares about polarity 
@mck1117 just to confirm

mi-hol:
@rusefillc thanks for " yes VR cares about polarity", will fix this at a later stage (when I have access to the bike)
Please merge the current version

PR #4989, opened 2023-01-21
LTS - one way of doing things

chuckwagoncomputing:
This expects/requires a 'lts' subdirectory of build-server.
How to use:
1. Go to the '[Firmware at GHA](https://github.com/rusefi/rusefi/actions/workflows/build-firmware.yaml)' action
2. Click "Run Workflow"
3. Select your LTS branch (assuming you have a special branch; if not, master works too)
4. Check "LTS Build" toggle
5. Click green "Run Workflow" button
Screenshot_20230121_113257

If this workflow isn't to your liking, we can probably find a different way. This just seemed simplest to me.

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -374,12 +374,12 @@ jobs:
 
     # Build rusEFI console
     - name: Build console
-      if: ${{( github.event_name == 'push' && github.ref == 'refs/heads/master' && env.skip != 'true' ) || inputs.lts == 'true' }}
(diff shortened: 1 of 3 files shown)


mck1117:
can we check if the branch starts with `lts` instead? that way, all you'd have to do would be `git checkout -b lts-2301 && git push` and we'd have an LTS

chuckwagoncomputing:
That could be done. I didn't do it that way because I imagined a scenario in which we realize that another change needs to be made after initially pushing the branch. I'm not exactly sure how bad that would be. Worst case scenario is probably that someone would download a broken LTS build from the server.

Issue #3566, opened 2021-11-17
specific PN F7 boards bricking on incremental flash due to ST bug


rusefillc:
on some/many F7 units incremental DFU flash via command line stm32 causes bricking

only full erase and flashing of same helps

rusefillc:
https://www.rusefi.com/forum/viewtopic.php?p=43239#p43239 has a failing dump

one day someone should try bricking by flashing same exact firmware again, maybe after changing settings?

i.e.
step 1) erase and flash of release
step 2) change settings
step 3) incremental flash of same exact release

rusefillc:
https://github.com/rusefi/rusefi/wiki/F7-requires-full-erase

NMSTEC:
Good news, I might as well as I have an STLink already connected.

NMSTEC:
Having issues replicating this with ST-Link only. Should I try bricking it via console? Boots everytime normally. No file corruption.
![image](https://user-images.githubusercontent.com/45252464/143719751-6fa5e456-0a22-4274-b68a-dfe7f989f174.png)

NMSTEC:
![image](https://user-images.githubusercontent.com/45252464/143719775-4b653e88-e998-4241-892c-81782b0fa259.png)

NMSTEC:
Attempted to change plenty of settings that require reboot on ECU. Then reconnect, confirm ok. shutdown TS, AutoUpdate console, works fine.
No flash differences detected. Size 0x8399C

Recommend checking if OB options are set to dual bank.

![image](https://user-images.githubusercontent.com/45252464/143720026-c409e205-8f58-48e7-b587-c8da903633e2.png)

NMSTEC:
Tried single banking it. Still works fine. TS also detects that, thats good.

mck1117:
@NMSTEC saw

![image](https://user-images.githubusercontent.com/568254/143927612-f4258e9d-4193-4038-a543-0a8bc095ace6.png)

mck1117:
0x0808 0008 is just after 0x0808 0000, which is the start of sector 12, the start of the second flash bank.

![image](https://user-images.githubusercontent.com/568254/143927757-aea75f35-a666-4894-924b-c009ec522d48.png)

I wonder if this is a bug in STM32CubeProgrammer where it has trouble doing a contiguous erase+write across the bank boundary.

mck1117:
[...]

PR #4996, opened 2023-01-22
build f7 with -Os, fits in 512k

mck1117:
go back to 512k for #3566 since stm programmer is still not fixed

Os vs O2 performance hit is very small, F7 is plenty fast, and it knocks ~53k off the binary. Probably mostly due to loop unrolling not happening like it does at O2.


rusefillc:
Allow me not to merge until I am back in US in a week, want to double-check which workload works which one does not

PR #5004, opened 2023-01-24
Let's admit it's time to move to gradle #3014

rusefillc:
running HW CI explicitly without ant junit runner as a step on our ant to gradle migration

expected: this change makes zero functional difference
actual: HW CI does not like this change?

### DIFF
--- a/.github/workflows/hardware-ci.yaml
+++ b/.github/workflows/hardware-ci.yaml
@@ -78,5 +78,5 @@ jobs:
 
     # This both compiles and runs HW CI tests
     - name: Run Hardware CI
-      run: java -cp java_console/autotest/build/libs/autotest-all.jar com.rusefi.HwCiF4Proteus ${{matrix.test-suite}}
+      run: java -cp java_console/autotest/build/libs/autotest-all.jar ${{matrix.test-suite}}
 
--- a/.github/workflows/hardware-ci.yaml
+++ b/.github/workflows/hardware-ci.yaml
@@ -12,7 +12,6 @@ jobs:
 
         include:
           - build-target: f407-discovery
-            script: hardware_ci_f4_discovery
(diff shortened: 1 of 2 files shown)


rusefillc:
@nmschulte @chuckwagoncomputing @mck1117 any chance you guys could help take this further?

rusefillc:
@chuckwagoncomputing any chance you could help with this one? I guess part of the context is if you would be interested to java more :) I know you have either discovery or Proteus or maybe even both, just one HW piece would probably help troubleshoot.

chuckwagoncomputing:
Is it possible that HARDWARE_CI is unset when built with gradle?
```
#ifdef HARDWARE_CI
        case PROTEUS_ANALOG_PWM_TEST:
                setProteusAnalogPwmTest();
                break;
#endif // HARDWARE_CI
```

mck1117:
@chuckwagoncomputing gradle doesn't build the firmware, just the java, afaik

chuckwagoncomputing:
I'm pretty stumped about what's going on. Hopefully my findings will give someone an idea.

On both Proteus and Discovery, there is output that appears to be looping infinitely.
30-45 seconds after the looping starts, test summary is printed, as though the tests are complete, but then the looping continues.

Looping output on Disco starts after setting the RPM to 6000 in testVW, which runs after testV12 when built with Gradle. (With Ant, they are run in the same order as they are in the file.)

Looping output on Proteus starts after the first `setIdlePositionAndAssertTps` in `testTpsAnalogInput`, and the other tests continue to run - the looping output isn't blocking.

rusefillc:
Wow, interesting :(

rusefillc:
@chuckwagoncomputing this turned out to be about java daemon threads
ant seems to be doing a more explicit System.exit(0) in case of success, we now do same from our own junit invoker

PR #5005, opened 2023-01-24
Don't download ARM Gnu toolchain if present

chuckwagoncomputing:
This enables self-hosted runners to pre-download the arm-none-eabi toolchain at build time.
The script checks the checksum of one of the executables, so that we are sure to be running the correct toolchain version.

### DIFF
--- a/firmware/provide_gcc.sh
+++ b/firmware/provide_gcc.sh
@@ -6,7 +6,7 @@ set -e
 
 # URL to download original toolchain from
 URL="https://github.com/rusefi/build_support/raw/master/rusefi-arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi.tar.xz"
-# This is the md5sum of the /bin/ld executable within the archive, used for verifying we have the proper version.
+# This is the md5sum of the /bin/arm-none-eabi-ld executable within the archive, used for verifying we have the proper version.
 # If you change the above URL, you will need to update this checksum as well.
 MANIFEST_SUM="8e50ee1adb41acfd56fc38d74d6bb18e"
 # colloquial directory name, to afford re-use of script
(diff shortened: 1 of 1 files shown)


mck1117:
> This enables self-hosted runners to pre-download the arm-none-eabi toolchain at build time.

what's the actual observable change/purpose?

chuckwagoncomputing:
Currently (pre-this-pr), the firmware workflow always downloads the toolchain, deleting it if it already exists.
This PR makes it so that if the toolchain already exists, it isn't downloaded again.

mck1117:
I must be missing some piece, is there a way to have it pre-placed on a GHA runner?

chuckwagoncomputing:
It's possible to do so by providing a Docker container, but that's beside the point of this PR.

Right now, this is just a time-save for self hosted runners. The current Dockerfile of https://github.com/ZHoob2004/rusefi-ci/ downloads the toolchain, but doing so is useless as it just gets deleted and re-downloaded on every job.

mck1117 (approved):

chuckwagoncomputing:
If we run our own Docker container on github-hosted runners, we won't need to run this script in the workflow.
We currently run this script in the workflow because we don't have arm-none-eabi present on github-hosted runners.
We have the power to have arm-none-eabi present on self-hosted runners.
Currently, this does us no good because the workflow just deletes it and re-downloads it.
Merging this PR will correct this, saving time for self-hosted runners.

nmschulte:
Do the github-hosted runners get re-used or something; is that why this is a win w/o self-hosted?

chuckwagoncomputing:
They don't unfortunately. This is a win because I can self-host my own runners and reap the advantages even if there aren't any on the main repository.

PR #5007, opened 2023-01-27
dual ADC wideband support

opelpanfan:
Initial work on dual ADC wideband support under single calibration (same type controllers)

Have Audi 4.2 v8 and MB m113 NA engines with each bank exhausts merging very far so no option to run single wideband support. I know FW can support dual CAN wideband so would be good to add support for dual ADC wideband based on single calibration set. My core8 ECU has built-in dual 14point7 controllers so they can share the same calibration ranges. 

Done just initial work:
1. TS rusefi.input added EGO 2 panel
2. rusefi_config.txt added hwChannel2 to store input pin
3. updated getAfr in ego.cpp/.h to provide sensor as a parameter to read ADC from 
4. used sensor to read volts from hwChannel or hwChannel2 based on provided sensor type.
5. init_lambda.cpp updated getLambda wrapper to use Lambda1 senor also added getLambda2 wrapper to do same for Lambda2 sensor.
image

So now is a question of how and where to go from here. 

I can see FunctionPointerSensor lambdaSensor is getting lambda reading from sensor 1. so I think I need to replicate this for sensor 2. But I'm stuck on where and how to read it per bank now.

If you think that is a useful addon could you please point me in the right direction?

UPDATED:::::::

I think it works at least on the bench reads correctly the second sensor.

image

Thank you.

### DIFF
--- a/firmware/controllers/sensors/ego.cpp
+++ b/firmware/controllers/sensors/ego.cpp
@@ -106,7 +106,7 @@ bool hasAfrSensor() {
 
 extern float InnovateLC2AFR;
 
-float getAfr() {
+float getAfr(SensorType type) {
 #if EFI_AUX_SERIAL
 	if (engineConfiguration->enableInnovateLC2)
 		return InnovateLC2AFR;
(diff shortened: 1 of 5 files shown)


mck1117:
Looks good to me! Interested in tackling the remaining AFR piece of #2667? It would entail switching the AFR sensors to use `LinearFunc`/`FunctionalSensor` like is done in `init_fluid_pressure.cpp`.

But this PR looks great for now!

mck1117 (approved):

opelpanfan:
Yes, I'll take look at #2667.

Issue #4993, opened 2023-01-22
LTS: bundle autoupdate to use proper branch source


rusefillc:
We have java autoupdate logic, it needs to embrace branches

chuckwagoncomputing:
Could you give an example?

rusefillc:
Download bundle from build server lts folder 
Wait a week
Hit "autoupdate"

Expected: nothing
Actual result: I assume it would override content with master stuff?

chuckwagoncomputing:
Thanks, I understand now. It seems to me that #4992 is a prerequisite to this.

rusefillc:
Yes absolutely!

chuckwagoncomputing:
Could we safely assume that if the branch name (stored in the file name or otherwise) is not master, then we should find the files at `rusefi.com/build_server/lts/<branch name>`?

rusefillc:
yes, since we own the convention?

PR #5008, opened 2023-01-28
[LTS] Put branch in bundle name (please read caveats)

### LINKED ISSUE #4992: LTS: folder inside bundle and self identifition
At the moment all .zip bundles have files within "snapshot_timestamp" folder. While we need a way to say what bundle we look at, current approach is broken

Timestamp in folder makes no sense in ligh of bundle autoupdate, looks like we shall remove timestamp

For lts branches shall we use branch name as folder name?

Open question if we need some metainfo.txt file with original bundle timestamp, and maybe separate entry somewhere with latest auto update timestamp.

chuckwagoncomputing:
This will fix #4992 and #4993.
- Removed date from bundle name
- I chose the approach of using the name of the folder rather than putting the info in bundle_name.txt
  I chose this because I thought it would be simpler, but the exception to that is switch_bundle.sh, which got a little more complicated - see Caveats below. I'm not set on doing things this way.
- The bundle names are `rusefi.<branch name>.<bundle>`
  If the branch is master, then the `<branch name>` is "snapshot".
  Examples:
  `rusefi.snapshot.proteus_f7`
  `rusefi.lts_foo.mre_f4`
- I chose to use `.` instead of `_` because `_` is often used in board names which makes parsing difficult.
- I chose to move `rusefi` to the beginning of the bundle name to make them easier for the end user to find. By this I mean that they'll be together in an alphabetically-sorted directory listing/file explorer window.

Caveats:
- I haven't tested the autoupdater, because I'm too lazy to set up a webserver etc.
- I don't know Java, so I don't know if my "string templating" method is the correct one. See "BASE_URL_LTS" in the diff.
- Likewise, I don't know if "FileSystemNotFoundException" is the correct one to catch in BundeUtil.java
- switch_bundle.sh moves the directory to the new name, and has to launch a subshell to put the user in the correct (new) directory. I'm not sure what the use case of this script is, so I don't know if that could be a problem. As far as I can think, the only things the user would notice would be not having history from the original shell, and having to exit twice to close the shell.
- Because I use the `.` for parsing, we can't have `.` in branch names. E.g. `lts-2023.01` will break things. Maybe we could just use a space instead?

Questions:
- Using the name of the directory gives us the power to switch branches as well - e.g. switch from one LTS branch to another. Would you like me to write a script to do that?

### DIFF
--- a/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
+++ b/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
@@ -82,7 +82,7 @@ public class Autoupdate {
 
     private static void handleBundle(String bundleFullName, UpdateMode mode, String baseUrl) {
         try {
-            String boardName = bundleFullName.split(".")[2]
+            String boardName = bundleFullName.split(".")[2];
             String zipFileName = "rusefi_bundle_" + boardName + "_autoupdate" + ".zip";
             ConnectionAndMeta connectionAndMeta = new ConnectionAndMeta(zipFileName).invoke(baseUrl);
             System.out.println("Remote file " + zipFileName);
--- a/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
+++ b/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
@@ -40,7 +40,12 @@ public class Autoupdate {
                 System.out.println("Snapshot requested");
                 if (bundleFullName != null) {
                     System.out.println("Handling " + bundleFullName);
-                    handleBundle(bundleFullName, mode, ConnectionAndMeta.BASE_URL_LATEST);
+                    String branchName = bundleFullName.split(".")[1];
+                    if ( branchName == "snapshot" ) {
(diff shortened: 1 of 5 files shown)


rusefillc:
My first thought is to split pr into two, let me try cherry picking

chuckwagoncomputing:
The last 4 commits are pertaining to autoupdater and switch_bundle.sh, all those before are for changing the bundle name. (Up to "encapsulate ini file" inclusive)

rusefillc:
https://github.com/rusefi/rusefi/pull/5009

rusefillc:
@chuckwagoncomputing can you please rebase? want to see second hand diff

rusefillc:
The dot in branch name I am happy to use underscore

But as a human I will be forgetting, need CI to fail quickly and obviously. If it does not fail on CI with dots @chuckwagoncomputing please make it fail CI with some Java exception referencing relevant Java constant?

chuckwagoncomputing:
> @chuckwagoncomputing please make it fail CI with some Java exception referencing relevant Java constant?
I'm failing to understand. In my mind, the build-firmware workflow should fail so that packages with dots aren't uploaded.
There's no Java involved there.
Have I misunderstood?

rusefillc:
Well, I guess any code could fail as long as comments of message point of reason behind failing.

I guess passing branch name to Java would not be very natural

PR #5013, opened 2023-01-29
Lean submodule fetching (don't merge until all green)

### LINKED ISSUE #3494: GHA: submodules: recursive
Do we really need ``submodules: recursive`` or can we save the planet by not?

image

chuckwagoncomputing:
I haven't tested HW-CI, so we for sure need to wait for that.

This will fix #3494. I removed all `submodules: recursive` statements and figured out, largely by trial and error, which submodules were required for which workflows.

I also removed some `fetch-depth: 0` statements. Somewhat counterintuitively, this means that the whole history is checked. The default is to only fetch HEAD.

I also haven't tested:
- gen-ibom (currently broken green)
- gen-docs (only removed fetch-depth)
- set-date (only removed fetch-depth)

### DIFF
--- a/.github/workflows/build-android.yaml
+++ b/.github/workflows/build-android.yaml
@@ -12,6 +12,7 @@ jobs:
       - name: Checkout Submodules
         run: |
           git submodule update --init --depth=1 unit_tests/googletest
+          git submodule update --init --depth=1 firmware/libfirmware
 
       - uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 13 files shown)


mck1117:
> I haven't tested HW-CI, so we for sure need to wait for that.

HW CI actually doesn't need to do a lean fetch, since it doesn't wipe the checkout between runs like the GH runners do. A full fetch is just as fast, because it just hits each remote and finds no new deltas to download.

PR #5014, opened 2023-01-29
Make autoupdate work

### LINKED ISSUE #4993: LTS: bundle autoupdate to use proper branch source

chuckwagoncomputing:
Tested and seems to be working, though I didn't actually flash a board. This should fix #4993.

### DIFF
--- a/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
+++ b/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
@@ -40,8 +40,8 @@ public class Autoupdate {
                 System.out.println("Snapshot requested");
                 if (bundleFullName != null) {
                     System.out.println("Handling " + bundleFullName);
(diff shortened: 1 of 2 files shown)


java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
```diff
@@ -40,8 +40,8 @@ public static void main(String[] args) {
                 System.out.println("Snapshot requested");
                 if (bundleFullName != null) {
                     System.out.println("Handling " + bundleFullName);
-                    String branchName = bundleFullName.split(".")[1];
-                    if ( branchName == "snapshot" ) {
+                    String branchName = bundleFullName.split("\\.")[1];
+                    if ( branchName.equals("snapshot") ) {
```
rusefillc, line 44: https://en.wikipedia.org/wiki/JUnit :)

PR #5017, opened 2023-01-31
apt-spy is doing more harm than good

chuckwagoncomputing:
apt-spy keeps picking repo mirrors that aren't reliable, causing workflows to fail.

### DIFF
--- a/.github/workflows/build-android.yaml
+++ b/.github/workflows/build-android.yaml
@@ -24,7 +24,6 @@ jobs:
 
       - name: Install Tools
         run: |
-          sudo gem install apt-spy2 && sudo time apt-spy2 fix --commit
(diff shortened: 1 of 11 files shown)


nmschulte:
I was worried about this; thanks for fixing.

nmschulte:
ref: https://github.com/rusefi/rusefi/pull/4967

chuckwagoncomputing:
It would be lovely if you could provide apt with a list of mirrors, and it would automatically use the next in the list if it had a failure with one. I googled around and couldn't find any way to do that.

nmschulte:
I think apt will automatically fallback if a mirror fails and multiple are listed:
https://askubuntu.com/a/587529

Issue #5016, opened 2023-01-30
LTS: all vs stm32f407 mess


rusefillc:
We have technical debt/historical mess around "default" bundle

* bundle subfolder is rusefi.snapshot.stm32f407
* signature says "all" ``#define TS_SIGNATURE "rusEFI master.2023.01.30.all.2319868418"``
* java has custom code

```
        RusEfiSignature s = SignatureHelper.parse(signature);
        String bundleName = BundleUtil.readBundleFullName();
        if (bundleName != null && s != null) {
            String signatureWithPrefix;
            if ("all".equals(s.getBundle())) {
                signatureWithPrefix = PREFIX;
            } else {
                signatureWithPrefix = PREFIX + "_" + s.getBundle();
            }
```

Does discovery auto-update work at the moment? Shall we reduce the amount of technical debt?

mck1117:
maybe it's best for `all` to go away, since it isn't actually `all`

chuckwagoncomputing:
yep, I'm working on this. I'm aggressively removing stuff and finding out what happens.
https://github.com/rusefi/rusefi/compare/master...chuckwagoncomputing:rusefi:yeet?expand=1

Issue #3367, opened 2021-10-19
partial pinout workflow touches default generated files


rusefillc:
very unexpected diff coming from GHA https://github.com/rusefi/rusefi/commit/a1ed2018779e483f249cf4c72fed2711b03fb9e0

looks like depending on sequence or SOMETHING we get random bundle info in "default" bundle?

![image](https://user-images.githubusercontent.com/48498823/137831547-fa724235-b537-4762-bf56-22c7224cef0a.png)

rusefillc:
@chuckwagoncomputing looks like this is about partial pinout push

whatever pinout file is the only single one to be update overrides generated files for "main" bundle

this is probably related to the fact that we use same path and at least some work flow depends on "default" bundle being handles _last_

in case of partial incremental generation that last default is skipped leaving whatever pinout was touched in default generated files

chuckwagoncomputing:
If I'm understanding correctly, the easy way to fix this would be to reset all non-board-specific files before running the generation for default. That way if the default hasn't changed, it'll leave the files unchanged.
This solution seems kind of ugly to me, but I'm not sure how to do it better.
Maybe we could remove all non-board-specific file paths from gen_config_common.sh and gen_config_board.sh, and give gen_config_default.sh it's own java call instead of just calling gen_config_board.sh in it, but then we introduce more code duplication, which is something we're trying to reduce in the build scripts.

Issue #3182, opened 2021-08-20
Reduce scripts code duplication between matrix and primary build


rusefillc:
build-firmware.yaml has matrix for everything and separate build-primary-bundle shall we reduce duplication/reduce custom code/ make this more universal?

mck1117:
does primary bundle need to exist at all?

Issue #3085, opened 2021-07-31
build-firmware.yaml duplication build-primary-bundle


rusefillc:
gen_config_default.sh is no longer a copy/paste of gen_config_board.sh

next step would be reducing code duplication within build-firmware.yaml see build-primary-bundle

See also #2451

rusefillc:
Dup of #3182?

PR #5018, opened 2023-01-31
Replace "primary" and "default" builds with "f407-discovery"

chuckwagoncomputing:
It has been asked whether the primary build serves any purpose, and there hasn't been any answer. Therefore we shall yeet with extreme prejudice.

- Remove "rusefi_bundle"
- Remove "rusefi_bundle_default"
- Introduce "rusefi_bundle_f407-discovery"
- Reset one more file after board-specific gen-config
- Get rid of a gen_config_default call of unknown purpose (maybe for #3367)
- Don't check for "all"
- Don't run gen_config_default in gen_config

This should fix:
#5016
#3182
#3085
#3367


rusefillc:
let's break this PR apart

for first change can we go only incompatible rename of rusefi_bundle.zip which would break some old links but we should not care about those old links, that rename of "default" bundle should have CHANGELOG

.github/workflows/build-firmware.yaml
```diff
@@ -335,13 +339,6 @@ jobs:
       working-directory: ./firmware/
       run: ./gen_enum_to_string.sh
 
-      # depends on gen_live_documentation.sh for instance TS_TOTAL_OUTPUT_SIZE
-    - name: Generate Default config
-      if: ${{ env.skip != 'true' }}
```
rusefillc, line 340: i have a feeling that this would break something major, let' start breaking the change apart

rusefillc (changes requested):

misc/jenkins/functional_test_and_build_bundle/run.sh
```diff
@@ -1,8 +1,6 @@
 #!/bin/bash
 
 git submodule update --init
-bash misc/jenkins/functional_test_and_build_bundle/build_current_bundle.sh
```
rusefillc, line 4: huh? just like that - just do not build for HW CI? let's extract into separate PR?

chuckwagoncomputing: HW CI doesn't run that script; I don't know what it's for.

rusefillc: Not trying to attach just want to confirm: "i am not sure what this script does but I have confidence to make interesting changes"?

rusefillc: I am fine doing https://github.com/rusefi/rusefi/commit/d3c12dc6f53c77918de227bb7cb997ce2e4faabe and https://github.com/rusefi/rusefi/commit/33a26d05f7e209f4894557fa16dc9225aae04d18

chuckwagoncomputing:
Cherry pick as you please

rusefillc:
@chuckwagoncomputing need your help with all this :( we do need to reduce the madness we just need to do it in smaller chunks

chuckwagoncomputing:
OK I'll see what I can come up with.

.github/workflows/build-firmware.yaml
mck1117: there's always the scream test of "do it and see what screams"

mck1117 (approved):

java_console/io/src/main/java/com/rusefi/io/DfuHelper.java
```diff
@@ -34,15 +34,8 @@ public static boolean sendDfuRebootCommand(JComponent parent, String signature,
...
-/*
-todo: fix https://github.com/rusefi/rusefi/issues/5016
             String signatureWithPrefix;
-            if ("all".equals(s.getBundleTarget())) {
-                signatureWithPrefix = PREFIX;
-            } else {
-                signatureWithPrefix = PREFIX + "_" + s.getBundleTarget();
-            }
-*/
+            signatureWithPrefix = PREFIX + "_" + controllerSignature.getBundleTarget();
```
rusefillc, line 38: @chuckwagoncomputing are you adding an unused variable here?

rusefillc (changes requested):

.github/workflows/build-firmware.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
         # see also build-primary-bundle where default/primary bundle is build separately
         #
         build-target: [
-            all,
+            f407-discovery,
```
rusefillc: would this add second f407-discovery?

java_console/io/src/main/java/com/rusefi/io/DfuHelper.java
chuckwagoncomputing: I guess so. I just preserved it without checking to see if it was ever used.

.github/workflows/build-firmware.yaml
chuckwagoncomputing: That change isn't there anymore, I'm not sure when this review comes from.

rusefillc: i am reviewing commits one after another. if commit history is misleading can we please squash?

chuckwagoncomputing:
What's your current opinion of this?

PR #5019, opened 2023-01-31
Move docs to rusefi_documentation

chuckwagoncomputing:
I also removed SoftwareDevelopmentProcess.md which was very outdated. If it needs to exist, it needs to be rewritten.


chuckwagoncomputing:
https://github.com/rusefi/rusefi_documentation/pull/404

rusefillc:
i see three changes here
* map images total nonse
* weird images which seem not totally insane. those have no value by itself, those only have value if there are links - and we are clearly breaking links. I would guess this is source forge old stuff probably without links - that means no reason to move?
* the .md files: we can debate if those have any value. if those have any value again half of the value is in links and discoverability - by moving content from not very visible "misc" folder to completely invisible www location we effectively delete. is there enough hard to justify removal of md?

overall I would appreciate some context on motivation.

taken to extreme do we want to move code comments into wiki? probably not. we probably agree that some documentation is best right embedded into source code?

mi-hol:
> overall I would appreciate some context on motivation.

and this is a communications topic, we lack a regular forum to bring up change topics in advance

mi-hol:
> * is there enough hard to justify removal of md?

seems a misunderstanding [fix links and typos](https://github.com/rusefi/rusefi_documentation/commit/af696c5170f1c480885506990c9465fe534022c2) adds them to doc repo

chuckwagoncomputing:
Closing these because I don't feel like swimming upstream today.

PR #5020, opened 2023-02-01
Build for F407 Discovery

chuckwagoncomputing:


rusefillc:
This would add discovery alongside "all" right?

Would it be better to modify current custom all/default to use future discovery file name/look more consistent as discovery /be more consistent as discovery? i.e. a smaller change of current artifact instead of new artifact and assumption of old artifact removal later?

chuckwagoncomputing:
Doesn't make sense to me, but I can do it if you want.

mck1117:
> Doesn't make sense to me

yeah, that seems ripe for more issues of "oops we forgot to move that part", much cleaner to rip the bandand off and see what breaks.

PR #5022, opened 2023-02-01
pass board dir to scripts instead of board name

### LINKED ISSUE #5033: build-firmware (stm32f767_nucleo) build is broken
```
make: *** No rule to make target 'board_configuration.cpp', needed by 'build/obj/board_configuration.o'.  Stop.
```

image

mck1117:
Allows for building boards that may not necessarily be in the `config/boards` directory. For example if you want to try rusEFI with some board but not merge yet, you can make a repo, submodule rusEFI, then put your board config files in your repo, then build without needing to modify rusEFI.

- Simplifies build scripts
- Simplifies java (java doesn't need hardcoded `config/boards` any more)
- Simplifies makefiles (same idea, future change to avoid duplicate lines in many `board.h` files)

I guess related to #5016 

also fix #5033 the right way


mck1117:
let's let #5018 go first

rusefillc:
@mck1117 we now have #5018

mck1117:
needs #5035

PR #5024, opened 2023-02-01
deduplicate $(BOARD_DIR) in makefiles

mck1117:
calculate $(BOARDS_DIR) and $(BOARD_DIR) exactly once, instead of many times, or repeating a board's name over and over again in its board.mk.

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -164,6 +164,7 @@ MAKEFLAGS += ${NUMJOBS}
 # Project, sources and paths
 #
 
+BOARDINC = $(BOARD_DIR)
 include $(BOARD_DIR)/board.mk
 
 # Include various ChibiOS mk files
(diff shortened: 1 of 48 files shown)


firmware/config/boards/common_make.sh
```diff
@@ -17,7 +13,7 @@ mkdir -p .dep
 make -j$(nproc) -r PROJECT_BOARD=$PROJECT_BOARD PROJECT_CPU=$PROJECT_CPU
 [ -e build/rusefi.hex ] || { echo "FAILED to compile by $SCRIPT_NAME with $PROJECT_BOARD $DEBUG_LEVEL_OPT and $EXTRA_PARAMS"; exit 1; }
 if [ "$USE_OPENBLT" = "yes" ]; then
-  make openblt
+  make openblt PROJECT_BOARD=$PROJECT_BOARD PROJECT_CPU=$PROJECT_CPU
```
mck1117, line 16: @dron0gus thoughts? I'd rather be explicit about what we pass to openblt makefile rather than implicit with exported vars.

Issue #5026, opened 2023-02-02
No load over 250kpa on Generic PWM


ElDominio:
![image](https://user-images.githubusercontent.com/5685108/216408107-ec9335a6-f7ee-4040-bb9d-0db740f59ecb.png)

Can I have more than 250 load? It's ok if it's  in multiples of 2

rusefillc:
@ElDominio how far would you like to go?

ElDominio:
We need negative numbers, so maybe multiples of 5?

-635 to 635 ?

rusefillc:
So either way it's an incompatible change.

We can expand this field to be two bytes not just play with scale

@mck1117 what do you think?

PR #5029, opened 2023-02-03
No load over 250kpa on Generic PWM #5026

rusefillc:


rusefillc:
@mck1117 looks like I need help if we go with 16 pin load axis

mck1117:
help is in #5032

By the way, mentioning the issue only in commits or in the PR title means that there's no link from the issue page to the PR. It has to be in the description.

Issue #3624, opened 2021-11-26
Any gauge should be an option for GPPWM either axis


rusefillc:
At the moment we focus on the dozen most fundamental sensors as axis options while really any of the 300 gauges should be available as an axis field

rusefillc:
Depends on #197

PR #5032, opened 2023-02-04
more flexible GPPWM

### LINKED ISSUE #5026: No load over 250kpa on Generic PWM

mck1117:
- Selectable X-axis (Y axis already selectable)
- Compute tunerstudio axis display values on the ECU instead of crazy switch statement in TS ini
- Expand Y axis limits to +-1000 with 0.1 resolution

fix #5026
#3624 for second axis selection option


mck1117:
needs #5035

PR #5034, opened 2023-02-05
Use external interactive pinout generator

chuckwagoncomputing:
I decided to make the interactive pinout generator usable by other people and projects.
I don't want to maintain two versions, so this pull request will make this repo use the new version, which is a Github Action.
If you would rather keep using the current version to keep things in control of the rusEFI organization, that's fine.

Also, I didn't do it in this PR, but we can freeze the version, e.g. `uses: chuckwagoncomputing/interactive-pinout@1.0`


rusefillc:
Would submodule give us better control over what revision of dependency we would be using? @mck1117 what do you think, merge as is in PR or suggest submodule approach?

mck1117:
> better control over what revision of dependency we would be using?

freezing a version of the action step does the same thing, more submodules will increase clone time

PR #5038, opened 2023-02-06
GitHub workflows: more ubuntu-latest mirrors

nmschulte:
re: https://github.com/rusefi/rusefi/pull/5037#issuecomment-1419135471, #4967.

Attempts to resolve issues with `archive.ubuntu.com` mirrors failing, without `apt-spy2`, by adding more mirrors to the `apt` configuration, which sources say will allow apt to use these additional mirrors in the case of failure.

### DIFF
--- a/.github/workflows/gen-diffs.yaml
+++ b/.github/workflows/gen-diffs.yaml
@@ -20,6 +20,10 @@ jobs:
 
     - name: Install sshpass, kicad, and tk bindings
       run: |
+        sudo cp /etc/apt/sources.list /etc/apt/sources.list.d/ubuntu-latest-mirrors.list
+        sudo sed 's#http://azure.archive.ubuntu.com/ubuntu/#http://ubuntu.osuosl.org/ubuntu/#' -i /etc/apt/sources.list.d/ubuntu-latest-mirrors.list
+        cat /etc/apt/sources.list | sudo tee -a /etc/apt/sources.list.d/ubuntu-latest-mirrors.list > /dev/null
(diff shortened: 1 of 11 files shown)


.github/workflows/gen-diffs.yaml
```diff
@@ -20,6 +20,10 @@ jobs:
 
     - name: Install sshpass, kicad, and tk bindings
       run: |
+        sudo cp /etc/apt/sources.list /etc/apt/sources.list.d/ubuntu-latest-mirrors.list
+        sudo sed 's#http://azure.archive.ubuntu.com/ubuntu/#http://ubuntu.osuosl.org/ubuntu/#' -i /etc/apt/sources.list.d/ubuntu-latest-mirrors.list
+        cat /etc/apt/sources.list | sudo tee -a /etc/apt/sources.list.d/ubuntu-latest-mirrors.list > /dev/null
+        sudo sed 's#http://azure.archive.ubuntu.com/ubuntu/#http://mirrors.ocf.berkeley.edu/ubuntu/#' -i /etc/apt/sources.list.d/ubuntu-latest-mirrors.list
         sudo add-apt-repository --yes ppa:kicad/kicad-6.0-releases
```
nmschulte, line 27: Maybe this `add-apt-repository` command could be used instead of the copy/modify logic?  The copy/modify logic keeps all of the listed components and codename of the currently used mirror without having to specify them.

rusefillc:
@nmschulte I merge as is since I am desperate but I see copy paste, looks like a shared script would be more elegant?

nmschulte:
I agree, I'll look into factoring into a script or so.

Issue #5036, opened 2023-02-06
Antilag switch pin mode was ignored


Benjamin-neo:
Antilag
Antilag activation by pin state always on no. (test on different pin, pullup, pulldown)
LaunchControl have same fonctionnality and work.

rusefillc:
@Benjamin-neo is this tested on bench or real car? Any content you can share build thread pictures logs videos?

Benjamin-neo:
tested on car, and on bench to tested on different pin to known if problem is hadrware or firmware. have look to compare with lauch same fonction, bug is in only on antilag ALSactivate pin state doesn't change .I dont tell antilag doesnt work , I report bug only in antilag activation by switch pin, all work expect ALSactivationPinstate

rusefillc:
@Benjamin-neo glad to hear tested on car, looking to see more info on the forum

rusefillc:
@Benjamin-neo will look into your bug report

Benjamin-neo:
@rusefillc yes will post more on forum when project will ready.
For the bug, I have look at more, the problem is in activation mode ( switch Input/ always active). it seem to be simple inverted. maybe just need correct .ini
I let you see, I hope this info will help you

rusefillc:
@TurboMarian looks like ALSActivatePinMode field was never used?

rusefillc:
@Benjamin-neo I've made a change please try https://rusefi.com/build_server/ no sooner than 20 minutes from now :)

Benjamin-neo:
@rusefillc 
I have look at and test Turbomarian , here is solution for switch antilag
![soluce](https://user-images.githubusercontent.com/85869292/217358514-18b58c03-8f1d-4905-8848-eccf0a39e87d.png)

Benjamin-neo:
Have test f407 from bundle, doesnt work.
test with previous version and just != instead of == . and work perfectly

rusefillc:
Well I do not get it

ALSActivatePinMode was not used I assume it's wrong for ALSActivatePinMode to be not used?

Also the != or == please someone explain in human words

Benjamin-neo:
yes it's wrong for ALSActivatePinMode.
In file [firmware/controllers/algo/antilag_system.cpp]
turbomarian has only made on change == to !=

isALSSwitchActivated = engineConfiguration->antiLagActivationMode != SWITCH_INPUT_ANTILAG;

[...]

PR #5039, opened 2023-02-06
Anitlag update

TurboMarian:
Added: Timer, Spark skip table
Antilag bug #5036 resolved


rusefillc:
Good: we now use PR progress

Less good: multiple good changes are piled together

Next step would be using branches which would allow multiple separate PRs all with separate review process

rusefillc:
squashing commits on https://github.com/TurboMarian/rusefi would make separating the PR harder. Sad face.

PR #5040, opened 2023-02-06
Als timer

rusefillc:
@TurboMarian let's see if this one would be green

seems like those three are all about timer, not sure if that's the complete change

### DIFF
--- a/firmware/controllers/algo/antilag_system.cpp
+++ b/firmware/controllers/algo/antilag_system.cpp
@@ -59,6 +59,12 @@ bool AntilagSystemBase::isALSMaxThrottleIntentCondition() const {
 	return engineConfiguration->ALSMaxTPS > throttleIntent;
 }
 
+bool AntilagSystemBase::isInsideALSTimerCondition() {
+	auto ALStime = ALStimer.getElapsedSeconds();
+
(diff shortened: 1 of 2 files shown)


rusefillc:
@TurboMarian an example of small pr which probably works on real firmware but seem to break unit tests. We care about unit tests.

rusefillc:
@TurboMarian how can I help revive all that? Would a video conf call? Should I fix this PR? anything else?

TurboMarian:
Could you give me a hint on how to do units tests on my computer? 

Also that bit need to be changed 

to if (!ALSMaxRPMCondition) {ALStimer.reset();}

nmschulte:
To run the unit tests I follow the GitHub Workflow (https://github.com/rusefi/rusefi/blob/master/.github/workflows/build-unit-tests.yaml#L13), which is like:

`pushd firmware/; ./gen_default_everything.sh; popd`
`pushd unit_tests/; make -j4 COVERAGE=yes; popd`
`pushd unit_tests/; ASAN_OPTIONS=detect_stack_use_after_return=1 build/rusefi_test; popd`

rusefillc:
@TurboMarian I've just picked some code from your fork. please rebase and go over missing parts

PR #5041, opened 2023-02-06
Als timer smaller

rusefillc:


rusefillc:
@TurboMarian an example of small progress which does not break things

TurboMarian:
> @TurboMarian an example of small progress which does not break things

OK. Whole code is working but from today I will PR only small pieces of code. One new function or idea = one PR. Thanks

PR #5049, opened 2023-02-10
Update to interactive-pinout@2.0

chuckwagoncomputing:
# TL;DR

Small breaking changes to yaml schema
I have update all YAMLs to reflect this
I'll consider myself responsible for updating the Connector-Mapping wiki page

# YAML Change 1

Older versions of interactive pinouts assumed that the pinouts were at `firmware/config/boards/<something>/connectors/<whatever>.yaml`, and made the `<something>` the directory the generated pinout was placed in within the pinouts directory.
e.g. `firmware/config/boards/microrusefi/connectors/main.yaml` -> `pinouts/microrusefi`

Now a field has been added to the `info` section of the yaml to specify the directory.
If it doesn't exist, the `title` field will be used as backup, and if that doesn't exist, the full path will be used, e.g. `pinouts/firmware/config/boards/microrusefi/connectors`

I have switched all yamls over, so the final directory structure hasn't changed. This required adding the info section to some boards.

# YAML Change 2

The `id` field in the `info` section has been changed to `cid`. I have updated all yamls to reflect this.
I think this will satisfy one of @mi-hol's complaints in https://github.com/rusefi/rusefi_documentation/issues/388

### DIFF
--- a/.github/workflows/gen-pinouts.yaml
+++ b/.github/workflows/gen-pinouts.yaml
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v3
 
     - name: Generate Pinouts
-      uses: chuckwagoncomputing/interactive-pinout@1.2
+      uses: chuckwagoncomputing/interactive-pinout@2.0
       with:
         mapping-path: ./firmware/config/boards/*/connectors/*.yaml
         warnings: "false"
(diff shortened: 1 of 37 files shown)


chuckwagoncomputing:
Compare changes in interactive-pinout:
https://github.com/chuckwagoncomputing/interactive-pinout/compare/1.2...2.0

PR #5054, opened 2023-02-11
remove useETBforIdleControl

mck1117:
Everyone with an ETB wants the ETB to do idle control, so why do we offer the option?

We've had a few users confused by this recently, so it's much clearer to simply remove the option. If you don't want ETB idle (for some weird reason), then just set the control range to 0.


firmware/integration/rusefi_config.txt
```diff
@@ -782,7 +782,7 @@ bit is_enabled_spi_2
 	bit is_enabled_spi_4
 	bit pauseEtbControl;Disable the electronic throttle motor and DC idle motor for testing.\nThis mode is for testing ETB/DC idle position sensors, etc without actually driving the throttle.
 	bit alignEngineSnifferAtTDC
-	bit useETBforIdleControl;This setting allows the ETB to act as the idle air control valve and move to regulate the airflow at idle.
+	bit unused764b29;
```
nmschulte, line 785: :tada:

Issue #5061, opened 2023-02-13
ETB auto-calibrate usability: both buttons should not be enabled at the same time


rusefillc:
At the moment both "start ETB pin autotune" and "stop ETB pin autotune" buttons are always or often enabled at the same time.

Would be nice for buttons to know current state of the process. If we are NOT tuning - start should be enabled, stop should be disabled

while we are tuning start should be DISABLED and stop should be ENABLED

![image](https://user-images.githubusercontent.com/48498823/218373860-1771032f-f25b-41d4-bb24-31e86d398e89.png)

rusefillc:
@nmschulte is this by any chance something you could help with?

nmschulte:
On it; free to assign to me.

rusefillc:
@nmschulte you would need to expose "isAutoTuneRunning" bit output for UI to use it for conditional logic

see into existing .txt files maybe we even have the bit already. maybe only UI change is needed

nmschulte:
Cool.  While researching, I found this, semi-related ...
https://github.com/rusefi/rusefi/blob/master/firmware/controllers/actuators/electronic_throttle.cpp#L1036

which lead to this (ex. `isFanOn`) ... is this file still deprecated; what's the alternative?
https://github.com/rusefi/rusefi/blob/master/firmware/console/binary/output_channels.txt#L4

rusefillc:
https://github.com/rusefi/rusefi/commit/ce3cfc89c19286b16b1e9790d6ac30f457b8a678

rusefillc:
DC vs ETB is a much much bigger fish to fry :(

nmschulte:
It looks like `calibrationMode` already outputs the status, so that's a win.

"DC" means "deprecated" or ?

rusefillc:
DC as https://en.wikipedia.org/wiki/DC_motor


like DC motor wastegate code flow should probably NOT go through electronic_throttle.cpp right?

PR #5063, opened 2023-02-13
Fixup ETB auto-calib/tune buttons

nmschulte:
for #5061

### DIFF
--- a/firmware/controllers/actuators/electronic_throttle.cpp
+++ b/firmware/controllers/actuators/electronic_throttle.cpp
@@ -447,19 +447,15 @@ expected<percent_t> EtbController::getClosedLoopAutotune(percent_t target, perce
 		float kd = 0.08f * ku * m_tu;
 
 		// Every 5 cycles (of the throttle), cycle to the next value
-		if (m_autotuneCounter == 5) {
+		if (m_autotuneCounter >= 5) {
 			m_autotuneCounter = 0;
-			m_autotuneCurrentParam++;
-
-			if (m_autotuneCurrentParam >= 3) {
-				m_autotuneCurrentParam = 0;
-			}
+			m_autotuneCurrentParam = (m_autotuneCurrentParam + 1) % 3; // three ETB calibs: P-I-D
 		}
 
 		m_autotuneCounter++;
 
 		// Multiplex 3 signals on to the {mode, value} format
-		engine->outputChannels.calibrationMode = (uint8_t)static_cast<TsCalMode>(m_autotuneCurrentParam + 3);
+		engine->outputChannels.calibrationMode = (uint8_t)static_cast<TsCalMode>((uint8_t)TsCalMode::EtbKp + m_autotuneCurrentParam);
 
 		switch (m_autotuneCurrentParam) {
 		case 0:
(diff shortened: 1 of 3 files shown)


nmschulte:
I don't immediately have access to hardware to test this.

firmware/controllers/actuators/electronic_throttle.cpp
```diff
@@ -447,19 +447,15 @@ expected<percent_t> EtbController::getClosedLoopAutotune(percent_t target, perce
 		float kd = 0.08f * ku * m_tu;
 
 		// Every 5 cycles (of the throttle), cycle to the next value
-		if (m_autotuneCounter == 5) {
+		if (m_autotuneCounter >= 5) {
```
mck1117, line 450: the idea with the `>` was that it's more robust against an extra `++` slipped in somewhere, but it's not a big deal.

nmschulte: That's why I added it ;)

firmware/tunerstudio/rusefi.input
```diff
@@ -4028,11 +4028,11 @@ dialog = tcuControls, "Transmission Settings"
...
-		commandButton = "Auto Calibrate ETB 2", cmb_etb_auto_calibrate_2, {etb2configured == 1 && tps2_1AdcChannel != @@ADC_CHANNEL_NONE@@}
+		commandButton = "Auto Calibrate ETB 1", cmd_etb_auto_calibrate, {etb1configured == 1 && tps1_1AdcChannel != @@ADC_CHANNEL_NONE@@ && calibrationMode == @@TsCalMode_None@@}
+		commandButton = "Auto Calibrate ETB 2", cmd_etb_auto_calibrate_2, {etb2configured == 1 && tps2_1AdcChannel != @@ADC_CHANNEL_NONE@@ && calibrationMode == @@TsCalMode_None@@}
 		field = "Second step"
-		commandButton = "Start ETB PID Autotune", cmd_etb_autotune,       { etb1configured == 1 || etb2configured == 1 }
-		commandButton = "Stop ETB PID Autotune",  cmd_etb_autotune_stop,  { etb1configured == 1 || etb2configured == 1 }
+		commandButton = "Start ETB PID Autotune", cmd_etb_autotune,       {(etb1configured == 1 || etb2configured == 1) && calibrationMode = @@TsCalMode_None@@}
+		commandButton = "Stop ETB PID Autotune",  cmd_etb_autotune_stop,  {(etb1configured == 1 || etb2configured == 1) && calibrationMode >= @@TsCalMode_EtbKp@@ && calibrationMode <= @@TsCalMode_EtbKd@@}
```
nmschulte, line 4035: The condition in parenthesis is redundant now, as 3-5 imply ETB calib underway.  It also may cause a problem if somehow the calibration is enabled and they change: cannot be disabled (though that's always been the case): shall I remove it?

firmware/controllers/actuators/electronic_throttle.cpp
mck1117: oh you ADDED it, apparently I can't read a diff

PR #5072, opened 2023-02-13
add 50/83.3/125kbit CAN

mck1117:
- Add 50kbit/s, 83.33kbit/s, 125kbit/s
- Fix 100kbit on F7
- Fix 100kbit on H7
- tweak 250/500/1000kbit on H7
- reorder CAN bitrate options so they're sorted with the new ones

### DIFF
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -93,7 +93,7 @@
 ! Any time an incompatible change is made to the configuration format stored in flash,
 ! update this string to the current date! It is required to also update TS_SIGNATURE above
 ! when this happens.
-#define FLASH_DATA_VERSION 10019
+#define FLASH_DATA_VERSION 10020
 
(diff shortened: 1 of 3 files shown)


Er0pification:
> reorder CAN bitrate options so they're sorted with the new ones
Is this the right thing to do? Kinda messes with existing configs

firmware/integration/rusefi_config.txt
```diff
@@ -93,7 +93,7 @@
 ! Any time an incompatible change is made to the configuration format stored in flash,
 ! update this string to the current date! It is required to also update TS_SIGNATURE above
 ! when this happens.
-#define FLASH_DATA_VERSION 10019
+#define FLASH_DATA_VERSION 10020
```
rusefillc, line 96: @Er0pification that's why FLASH_DATA_VERSION change to force configuration migration via TS

PR #5077, opened 2023-02-14
Rearrange & clarify trigger UI

mck1117:
Reorganize trigger settings in to two windows;

- One for all "normal things": Primary trigger shape, cam trigger shapes, VVT offsets, all 6 inputs. Hopefully this helps a bit of user confusion. Previously, you had two switch back and forth between two windows, and it really wasn't clear why the cam inputs were greyed out, as the setting to enable them was **hidden on a different page**.
- One for "weird things": Advanced trigger settings, gap override, and comparator settings (if so equipped)

image


mi-hol:
should this change have an impact on current documentation?
i.e. https://wiki.rusefi.com/How-Do-I-Set-My-Trigger-Offset

mck1117:
> should this change have an impact on current documentation?

I think it should be OK, all parameters are still named the same thing, just moved around a little

firmware/integration/rusefi_config.txt
```diff
@@ -467,7 +467,7 @@ bit useTLE8888_cranking_hack;
 bit kickStartCranking
 bit useSeparateIdleTablesForCrankingTaper;This uses separate ignition timing and VE tables not only for idle conditions, also during the postcranking-to-idle taper transition (See also afterCrankingIACtaperDuration).
 bit launchControlEnabled;
-bit doNotFilterTriggerEdgeNoise,"without filter","with filter";"Detect double trigger edges"
```
rusefillc, line 470: Can we please keep the bit? Implementing the logic would also be great!

rusefillc:
Just cherry-picked the sweet part!

firmware/integration/rusefi_config.txt
mck1117: Do we think there is any value in turning it off? It's computationally very very cheap, and I don't think has any risk in it always being on. If we can be smarter (ie, detect that error) and it's not dangerous to do, why shouldn't we just always do it?

rusefillc: The common pattern I see is "we should be smart by default but have advanced feature to act simpler"

rusefillc:
@mck1117 rebase or close or else?

rusefillc:
@mck1117 rebase or close or else?

mck1117:
is there anything left after your rude cherry pick move?

PR #5082, opened 2023-02-15
[HW CI Gradle] remove argument

chuckwagoncomputing:
I *think* this PR will make the workflow run to see if it works.


rusefillc:
i would hope that it does not work

my understandin
"java -jar XXX" looks into manifest for entry point class name

"java -cp XXX" needs start class argument

chuckwagoncomputing:
It's getting the class name, held in the `${{matrix.test-suite}}` variable.

rusefillc:
@chuckwagoncomputing thank you! finally I get it - I was specifying two parameters

PR #5086, opened 2023-02-16
update Toyota ETCS-i partial redundancy secondary-max values

nmschulte:


rusefillc:
See also https://github.com/rusefi/rusefi/pull/4898

PR #5087, opened 2023-02-16
TPS, meld FordRedundantSensor

nmschulte:
- melds `RedundantFordTps`'s partial redundancy feature into `RedundantSensor` (w/ unit tests)
- alters Broken / Experimental `useFordRedundant*`/`*SecondaryMaximum` configs to a single scaler `*SecondaryMaximum` (default: `100`%), which when changed implies partial redundancy


mck1117:
My concern with merging these two is that the existing `RedundantSensor` makes no assumptions about the nature of the data it's handling, but this adds that assumption even in case the fordoyota mode is disabled.  We could end up with other types of sensor we want redundant (some cars have redundant fuel pressure sensor inputs, for example), and then we'd have to split them back apart.

Thoughts on leaving the existing redundant sensor as-is, no meld? I'm happy to be convinced either way.

nmschulte:
My motivations for combining were that:
- `RedundantSensor` has already an "ignore second sensor," and ...
- the "monitors vs averages" logic would need split across the classes if accepted

Your concern about "the nature of the data" is the"secondary maximum", and/or that it maxes at `100` (and there are now literals about it)?  Possibly that can be agnosticized somehow into percentage terms if so?

nmschulte:
reverted to draft status, pending agnosticizing the meld

nmschulte:
@mck1117:
> this adds that assumption even in case the fordoyota mode is disabled.

I've been thinking about this more and I think I don't understand; can you point to the code that makes this assumption?  `m_secondMaximum` is just a percent that the code uses to proportion the secondary sensor to the first: it doesn't make any assumption about the nature of the data, it just allows the two to have different scales.

Otherwise: does it make sense to alter `FuncSensPair` to handle this scaling instead (in `init_tps.cpp`)?

rusefillc:
Not alive

PR #5088, opened 2023-02-16
TPS, "monitored" RedundantSensor

nmschulte:
_follows #5087_

- tidies/clarifies some `init_tps` logic/code
- adds Broken / Experimental `*SecondaryMonitors` config bits to configure the redundancy to purely monitor (second/secondary sensor checks the first/primary) and resolve directly the primary sensor

The "monitored" feature allows to avoid the step-wise transition at the partial redundancy switch-point, which is difficult/impossible to remove even in static/bench conditions (and likely to vary in dynamic/real-world conditions).  Ex. Toyota ETCS-i:

Averaging (`tpsSecondaryMaximum`: `64.0`% [60.8% threshold @ 4.81325 V]; stepwise highlighted):
image

Monitoring (`tpsSecondaryMaximum`: `64.0`% [60.8% threshold @ 4.81325 V]; stepwise eliminated):
image

The stepwise transition can be minimized by choosing a[n unachievable] "perfect" `tpsSecondaryMaximum`, but there will always be a small step due to the 0.5% resolution of this field and that the sensors can/will vary and drift.  More precision can be had by tuning the `tps#SecondaryMax` config in coordination, but this makes the auto-tune features useless, and isn't ideal.


firmware/controllers/sensors/redundant_sensor.cpp
```diff
@@ -9,33 +9,66 @@ RedundantSensor::RedundantSensor(SensorType outputType, SensorType first, Sensor
...
+		float threshold = m_secondMaximum * 0.95;
+
+		// The scaled second sensor, proportioning it to the first sensor
+		float scaledSecond = sensor2.Value * m_secondMaximum / 100;
+
+		// Check second sensor is below partial redundancy switch-over threshold
+		if (sensor2Value > threshold) {
+			// Check first sensor is at or above partial redundancy switch-over threshold
+			if (sensor1.Value > threshold - m_maxDifference) {
+				return sensor1.Value;
```
nmschulte, line 58: This could instead, with an inverted `if` condition, set `sensor2Value = sensor1.Value`, and fall-through to the averaging case if for some reason that was preferred.
Similarly could force a deviant `sensor2Value` to blow the check and leverage the catch-all `return UnexpectedCode::Inconsistent;` as well.

firmware/init/sensor/init_tps.cpp
```diff
@@ -13,6 +12,8 @@
 #define MAX_TPS_PPS_DISCREPANCY 5.0f
 #endif
 
+#define TPS_ADC_UNITS_PER_VOLT 200
```
nmschulte, line 15: Maybe this should live in a higher scope, for other (existing?) use?

rusefillc:
Not active?

Issue #5079, opened 2023-02-14
Native SENT TPS input


rusefillc:
While we have a Lua hack for TPS input we need proper native SENT TPS integration

rusefillc:
https://youtu.be/ZSk85Z0UYQw

PR #5114, opened 2023-02-21
Native SENT TPS input #5079 step forward

rusefillc:


unit_tests/tests/sensor/test_sensor_init.cpp
```diff
@@ -131,6 +131,7 @@ TEST(SensorInit, DriverIntentNoPedal) {
 
 	initTps();
 
+	ASSERT_TRUE(Sensor::hasSensor(SensorType::AnalogTps1));
```
mck1117, line 134: this doesn't need a different sensor type, does it?

mck1117: won't the SENT sensor just register itself?

rusefillc: sorry do not follow. not sure if https://github.com/rusefi/rusefi/pull/5114/commits/1d4e41bd2853ac09105138c1ce36a3235800084a answers any questions

mck1117: imo this is a misuse of the proxy sensor, since analog TPS and SENT tps are mutually exclusive, there is no reason to process both as separate sensors instead of registering one vs other

rusefillc: thank you

PR #5115, opened 2023-02-21
Build only one bundle

### LINKED ISSUE #5104: "Build only one bundle fast" hack mode of operation
We have so much amazing automation like code generation, ini upload etc that I am addicted to github actions now.

With 20 targets three commits would create a nasty queue of builds

We can either fix the root cause of agents shortage by throwing CPU at the problem or shall we implement a shortcut:

So when I am in the field and I need to fix stuff again and again I need to only build say 4chan. At the moment I do nasty https://github.com/rusefi/rusefi/commit/47a59294058480a67e26b0a12feab65ab96d1a69

Could be that it's best as is. Or could it be improved somehow? Looking for ideas.

chuckwagoncomputing:
To use this mode, put `only:<bundle name>` in your commit message.
There must not be a space between `only:` and the name, and the name is case-sensitive.
Fix #5104

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -298,8 +298,12 @@ jobs:
       run: echo '::error::Branch names must not contain ".", this breaks firmware autoupdates.' && exit 1
     - name: Execution throttle early exit
       # Don't skip any jobs if this workflow was run manually.
-      if: ${{ matrix.skip-rate && github.event_name != 'workflow_dispatch' }}
-      run: if (($(($RANDOM % 100)) < ${{ matrix.skip-rate }})); then echo "skip=true" >> $GITHUB_ENV; fi
+      if: ${{ matrix.skip-rate && github.event_name != 'workflow_dispatch' || contains(github.event.head_commit.message, 'only:') }}
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-firmware.yaml
```diff
@@ -298,8 +298,17 @@ jobs:
       run: echo '::error::Branch names must not contain ".", this breaks firmware autoupdates.' && exit 1
     - name: Execution throttle early exit
       # Don't skip any jobs if this workflow was run manually.
-      if: ${{ matrix.skip-rate && github.event_name != 'workflow_dispatch' }}
-      run: if (($(($RANDOM % 100)) < ${{ matrix.skip-rate }})); then echo "skip=true" >> $GITHUB_ENV; fi
+      if: ${{ matrix.skip-rate && github.event_name != 'workflow_dispatch' || contains(github.event.head_commit.message, 'only:') }}
```
rusefillc, line 301: our yaml is getting more and more complicated. please add usage documentation and/or example as a code comment right into yaml?

rusefillc:
@chuckwagoncomputing funny it has triggered to itself but also something about missing bracket in bottom line of failure?

image

chuckwagoncomputing:
Oh, I never thought about the possibility of newlines in a commit message.

PR #5124, opened 2023-02-23
can msg tx: do not ignore bus argument

dron0gus:
bus argument was introduces in 12c899ddcab89ff79a40782bc82d473f90ae6bfc but was ignored.

@mck1117 ping


mck1117:
oops, good catch!

mck1117 (approved):

PR #5133, opened 2023-02-25
Reduce missing hex errors

chuckwagoncomputing:
I'm not sure if hex is needed?
There will still be an error for boards that use an OpenBLT composite image. The comments in common_make.sh say not to deliver any format that can confuse users, so I didn't copy the .hex for those boards.
Before:
image
After:
image


rusefillc:
Not merging because not sure which way is the better way to fix it.

there is an argument that fewer files is smaller bundle and less confusion. Also funny connection https://github.com/mck1117/wideband/pull/212

Issue #5145, opened 2023-03-01
Shall we enable '-Werror=stringop-truncation'?


rusefillc:
-Werror=stringop-truncation sounds cool but it generates at least the following error

```
./console/binary_log/log_field.cpp:47:24: error: 'strncpy' specified bound 34 equals destination size [-Werror=stringop-truncation]
   47 |                 strncpy(&buffer[55], m_category, 34);
      |                        ^
```

nmschulte:
How bad is `-Werror=all`?

nmschulte:
Does MLV/`.mlq` require that the string field here include a null terminator?  `BinarySensorLog` seems to not need it/worry: https://github.com/rusefi/rusefi/blob/master/java_console/ui/src/main/java/com/rusefi/sensor_logs/BinarySensorLog.java#L148

nmschulte:
`strncpy` behavior is the ideal solution here -- are we opposed to judicious use of `#pragma GCC diagnostic ignored`?
```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wstringop-truncation"
    // strncpy's exclusion of null terminator is correct and appropriate behavior for this field: silence the warning
    strncpy(...);
#pragma GCC diagnostic pop
```

rusefillc:
I am all for it

PR #5146, opened 2023-03-02
silence -Wstring-truncation for correct use of strncpy

nmschulte:
- ensuring header is zerod

addresses #5145

### DIFF
--- a/firmware/console/binary_log/log_field.cpp
+++ b/firmware/console/binary_log/log_field.cpp
@@ -43,10 +43,13 @@ void LogField::writeHeader(Writer& outBuffer) const {
 	buffer[54] = m_digits;
 
 	// Offset 55, (optional) category string
+	memset(&buffer[55], 0, 34);
 	if (m_category) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
+		// strncpy's exclusion of null terminator is correct and appropriate behavior for this field
 		strncpy(&buffer[55], m_category, 34);
-	} else {
-		memset(&buffer[55], 0, 34);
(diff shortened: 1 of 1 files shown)


rusefillc:
I think it's null terminated or max length? I guess based on lack of length

firmware/console/binary_log/log_field.cpp
```diff
@@ -43,10 +43,9 @@ void LogField::writeHeader(Writer& outBuffer) const {
 	buffer[54] = m_digits;
 
 	// Offset 55, (optional) category string
+	memset(&buffer[55], 0, 34);
```
mck1117, line 46: we have `efi::clear`, maybe useful

nmschulte:
It's unclear to me why `stringop-truncated` is not treated as an error; there is some interplay between the `RUSEFI_OPT` values of `firmware/rusefi_rules.mk` (fed to `USE_OPTS` of ChibiOS IIUC) and `firmware/Makefile`'s `CWARN`/`CPPWARN`, and/or GCC behavior re: order of flags that I do not understand.  I saw also some `maybe-unitialized` warns/errors (from the linker) in my confusing trials that may be going unnoticed.

_(Does `USE_OPTS` only apply to ChibiOS code, leading to my confusion?  Is `-Werror` and `-Werror=all` the same to GCC?  Should strive for `CWARN`/`CPPWARN` to have `-Werror`[`=all`]?)_

firmware/console/binary_log/log_field.cpp
nmschulte: I poked at this but the generics confused me: does this support `char[]` yet?

PR #5147, opened 2023-03-02
Shall we enable '-Werror=stringop-truncation'? #5145

rusefillc:
@nmschulte here's a branch just to see where we are


rusefillc:
@nmschulte poke :)

Issue #5159, opened 2023-03-08
latest snapshots cause error "firmware format incompatible" in console 


mi-hol:
downloaded https://rusefi.com/build_server/rusefi_bundle_mre_f4_no_openblt.zip
gives

![2023-03-08 14_14_57-DISCONNECTED Console 20230227; firmware=N_A@COM4 not loaded](https://user-images.githubusercontent.com/22799428/223728583-97a01d22-8d26-4164-9025-48c267d9c1b2.png)

steps leading to it described in https://rusefi.com/forum/viewtopic.php?p=47341#p47341

rusefillc:
please confirm that you have flashed rusefi_bundle_mre_f4_no_openblt.zip into hardware

As is I wonder if you have used latest console to connect to very old firmware

mi-hol:
I connected MRE via USB for first time and then pressed "connect".
Will try to flash as suggested.

mi-hol:
flashing fixed error
![image](https://user-images.githubusercontent.com/22799428/223730535-7c8935fb-069f-49a5-a744-353e2c428a11.png)
![image](https://user-images.githubusercontent.com/22799428/223730820-d0644802-c5a9-48d9-b3ba-fca14d567b1b.png)

mi-hol:
I'll add this encountered issue as note for troubleshooting in doc

mi-hol:
@rusefillc actually a better approach would be to change the error message in code.
The recommendation to "use previous console version or flash new firmware" should be added to error message.

Benefit:
This change would allow issue and resolution to be understood by user instantly, improving user experience

mi-hol:
unfortunately searching with string "incompatible firmware format" had no hits in GitHub, so I don't know where to change it for a PR

mi-hol:
found string  now via VSCode in \rusefi\java_console\io\src\main\java\com\rusefi\binaryprotocol\BinaryProtocol.java

PR #5161, opened 2023-03-08
improve error message

mi-hol:
My JAVA skills are minor and I can't compile it yet in my environment, hence this is a test only


mi-hol:
fixes #5159

PR #5164, opened 2023-03-08
fix #5160

### LINKED ISSUE #5160: CAN bus AFR value inverted
https://rusefi.com/forum/viewtopic.php?f=17&t=2507

mck1117:
send lambda properly over OBD2

fix #5160

### DIFF
--- a/firmware/controllers/can/obd2.cpp
+++ b/firmware/controllers/can/obd2.cpp
@@ -156,11 +156,9 @@ static void handleGetDataRequest(const CANRxFrame& rx, size_t busIndex) {
 		obdSendValue(_1_MODE, pid, 1, Sensor::getOrZero(SensorType::Tps1) * ODB_TPS_BYTE_PERCENT, busIndex);	// (A*100/255)
 		break;
 	case PID_FUEL_AIR_RATIO_1: {
-		float lambda = Sensor::getOrZero(SensorType::Lambda1);
-		// phi = 1 / lambda
-		float phi = clampF(0, 1 / lambda, 1.99f);
+		float lambda = clampF(0, Sensor::getOrZero(SensorType::Lambda1), 1.99f);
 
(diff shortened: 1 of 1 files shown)


rusefillc:
this PR is red

mck1117:
looks like a fluke, and one I've never seen before

image

firmware/controllers/can/obd2.cpp
```diff
@@ -156,11 +156,9 @@ static void handleGetDataRequest(const CANRxFrame& rx, size_t busIndex) {
 		obdSendValue(_1_MODE, pid, 1, Sensor::getOrZero(SensorType::Tps1) * ODB_TPS_BYTE_PERCENT, busIndex);	// (A*100/255)
 		break;
 	case PID_FUEL_AIR_RATIO_1: {
-		float lambda = Sensor::getOrZero(SensorType::Lambda1);
-		// phi = 1 / lambda
-		float phi = clampF(0, 1 / lambda, 1.99f);
+		float lambda = clampF(0, Sensor::getOrZero(SensorType::Lambda1), 1.99f);
```
nmschulte, line 159: Just curious, why `1.99f` vs e.g. `1.9999f` -- I realize "< 2 ratio" is the concern; is the extra "top end" not really worth?

mck1117: Nope. Not much reason to unless you're doing "real" diagnosis using OBD2. We support OBD2 pretty much exclusively for gauges/dashes/apps/etc, not real diagnosis.

mi-hol: > We support OBD2 pretty much exclusively for gauges/dashes/apps/etc, not real diagnosis.

This statement is an important fact but it seems not captured in doc yet.
Is it ok that I add it to https://wiki.rusefi.com/CAN/ ?

PR #5183, opened 2023-03-16
Do I see dead code?

rusefillc:


mck1117:
Kinetis uses that implementation. Very much not dead.

rusefillc:
I will add some comments instead of killing code

Issue #5117, opened 2023-02-21
Daily full build


rusefillc:
We have amazing skip-rate

With conditional compilation skip-rate hides errors in specific not frequently compiled binaries like https://github.com/rusefi/rusefi/actions/runs/4235793399/jobs/7359853821

We need a daily full compilation at 4AM EST unless I am overthinking that all

chuckwagoncomputing:
related: #5098

chuckwagoncomputing:
#5127 now creates a Github release every day.

rusefillc:
Wow @chuckwagoncomputing thank you! Would you be interested to take the honors and announce it around https://rusefi.com/forum/viewtopic.php?f=5&t=1509 once we have first daily?

Damn that's a leap towards proper _LTS_ and to some extent we are at a point of major potential confusion for the users. We now have what three things?
* release? LTS release? proper release? I guess let's use _release_ only for that
* daily... snapshot?
* xxx snapshot? what do we call that momentary snapshot?

Wow now we need to update some documentation somewhere @mi-hol :)

@eldominio @mck1117 FYI

chuckwagoncomputing:
Maybe...
-Release
-Nightly
-Continuous or Development, not sure which I prefer.

To me, "Snapshot" infers that it will stick around. "Snapshot" would be a better name for what I call "Nightly" than it would be for the Continuous build.

rusefillc:
https://stackoverflow.com/questions/5901378/what-exactly-is-a-maven-snapshot-and-why-do-we-need-it has too much answer

nmschulte:
Is it possible to split the cron about `build-firmware.yml` to a separate workflow/action, so that non-rusefi forks don't run automatically and report failures (https://github.com/nmschulte/rusefi/actions/runs/4328501687/jobs/7558258610#step:3:10)?

chuckwagoncomputing:
It's not failing because it's a fork, it's failing because I'm an idiot.
But I could add a check to only release if the repo is rusefi/rusefi

mi-hol:
@chuckwagoncomputing could this issue be closed?

chuckwagoncomputing:
Yep I'm not sure why it was re-opened.

PR #5187, opened 2023-03-17
firmware workflow schedule only for rusefi/rusefi

nmschulte:
for https://github.com/rusefi/rusefi/issues/5117#issuecomment-1454313413

avoids failing release scheduled runs for non-golden repos, allowing to keep the workflow and associated notifications and avoid spam about unconcerning failures: https://github.com/nmschulte/rusefi/actions/runs/4443042591/jobs/7799915165#step:4:15

please double check my work:
- condition is correct? https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
- concern in all the correct places?

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -18,7 +18,7 @@ jobs:
 
     - name: Set run condition variables
       run: |
-        if [ "${{github.event_name}}" = "schedule" ]; then
+        if [ "${{github.event_name}}" = "schedule" ] && [ "${{github.repository}}" = "rusefi/rusefi" ]; then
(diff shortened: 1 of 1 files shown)


chuckwagoncomputing:
I would do the check in lines [21](https://github.com/rusefi/rusefi/blob/15576819d269acacc03144b3709fd77b2a0f2ff3/.github/workflows/build-firmware.yaml#L21) and [363](https://github.com/rusefi/rusefi/blob/15576819d269acacc03144b3709fd77b2a0f2ff3/.github/workflows/build-firmware.yaml#L363) instead.

chuckwagoncomputing (approved):

Issue #5193, opened 2023-03-21
vvttoothDurations0 NT2US


rusefillc:
https://rusefi.com/online/view.php?log=1190

vvttoothDurations0 is currently in NT should be in proper units

![image](https://user-images.githubusercontent.com/48498823/226605720-4d5a9b34-a9a9-448a-a056-6d6a0e293a07.png)

PR #5194, opened 2023-03-21
vvttoothDurations0 NT2US #5193

andreika-git:


firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -311,7 +311,7 @@ void hwHandleVvtCamSignal(TriggerValue front, efitick_t nowNt, int index) {
 			front == TriggerValue::RISE ? SHAFT_PRIMARY_RISING : SHAFT_PRIMARY_FALLING, nowNt);
 		// yes we log data from all VVT channels into same fields for now
 		tc->triggerState.vvtSyncGapRatio = vvtDecoder.triggerSyncGapRatio;
-		tc->triggerState.vvttoothDurations0 = vvtDecoder.toothDurations[0];
+		tc->triggerState.vvttoothDurations0 = (float)NT2US(vvtDecoder.toothDurations[0]);
```
rusefillc, line 314: why do we have to cast to float? would this value ever have numeric in it?

i.e. would it be better to just change data type to int32?

andreika-git: I don't think we need sub-microsecond precision, so I'll make it uint32_t...

PR #5215, opened 2023-04-02
fix typo

mi-hol:


mck1117:
the rest of that comment is horribly out of date

mi-hol:
> the rest of that comment is horribly out of date

@mck1117 unfortunately I'm not in a position to correct it, would you mind to step-in?

PR #5220, opened 2023-04-05
Cleanup 2023

dron0gus:


firmware/hw_layer/adc/adc_inputs.cpp
```diff
@@ -398,6 +398,7 @@ class SlowAdcController : public PeriodicController<256> {
 };
 
 void addChannel(const char *name, adc_channel_e setting, adc_channel_mode_e mode) {
+	(void)name;
```
mck1117, line 401: why not just remove the parameter?

dron0gus: it is used in many places. not sure about long term plans. may be we will have adc users list like we have for gpio...

firmware/ext/FatFS/ff.c
```diff
@@ -3094,7 +3094,7 @@ FRESULT find_volume (	/* FR_OK(0): successful, !=0: any error occurred */
 	/* Find an FAT partition on the drive. Supports only generic partitioning rules, FDISK and SFD. */
 	bsect = 0;
 	fmt = check_fs(fs, bsect);			/* Load sector 0 and check if it is an FAT-VBR as SFD */
-	if (fmt == 2 || (fmt < 2 && LD2PT(vol) != 0)) {	/* Not an FAT-VBR or forced partition number */
+	if (fmt == 2 || ((fmt < 2) && (bool)(LD2PT(vol) != 0))) {	/* Not an FAT-VBR or forced partition number */
```
mck1117, line 3097: why do we edit imported libraries?

rusefillc: Git submodule but yes

PR #5221, opened 2023-04-07
set aim dash CAN gear frame to current gear

m888r:
Small edit sending gear over CAN with AIM dashboards, I want it to update with rpm/vss gear calculations. 
From the AIM documentation, it seems like it's supposed to be a signed 16 bit int:
image

Currently the frame struct is using `scaled_channel<uint16_t, 10>`, so I imagine it should be changed to `scaled_channel<int16_t, 1>`?


rusefillc:
``scaled_channel<uint16_t, 10> Gear;`` sure looks weird. Do you know a URL for any relevant AIM documentation?

m888r:
[yep, here it is, page 3](https://www.aimtechnologies.com/support/racingecu/AiM_CAN_101_eng.pdf)

rusefillc:
oops we even have that PDF linked in the source code already

PR #5269, opened 2023-05-08
flash_main: more checks when loading config from MFS

dron0gus:


firmware/controllers/flash_main.cpp
```diff
@@ -260,11 +260,27 @@ static FlashState readConfiguration() {
 	mfs_error_t err = mfsReadRecord(&mfsd, EFI_MFS_SETTINGS_RECORD_ID,
 						&settings_size, (uint8_t *)&persistentState);
 
-	// TODO: check err result better?
 	if (err == MFS_NO_ERROR) {
+		// readed size is not exactly the same
```
rusefillc, line 264: @dron0gus is this code duplication?

PR #5270, opened 2023-05-08
usb: wait 15 sec for usb to spin up

dron0gus:
Wait 15 seconds for USB to become CONFIGURED.
Previous timeout was 5 sec. Windows does usb enumeration quite slow lets give some more room


mck1117:
Really, you got it to take that long? I’ve never seen it take longer than a second.

PR #5271, opened 2023-05-08
Ms broadcast

dron0gus:


firmware/controllers/can/can_dash_ms.cpp
```diff
@@ -0,0 +1,130 @@
...
+#if EFI_CAN_SUPPORT
+#include "can.h"
+#include "can_msg_tx.h"
+
+#include "rusefi_types.h"
+#include "rtc_helper.h"
+#include "fuel_math.h"
+
+/* TODO:
+ * use engine->outputChannels instead of Sensor::getOrZero as it cheaper */
```
rusefillc, line 21: no, no-no-no!!! Definitely not read from engine->outputChannels

dron0gus: Don't we care about coherency and cpu time? Why do same job twice?

rusefillc: it's a balance between design risks and costs. the code here is not on the mission critical path so we can totally afford x5 CPU consumption for 20% readability improvement.

humans are expensive, CPU is cheap

mck1117: Using `outputChannels` is not cheaper at all. If TS isn't connected, there shouldn't be a requirement to populate it at all! There is no guarantee that it is populated, in fact.

PR #5286, opened 2023-05-24
Fix invoking virtual functions from constructors

szsam:
Virtual functions should not be invoked from a constructor or destructor of the same class. Confusingly, virtual functions are resolved statically (not dynamically) in constructors and destructors for the same class. The call should be made explicitly static by qualifying it using the scope resolution operator.


rusefillc:
@szsam is this something GCC could or should be preventing with a relevant compiler setting?

PR #5287, opened 2023-05-24
Fix offset used before range check

szsam:
Accessing an array offset before checking the range means that the program may attempt to read beyond the end of a buffer.


Machiry:
Hey @rusefillc,

I am Aravind Machiry, an Assistant Professor at Purdue's ECE Department.

This pull request (along with several others) is the result of our ongoing research work (along with @szsam) to improve the security of open-source embedded projects.

We are doing a short (5-minute) survey to understand the use of static analysis tools in embedded projects.

It would greatly benefit our research if you could fill out this anonymous survey: https://purdue.ca1.qualtrics.com/jfe/form/SV_0OnXfr5plPe1QCa

Thank you,
Aravind

PR #5289, opened 2023-05-29
Trim EEC-IV breakout so it fits in case

chuckwagoncomputing:
Wallace over on The Turbo Forums had to trim his EEC-IV breakout to fit in his case. I don't know if this is necessary for all EEC-IV cases, or just specific ones, but I went ahead and trimmed the breakout so it would fit.
https://www.theturboforums.com/threads/91-lincoln-lsc.388470/#post-2058417


rusefillc:
Can we add changelog.txt and gerbers.zip for new revision to be complete?

chuckwagoncomputing:
Done.

Issue #5283, opened 2023-05-23
test_trigger_input_adc progress


rusefillc:
We need more scenarios to be covered in test_trigger_input_adc

andreika-git:
Currently the following datasets are tested:

- trigger_adc_1.csv 
- trigger_adc_750.csv 
- trigger_adc_1000.csv 

Remaining are "decr/incr" and "real*"

rusefillc:
We now have a unit test with RPM! Progress was made.

PR #5298, opened 2023-06-01
test_trigger_input_adc progress

andreika-git:
https://github.com/rusefi/rusefi/issues/5283


mck1117:
have we done the real oscilloscope hardware test to check timing accuracy?

firmware/hw_layer/digital_input/trigger/trigger_input_adc.cpp
```diff
@@ -16,9 +16,7 @@
 #define DELTA_THRESHOLD_CNT_LOW (GPT_FREQ_FAST / GPT_PERIOD_FAST / 32)		// ~1/32 second?
 #define DELTA_THRESHOLD_CNT_HIGH (GPT_FREQ_FAST / GPT_PERIOD_FAST / 4)		// ~1/4 second?
 
-#if HAL_USE_ADC
```
rusefillc, line 19: please do not remove ADC guard in firmware code?

Please rebase on master now that it has https://github.com/rusefi/rusefi/pull/5299/files - the idea is to be green without ADC.

this line here looks like it should be #if HAL_USE_ADC || EFI_UNIT_TEST unless unit test defines EFI_UNIT_TEST ?

overall when having unit tests considerations always safer to add " || EFI_UNIT_TEST " than to remove guards

PR #5334, opened 2023-06-17
Add Transmission range input

chuckwagoncomputing:
I haven't PR'd this until now because I'm not terribly pleased with how I ended up doing the TS UI, but I can't think of a better way.

image

Being able to hide the graph by default would be an improvement, but I don't think it's possible.

I also put the transmission settings back in the menu with an "(alpha)" label. If this isn't acceptable, I'm open to suggestions.


rusefillc:
@chuckwagoncomputing simulator failure is related to my recent GHA changes

I was blinding copy-pasting from gen-configs.yaml did I not paste enought?

PR #5336, opened 2023-06-17
Add check for push to master for simulator workflow

chuckwagoncomputing:
The gen-config workflow only runs on push to master, which is why it doesn't cause errors.


rusefillc:
@chuckwagoncomputing looks like original was always disabling the commit. I've attempted re-using snippets of code to fix (see too much history on the job yaml) but still struggling see https://github.com/rusefi/rusefi/pull/5343

PR #5393, opened 2023-07-06
Chibios 21 nice

rusefillc:
Here we have the cleanest commits
F4 compiles but does not actually run

See also https://github.com/rusefi/rusefi/pull/5394


rusefillc:
@dron0gus @mck1117 what is proper way around https://github.com/rusefi/ChibiOS/commit/d30a6924ef8b262e0375fcc383e78664e223fb19? What is the problem statement / shall we PR something into upstream ChibiOS?

rusefillc:
@dron0gus @mck1117 FYI - looks like rusEFI mostly compiles with ChibiOS 21 now but firmware does not start. Have not yet picked most of rusEFI changes in ChibiOS but I would expect firmware at least to start without those?

rusefillc:
Some write-up https://github.com/FOME-Tech/ChibiOS/pull/6

Issue #5354, opened 2023-06-25
proteus F7 hard fault on "enable self stimulation"


rusefillc:
* take F7 snapshot
* open console, hit "enable self stimulation"
* result: RED LED :(

on next boot
```
2023-06-25_01_33_34_046: EngineState: Last boot had hard fault type: 3 addr: 0 CSFR: 10000
2023-06-25_01_33_34_047: EngineState: Fault ctx 0: 29B55C0
2023-06-25_01_33_34_048: EngineState: Fault ctx 1: 0
2023-06-25_01_33_34_049: EngineState: Fault ctx 2: 29B55C0
2023-06-25_01_33_34_050: EngineState: Fault ctx 3: 200303B0
2023-06-25_01_33_34_051: EngineState: Fault ctx 4: 0
2023-06-25_01_33_34_052: EngineState: Fault ctx 5: 255F2D
2023-06-25_01_33_34_053: EngineState: Fault ctx 6: 255F2C
2023-06-25_01_33_34_054: EngineState: Fault ctx 7: 41000042
2023-06-25_01_33_34_055: EngineState: Fault ctx 8: 221401
2023-06-25_01_33_34_055: EngineState: Fault ctx 9: 200290B4
2023-06-25_01_33_34_056: EngineState: Fault ctx 10: 20021690
2023-06-25_01_33_34_057: EngineState: Fault ctx 11: 20029408
2023-06-25_01_33_34_057: EngineState: Fault ctx 12: 200306A8
2023-06-25_01_33_34_058: EngineState: Fault ctx 13: E0000000
2023-06-25_01_33_34_059: EngineState: Fault ctx 14: 0
2023-06-25_01_33_34_059: EngineState: Fault ctx 15: 20020EF0
2023-06-25_01_33_34_060: EngineState: Fault ctx 16: 200306A8
2023-06-25_01_33_34_061: EngineState: Fault ctx 17: E0000000
2023-06-25_01_33_34_062: EngineState: Fault ctx 18: 0
2023-06-25_01_33_34_062: EngineState: Fault ctx 19: 206D5B
2023-06-25_01_33_34_063: EngineState: Fault ctx 20: 20029F00
2023-06-25_01_33_34_064: EngineState: Fault ctx 21: FFFFFFE1
2023-06-25_01_33_34_064: EngineState: Fault ctx 22: 0
2023-06-25_01_33_34_065: EngineState: Fault ctx 23: 24F851
[...]

PR #5454, opened 2023-07-24
proteus F7 hard fault on "enable self stimulation" #5354

rusefillc:
only:coverage, that's easy!


rusefillc:
@mck1117 great news the issue is reproducible with HW CI and it's somewhere around composite logger which console uses mostly same as TS right?

PR #5455, opened 2023-07-25
Random6 local O0 does not help

rusefillc:


firmware/console/binary/tooth_logger.cpp
```diff
@@ -10,6 +10,9 @@
 
 #include "pch.h"
 
+#pragma GCC push_options
+#pragma GCC optimize ("O0")
```
mck1117, line 14: this likely won't work with LTO turned on

PR #5468, opened 2023-07-25
bigBuffer should be aligned

dron0gus:
Fixes https://github.com/rusefi/rusefi/issues/5354


mck1117:
what's the bug, and why does it only fail on some configurations?

rusefillc:
this PR here has two commits

Only one line is the bugfix - https://github.com/rusefi/rusefi/commit/1e43299eab2b2af033cb79d03446dbf07e4bd2c3

HW CI agrees that this line changed the behavior https://github.com/rusefi/rusefi/commit/7fb3f31c61586f3e0f6118abb3b694cffda02989

mck1117:
so what was the bug?

PR #5477, opened 2023-08-01
Update 8chan-C.yaml

ElDominio:
updated pins for new functions


rusefillc:
@ElDominio notice red due to "ID used multiple times with different ts_name: F14"

Issue #5437, opened 2023-07-23
CAN bench test: broadcast raw analog values


rusefillc:
Every 100ms we should send following seven one byte raw analog values on BENCH_TEST_RAW_ANALOG extended CAN ID:
* tps1_1
* tps1_2
* pps1
* pps2
* map
* clt
* iat

andreika-git:
+ Add vbatt raw value

PR #5512, opened 2023-08-14
CAN bench test: broadcast raw analog values #5437

andreika-git:


firmware/controllers/can/can_bench_test.cpp
```diff
@@ -48,6 +51,28 @@ void sendEventCounters() {
...
+		engine->outputChannels.rawMap,
+		engine->outputChannels.rawClt,
+		engine->outputChannels.rawIat,
+		engine->outputChannels.rawBattery,
+	};
+
+	// send the first packet
+	{
+		CanTxMessage msg(CanCategory::BENCH_TEST, BENCH_TEST_RAW_ANALOG, 8);
+		for (int valueIdx = 0; valueIdx < 8; valueIdx++) {
```
mck1117, line 69: static assert that `values` has a reasonable size?

PR #5628, opened 2023-10-22
Changed the way AC idle RPM bump is handled

pchmura4:
Changed the way AC idle RPM bump is handled. Before it was always adding fixed RPM value.
In my example, the A/C compressor is happy with 900 RPM, fully warmed-up engine idles at 600, so I'm adding 300 RPM to have 900 RPM.
But... before warming up I'm Idling at more than 1000 RPM, so adding 300 RPM results with 1300 RPM idle that is not necessary...
This commit changes that, so the value is actually not an adder, but a target RPM for the A/C compressor so it works efficiently behaving like OEM system.


firmware/integration/rusefi_config.txt
```diff
@@ -1060,7 +1060,7 @@ bit tempBooleanForVerySpecialLogic
 	float turboSpeedSensorMultiplier;;"mult", 1, 0, 0, 7000, 3
 
 	Gpio[CAM_INPUTS_COUNT iterate] camInputsDebug;
-	int16_t acIdleRpmBump;Extra idle target speed when A/C is enabled. Some cars need the extra speed to keep the AC efficient while idling.;"RPM", 1, 0, 0, 1000, 0
+	int16_t acIdleRpmBump;Idle target speed when A/C is enabled. Some cars need the extra speed to keep the AC efficient while idling.;"RPM", 1, 0, 0, 2000, 0
```
rusefillc, line 1063: "bump" in the method name is no longer appropriate

while renaming variable hopefully we would find not-zero default value which would need to be updated

rusefillc (changes requested):

rusefillc:
Please mention the change in CHANGELOG (as a breaking change I guess?)

ElDominio:
This works by guaranteeing a minimum RPM when AC is on, but not changing the RPM if it is over the guaranteed minimum correct? So in cold idle, you don't get extra RPMs you don't need? That seems good

firmware/integration/rusefi_config.txt
pchmura4: Hi, I have changed the names, default value and CHANGELOG, just not sure how to add this to this pull request 😅
https://github.com/pchmura4/rusefi/commit/5a948f2ae6bfcdb6730518ff937599c21118e1ea

pchmura4:
> This works by guaranteeing a minimum RPM when AC is on, but not changing the RPM if it is over the guaranteed minimum correct? So in cold idle, you don't get extra RPMs you don't need? That seems good

Correct, probably the idle % adder would need to be a curve, but in my case I dont' use it at all, as idle PID does it's job perfectly

rusefillc:
@pchmura4 note compilation errors

pchmura4:
now it builds, checked on my PC

rusefillc:
Let me attempt https://github.com/rusefi/rusefi/pull/5629 since this PR definitely does not need to change 73 files

pchmura4:
Sure, thanks

rusefillc:
Fixed by https://github.com/rusefi/rusefi/pull/5629

rusefillc:
Thank you for the contribution!

@pchmura4 do you have a build thread on the forum?

pchmura4:
I will create one when I'll finish my project 😄

PR #5632, opened 2023-10-23
After-start enrichment handling more similar to OEM

pchmura4:
My LPG-converted engine was not happy with fixed ASE for all temperatures.
That is because while cold it needs lots of fuel after start, but when the LPG evaporator is hot, and I'm using the hot-start feature it's giving too rich mixture.
LPG doesn't get along too well with overly rich mixtures (and cylinder head really hates this!) like petrol does.
I have implemented this as 3D table like in OEM and other Standalone EFI systems do.
Ideally the axis for duration should not be in seconds, but engine cycles... I'm still getting to know how rusEFI works...

### DIFF
--- a/firmware/config/boards/hellen/hellen121nissan/board_configuration.cpp
+++ b/firmware/config/boards/hellen/hellen121nissan/board_configuration.cpp
@@ -139,16 +139,6 @@ void setBoardDefaultConfiguration() {
 	setPPSCalibration(0.75, 4.45, 0.43, 2.20);
 
 	engineConfiguration->startUpFuelPumpDuration = 4;
-	
-	static const float defaultPostCrankingCLTBins[] = {
-		-20.0f, 0.0f, 30.0f, 60.0f
-	};
-	static const uint16_t defaultPostCrankinDurationBins[] = {
-		0, 2, 4, 6, 8, 10, 12, 15
-	};
-	copyArray(engineConfiguration->postCrankingCLTBins, defaultPostCrankingCLTBins);
-	copyArray(engineConfiguration->postCrankingDurationBins, defaultPostCrankinDurationBins);
-	setTable(engineConfiguration->postCrankingFactor, 1.0f);
 
     setEtbPID(6.1350, 87.7182, 0.0702);
 
--- a/firmware/config/boards/hellen/hellen121nissan/board_configuration.cpp
+++ b/firmware/config/boards/hellen/hellen121nissan/board_configuration.cpp
@@ -139,7 +139,16 @@ void setBoardDefaultConfiguration() {
 	setPPSCalibration(0.75, 4.45, 0.43, 2.20);
 
 	engineConfiguration->startUpFuelPumpDuration = 4;
-	engineConfiguration->postCrankingFactor = 1.05;
+	
+	static const float defaultPostCrankingCLTBins[] = {
+		-20.0f, 0.0f, 30.0f, 60.0f
+	};
+	static const uint16_t defaultPostCrankinDurationBins[] = {
+		0, 2, 4, 6, 8, 10, 12, 15
+	};
+	copyArray(engineConfiguration->postCrankingCLTBins, defaultPostCrankingCLTBins);
+	copyArray(engineConfiguration->postCrankingDurationBins, defaultPostCrankinDurationBins);
+	setTable(engineConfiguration->postCrankingFactor, 1.0f);
 
     setEtbPID(6.1350, 87.7182, 0.0702);
 
--- a/firmware/config/engines/custom_engine.cpp
+++ b/firmware/config/engines/custom_engine.cpp
@@ -889,17 +889,6 @@ void detectBoardType() {
 void fuelBenchMode() {
     engineConfiguration->cranking.rpm = 12000;
     setFlatInjectorLag(0);
-
-	static const float defaultPostCrankingCLTBins[] = {
-		-20.0f, 0.0f, 30.0f, 60.0f
-	};
-	static const uint16_t defaultPostCrankinDurationBins[] = {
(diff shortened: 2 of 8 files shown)


firmware/config/engines/custom_engine.cpp
```diff
@@ -889,7 +889,17 @@ void detectBoardType() {
 void fuelBenchMode() {
     engineConfiguration->cranking.rpm = 12000;
     setFlatInjectorLag(0);
-    engineConfiguration->postCrankingFactor = 1;
+
+	static const float defaultPostCrankingCLTBins[] = {
```
rusefillc, line 893: why same (or not same?) code three times? why in such odd places?

rusefillc (changes requested):

firmware/config/engines/custom_engine.cpp
pchmura4: so it should be just in firmware/config/engines/custom_engine.cpp, correct?

rusefillc: * should never be copy-paste
* custom_engine is some specific configuration, default_cranking.cpp seems most reasonable

pchmura4: done!

firmware/config/engines/honda_k_dbc.cpp
```diff
@@ -42,8 +42,7 @@ void setHondaK() {
 	engineConfiguration->injectionMode = IM_SEQUENTIAL;
 
 	engineConfiguration->crankingIACposition = 70;
-    engineConfiguration->postCrankingFactor = 1.25;
```
rusefillc, line 45: would setTable(engineConfiguration->postCrankingFactor, 1.25f); be a more compatible change here?

firmware/controllers/algo/defaults/default_cranking.cpp
```diff
@@ -23,11 +23,17 @@ void setDefaultCranking() {
 	// After start enrichment
 #if !EFI_UNIT_TEST
 	// don't set this for unit tests, as it makes things more complicated to test
-	engineConfiguration->postCrankingFactor = 1.2;
```
rusefillc, line 26: setTable(engineConfiguration->postCrankingFactor, 1.2f); for more compatible change?

firmware/config/boards/hellen/hellen121nissan/board_configuration.cpp
```diff
@@ -139,7 +139,6 @@ void setBoardDefaultConfiguration() {
 	setPPSCalibration(0.75, 4.45, 0.43, 2.20);
 
 	engineConfiguration->startUpFuelPumpDuration = 4;
-	engineConfiguration->postCrankingFactor = 1.05;
```
rusefillc, line 142: setTable(engineConfiguration->postCrankingFactor, 1.05f); ?

rusefillc (changes requested):
We are close please make defaults a bit more compatible?

firmware/config/engines/honda_k_dbc.cpp
pchmura4: done, that one was ending at different time, re-scaled bins for that

firmware/config/boards/hellen/hellen121nissan/board_configuration.cpp
pchmura4: done

firmware/config/engines/honda_k_dbc.cpp
```diff
@@ -42,8 +43,11 @@ void setHondaK() {
...
-    engineConfiguration->postCrankingDurationSec = 15;
+	static const uint16_t defaultPostCrankinDurationBins[] = {
```
rusefillc, line 46: say honda K overrides what we have in setDefaultCranking()

meaning you already have axis, you only need to override table

rusefillc (changes requested):

firmware/config/engines/honda_k_dbc.cpp
pchmura4: yes, the axis was changed, as Honda used here 15 instead of default 10 seconds

pchmura4: we can either have different Z values in table, or change axis to reproduce what was done before

firmware/controllers/algo/engine2.cpp
```diff
@@ -119,14 +119,17 @@ void EngineState::periodicFastCallback() {
...
+	);
 	// for compatibility reasons, apply only if the factor is greater than unity (only allow adding fuel)
-	if (engineConfiguration->postCrankingFactor > 1.0f) {
-		// use interpolation for correction taper
-		engine->fuelComputer.running.postCrankingFuelCorrection = interpolateClamped(0.0f, engineConfiguration->postCrankingFactor,
-			engineConfiguration->postCrankingDurationSec, 1.0f, engine->fuelComputer.running.timeSinceCrankingInSecs);
-	} else {
-		engine->fuelComputer.running.postCrankingFuelCorrection = 1.0f;
+	// if the engine run time is past the last duration bin, disable ASE in case the table is filled with values more than 1.0, helps with compatibility with older solutioin
+	if ((m_postCrankingFactor < 1.0f) || (engine->fuelComputer.running.timeSinceCrankingInSecs > engineConfiguration->postCrankingDurationBins[(sizeof(engineConfiguration->postCrankingDurationBins)/sizeof(engineConfiguration->postCrankingDurationBins[0]))-1])) {
```
rusefillc, line 129: efi::size() or CRANKING_ADVANCE_CURVE_SIZE to improve readability please, and let's merge it!

pchmura4: done!

PR #5638, opened 2023-10-25
ASE and CL Ignition fixes confirmed on the real car

pchmura4:
Thanks to the last tip, I was able to get it to build myself to test on my Lexus
Idle advance is entering slowly now upon starting the car. That can be changed to work always when entering idle by setting m_crankTaperEndTime to 0 below m_timingPid.reset();
For me I did not see reason to do it.

obraz

I have fixed both previous feature commits, unfortunately I had to break the TunerStudio again.
Also, it looks to me like the interpolate3d function, or TunerStudio generation swaps X an Y axis, if the axes are of not equal lengths like it was 4x8 instead of 6x6, then the axes are swapped.

### DIFF
--- a/firmware/config/engines/honda_k_dbc.cpp
+++ b/firmware/config/engines/honda_k_dbc.cpp
@@ -44,7 +44,7 @@ void setHondaK() {
 
 	engineConfiguration->crankingIACposition = 70;
 	static const uint16_t defaultPostCrankinDurationBins[] = {
-		0, 35, 65, 100, 125, 150, 200, 250
+		0, 65, 100, 125, 150, 250
 	};
 	copyArray(engineConfiguration->postCrankingDurationBins, defaultPostCrankinDurationBins);
 	setTable(engineConfiguration->postCrankingFactor, 1.25f);
--- a/firmware/controllers/actuators/idle_thread.cpp
+++ b/firmware/controllers/actuators/idle_thread.cpp
@@ -76,6 +76,12 @@ IIdleController::Phase IdleController::determinePhase(int rpm, int targetRpm, Se
 	}
 #endif // EFI_SHAFT_POSITION_INPUT
 
+	// If we are entering idle, and the PID settings are aggressive, it's good to make a soft entry upon entering closed loop
+	if (m_crankTaperEndTime == 0.0f) {
+		m_crankTaperEndTime = engine->fuelComputer.running.timeSinceCrankingInSecs;
+		m_idleTimingSoftEntryEndTime = m_crankTaperEndTime + engineConfiguration->idleTimingSoftEntryTime;
+	}
+
 	// No other conditions met, we are idling!
 	return Phase::Idling;
 }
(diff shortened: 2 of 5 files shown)


rusefillc:
> looks to me like the interpolate3d function, or TunerStudio generation swaps X an Y axis, if the axes are of not equal lengths like it was 4x8 instead of 6x6, then the axes are swapped.

See good news there is one example of a not-square table https://github.com/rusefi/rusefi/commit/8f9363e2215f103999f03e0424e1c0e5f7030f1d

We can use kinetis, cypress, prometheus and atlas to try non-square tables effectively for everything. Do you have time and motivation to make progress in this area?

rusefillc:
@pchmura4 actually https://github.com/rusefi/rusefi/pull/5639

pchmura4:
I think I can give it a try

pchmura4:
I see there are already some changes merged?
For me it looks like the issue with Java tool that generates *.ini files
When I manually swapped the axis sizes in produced ini like below, it was working as expected.
If the axis were swapped in rusefi_config.txt before generating, then there was an issue with interpolate3d expecting the axes to be swapped there as well with the issue still there

obraz

rusefillc:
Now I am officially confused could this be an indication of too many things mixed together?

Let's start from the beginning. Master has postCrankingFactor = array, F32, [4x8]. If anything is broken in master shall we create a separate ticket? Are there any relevant screen shots? Can we add a relevant unit test showing a problem?

rusefillc:
See also https://github.com/rusefi/rusefi/blob/master/unit_tests/test_basic_math/test_interpolation_3d.cpp

pchmura4:
Yes, current master still has the axis swapped because the resulting .ini is generated incorrectly. That is why in this commit I have changed it to square map to make it work correctly

pchmura4:
No matter what you do, the Interpolate3D template will not let you bind the axes in correct order (length mismatch)

rusefillc:
> No matter what you do, the Interpolate3D template will not let you bind the axes in correct order (length mismatch)

let's create a separate ticket. on the separate ticket please explain the defect in more details

pchmura4:
Will this pull request be merged to have the master functional for now?
If it comes to the issue we are discussing now, I would not touch interpolate3d as that would break too many things for now, the ini generation will probably need to be changed to swap X and Y bin count in square brackets

rusefillc:
@pchmura4 see https://rusefi.com/forum/viewtopic.php?f=5&t=2638

pchmura4:
Found a small bug during checks on simulator:
```
diff --git a/firmware/controllers/algo/engine2.cpp b/firmware/controllers/algo/engine2.cpp
index 412f36b040..aff972084f 100644
--- a/firmware/controllers/algo/engine2.cpp
+++ b/firmware/controllers/algo/engine2.cpp
@@ -126,7 +126,7 @@ void EngineState::periodicFastCallback() {
 	);
 	// for compatibility reasons, apply only if the factor is greater than unity (only allow adding fuel)
 	// if the engine run time is past the last bin, disable ASE in case the table is filled with values more than 1.0, helps with compatibility
-	if ((m_postCrankingFactor < 1.0f) || (engine->fuelComputer.running.timeSinceCrankingInSecs > engineConfiguration->postCrankingDurationBins[efi::size(engineConfiguration->postCrankingDurationBins)-1])) {
+	if ((m_postCrankingFactor < 1.0f) || (engine->rpmCalculator.getRevolutionCounterSinceStart() > engineConfiguration->postCrankingDurationBins[efi::size(engineConfiguration->postCrankingDurationBins)-1])) {
 		m_postCrankingFactor = 1.0f;
 	}
 	engine->fuelComputer.running.postCrankingFuelCorrection = m_postCrankingFactor;
```

rusefillc:
@pchmura4 you are putting me into a horrible situation when you report issues on closed tickets with out of date patches.

mck1117:
> bug during checks on simulator

@rusefillc this is on you for merging with no tests

pchmura4:
I'm sorry, that line somehow didn't make it into pull request...

rusefillc:
> I'm sorry, that line somehow didn't make it into pull request...

I think we are mixing two separate issues here.
* it's expected and acceptable to have defects
* it's not expected and is much less acceptable to be raising new issues on closed tickets/PRs

pchmura4:
Got it, noted

PR #5643, opened 2023-10-26
require LEDs

rusefillc:
Do we or do we not like weak?


dron0gus:
I don't like weak functions. :)

dron0gus (approved):

Issue #5644, opened 2023-10-26
simulator does not compile for pchmura4 on gcc13


rusefillc:
Do we need something similar to [gcc_version_check.c](https://github.com/rusefi/rusefi/blob/master/firmware/gcc_version_check.c) in simulator?

![image](https://github.com/rusefi/rusefi/assets/48498823/f2d4150d-351f-4368-958a-a5f471935a09)

rusefillc:
@pchmura4 can you please ``gcc -v``?

here is my happy debian-based mxlinux


```
Andrey@nuc1:~/rusefi/simulator
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/10/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa:hsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
[...]

PR #5645, opened 2023-10-26
simulator does not compile for pchmura4 on gcc13 #5644

rusefillc:
https://github.com/rusefi/rusefi/issues/5644


rusefillc:
@mck1117 what's your take on all this?

mck1117:
does this change the contents of the firmware binary?

PR #5662, opened 2023-10-29
Artery AT_START_F435 board support

dron0gus:


firmware/controllers/core/error_handling.cpp
```diff
@@ -29,6 +29,8 @@ const char* getCriticalErrorMessage(void) {
 void checkLastBootError() {
 #if EFI_BACKUP_SRAM
 	auto sramState = getBackupSram();
+	if (sramState == NULL)
```
rusefillc, line 32: why EFI_BACKUP_SRAM if on backup RAM? i kind of what to do in the opposite direction remove null check in other places

PR #5683, opened 2023-11-02
unused-variable is a nice error message to have

rusefillc:


rusefillc:
@andreika-git everything is happy with exception of kinetis. Let's make this conditional and disable for kinetic only?

PR #5699, opened 2023-11-07
Refactor AlternatorController to use ClosedLoopController<>

andreika-git:


rusefillc:
the onConfigurationChange change is technically separate from ClosedLoopController change is it?

andreika-git:
A copy of this PR will be created due to force-push + no checks problems.

PR #5725, opened 2023-11-14
Add a command to test a HardFault

andreika-git:


rusefillc:
@andreika-git can we reduce copy/paste make it commonArmHardFault?

PR #5732, opened 2023-11-17
was: everyone ignores me now: need to restore unused_ram guard

rusefillc:


rusefillc:
See also https://github.com/rusefi/rusefi/pull/5772

PR #5751, opened 2023-11-24
Watchdog impl.

andreika-git:


rusefillc:
@andreika-git shall we mark not draft? Looks like a safe step forward?

firmware/hw_layer/ports/stm32/stm32_common.cpp
```diff
@@ -103,11 +103,44 @@ BOR_Result_t BOR_Set(BOR_Level_t BORValue) {
...
+
+static efitimems_t watchdogResetPeriodMs = 0;
+
+void setWatchdogResetPeriod(int resetMs) {
+	watchdogResetPeriodMs = (efitimems_t)resetMs;
+}
+
+void tryResetWatchdog() {
+#if HAL_USE_WDG
+	static efitimems_t lastTimeWasResetMs = 0;
```
rusefillc, line 124: "Timer" helper class would make this code simpler

PR #5802, opened 2023-12-16
Benelli_Walbro connector mapping improvements

mi-hol:


firmware/config/boards/Benelli_Walbro/connectors/Benelli_Walbro.yaml
```diff
@@ -230,34 +233,38 @@ pins:
...
-    class: analog_inputs
+    function: 'side stand Sensor Input  (grounded or open circuit)'
+    class: input_ON_OFF
     color: white
 
   - pin: F7
     function: 'Injection warning light (MIL)'
-    class: analog_inputs
-    type: ??
+    class: analog_outputs
```
rusefillc, line 266: please clarify what is ``analog_outputs``?

see https://github.com/rusefi/rusefi/blob/master/.github/workflows/gen-pinouts.yaml for all official types. MIL is not one of those types. Happy to merge as is just need your awareness.

mi-hol: Sorry, can't seem to find "official types" in the referenced file :( 
Had already created https://github.com/chuckwagoncomputing/interactive-pinout/issues/16 to have a documented list for reference.
Any hint, how I would code MIL correctly?

rusefillc: @mi-hol please note lines 23 and below

mi-hol: Sorry, but even with this hint, I fail to see the list of valid "class" values to fix "class: analog_outputs".

1. "class" is not mentioned in
`        columns: |
          {
          "pin":"Pin Number",
          "ts_name":"TS Name",
          "type":"Type",
          "function":"Typical Function",
          "color":"Pigtail Color"
          }
`
Therefore to fix "class: analog_outputs"  I'm still unclear about "how to code MIL correctly"

2. the list of valid "type" values seems to start at line 44 but is hard to digest for an inexperienced user due to the preceeding "color" label. Maybe a simple comment would avoid this hurdle for now?

image

PR #5823, opened 2023-12-20
mc33810: moving init to thread in order to make incremental progress of at least retying until first success

rusefillc:


dron0gus:
I don't like this.
HW should be ready before we start to use it. Trying to do delayed initialization of HW is way to unpredictable errors and magic smoke.
This solves you current problem (which is not a problem) with ECU powered from USB.
Does this helps if ECU is powered from switchable 12V through ignition key and laptop through USB? What will happen is user switches ign off and then on?
Check how reset detection and re-initialization is done in TLE8888.

dron0gus (changes requested):
We should expect chip to be reset at runtime.
We need to detect this.
We need to reinit on each reset. Not only once.

PR #5831, opened 2023-12-28
Add ChibiOS-compatible Makefile

andreika-git:


rusefillc:
better version was merged already

PR #5840, opened 2023-12-30
Board c

dron0gus:
Allow F7 boards to have its own custom board.c file


rusefillc:
looks like @andreika-git did a similar fix two minutes earlier?

PR #5890, opened 2024-01-28
gpio diagnostic: use PIN_UNKNOWN for pins with no diagsotic

dron0gus:


rusefillc:
@dron0gus this PR needs some minor help

PR #5938, opened 2024-02-13
Makeover - build bundle with Make

chuckwagoncomputing:
turn convert_<build target>.sh scripts into a script that only exports variables, including SHORT_BOARD_NAME, which is removed from board.mk files.

Add a makefile target that creates the bundle zips, and remove the old shell scripts

Ignore invalid PCH warnings in the bootloader Makefile as @racer-coder did in the firmware Makefile in https://github.com/rusefi/rusefi/commit/37da55c9206328bfba2a3bc24ef26d0ad419d1ab

Revert meta-info.env stuff


chuckwagoncomputing:
Closing this for now so as to not clog your CI pipes while I make some changes.

PR #5941, opened 2024-02-14
Makeover - build bundle with Make (#2)

chuckwagoncomputing:
turn `convert_<board>.sh` scripts into a script that only exports variables, including `SHORT_BOARD_NAME`, which is removed from board.mk files.

Add a makefile target that creates the bundle zips, and remove the old shell scripts

Ignore invalid PCH warnings in the bootloader Makefile as @racer-coder did in the firmware Makefile in https://github.com/rusefi/rusefi/commit/37da55c9206328bfba2a3bc24ef26d0ad419d1ab

Revert meta-info.env stuff

A custom firmware build needs only to set `SHORT_BOARD_NAME` in compile_firmware.sh. If `BUNDLE_NAME` or `PROJECT_BOARD` are unset, they will be set to `SHORT_BOARD_NAME`.
If a custom firmware builds has several variants, each will need to pass a bundle name to the custom-build-board workflow, and have a corresponding `compile_<bundle name>.sh` script


chuckwagoncomputing:
One small issue I have is that the `compile_*.sh` scripts don't compile the firmware, just set some variables, and this could be confusing to a newcomer. Perhaps they should be renamed to something like `configure_*.sh` or `setenv_*.sh` or `prepare_*.sh`.

mck1117:
> One small issue I have is that the `compile_*.sh` scripts don't compile the firmware, just set some variables, and this could be confusing to a newcomer. Perhaps they should be renamed to something like `configure_*.sh` or `setenv_*.sh` or `prepare_*.sh`.

What's the point of not compiling from the script? This doesn't quite fix the issue of having all the "what board am I" information duplicated between the GHA yaml, shell scripts, and mk files. It [seems less unexpected](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) to have a script that handles the build (perhaps with no arguments) rather than one that just sets some environment variables, then requires running `make` anyway.

rusefillc:
The scipts are now a combination of "metainfo but also let's export into environment". I am curious about one universal script to export meta-info into bash environment to have best of both worlds

mck1117:
> The scipts are now a combination of "metainfo but also let's export into environment". I am curious about one universal script to export meta-info into bash environment to have best of both worlds

My point is to go the other way. Never export anything. Don't pollute the environment if you don't have to. No modern build system or project does this.

chuckwagoncomputing:
@mck1117 You have a point.
My reasoning is that there are other things that need those variables, in this case the ini upload script. Those things can source the script to get the info they need.

Some of the choices in this PR are slightly weird because they are part of a larger effort. This is an example of that; just looking at this PR you might think that the short-board-name from the matrix should be passed to the ini upload script. However, in the interest of data-deduplication, that value will be removed from the matrix in the future.
My justification for that design decision is that as little information as possible should be in the master list of build targets (currently the matrix); information should instead be stored locally to the board directory.

@rusefillc has pointed out that the meta-info.env format is better for the human side of things, and suggested that it can be parsed into environment variables. My only disagreement with this is that I like the power and flexibility that having it be a shell script provides, but I can't think of any places that would be useful right now, and it may just be an invitation for more complexity.

rusefillc:
@chuckwagoncomputing let's go meta-info text file with environment variables for now as a way to get the exciting parts i.e. everything via make going? 

Overall looks like we sre all pretty aligned on the big picture.

chuckwagoncomputing:
OK, closing for now. Thanks for the feedback.

PR #5943, opened 2024-02-15
name meta-info.env files according to build target

chuckwagoncomputing:
This only applies to board directories with more than one `compile_<build target>.sh` script in them. Boards with only one script use the default `meta-info.env`.


rusefillc:
@chuckwagoncomputing merged under https://github.com/rusefi/rusefi/pull/5945 with https://github.com/rusefi/rusefi/pull/5945/commits/a5d66e47a94308b73231940479a9d2985dc1b910 to keep things green

PR #5944, opened 2024-02-15
Update rusefi_f407-discovery.ini with 3D maps

zed65:
Added dedicated 3D maps

### DIFF
--- a/firmware/tunerstudio/generated/rusefi_f407-discovery.ini
+++ b/firmware/tunerstudio/generated/rusefi_f407-discovery.ini
@@ -5601,8 +5601,11 @@ menuDialog = main
 		groupChildMenu = fuelTrimTbl12,		"Fuel trim cyl 12"
 
 		subMenu = std_separator
+ 
+    subMenu = veTableMap,    "Fuel Table", 0, {isInjectionEnabled}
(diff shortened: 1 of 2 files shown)


zed65:
3dTuning

rusefillc:
@zed65 ohh you mean all the way top level maps. I am hesitant to add, I am not convinced that it's worth the cost of screen "real estate". Somewhere on a narrow screen that new button would be a major problem for someone

@ElDominio do you have an opinion?

zed65:
I understand, but what if we add them under "Fuel", my point is to get gauges and histogram to the left of the map

rusefillc:
> my point is to get gauges and histogram to the left of the map

that I would happily merge! Still a bit confused what does that have to do with 3D since 3D and flat table seems to be separate unrelated switch? anyway please create new or update this PR?

ElDominio:
But Tunerstudio generates 3d maps if you click the 3d map checkbox?

zed65:
Maybe I'm missing something but I'm used to be able to maximize the tune window and still have your favorite gauges there
Anyway I think I made the changes in this PR (I'm totally new to this fork and PR thing :))
3dTuning4

PR #5949, opened 2024-02-15
pull any variable from meta-info

chuckwagoncomputing:
I'm not sure where the PROJECT_CPU -> BOARD_CPU switch happened (was it intentional)?, but I reverted that.
Also, I moved the PROJECT_BOARD = SHORT_BOARD_NAME assignment to a check in the Makefile


rusefillc:
That's more than one change and I am not comfortable that I cannot review effectively since massive PR size requires massive review effort to distinguish between functional changes and automatic changes.

I am hesitant about massive renames which do not seem to provide value I can comprehend. @chuckwagoncomputing what's the value in PROJECT_CPU -> BOARD_CPU rename? if you want to touch it please provide more details on when was this different?

overall please have mercy on me.

chuckwagoncomputing:
The PROJECT_CPU -> BOARD_CPU rename happened in commit https://github.com/rusefi/rusefi/issues/5931
I have no idea why or what it's value is.
Rather than changing it everywhere else (makefiles), I reverted it.

chuckwagoncomputing:
This is a tiny commit. I could kinda break it up more, but barely.

chuckwagoncomputing:
Are you going to merge or do you want changes?

rusefillc:
cannot think and merge right now, in the middle of something else

rusefillc:
@chuckwagoncomputing merged!

mck1117:
wow it only took 3 minutes to get out of "the middle of something else"

chuckwagoncomputing:
don't stir the pot Matthew

PR #5953, opened 2024-02-16
Import board-specific meta-info

chuckwagoncomputing:
Don't call the compile script any more, we can just call common_script.sh directly. This gets rid of most of compile.sh.

Arguments can be removed from calls to scripts in build-firmware.yaml, as they can get those values from the environment variables. Also included are changes in upload_bundle.sh to use the environtment variables instead of the passed arguments.
There's no reason this has to be done in this commit, but I'm out of Github Actions minutes for today and am depending on my own self-hosted runners, which makes splitting up commits take way longer, so I left them in this commit.

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -532,18 +532,11 @@ jobs:
       if: ${{ env.skip != 'true' }}
       working-directory: ./firmware/
       run: |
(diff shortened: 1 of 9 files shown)


chuckwagoncomputing:
I always forget about HW-CI because I have it disabled

chuckwagoncomputing:
oh good grief it's an amazing typo `find_meta-info.sh`

PR #5973, opened 2024-02-17
Call make directly from workflow

chuckwagoncomputing:
Deleted unused common_script.sh
Add logic in Makefile to set PROJECT_BOARD to SHORT_BOARD_NAME if it is empty (this was in common_script.sh)


chuckwagoncomputing:
Note that fw-custom repos will need to be updated to not call common_script.sh

rusefillc:
Note that @chuckwagoncomputing has write access to https://github.com/rusefi/fw-custom-example/ and https://github.com/rusefi/fw-custom-core8/

chuckwagoncomputing:
I think this is red because of your stuff. I guess re-run the CI when you get that sorted.

PR #5974, opened 2024-02-17
Fix paralela https://github.com/rusefi/fw-custom-paralela/issues/2

andreika-git:


rusefillc:
@chuckwagoncomputing FYI

Issue #5983, opened 2024-02-18
revive ptrace


rusefillc:
we have https://github.com/rusefi/rusefi/wiki/Developer-Performance-Tracing

on master on Proteus F7 I am failing to get any data in the json file. Need to either improve documentation or revive the feature or else

See https://github.com/rusefi/rusefi/commit/0c8785d7d3518f27349d858670de337676981589

rusefillc:
@mck1117 stuff seems to be broken between perf_trace.cpp and BigBuffer

andreika-git:
The BigBuffer is occupied with ToothLogger.

PR #6029, opened 2024-02-22
Fix ptrace "empty buffer" error and bigbuffer

andreika-git:
https://github.com/rusefi/rusefi/issues/5983

Tested on Proteus-F7.

### DIFF
--- a/firmware/controllers/core/big_buffer.cpp
+++ b/firmware/controllers/core/big_buffer.cpp
@@ -32,13 +32,17 @@ BigBufferHandle::BigBufferHandle(BigBufferHandle&& other) {
 }
 
 BigBufferHandle& BigBufferHandle::operator= (BigBufferHandle&& other) {
-	// swap contents of the two objects
-	m_bufferPtr = other.m_bufferPtr;
-	other.m_bufferPtr = nullptr;
+	if (this != &other) {
+		if (m_bufferPtr) {
+			releaseBuffer(m_bufferPtr, m_user);
+		}
+		// swap contents of the two objects
+		m_bufferPtr = other.m_bufferPtr;
+		other.m_bufferPtr = nullptr;
 
-	m_user = other.m_user;
-	other.m_user = BigBufferUser::None;
-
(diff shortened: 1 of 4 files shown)


rusefillc:
Already in master thank you @andreika-git !

firmware/controllers/core/big_buffer.cpp
```diff
@@ -32,13 +32,17 @@ BigBufferHandle::BigBufferHandle(BigBufferHandle&& other) {
...
 BigBufferHandle& BigBufferHandle::operator= (BigBufferHandle&& other) {
-	// swap contents of the two objects
-	m_bufferPtr = other.m_bufferPtr;
-	other.m_bufferPtr = nullptr;
-
-	m_user = other.m_user;
-	other.m_user = BigBufferUser::None;
+	if (this != &other) {
+		if (m_bufferPtr) {
+			releaseBuffer(m_bufferPtr, m_user);
```
mck1117, line 37: You don't need this part of the change.

This operator is for the `rvalue assignment operator`, which means that the right side is an immediate that's about to poof out of existence. When it does, it'll call the destructor on the instance you're about to lose a reference to, which will clean up the buffer if it's holding it.

mck1117: ...it's also a code smell for the move constructor and move assignment operator to be different

rusefillc: > You don't need this part of the change.

I read https://github.com/rusefi/rusefi/commit/a0c61a0ebc5d18bf627d512977ccf7fff3b1ab54 + https://github.com/rusefi/rusefi/commit/701f3a073c3b3e1c2eaa217d88efcb4186e2aae3 differently

mck1117: haha the actual bug is that the comment says one thing, and the code did something else!

https://github.com/FOME-Tech/fome-fw/commit/5e1359c9192c607285ec514f63fa94e0636a9c28

PR #6031, opened 2024-02-23
Build console with make bundle

chuckwagoncomputing:
`flock` waits for a lock on a file; once it's unlocked, it locks it and holds the lock until the command is completed. This is necessary because parallel Gradle builds interfere with each other.

Green custom: https://github.com/chuckwagoncomputing/fw-custom-example/actions/runs/8020089442


rusefillc:
@chuckwagoncomputing do we expect this to be happy on Windows?

rusefillc:
> This is necessary because parallel Gradle builds interfere with each other.

shall we disable parallel gradle? can you elaborate on the problem statement what is interfering with what how and when?

chuckwagoncomputing:
When you try to run a Gradle task when another one is already running, they fail in all sorts of ways. Making each Gradle worker do its work sequentially wouldn't solve the problem because you would still have multiple Gradle workers running at the same time. I have solved this by using flock to make sure only one is running at any time.

rusefillc:
@chuckwagoncomputing please ELI5 for me. Who is invoking multiple gradle instances in parallel?

rusefillc:
So a variation of https://discuss.gradle.org/t/avoiding-gradle-locking-issues-with-concurrent-gradle-task-runs/30259

mck1117:
> hopefully gradle has _something_ official remedy for that?

The official guidance is "please for the love of god don't call `gradle` from `make`"

rusefillc:
@chuckwagoncomputing FYI pick https://github.com/rusefi/rusefi/commit/2e6f2189f89130f39b1579bb18c8745f4a2dd90b

chuckwagoncomputing:
> do we expect this to be happy on Windows?

I expect it to work with Cygwin or WSL, if that's what you mean.


> The official guidance is "please for the love of god don't call `gradle` from `make`"

@mck1117 Do you have a source, or can you explain why not?

PR #6039, opened 2024-02-24
Update can_dash.cpp to match the Haltech standard

zed65:
https://support.haltech.com/portal/en/kb/articles/haltech-can-ecu-broadcast-protocol


zed65:
The Haltech dash output should only send Lambda not AFR

PR #6047, opened 2024-02-25
Call gen_config_board with make

chuckwagoncomputing:
Instead of having hard-coded checks for calling custom gen_config scripts, I have added a variable to the meta-info of those boards that require it.

As of this PR, configs are generated every time make is run. This is not optimal, it's a small-PR stepping stone.

I had to add flock to the call to Gradle in gen_config_board.sh. Eventually, this call will be removed from this file, but it has to stay for now.
Because of this, unit tests on macOS require installing flock.  Flock is a BSD standard system call which Linux has also adopted, but macOS doesn't ship with a command line utility for using it. In linux, the command line tool is a part of the util-linux package. Util-linux is shipped with every linux distro and cygwin.

All the config-related stuff is in a new file, rusefi_config.mk. This allows us to claim the configs as prerequisites in different makefiles, e.g. the unit tests makefile.

There's one thing I don't really like here.

`$(ACOBJS): $(CONFIG_FILES)`

This is kinda following the example of rusefi_pch.mk

https://github.com/rusefi/rusefi/blob/44ebfc633f5c5d29b583b38a130f25ce64031cd6/firmware/rusefi_pch.mk#L20-L21

This is required because of the way the ChibiOS makefile is set up. The automatic prerequisite generation is done in the same step as the compilation. Therefore, if the .dep folder isn't already in the correct state, Make doesn't know which .o files require the config headers until it has already tried and failed to compile them. So to work around this, we just tell Make that all C .o files require the config headers. The downside of this is that if a config header changes, all C .o files will be rebuilt, not jut those that require the config headers. The lines in rusefi_pch.mk take care of most things that require the config headers, but there's an exception I ran into: init.c includes lwip/opts.h which includes console/lwipopts.h, which includes generated_lookup_meta.h. It would also work to make CONFIG_FILES a prerequisite of just init.o, and not ACOBJS, but who know what other edge cases are lurking.

Example of lwipopts failure: https://github.com/chuckwagoncomputing/rusefi/actions/runs/7942940928/job/21687118848

Green custom: https://github.com/chuckwagoncomputing/fw-custom-example/actions/runs/8035615243

### DIFF
--- a/.github/workflows/build-firmware-windows.yaml
+++ b/.github/workflows/build-firmware-windows.yaml
@@ -41,9 +41,9 @@ jobs:
       - name: Test Java Compiler
         run: javac -version
 
-      - name: Generate Configs, Enums & Live Documentation
+      - name: Generate Enums & Live Documentation
         working-directory: ./firmware/
-        run: bash gen_default_everything.sh
+        run: bash gen_live_documentation.sh
 
       - name: Print arm GCC version
         working-directory: .
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -513,30 +513,6 @@ jobs:
       working-directory: ./firmware/
       run: ./gen_live_documentation.sh
 
-    - name: Generate Configs for build-target
-      if: ${{ env.skip != 'true' }}
-      working-directory: ./firmware/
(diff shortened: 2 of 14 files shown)


chuckwagoncomputing:
I see  we have a `flock: command not found` on Windows, so I was wrong.

PR #6065, opened 2024-02-27
Include config makefile for bootloader

chuckwagoncomputing:
I was unable to reproduce the issue locally, but I'm pretty sure this will resolve it.


chuckwagoncomputing:
Argh I've got red

PR #6068, opened 2024-02-27
Update gen_signature.sh

chuckwagoncomputing:


rusefillc:
@chuckwagoncomputing please add _why_ in the code comment

Do I guess right that the whole conditional is to help incremental compilation? i.e. 
* GHA scenario: .txt does not exists
* local incremental: we do not update .txt unless we have to?

PR #6070, opened 2024-02-27
Enable SPI5 for Proteus?

pchmura4:
In the latest compilation only SPI1 up to SPI4 are visible in Tuner studio.
Not sure if the added define is in correct place


rusefillc:
@pchmura4 still need to add UI for pin selection

I agree with @mck1117 that pin MODES are on the way out, definitely do not add UI for pin MODES

rusefillc:
@pchmura4 please rebase or merge https://github.com/rusefi/rusefi/commit/c6d883cf5bc80009014930a7ee7594a90b512c4c

PR #6085, opened 2024-02-28
Move ramdisk creation to its own script and make target

chuckwagoncomputing:
Previous to this PR, ramdisk headers are always touched, even when no other config files are. Now the ramdisk headers will only be built if the .ini file was touched.
Currently the configs are always re-generated, which prior to this PR also caused most of the firmware to be rebuilt.
A future PR (hopefully next one) will only re-generate configs if one of their input files changed, but we still want to avoid rebuilding the firmware if one of those input files changed in a way that doesn't actually effect the output of config generation.

Green custom: https://github.com/chuckwagoncomputing/fw-custom-example/actions/runs/8087460266


chuckwagoncomputing:
Not working :(

PR #6093, opened 2024-02-29
always try to rebuild ramdisk, but check and save SHORT_BOARD_NAME

### LINKED ISSUE #6088: Ramdisk images don't rebuild when switching targets
Currently ramdisk images only rebuild if the .ini is newer than the images.
This isn't a problem for CI because the signature is always different.
It is a problem if for example I build mre, then proteus, the mre again.

I will fix this, but I haven't decided how yet, and I decided to create an issue so if someone else noticed, they wouldn't have to.

chuckwagoncomputing:
Fix #6088

Don't worry, comments are coming. I'm waiting until files are no longer in a state of flux so I don't have to waste so much time rebasing comments.

Basically we check if the .ini file was what triggered this rule, or if the contents of .ramdisk-sentinel don't match SHORT_BOARD_NAME, and if either of those things are true, rebuild the ramdisk images and write the SHORT_BOARD_NAME to .ramdisk-sentinel.

### DIFF
--- a/firmware/rusefi_config.mk
+++ b/firmware/rusefi_config.mk
@@ -39,9 +39,11 @@ $(SIG_FILE): .FORCE
 
 $(RAMDISK): .ramdisk-sentinel ;
 
-.ramdisk-sentinel: $(INI_FILE)
-	bash $(PROJECT_DIR)/bin/gen_image_board.sh $(BOARD_DIR) $(SHORT_BOARD_NAME)
-	@touch $@
+.ramdisk-sentinel: $(INI_FILE) .FORCE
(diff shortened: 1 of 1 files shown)


chuckwagoncomputing:
Don't merge yet, I think there's a problem.

chuckwagoncomputing:
OK should be good now if it's green

chuckwagoncomputing:
I'm having some second thoughts about this. I think I should make a separate .target-sentinel that can be used for both this and some other things like pch.h which is also broken in the same way. You can either merge this or not, I don't care.

rusefillc:
@chuckwagoncomputing please decide tomorrow?

PR #6096, opened 2024-02-29
Use file with short board name to trigger rebuild on target change

### LINKED ISSUE #6088: Ramdisk images don't rebuild when switching targets

chuckwagoncomputing:
Caveat: SHORT_BOARD_NAME doesn't cover all possible changes. For example, if I build Proteus F7, then build Proteus F7 Debug, it won't actually rebuild. BUNDLE_NAME would be a better specifier, but it's currently only available from the build_firmware GHA workflow. Another option would be to use BOARD_META_PATH, and export it in config.sh.
@rusefillc Let me know what you think.

Fix #6088
Green custom: https://github.com/chuckwagoncomputing/fw-custom-example/actions/runs/8102640677

### DIFF
--- a/firmware/rusefi_config.mk
+++ b/firmware/rusefi_config.mk
@@ -37,15 +37,19 @@ $(TCPPOBJS): $(RAMDISK)
 $(SIG_FILE): .FORCE
 	bash $(PROJECT_DIR)/gen_signature.sh $(SHORT_BOARD_NAME)
 
+.target-sentinel: .FORCE
+	if [ "$$(cat $@ 2>/dev/null)" != $(SHORT_BOARD_NAME) ]; then \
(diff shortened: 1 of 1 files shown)


rusefillc:
@chuckwagoncomputing I think it's totally fine for right now

the only improvement I would suggest is moving the note about the limitations into the actual .mk file

chuckwagoncomputing:
The only red is a network glitch.

Comments coming in next PR.

Issue #6079, opened 2024-02-28
bad fix of sensors reinit has caused a forever CAN listener loop


runesmdk:
Board looses comms (fail to burn page) if set to daihatsu 4cyl trigger using c5 cam1/hall1 input. 

root cause https://github.com/rusefi/rusefi/commit/ee20cbd33b114dd05782faf02728f1966216ae9b
fix https://github.com/rusefi/rusefi/pull/6119
follow up https://github.com/rusefi/rusefi/issues/6120

runesmdk:
> Yes, I am able to reproduce the issue, thank you for the report!
> 
> @runesmdk please use https://github.com/rusefi/rusefi/releases/tag/2024-02-18 as a workaround that one seems to work

I downloaded the fresh snapshot today. This link is to nightly's thats 2 weeks old?

rusefillc:
> This link is to nightly's thats 2 weeks old?

yes

https://github.com/rusefi/rusefi/releases/tag/2024-02-22 does not have burn issue but has 5VP issue

actually https://github.com/rusefi/rusefi/releases/tag/2024-02-27 seems to be OK as well so whatever defect we have introduced seems really recent?

runesmdk:
> > This link is to nightly's thats 2 weeks old?
> 
> yes
> 
> https://github.com/rusefi/rusefi/releases/tag/2024-02-22 does not have burn issue but has 5VP issue
> 
> actually https://github.com/rusefi/rusefi/releases/tag/2024-02-27 seems to be OK as well so whatever defect we have introduced seems really recent?

Will try later, thanks

runesmdk:
> > > This link is to nightly's thats 2 weeks old?
> > 
> > 
> > yes
> > https://github.com/rusefi/rusefi/releases/tag/2024-02-22 does not have burn issue but has 5VP issue
> > actually https://github.com/rusefi/rusefi/releases/tag/2024-02-27 seems to be OK as well so whatever defect we have introduced seems really recent?
> 
> Will try later, thanks

Updated to 2024-02-27. Kept getting "Data write failed for 2 attempts, local data will be refreshed from controller" and "Controller error: Burn page failed crc check" while using "slow comms (wireless)" on usb cable. Switched to "fast comms" and worked. Dont know if i its just because you cant use slow on usb? But daihatsu trigger is now set atleast.

rusefillc:
Does not reproduce with f407 discovery or proteus F7 firmware.

[...]

PR #6119, opened 2024-03-02
Do not insert aem can listeners to list on each flash write

dron0gus:
reconfigureSensors()->initLambda()->registerCanSensor() ...

This cause linked list loop.

Proper solution is to remove AEM sensors from can listener list before inserting them again.

Revert "missing initLambda in reconfigureSensors"

This reverts commit ee20cbd33b114dd05782faf02728f1966216ae9b.


dron0gus:
Related to https://github.com/rusefi/rusefi/issues/6079 ? CAN RX thread cause 100% CPU load...

mck1117:
You can't just remove that line, it will break analog widebands after a burn.

Issue #5990, opened 2024-02-19
auto-clean if you switch between boards


rusefillc:
if I compile mre

and then go into uaefi folder and invoke compilation without clean I get a confused compiler

would it be easy to get automatic clean if I move from MRE to uaEFI compilation?

chuckwagoncomputing:
It's coming, only 17 more pull requests to go.

PR #6124, opened 2024-03-03
Generate .deps before compiling

chuckwagoncomputing:
Explained in deps.mk

In theory this should take care of #5990, but I think it's likely there are edge cases. Also, .target-sentinel is still using SHORT_BOARD_NAME, so it will really only work when switching between boards, not board variants.

I think this change is more likely than some to cause issues, so I'm making this a draft until I have the time to deal with any ensuing problems. Feel free to check it out and play around.

Note that live docs and enums aren't auto generated yet.

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -396,6 +396,10 @@ include rusefi_pch.mk
 
 include bundle.mk
 
+ifeq (,$(filter clean,$(MAKECMDGOALS)))
+include deps.mk
+endif
+
 .PHONY: CLEAN_RULE_HOOK CLEAN_PCH_HOOK CLEAN_BUNDLE_HOOK
 
 CLEAN_RULE_HOOK: CLEAN_PCH_HOOK CLEAN_BUNDLE_HOOK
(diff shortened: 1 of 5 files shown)


chuckwagoncomputing:
The biggest drawback of this is that initial compile time is significantly higher. Just looking at some GHA workflow runs it looks like about double.
I expect the payback to come on rebuilds with small changes. The alternative approach to this would be to tell make that all .o files depend on live docs, enums, and configs. Changing any one live doc, enum, or config input file would result in all .c and .cpp files being rebuilt.
What we're doing here is paying the time up front, letting GCC tell us which .c and .cpp files need which generated files, then when the generated files are changed, make knows which files it needs to rebuild.

rusefillc:
There is fragility/mainenability side. Fill rebuild for any meta change was that we had itvwas accepable.

But let's give it a try with smarter why not

chuckwagoncomputing:
Also, how much do we we build locally vs. letting GHA do it? When working on infrastructure, I'm often letting GHA do it because it's GHA workflows that I'm modifying. I expect if I were actually working on firmware I'd be compiling locally 90% of the time. What's your balance Andrey?

rusefillc:
I do more than 10% of local 

But 95% of local is touching only c/cpp not .txt

chuckwagoncomputing:
Closing this for now

PR #6126, opened 2024-03-03
Dedup ChibiOS configs

dron0gus:
stm32f4, stm32f7, stm32h7 and at32f4 all have almost the same ChibiOS config.
Use same common config for all of them.
@mck1117 do you know why stm32h7 has `CORTEX_ENABLE_WFI_IDLE` set to `FALSE`? Can we enable and have 100% same config?


rusefillc:
@dron0gus can we rename to chconf_common.h or something just for sanity?

would that allow for a simpler include path?

dron0gus:
`hw_layer/ports/stm32/cfg` is not in include paths. So we already have:
```
$ grep 'halconf_community.h' -R ./
./hw_layer/ports/at32/at32f4/cfg/halconf.h:#include "../../cfg/halconf_community.h"
./hw_layer/ports/stm32/stm32f4/cfg/halconf.h:#include "../../cfg/halconf_community.h"
```

dron0gus:
@rusefillc btw we already have chconf_common.h https://github.com/rusefi/rusefi/blob/master/firmware/hw_layer/ports/chconf_common.h

rusefillc:
@chconf_stm32_commom.h and include path into lookup?

PR #6143, opened 2024-03-05
Call gen_live_documentation.sh from make

chuckwagoncomputing:
- call gen_live_documentation if any input files or the jars were updated
- don't build config_definition_base or enum_to_string in gen_live_documentation or gen_enum_to_string because make will do that if needed
- build config_definition_base and enum_to_string in places that are calling gen_live_documentation directly


chuckwagoncomputing:
The 1 red appears to be a network glitch.
Green custom: https://github.com/chuckwagoncomputing/fw-custom-example/actions/runs/8150988943

Issue #6168, opened 2024-03-08
two problems "rusefi_bundle_.zip: Cannot stat: No such file or directory"


rusefillc:
https://github.com/rusefi/rusefi/actions/runs/8206075653/job/22444372426

* should have failed due to "Cannot stat: No such file or directory"
* should have used better name and not fail?

chuckwagoncomputing:
Appears to be fixed.

PR #6172, opened 2024-03-08
Use custom board build for main firmware workflow

chuckwagoncomputing:
Pro: Less workflow to maintain
Cons:
- In the Github interface, the steps don't have an individual collapsible section, they all share one section.
- Actions don't have specific inputs for secrets, so secrets are passed as normal inputs then masked with the `::add-mask` GHA command. This seems a little sketchy, but I guess it's OK.

Incidentally fixes half of #6168. If you don't like this PR I can submit a different fix for that.


rusefillc:
Not ready for new risk, van we pleaee fix before we enhance. 

Desperately need reliable toolchain. Lets

Issue #6190, opened 2024-03-09
only:not_a_board generates an error build/email


rusefillc:
only:debug

would do https://github.com/rusefi/rusefi/actions/runs/8217366451

PR #6191, opened 2024-03-09
Check for empty matrix

chuckwagoncomputing:
This is a funny little `contains()` check because GHA expressions don't have anything like `length()`


chuckwagoncomputing:
Example: https://github.com/chuckwagoncomputing/rusefi/actions/runs/8217423815/job/22473119674
Fixes #6190

PR #6204, opened 2024-03-12
max3185x: fix max31855 detection

dron0gus:


dron0gus:
@zed65 tested on my side with max31855 - works fine.
This was my stupidest typo.

PR #6238, opened 2024-03-17
only:simulator has rules

rusefillc:
regardless of boards let's force simulator to be clean?

### DIFF
--- a/simulator/Makefile
+++ b/simulator/Makefile
@@ -25,6 +25,7 @@ endif
 endif
 
 include ../firmware/rusefi.mk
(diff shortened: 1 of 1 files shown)


chuckwagoncomputing:
I don't understand what you are trying to achieve. Does `make clean` not clean simulator?

rusefillc:
@chuckwagoncomputing this is for simulator to have same gcc warning level as firmware

rusefillc:
@chuckwagoncomputing but side effect first or yet another piece of shared .mk

PR #6239, opened 2024-03-17
Get default tune and include meta output folder

chuckwagoncomputing:
Draft because I have a red custom board with an error I haven't seen before.
https://github.com/chuckwagoncomputing/fw-custom-paralela/actions/runs/8312666260/job/22747852617


chuckwagoncomputing:
Oh I see that the size is off by the same amount that LUA_SCRIPT_SIZE increased.

Issue #6247, opened 2024-03-18
"Switch to OpenBLT" command doesn't work on Proteus-F7


andreika-git:
The "reboot_openblt" command reboots the ECU, but OpenBLT doesn't switch to "backdoor" mode on Proteus-F7...
Something around SharedParamsReadByIndex(0, &value): the value != 1 (but the function returns BLT_TRUE)
The same build works on 2chan (recompiled locally or from bundle).

mck1117:
flush the cache before reboot

rusefillc:
> flush the cache before reboot

@mck1117 do you have a relevant FOME line? How does this work for you on F7?

dron0gus:
https://github.com/rusefi/rusefi/pull/6291

mck1117:
> @mck1117 do you have a relevant FOME line? How does this work for you on F7?

it works and I'm not sure why

andreika-git:
Fixed by https://github.com/rusefi/rusefi/commit/4c7cd85f8bf7b8f6137af45f45c09b30b0dc51fe

Issue #6246, opened 2024-03-18
OpenBLT doesn't accept connections during the start if the firmware is already present


andreika-git:
If PC fails to connect to the firmware, or for some reason the command "Switch to OpenBLT" cannot be executed, then there's no way to use OpenBLT to update the firmware.

PR #6248, opened 2024-03-18
OpenBLT will accept connections during the start if the firmware is already present

andreika-git:
https://github.com/rusefi/rusefi/issues/6246

Tested on small-can-board

Also tested on Proteus-F7 (except for https://github.com/rusefi/rusefi/issues/6247)


mck1117:
so this increases the boot time by 500ms?

andreika-git:
The default 0.5 sec delay is usually NOT ENOUGH to connect via USB during the normal boot mode, but 1 second is enough.
We should decide is 1 sec will be the new default value. Also need to test CAN connection.

andreika-git:
> so this increases the boot time by 500ms?

Yes, otherwise the bootloader is useless as a recovery tool with failed firmware.

andreika-git:
I've tested BOOT_BACKDOOR_ENTRY_TIMEOUT_MS=0 on 2chan. It doesn't work as expected.
We cannot skip BootTask() if BOOT_BACKDOOR_ENTRY_TIMEOUT_MS==0, because we still need to use the bootloader to update the firmware. We cannot even call it only for stayInBootloader==true, because there's a case when there's no firmware (only bootloader), and we cannot detect it easily, except for calling CpuStartUserProgram() and BootTask() repeatedly.

PR #6295, opened 2024-03-24
Use make output-sync option

### LINKED ISSUE #6259: Usability and using github action for documentation
* my eye lid is twitching every time *simulator* build fails under "Build Firmware" banner like at https://github.com/chuckwagoncomputing/rusefi/actions/runs/8336663644/job/22814262567
* multi-threaded output is harder to read - C/C++ error is buried somewhere in the middle of gradle output

Can we consecutively invoke a series of make commands in compile.sh before invoking the final make command? It would probably have only minor performance consequences while major usability improvement.

chuckwagoncomputing:
This keeps log output together on a per-recipe basis, and all the output of a submake (e.g. bootloader) together.
There is still some stuff that seems like output is mixing but isn't actuallly, in particular, output from other recipes can happen between compilation of .c files and compilation of .cpp files.
This is my preferred way to fix #6259, though it doesn't address the concern of "simulator build failing under Build Firmware banner".

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -237,25 +237,11 @@ jobs:
 #        name: libopenblt-macos
 #        path: ./firmware/ext/openblt/Host/
 
-    - name: Clean (Just in Case)
-      working-directory: ./firmware/
-      run: |
-        make clean
-
-    - name: Handle configs separately just to make github logs more readable
-      working-directory: ./firmware/
-      run: |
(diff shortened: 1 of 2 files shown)


rusefillc:
just cherry-picked https://github.com/rusefi/rusefi/commit/735d10243ce21aa495d18d706c3656e8ef4cf1a1 need a couple of days to make up my mind on the rest please?

chuckwagoncomputing:
Yep sounds good.

PR #6363, opened 2024-04-14
Add support for analog transmission range sensors

chuckwagoncomputing:


firmware/controllers/sensors/sensor_type.h
```diff
@@ -113,6 +113,13 @@ enum class SensorType : unsigned char {
 
 	DetectedGear,
 
+	RangeInput1,
```
mck1117, line 116: does it make sense to put the un-decoded range inputs in the sensor model, or to put only the "decoded" PRNDL21whatever?

chuckwagoncomputing: It looked to me like the way to read an analog pin was to create a Sensor for it. Is there a better way?

chuckwagoncomputing: Also I found ResistanceFunc and wanted to use it.

chuckwagoncomputing:
kinetis build failed with "cannot move location counter backwards (from 20008058 to 20008000)"
I may have deducted the wrong amount from `unusedOftenChangesDuringFirmwareUpdate`.

chuckwagoncomputing:
Oh I see Kinetis has it's own RAM_UNUSED_SIZE, which is 1.
https://github.com/rusefi/rusefi/blob/master/firmware/config/boards/kinetis/efifeatures.h#L329

What do?

firmware/controllers/sensors/sensor_type.h
chuckwagoncomputing: Andrey's suggestion was to reuse the Lua aux analog input Sensors, but that seemed like a bad idea to me.

rusefillc:
> kinetis build failed with "cannot move location counter backwards"

I suggest exploring https://github.com/rusefi/rusefi/commit/12fb012321b31cfc234b032545fa6640dc3bb9d8 further if needed

firmware/controllers/sensors/sensor_type.h
rusefillc: > Andrey's suggestion was to reuse the Lua aux analog input Sensors, but that seemed like a bad idea to me.

that was suggested only as a dev time shortcut

chuckwagoncomputing:
Maybe init_range.cpp should get EFI_TCU wrapper

PR #6378, opened 2024-04-20
Automatic shifting

chuckwagoncomputing:


firmware/controllers/algo/rusefi_enums.h
```diff
@@ -699,7 +699,8 @@ enum class TsCalMode : uint8_t {
 enum class GearControllerMode : uint8_t {
 	None = 0,
 	ButtonShift = 1,
-	Generic = 2,
+	Automatic = 2,
+	Generic = 3,
```
rusefillc, line 703: is now a good time to remove or at least document that weird 'Generic'? a separate follow-up ticket would also work

Issue #5371, opened 2023-07-02
ini file weight drop


dron0gus:
Define and reuse string lists for inputs, outputs, analog inputs, output modes, etc.
See 7.3 - Additional Directives -- 7.3.1 #define of "EFI Analytics ECU Definition files.pdf"

rusefillc:
uncompressed 650K was reduced to 500K
no idea if compressed size has similar rate of reduction

PR #6391, opened 2024-04-20
rusefi_config: who use output_pin_e_enum ?

dron0gus:
@rusefillc is it safe? do we need `output_pin_e_fullenum`?


rusefillc:
could be dead code after #5371

PR #6410, opened 2024-04-24
macOS test debug

chuckwagoncomputing:
I tried on my fork and it succeeded. I have no idea why.

### DIFF
--- a/.github/workflows/build-unit-tests.yaml
+++ b/.github/workflows/build-unit-tests.yaml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-12]
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


chuckwagoncomputing:
My run for comparison: https://github.com/chuckwagoncomputing/rusefi/actions/runs/8822828017/job/24221888004
The location of realpath is different - `/bin/realpath` on yours and `/usr/local/bin/realpath` on mine. I have no idea why.

chuckwagoncomputing:
Oh mine's on macOS 12.7.4, and your's is on 14.4.1. But WHY? They're both calling for `macos-latest`!

rusefillc:
I have a failing PR askng for macos 13, sounds like i shall be asking for 12?

chuckwagoncomputing:
It seems that `/bin/realpath` is an Apple-included version of realpath starting in macOS 13 that doesn't support all the options. Maybe we should change the alias check to always use grealpath if it exists.

PR #6433, opened 2024-04-29
drivers: gpio: add initial support for MSIOBox

dron0gus:
Works for outputs only.
Refresh intervals are hardcoded.


rusefillc:
@dron0gus are we good to close this one? sorry for all the force push drama on my end :(

PR #6439, opened 2024-04-30
TestExecutor::executeAll accepts ticks, not microseconds

kifir23917:
I'm sure that this fix is correct, but I'm not sure how to fix broken tests


rusefillc:
@kifir23917 I am pretty sure it's not right

PR #6440, opened 2024-04-30
let's use Nt for EventQueue in unit tests same as prod code

rusefillc:


rusefillc:
@kifir23917 looks like unit tests are handing with this innocent change?!

I guess we shall add a counter in one of the while(true) loops :)

rusefillc:
@kifir23917 https://github.com/rusefi/rusefi/commit/0b5495e19121fe1f34f7da96f12b75707e70e2bf has helped with hanging up

PR #6449, opened 2024-05-02
SimplePWM: utilize GpioChip PWM capabilities

dron0gus:
@mck1117 @rusefillc can you check that I did not break some C++ magic with my dirty hands?


rusefillc:
@dron0gus looks like legit red kinetis?

PR #6495, opened 2024-05-11
definitely broken TEST BRANCH STILL WITH DEFECT, also starts at 268K uptime

rusefillc:
tests builds which start with 200,000 second uptime right from boot

image


rusefillc:
Artifacts https://github.com/rusefi/rusefi/actions/runs/9045410526

PR #6500, opened 2024-05-12
just a set of questions - I'm not sure that we need to merge this branch into master

kifir23917:


rusefillc:
-fsingle-precision-constant

PR #6518, opened 2024-05-15
 interpret value from openLoopMap according to boostOpenLoopYAxis constant (fixes #4778)

### LINKED ISSUE #4778: Allow MAP as Open Loop Boost vertical axis
I think boost control should have a few changes implemented.
First, open lood should have selcteable Y axis between TPS and MAP. Using MAP allows to do pseudo closed loop boost ( I KNOW IT IS NOT TRUE CLOSED LOOP) but allows for quick somewhat reliable results.

kifir23917:


unit_tests/tests/actuators/test_boost.cpp
```diff
@@ -64,7 +64,7 @@ TEST(BoostControl, OpenLoop) {
 	bc.init(nullptr, &openMap, nullptr, nullptr);
 
 	// Should pass TPS value thru
-	Sensor::setMockValue(SensorType::DriverThrottleIntent, 47.0f);
+	Sensor::setMockValue(SensorType::Tps1, 47.0f);
```
kifir23917, line 67: @rusefillc, could you please confirm that this change in boost control open loop behavior is expected?

PR #6535, opened 2024-05-21
TS mem and speed optimization

dron0gus:
No need to have big scratch buffer.
No need to have big usb buffers.
BLOCKING_FACTOR is now not related to scratch buffer size.
Minor cleanups.

### DIFF
--- a/firmware/console/binary/ts_can_channel.cpp
+++ b/firmware/console/binary/ts_can_channel.cpp
@@ -34,7 +34,7 @@ public:
 	void stop() override;
 
 	// Special override for writeCrcPacket for small packets
-	void writeCrcPacket(uint8_t responseCode, const uint8_t* buf, size_t size, bool allowLongPackets = false) override;
+	void writeCrcPacket(uint8_t responseCode, const uint8_t* buf, size_t size) override;
 };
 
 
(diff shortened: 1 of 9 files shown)


mck1117:
Why remove the parameter? The point of the parameter and matching runtime assertion is to avoid no-copy transmitting a buffer that can change during/after the CRC computation.

dron0gus:
@mck1117 remove because:
1. `allowLongPackets` is not clear name for parameter saying if source data is volatile or not.
2. mess of additional checks if allowLongPackets == false and if packet fits into buffer, need to have bit scratch buffer.
3. simplify code. 99% of data is livedata that is handled separately. Additional memcpy once per settings read or fast logger read makes zero sense.

dron0gus:
Ops,

> 3.4.2 blockingFactor
> Limits the largest chunk to be **written** or read.

This can cause problems, because same scratchBuffer is used for receive to. And blockingFactor >  sizeof(scratchBuffer) can cause overflow of this buffer https://github.com/rusefi/rusefi/blob/d8a427a5451cb78832f3718f06be312e8ee4348b/firmware/console/binary/tunerstudio.cpp#L534

rusefillc:
merge conflict

@dron0gus do you remember the overall state of all this?

dron0gus:
Some regression had happened.
Screenshot from 2025-08-08 16-30-07

mck1117:
min_query_time = 1ms * floor(output_channel_size / blocking_factor)

not sure why tunerstudio can't do a transaction faster than exactly 1ms

rusefillc:
https://github.com/rusefi/rusefi/pull/8996

PR #6536, opened 2024-05-21
TS: clear naming, isBigPacket() against actual buffer size

rusefillc:
@dron0gus just to confirm, this change is a bugfix, not just refactoring, is that right?

### DIFF
--- a/firmware/console/binary/tunerstudio_io.cpp
+++ b/firmware/console/binary/tunerstudio_io.cpp
@@ -19,7 +19,7 @@ size_t TsChannelBase::read(uint8_t* buffer, size_t size) {
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


dron0gus:
@rusefillc with current code this is more cosmetic change.
While `scratchBuffer` size is `BLOCKING_FACTOR + 30` this check is correct, but look unclear and wierd. See https://github.com/rusefi/rusefi/blob/d8a427a5451cb78832f3718f06be312e8ee4348b/firmware/console/binary/tunerstudio_io.h#L56
With other changes from https://github.com/rusefi/rusefi/pull/6535 this becomes bugfix because `scratchBuffer` size is no longer related to `BLOCKING_FACTOR`.

dron0gus (approved):

PR #6539, opened 2024-05-23
variable shadowing should be avoided fix #5676

### LINKED ISSUE #5676: variable shadowing should be avoided
variable shadowing is a pitfall

rusefillc:
#5676

boards with Ethernet fail due to
```
ChibiOS/os/various/lwip_bindings/lwipthread.c: In function 'lwip_thread':
Compiling lwipthread.c
ChibiOS/os/various/lwip_bindings/lwipthread.c:414:20: error: declaration of 'p' shadows a parameter [-Werror=shadow]
  414 |       struct pbuf *p;
```

desire is to not do -Werror=shadow for such configurations while having it everywhere else

### DIFF
--- a/firmware/Makefile
+++ b/firmware/Makefile
@@ -42,8 +42,6 @@ ifneq ("$(wildcard $(../.git))","")
 $(error "../.git not found. We expect source code to be cloned not downloaded as zip file.")
 endif
 
-include rusefi_rules.mk
-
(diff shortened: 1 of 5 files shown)


rusefillc:
@kifir23917 see proposed "ALLOW_SHADOW = yes" but it does not seem to work as intended. What am I missing?

PR #6599, opened 2024-06-09
TS: handle ochGetCommand with no arguments (read all)

dron0gus:
Get all livedata with one command with no arguments. Preparation for blockingFactor = infinity.

### DIFF
--- a/firmware/console/binary/tunerstudio.cpp
+++ b/firmware/console/binary/tunerstudio.cpp
@@ -693,9 +693,11 @@ int TunerStudio::handleCrcCommand(TsChannelBase* tsChannel, char *data, int inco
 	uint16_t offset = 0;
 	uint16_t count = 0;
 
+	// command may not have offset field - keep safe default value
 	if (incomingPacketSize >= 3) {
 		offset = data16[0];
 	}
+	// command may not have count/size filed - keep safe default value
 	if (incomingPacketSize >= 5) {
 		count = data16[1];
 	}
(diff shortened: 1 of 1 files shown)


firmware/console/binary/tunerstudio.cpp
```diff
@@ -686,12 +686,23 @@ int TunerStudio::handleCrcCommand(TsChannelBase* tsChannel, char *data, int inco
 
 	const uint16_t* data16 = reinterpret_cast<uint16_t*>(data);
 
-	uint16_t offset = data16[0];
-	uint16_t count = data16[1];
+	uint16_t offset = 0;
+	uint16_t count = 0;
+
+	if (incomingPacketSize >= 3) {
```
mck1117, line 692: You don't need this check, data16[0]/[1] are always guaranteed to be inside the write buffer even if they didn't get written this time.

mck1117: (doing it this way adds extra branches whose result is always ignored)

firmware/console/binary/tunerstudio.cpp
```diff
@@ -686,12 +686,23 @@ int TunerStudio::handleCrcCommand(TsChannelBase* tsChannel, char *data, int inco
...
+	}
+	if (incomingPacketSize >= 5) {
+		count = data16[1];
+	}
 
 	switch(command)
 	{
 	case TS_OUTPUT_COMMAND:
+		if (incomingPacketSize == 1) {
+			/* Read command with no offset and size - read whole livedata */
```
mck1117, line 703: style here is `//` comments

firmware/console/binary/tunerstudio.cpp
dron0gus: Yes. It is guaranteed by current design. Not by function definition.
It is safe to access this memory. But it is not safe to use values that can be random.
Good compiler will not insert branching here, just a few instructions with conditional execution.

PR #6620, opened 2024-06-16
Fix Bluetooth programing order

jcor25:
moved restarting on different baud rate after baud rate change command is sent, also commented set default out, if reset to default wanted rescanning for baud rate needed after reset to default.

### DIFF
--- a/firmware/controllers/generated/rusefi_generated_alphax-2chan.h
+++ b/firmware/controllers/generated/rusefi_generated_alphax-2chan.h
@@ -1274,7 +1274,7 @@
 #define show_microRusEFI_presets false
 #define show_Proteus_presets false
 #define show_test_presets false
-#define SIGNATURE_HASH 2933307290
+#define SIGNATURE_HASH 1969627505
 #define SIMULATOR_TUNE_BIN_FILE_NAME "generated/simulator_tune_image.bin"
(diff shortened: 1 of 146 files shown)


jcor25:
Verified working on JDY-31 and JDY-33.

firmware/console/binary/bluetooth.cpp
```diff
@@ -108,11 +108,11 @@ static int btVerOk(SerialTsChannelBase* tsChannel)
 {
 	int len;
 	int ret = -1;
-	char tmp[38];
+	char tmp[46];
 
 	/* wait for resposne  */
 	len = btReadLine(tsChannel, tmp, sizeof(tmp));
-	if ((len >= 36) && (len <= 40)) {
+	if ((len >= 36) && (len <= 46)) {
```
scfrontier, line 115: Hi Jcor25, what did you discover here?

jcor25: this is for JDY-31 only. jdy31 do not respond to AT or AT+DISCOVERY, only AT+VERSION, update is for the newer batches that have longer response. now that I'm thinking about it you can also check for a known response such as AT+BAUD, response wound be AT+BAUD(?) so same amount of characters.

scfrontier:
I think jcor23's solution is simpler than mine and solves the bug. If removing AT+DEFAULT creates problems for other users we can re-evaluate finding the "working" baud.

PR #6622, opened 2024-06-16
Update bluetooth.h

scfrontier:
Moved some code to create a function call. Not necessary but helped me stay organized.


rusefillc:
@scfrontier this with only header change does not look right?

scfrontier:
I need to move this. Working on it now.

PR #6657, opened 2024-06-23
Update bluetooth.cpp for using at+baud instead of at+version.

jcor25:
update using at+baud to ping jdy-31 instead of at+version. at+baud replay is consistent char length vs different version char length reply.


jcor25:
confired working for jdy31 and jdy 33

rusefillc:
``if (len == 9)) {`` legit does not compile

rusefillc:
https://github.com/rusefi/rusefi/commit/10c6f78b2e77022d71a74f035d9df220f09137da

PR #6709, opened 2024-07-13
MMC/SD fixes: activity indicators

dron0gus:


dron0gus:
Show read/write activity of SD card. Works for both USB access and internal use of SD card.

PR #6737, opened 2024-07-22
Mc33810 startup issue

dron0gus:


firmware/hw_layer/drivers/gpio/mc33810.cpp
```diff
@@ -189,6 +186,35 @@ inline bool isCor(uint16_t rx) {
...
+ * @details Will wait until SCK = low before releasing CS
+ */
+
+int Mc33810::spi_unselect()
+{
+	int retry = 0;
+	SPIDriver *spi = cfg->spi_bus;
+
+	if (cfg->sck.port) {
+		while (palReadPad(cfg->sck.port, cfg->sck.pad) && (++retry < 1000)) {
```
rusefillc, line 200: q1: shall we document the fact that we are busy-waiting?
q2: is buzy-waiting an acceptable route at all? what is the engine is running and we are busy waiting?
q3: would Q2 require that we explicitly lower priority of relevant threads here at least?

dron0gus: 1. will do
2. My test shows it does not takes much time:
```
2024-07-22_23_09_21_365: EngineState: mc33810wait done after 21
2024-07-22_23_09_21_365: EngineState: mc33810SPI [F00][800F]
2024-07-22_23_09_21_365: EngineState: mc33810wait done after 23
2024-07-22_23_09_21_367: EngineState: mc33810SPI [AF0][D0A]
2024-07-22_23_09_21_367: EngineState: mc33810wait done after 23
2024-07-22_23_09_21_370: EngineState: mc33810SPI [A00][201]
2024-07-22_23_09_21_370: EngineState: mc33810wait done after 21
2024-07-22_23_09_21_370: EngineState: mc33810SPI [450D][F]
2024-07-22_23_09_21_372: EngineState: mc33810wait done after 23
2024-07-22_23_09_21_373: EngineState: mc33810SPI [688A][F]
2024-07-22_23_09_21_375: EngineState: mc33810wait done after 23
2024-07-22_23_09_21_375: EngineState: mc33810SPI [1000][F]
2024-07-22_23_09_21_376: EngineState: mc33810 Successful Init
```
So it is about 22 iterations.
This is done in non-critical section, so any higher-priority task can preempt this task.
3. not sure if we need it to save 22 * x ?= 100..200 ticks of CPU?

rusefillc: @mck1117 what do you think about 2ns busy wait?

mck1117: > So it is about 22 iterations

design for worst case, not typical case. What if the chip is dying? what if the pin is loose? what is the wrong pin is selected?

dron0gus: > > So it is about 22 iterations
> 
> design for worst case, not typical case. What if the chip is dying? what if the pin is loose? what is the wrong pin is selected?

Then we exit after 1000 retries and emit error. This does not block execution of other tasks or ISR.

dron0gus: FYI, `spiPolledExchange()` used in this driver:
```
uint16_t spi_lld_polled_exchange(SPIDriver *spip, uint16_t frame) {

  spip->spi->DR = frame;
  while ((spip->spi->SR & SPI_SR_RXNE) == 0)
    ;
  return spip->spi->DR;
}
```

Issue #6738, opened 2024-07-22
small-can-board does not start


rusefillc:
hellen/small-can-board hellen F7 firmware does not start

expected: blue LED and USB
actual  result: no blue LED, no USB

rusefillc:
@dron0gus any chance you can try on uaefi?

rusefillc:
We even have https://github.com/rusefi/rusefi/blob/50a698fdafd2a660900f71c63d7189e90df56e70/firmware/config/boards/hellen/small-can-board/compile_small-can-board_debug.sh

rusefillc:
![image](https://github.com/user-attachments/assets/53f99449-5790-461b-944b-9656b23dc754)

![image](https://github.com/user-attachments/assets/4b7d8578-8082-4c99-b2f5-8d615357519f)

dron0gus:
Done in https://github.com/rusefi/rusefi/pull/6740 , @rusefillc can we close?

PR #6739, opened 2024-07-23
small-can-board does not start #6738

rusefillc:
only:small-can-board_debug


rusefillc:
@mck1117 any chance you could explain this compiler error message to me?

PR #6740, opened 2024-07-23
small-can-board: PERSISTENT_LOCATION do not use ram1

dron0gus:
ram1 is a part of ram0 region. ram0 region concatinate ram1, ram2 and optional ram3. linker don't care if regions are overlapped. So persistentState manualy put into ram1 was overlapping stack located in ram0.

```
(gdb) info symbol 0x20000000
__main_stack_base__ in section .mstack
persistentState in section .ram1
(gdb) p/x &persistentState
$24 = 0x20000000
(gdb) p/x &__main_stack_base__
$26 = 0x20000000
```


dron0gus:
Three memory corruption issues in a row.

mck1117:
what is the point of PERSISTENT_LOCATION?

rusefillc:
@mck1117 PERSISTENT_LOCATION is an attempt to place persistentState into a less common RAM section in order to gain as large Lua heap as possible

Issue #6424, opened 2024-04-26
CLT/IAT boost duty/target compensation


ElDominio:
**Is your feature request related to a problem? Please describe.**
Boost duty/target should have a 5x1 table to compensate for changes in temperatures (colder weather means air is more dense = same amount of DC get you more air; and vs CLT some people like to not have full power until operating temps are reached)

**Would we need new configuration fields?**
Two new tables to alter the duty (if in open loop boost mode) or target (if closed loop boost mode) to allow for higher/lower targets depending on temperature

**Describe the solution you'd like**
The said table to be added

**Describe alternatives you've considered**
I am currently using a blend table for this

kifir23917:
@ElDominio, am I correct that we need to add couple of tables similar to the following, but these two tables should contain 5 values instead of 16?
![Warmup fuel manual Multiplier](https://github.com/user-attachments/assets/045e7bdf-c88e-4ffd-8182-381c5216c23d)
![Intake Iir temperature fuel Multiplier](https://github.com/user-attachments/assets/b0e4b55c-7637-4506-8b40-e6f140b7c828)

What should be names of these new tables that will be displaed on the top like `Intake air temperature fuel Multiplier` and `Warmup fuel manual Multiplier` on the tables above?

What should be default values in `Air Temp` and `Coolant` columns?

As I understand, in the one new table `Multiplier` column should be renamed to `Duty`, in the another new table `Multiplier` column should be renamed to `Target`, but I'm not sure which of these columns should be alongside with `Air Temp` and which of these columns should be alongside with `Coolant`. Could you please clarify it?

ElDominio:
Intake Air Temperature Boost Multiplier
Coolant Temperature Boost Multiplier

Second columns of should BOTH still say Multiplier, since they are multipliers to the base duty and changing the name to duty will only result in confusion.
Both of these should be defaulted to 1.00, and they should work only with Boost Duty, and not touch target.

Default values should be
20 - 1.00
40 - 1.00
60 - 1.00
80 - 1.00
100 - 1.00

rusefillc:
![image](https://github.com/user-attachments/assets/c9caf3d8-055a-4c97-a93f-7bdb21cc7e32)

rusefillc:
@ElDominio in "open plus closed" which should we multiply? both open duty and closed target by same multiplier?

kifir23917:
Current implementation uses the following formulas:
### closed loop target
[...]

PR #6746, opened 2024-07-26
default `Coolant Temperature Boost Multiplier` and `Intake Air Temperature Boost Multiplier` curves and their validation #6424

kifir23917:


firmware/controllers/engine_controller.cpp
```diff
@@ -558,6 +558,9 @@ bool validateConfigOnStartUpOrBurn() {
 		ensureArrayIsAscending("Fuel CLT mult", config->cltFuelCorrBins);
 		ensureArrayIsAscending("Fuel IAT mult", config->iatFuelCorrBins);
 
+		ensureArrayIsAscending("Boost CLT mult", config->cltBoostCorrBins);
```
rusefillc, line 561: or should it be ensureArrayIsAscendingOrDefault? we have poor and inconsistent approach towards compatible changes

and we can only afford a cheap solution here at this point

kifir23917: Now I use `ensureArrayIsAscendingOrDefault` function instead of `ensureArrayIsAscending`

PR #6778, opened 2024-08-07
flash_main: compile time checks

dron0gus:


firmware/controllers/flash_main.cpp
```diff
@@ -97,7 +107,7 @@ bool getNeedToWriteConfiguration() {
 }
 
 void writeToFlashIfPending() {
-#if EFI_FLASH_WRITE_THREAD
+#if (EFI_FLASH_WRITE_THREAD == TRUE)
```
rusefillc, line 110: that's unusual do we really need? what's special specifically here?

rusefillc:
@dron0gus I've picked some lines https://github.com/rusefi/rusefi/commit/15649b4ff8f0847163817d6fff35672852f327d2

firmware/controllers/flash_main.cpp
dron0gus: Just a style alignment to ChibiOS.

PR #6792, opened 2024-08-09
Light refactorings

kifir23917:


rusefillc:
@kifir23917 please confirm that serial ports are still sorted when displayed on the combo box?

PR #6855, opened 2024-08-24
SPI_CR1_16BIT_MODE vs SPI_CR2_16BIT_MODE on F4 vs F7

rusefillc:


rusefillc:
@NOx-z hold on it's in rusEFI device_mpu_util.h files which are different between F4 and F7?

NOx-z:
> @NOx-z hold on it's in rusEFI device_mpu_util.h files which are different between F4 and F7?

sorry, the IDE followed SPI_CR1_16BIT_MODE from kinetis for bit 11.. but be that the same.. bit 0 in CR1 is Clock Phase for F7
SPI_CR2_16BIT_MODE in device_mpu_util.h does seem accurate for the F7

I do not have datasheets for the F4 in front of me at the moment to compare

Issue #4843, opened 2022-11-27
sim doesn't run on a mac


rusefillc:
@mck1117 says that sim doesn't run on a mac

rusefillc:
@mck1117 does it not compile or does it not run? is that a ChibiOS problem or rusEFI problem?

mck1117:
nowhere close to compiling

I wouldn't mind "macos sim" being simply out of scope - I'm not sure we really care

rusefillc:
@nmschulte @chuckwagoncomputing any interest? Worst case scenario this could be investigated using a github action.

nmschulte:
I don't have any Apple hardware so cannot (will not) help.

However ... possibly the current sim can run in a container on MacOS?

mck1117:
A container would probably work, but again, what do we think the use case is here? How much do we actually care?

Running in a container would exclude its use for any “regular” user, limiting its use to the one weirdo (me) who sometimes does rusEFI dev work on a Mac mostly while traveling.

PR #6869, opened 2024-08-29
fix: Remove a space to the generated board name

mirusu400:
### Current issue

On both `firmware/controllers/algo/generated_lookup_engine_configuration.h` and `firmware/controllers/algo/generated_lookup_meta.h`, They concatenate string using only `QUOTE`

However, it might be adding a space between both `ENGINE_GENERATED_NAME`, `SHORT_BOARD_NAME`, `ENGINE_GENERATED_EXT`.

for Example:
https://godbolt.org/z/zWzsKr9W4

This seems normally works on `arm-none-eabi` compiler , but other compilers such as clang might got an unexpected results:
```
<source>:12:10: fatal error: 'rusefi_generated_ f407-discovery .h' file not found
   12 | #include META_GENERATED_H_FILENAME
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
<source>:11:35: note: expanded from macro 'META_GENERATED_H_FILENAME'
   11 | #define META_GENERATED_H_FILENAME QUOTE(META_GENERATED_NAME SHORT_BOARD_NAME META_GENERATED_EXT)
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<source>:4:18: note: expanded from macro 'QUOTE'
    4 | #define QUOTE(x) Q(x)
      |                  ^~~~
<source>:3:14: note: expanded from macro 'Q'
    3 | #define Q(x) #x
      |              ^~
<scratch space>:4:1: note: expanded from here
    4 | "rusefi_generated_ f407-discovery .h"
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
Compiler returned: 1
```

### Resolve

So, I fixed them to removing a space between  `ENGINE_GENERATED_NAME`, `SHORT_BOARD_NAME`, `ENGINE_GENERATED_EXT`.

Sample compiler result:
https://godbolt.org/z/fa15ro89h

### DIFF
--- a/firmware/controllers/algo/generated_lookup_engine_configuration.h
+++ b/firmware/controllers/algo/generated_lookup_engine_configuration.h
@@ -3,12 +3,12 @@
 #pragma once
 
 #define META_ENGINE_GENERATED_NAME engine_configuration_generated_structures_
-#define META_ENGINE_GENERATED_EXT .h
+#define META_ENGINE_GENERATED_EXT h
 
 // todo: sad technical debt: failing to define SHORT_BOARD_NAME for unit_tests and concatenate specifically for mac os?!
 #if defined(SHORT_BOARD_NAME)
 
-#define META_ENGINE_GENERATED_H_FILENAME QUOTE(META_ENGINE_GENERATED_NAME SHORT_BOARD_NAME META_ENGINE_GENERATED_EXT)
+#define META_ENGINE_GENERATED_H_FILENAME QUOTE(CONCATENATE(META_ENGINE_GENERATED_NAME, SHORT_BOARD_NAME)) "." QUOTE(META_ENGINE_GENERATED_EXT)
 #include META_ENGINE_GENERATED_H_FILENAME
 
 #else
--- a/firmware/controllers/algo/generated_lookup_meta.h
+++ b/firmware/controllers/algo/generated_lookup_meta.h
@@ -3,12 +3,12 @@
 #pragma once
 
 #define META_GENERATED_NAME rusefi_generated_
-#define META_GENERATED_EXT .h
+#define META_GENERATED_EXT h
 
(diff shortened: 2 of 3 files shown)


rusefillc:
I vaguely recall being affected by some drama with mac builds this issue here could be the reason!

See also https://github.com/rusefi/rusefi/issues/4843 which could be another occurrence

PS: looks like this PR is red? :(

mirusu400:
I think I missed something 😢 (it missed `.h` on the trail)

```
In file included from ./controllers/algo/rusefi_types.h:18,
                 from ./controllers/core/common_headers.h:21,
                 from ./global.h:18,
                 from ./pch/pch.h:20:
./controllers/algo/generated_lookup_meta.h:12:1: error: extra tokens at end of #include directive [-Werror]
   12 | #include META_GENERATED_H_FILENAME
      | ^~~~~~~~~~~~~~~~~~~~~
./controllers/algo/generated_lookup_meta.h:12:1: fatal error: rusefi_generated_f407-discovery: No such file or directory
cc1plus: all warnings being treated as errors
compilation terminated.
make: *** [rusefi_pch.mk:10: pch/pch.h.gch/firmware] Error 1
```
I'll do a quick review and fix it

mirusu400:
@rusefillc 

This seems to be a more complicated issue than I thought.
Putting a macro string in `#include` seems more complicated than I thought, and I've tried to fix this, but I'm failing to include except for the current method.

I'll just close this PR and move it to issue.

Reference:
https://stackoverflow.com/questions/39038867/string-concatenation-for-include-path
https://stackoverflow.com/questions/4743822/generate-include-file-name-in-a-macro

Issue #6868, opened 2024-08-29
odd fire engines should crank in wasted spark


rusefillc:
Historically wasted spark for us was about sharing a coil between two cylinders. Individually wired wasted spark was just a virtual coil with two wires

odd file engines would not be happy with shared coils but as long as coils are individual, a per-cylinder wasted spark is totally a possibility for an odd fire engine

rusefillc:
https://rusefi.com/forum/viewtopic.php?f=2&t=2693 would be an example of such engine

rusefillc:
Dup of #4195

Issue #6662, opened 2024-06-25
single cylinder, 4 stroke wasted spark


rusefillc:
https://rusefi.com/forum/viewtopic.php?f=17&t=2768 states that single cylinder, 4 stroke does not operate in wasted spark mode

rusefillc:
@ElDominio do you know if we have wasted one cylinder?

ElDominio:
On chumas i run "single coil" without sync, so something works somehow

rusefillc:
related to #6868?

rusefillc:
Dup of #4195

PR #6875, opened 2024-08-31
Odd cylinder engine wasted spark

mck1117:
port of https://github.com/FOME-Tech/fome-fw/pull/479

support wasted spark on odd cylinder count engines, including during startup before the cam position is resolved.

5 cylinder firing wasted spark (every cylinder fires every 360 degrees):

image

https://github.com/rusefi/rusefi/issues/4195 and also https://github.com/rusefi/rusefi/issues/6662

Also supports odd-fire but even-cylinder engines, as that's the same fundamental problem!

https://github.com/rusefi/rusefi/issues/6868

Here's an odd-fire 2 cyl running in wasted spark:

image

### DIFF
--- a/firmware/controllers/algo/engine_state.h
+++ b/firmware/controllers/algo/engine_state.h
@@ -26,6 +26,8 @@ public:
 	 */
 	angle_t engineCycle;
 
+	bool useOddFireWastedSpark = false;
+
 	/**
 	 * this is based on sensorChartMode and sensorSnifferRpmThreshold settings
 	 */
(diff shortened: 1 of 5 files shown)


mck1117:
that failed build is either a fluke or preexisting, there's no way this caused `region 'flash0' overflowed by 3664 bytes`

rusefillc:
> that failed build is either a fluke or preexisting, there's no way this caused `region 'flash0' overflowed by 3664 bytes`

fixed https://github.com/rusefi/rusefi/issues/6876

PR #6918, opened 2024-09-15
Extend CAN EGT to six channels

LiQuiD04:
Using AEM X Series EGT CAN protocol we can make six channels EGT extension board.

PR #6943, opened 2024-09-21
Add option to configure CAN EGT channels count

LiQuiD04:
In reference to #6918 I add the option to select number of CAN EGT channels to reduce CPU resource usage


rusefillc:
Not having CPU usage measured before and after the change reduces the weight of CPU usage argument for this feature

on the other hand we get extra complexity in the UI and in the code

Issue #5608, opened 2023-10-14
Shift Torque Reduction (Flat Shift)


ElDominio:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I am shifting between gears (in any kind of transmission), and I need to let go of the accelerator to reduce torque to the transmission/rear wheels.

**Would we need new configuration fields?**
There would have to be a new menu under "Advanced" called "Shift Torque Reduction (Flat Shift)", containing **at least** the following settings:
Torque Reduction Trigger Pin- pin that activates the reduction/cut for shifting; **sometimes shared with the Launch Control pin.**
Torque Reduction Time - For how long after the pin has been triggered will the cut/reduction stay active (usually about 200ms). After that, even if the pin is still triggered, torque is re-introduced
Torque Reduction Arming RPM - Since this pin is usually shared with launch control, most people have an RPM where behavior under that is Launch Control, over that is Flat Shift/Torque Reduction
Torque Reduction Arming APP - Pedal position to realize that we need to reduce torque when the trigger pin is uuuh triggered
Torque Reduction Ignition Retard - How many degrees of timing advance will be reduced during the Torque Reduction Time.
Torque Reduction Ignition Cut - How many % of ignition events will be cut (similar to setSparkHardSkipRatio in Lua). 0-100%

This can later be complicated with much more things, but for now this would work good

**Describe the solution you'd like**
I want torque to be reduced by using timing controls when I trigger a pin, for a set amount of time and over a certain RPM and TPS, to assist in transmission shifting.

**Describe alternatives you've considered**
Writing this in Lua, but time constraints and possible memory constraints

ElDominio:
I did this with Lua, https://rusefi.com/forum/viewtopic.php?f=5&t=2670

ElDominio:
> [@ElDominio](https://github.com/ElDominio), above you listed the following settings on `Shift Torque Reduction (Flat Shift)` dialog:
> 
> * `Torque Reduction Trigger Pin`
> * `Torque Reduction Time`
> * `Torque Reduction Arming RPM`
> * `Torque Reduction Arming APP`
> * `Torque Reduction Ignition Retard`
> * `Torque Reduction Ignition Cut`.
> 
> Could you please map them on variables in your Lua scripts?
> 
> For example am I correct that `Torque Reduction Time` corresponds to `flatShiftTotalTimeMs`?

Sorry for the late reply

you are correct on torque reduction time

[...]

PR #6950, opened 2024-09-26
define fields for Flat Shift functionality #5608

kifir23917:


rusefillc:
@kifir23917 "torqueReductionArmingRpm name already used" do we add torqueReductionArmingRpm twice?

kifir23917:
> @kifir23917 "torqueReductionArmingRpm name already used" do we add torqueReductionArmingRpm twice?

My fault - I've already fixed it.

PR #6995, opened 2024-10-22
skoda trigger wheel - direction and ratio change

Tomika012345:
updated trigger angles according to measurement in working engine.
Falling edge sync proposal - ratio for rising edge would be above 20 (too short to too long gap after tdc)

### DIFF
--- a/firmware/controllers/trigger/decoders/trigger_vw.cpp
+++ b/firmware/controllers/trigger/decoders/trigger_vw.cpp
@@ -10,19 +10,19 @@
 #include "trigger_universal.h"
 
 void setSkodaFavorit(TriggerWaveform *s) {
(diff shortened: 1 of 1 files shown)


rusefillc:
@Tomika012345 please confirm successful local unit tesr run before third attemp.

Tomika012345:
i can not get it working. the measured edges on a live car are as follows: 

Rising 3° (3° after TDC)
Falling 130° falling (133° after TDC)
Rising 46° rising (180° after TDC)
Falling 134° falling (314° after TDC)
Rising 44° rising (348° after TDC)
Falling 3° falling (360°/TDC)

rusefillc:
See also https://rusefi.com/forum/viewtopic.php?f=17&t=2827

Issue #7002, opened 2024-10-28
3-0 dies for no reason at idle speeds


ElDominio:
**Describe the bug**
3-0 car with 6G72 cam is idling fine, then decides to quit. Logic logs shows no reason for this?

**Hardware**
 - Board: custom f407
 - Engine info 6g72

**To Reproduce**
1. turn negine on
2. idle for a while
3. watch triggerErrorCounter rise and engine die

**Expected behavior**
stay running

**Additional context**
ref: #6798

rusefillc:
MLV

[29scynerror.zip](https://github.com/user-attachments/files/17483548/29scynerror.zip)

at 29.6s ish,  see trigger error rise, but triggerPrimaryRise seems logical? who caused this error?

![image](https://github.com/user-attachments/assets/825f0151-e97e-4e80-9e71-904700cfdfe3)

rusefillc:
logic data

[dyinglogiclogs.zip](https://github.com/user-attachments/files/17512243/dyinglogiclogs.zip)

I took a bunch of logs of the car having difficulty with cranking, and specifically, in "diedoutlog2.logicdata" at about 12.5s it has a sync loss event that leads to shutdown of the engine.

rusefillc:
![image](https://github.com/user-attachments/assets/84155492-cc66-4761-bf77-8d313e4f3c12)

rusefillc:
problem one: we seem to have (electrical?) noise

problem two: we've inherited unusually forgiving gap ratio settings from 12/0 trigger wheel while re-using 12/0 code

* recommending to fix root cause of electrical noise
* recommending to try custom less forgiving gap settings

length=2
from 0.7 to 1.3? that would hopefully handle noise differently enough?

PR #7006, opened 2024-10-29
3-0 dies for no reason at idle speeds #7002

rusefillc:
different gaps for different wheels

@ElDominio wanna try or should I just merge?


rusefillc:
@ElDominio wanna try or should I merge?

ElDominio:
I was thinking to check hardware first?

rusefillc:
> I was thinking to check hardware first?

fair point, we need hardware fix before we are in a good position to test firmware. Let me merge to keep matter simpler

PR #7017, opened 2024-10-31
- knock fuel trim + diasble suppression then deceleration 

3eR01ive:
- knock disable suppression then deceleration


rusefillc:
@ElDominio FYI

PR #7018, opened 2024-10-31
WIP: updates on hardware-ci job

FDSoftware:
minor change in hardware-ci, I update the runner name to the ones generated by the script (for now to test only with f407)


rusefillc:
I see a lot of incompatible changes which are totally breaking (questionable) status quo.

FDSoftware:
> I see a lot of incompatible changes which are totally breaking (questionable) status quo.

I reverted the changes and created a workflow just for the new HW-CI until the existing runners can be migrated, if you need me to add another change let me know

PR #7023, opened 2024-11-01
Update trigger_vw.cpp - skoda trigger wheel

Tomika012345:
Trigger wheel was mirrored(small tooth on other end of gap). Setting up falling edge trigger in TS gives the same tooth pattern as originally coded with different gap lengths, so only length and gap ratio had to be modified. Tested on working car.
Untitled


rusefillc:
@Tomika012345 nice!

Issue #6595, opened 2024-06-09
data write failed between older project and newer firmware


rusefillc:
Looks like we are so used that we have stopped reporting  :(

reproducible with just a discovery board

[780rusEfi-min.zip](https://github.com/user-attachments/files/15829308/780rusEfi-min.zip)

![image](https://github.com/rusefi/rusefi/assets/48498823/19adb000-4feb-48ef-a598-db1e3357725b)

* grab fresh firmware, reset to default calibration
* open attachd 78-rusEfi-min project
* connect
* hit left button "send current TS settings"
* expected: happy
* actual result: "data write failed"

![image](https://github.com/rusefi/rusefi/assets/48498823/af2ced6c-8773-474a-b872-85e4fa8abe3b)

rusefillc:
is this similar to https://github.com/rusefi/rusefi/issues/6116 or different story?

zed65:
I erased my board with STM32Cube and flashed with fresh build, when I connect with TS it wanted to update firmware.ini, did that and then it starts to load tune to board and I got the same error.
![loss](https://github.com/rusefi/rusefi/assets/23425941/c408fd0c-9481-43aa-b4f5-b59e00bf5ff2)

rusefillc:
@zed65 please attach your project folder - both main controller and current tune files

zed65:
This is probably not related but I also got this error even though I got a fuelpress sensor configured. After a reboot of the board the error is gone.
![loss1](https://github.com/rusefi/rusefi/assets/23425941/55c91fa8-79ea-465f-8db2-24d57104acfd)

zed65:
Here is my folder (if that was the one you wanted).
[780rusEfi.zip](https://github.com/user-attachments/files/15751890/780rusEfi.zip)

rusefillc:
https://github.com/rusefi/rusefi/releases/tag/release_20240304

zed65:
If I do as you did (with hardware), everything is fine. Don't know what's going on :(

rusefillc:
> If I do as you did (with hardware), everything is fine. Don't know what's going on :(

i assume that it is some specific setting in the older project which triggers the problem

@zed65 @ggurov since we know that sometimes it does NOT happen - looks like the next step is to compare affected tune to new tune defaults and then keep reducing difference report until problem stops happening?

it could also be about the total amount of changes not any specific setting

zed65:
[...]

PR #7031, opened 2024-11-03
6595 burn failed

dron0gus:
https://github.com/rusefi/rusefi/issues/6595


dron0gus:
USB buffers size was:
768 * 2 for both Rx and Tx
Total 3072.
Now:
Rx: 64 * 2
Tx: 1024 * 2
Total: 2176.

PR #7039, opened 2024-11-04
TS: implement getCommand that have no params and requests all livedata

dron0gus:
But until ochBlockSize is bigger than blockingFactor TS will not use this command and prefer ochGetCommand.

But you can comment out blockingFactor in ini file and get:
Screenshot from 2024-11-04 16-06-47


rusefillc:
@dron0gus I've moved comment into source code https://github.com/rusefi/rusefi/commit/a5470c868a432229edb3b0b3f6461fc58f75a553

Issue #6046, opened 2024-02-25
config_definition-all.jar please support pages


dron0gus:
highSpeedOffsets should be located on separate settings page:
1. it should be excluded from settings CRC calculation.
2. it should not be saved to flash.

rusefillc:
@dron0gus highSpeedOffsets second page is going to be startic, right? I wonder what _exactly_ is needed on the code generation side of the effort.

I recall we were talking about some constant and well, I forgot :(

@dron0gus any chance you can make a branch with manual edits which we need?

PR #7040, opened 2024-11-04
TS: do we need pageActivate ?

dron0gus:


dron0gus:
@rusefillc is it some king of rudiment that can be safely removed?

rusefillc:
@dron0gus yes, we do not really use pages. Actually I think you were requesting that we ADD pages https://github.com/rusefi/rusefi/issues/6046?

rusefillc:
testing https://github.com/rusefi/rusefi/commit/fce0839ba4b47b49285b52744b38d21d43c2ec10

firmware/integration/ts_protocol.txt
```diff
@@ -59,6 +56,7 @@
 #define TS_RESPONSE_FRAMING_ERROR 0x8D
 ! 0x52 82
 #define TS_READ_COMMAND 'R'
+#define TS_PAGE_READ_COMMAND 'r'
```
rusefillc, line 59: Why is this line being added?

dron0gus:
Yes, I want to add pages, but we need no activate page command. We have enough resources to keep all pages in RAM.

firmware/integration/ts_protocol.txt
dron0gus: Fixed

Issue #7054, opened 2024-11-08
Allow up to 510 as Y axis in boost tables that use MAP as Y axis


ElDominio:
**Is your feature request related to a problem? Please describe.**
I want to use 500 as a Y value in a boost table that uses MAP as Y. Currently max is 255. I understand 510 is possible by multiplying *2

**Would we need new configuration fields?**
None

**Describe the solution you'd like**
Multiply Y axis by 2

**Describe alternatives you've considered**
None really

rusefillc:
```
! technical debt: some of these use uint8_t !!!
#define MAP_UPPER_LIMIT 255
```


![image](https://github.com/user-attachments/assets/ca55e746-9cde-41e0-b24a-fe6b93392171)

FDSoftware:
@ElDominio you can try now this with the build: [Nightly 2024-11-13](https://github.com/rusefi/rusefi/releases/tag/2024-11-13)

ElDominio:
@rusefillc this seems fixed?

PR #7064, opened 2024-11-11
MAP_UPPER_LIMIT, change from uint8 to uint16

FDSoftware:
- updated load-related configs from hardcoded 1000kpa to MAP_UPPER_LIMIT
- updated MAP_UPPER_LIMIT from 255 kpa to 1000kpa

related issue: https://github.com/rusefi/rusefi/issues/7054

### DIFF
--- a/firmware/integration/rusefi_config.txt
+++ b/firmware/integration/rusefi_config.txt
@@ -129,8 +129,8 @@ struct_no_prefix engine_configuration_s
 ! air cooled engines go up to 250, see community board firmware which has that
 ! PnP boards for liquid cooled engines should override with '120'
 #define CLT_UPPER_LIMIT 250
-! technical debt: some of these use uint8_t !!!
-#define MAP_UPPER_LIMIT 255
+
+#define MAP_UPPER_LIMIT 1000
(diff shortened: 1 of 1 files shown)


rusefillc:
Minor analysis paralysis: why do we sometimes use 8bit autoscale and sometimes user 16bit no scale? also different autoscale scale between different 8bit?

I do not know how to make that decision. if we keep autoscale we need to introduce new constant for shared scale value

FDSoftware:
> Minor analysis paralysis: why do we sometimes use 8bit autoscale and sometimes user 16bit no scale? also different autoscale scale between different 8bit?
> 
> I do not know how to make that decision. if we keep autoscale we need to introduce new constant for shared scale value

I think autoscale is fine for RPM fields, where jumps of 100rpm at a time are not a big deal. In VE, jumps of 10kpa (for example in lambdaProtection) are considerable for a naturally aspirated engine. Converting tables that already use 16b would also be negative for existing users, who will lose precision in the VE and Ign tables.

The ideal path for me would be to update the scaled 8b at the cost of a little more memory used

PR #7074, opened 2024-11-15
LTFT part 0/5: STFT tech debt

FDSoftware:
PR to fix STFT tech deb, as a first step to LTFT implementation, next PRs will be:

* RAM storage of STFT on a new table & interpolate, size of 6x6?
* LTFT table & storage of STFT on that table, also 6x6?
* LTFT & STFT, view tables on TS, clear, edit
* LTFT storage on SD Card
* LTFT implementation on fuel calculations


rusefillc:
@FDSoftware I see a PR which seems to be removing stuff while only stating as much as "technical debt"

I would suggest that we voice coordinate or discord coordinate and re-start this effort.

@FDSoftware can I ask NOT to change the PR and NOT to edit the ticket description? As is I see the best use of this PR as an example of how not to approach change announcement.

Issue #6974, opened 2024-10-06
BMW E46 Can Dash CLT gauge needle acting up


rusefillc:
he only small problem was CLT gauge needle acting up while using default E46 CAN dash. I recreated static void canDashboardBmwE46(CanCycle cycle) in lua using setTickRate(100) and that fixed it.

https://rusefi.com/forum/viewtopic.php?p=50024#p50024

PR #7078, opened 2024-11-16
Change BMW E46 CAN interval to 10 ms

jankovalski:
Fix for #6974


rusefillc:
@jankovalski thank you!

PR #7114, opened 2024-12-02
separating two different usages of variable

rusefillc:
prior to change we had two completely independent usages of same value


rusefillc:
@dron0gus poke :)

dron0gus (approved):

Issue #7034, opened 2024-11-03
Real hardware ICU coverage on nucleo-F767


rusefillc:
Once we have https://github.com/rusefi/rusefi/issues/7033

Same as discovery HW CI uses two jumper wires for continues integration of real pin toggle and real ICU peripheral we should add similar two jumpers on the nucleo 767

https://github.com/rusefi/rusefi/wiki/Dev-Quality-Control#stm32f407g-disc1

rusefillc:
@FDSoftware can you help?

rusefillc:
@FDSoftware we have ICU what about

```
//            MiscTest.class,
//            BurnCommandTest.class,
//            CommonFunctionalTest.class,
```

which one of these three is simplest to activate?

FDSoftware:
`MiscTest` It seems not, actual temp on the F7 is 25°C, i got 3° from `getMCUInternalTemperature`
 `BurnCommandTest` seems to be ideal, now I upload PR, It works without requiring any changes to the fw
log:  https://gist.github.com/FDSoftware/a33cc087f37b0b77fbad27cfeb02e289

PR #7141, opened 2024-12-09
Real hardware ICU coverage on nucleo-F767 #7034

FDSoftware:
- moved the `vw.cpp` change for the HWCI to `engine_type_impl.cpp` (since several switch cases need it in the tests) added with a #if to compile only for target nucleof767
- some tests uses "DEFAULT_FRANKENSO" as engine configuration, added the F7/H7 flags to the corresponding case statement
- created ` VssHardwareLoopTest.class` and `PwmHardwareTest.java` for F7 target, The last one currently does not work as expected, it is disabled on this PR
- added tests:  `MiscTest` , `BurnCommandTest`,  `CommonFunctionalTest`,  `VssHardwareLoopTest` to the F7 suite

### DIFF
--- a/firmware/controllers/algo/engine_type_impl.cpp
+++ b/firmware/controllers/algo/engine_type_impl.cpp
@@ -321,7 +321,7 @@ void applyEngineType(engine_type_e engineType) {
 		testEngine6451();
 		break;
 
-#if defined(HW_FRANKENSO) || EFI_SIMULATOR
+#if defined(HW_NUCLEO_F767) || defined(HW_NUCLEO_H743) || defined(HW_FRANKENSO) || EFI_SIMULATOR
 	case engine_type_e::DEFAULT_FRANKENSO:
 		setFrankensoConfiguration();
 		break;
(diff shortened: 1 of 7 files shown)


java_console/autotest/src/main/java/com/rusefi/f7nucleo/VssHardwareLoopTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static com.rusefi.binaryprotocol.BinaryProtocol.sleep;
+import static com.rusefi.config.generated.Fields.*;
+
+/**
+ * This test relies on jumpers connecting physical pins on f7-nucleo:
+ * PD1<>PC6
+ * PD2<>PA6
+ */
+public class VssHardwareLoopTest extends RusefiTestBase {
```
rusefillc, line 20: is this some copy-paste? :( can we please reuse code/share code?

firmware/controllers/algo/engine_type_impl.cpp
```diff
@@ -413,6 +413,22 @@ void applyEngineType(engine_type_e engineType) {
 	default:
 	  applyEngineTypeExt(engineType);
 	}
+
+#if defined(HW_NUCLEO_F767) || defined(HW_NUCLEO_H743)
```
rusefillc, line 417: no-no-no, this can go to many places but not here

java_console/autotest/src/main/java/com/rusefi/f7nucleo/VssHardwareLoopTest.java
FDSoftware: yup, now I see what I can do to make the pins used configurable

firmware/controllers/algo/engine_type_impl.cpp
FDSoftware: > no-no-no, this can go to many places but not here
Can I create a new engine type and put the initialization there? Or is it better with a console command? I need something similar to Jest.js `beforeEach`
I have to run that every time an engine type is set in the tests and before the test is executed

PR #7146, opened 2024-12-09
TLS115 Power Good sensing: prepare

dron0gus:


firmware/hw_layer/adc/adc_inputs.cpp
```diff
@@ -12,6 +12,11 @@ float PUBLIC_API_WEAK getAnalogInputDividerCoefficient(adc_channel_e) {
     return engineConfiguration->analogInputDividerCoefficient;
 }
 
+int PUBLIC_API_WEAK getAnalogInputStatus(adc_channel_e, float) {
```
rusefillc, line 15: let's have it simpler for now - just one global "fatal error: analog inputs power drain"?

PR #7168, opened 2024-12-14
SENT: move to libfirmware

dron0gus:


dron0gus:
Omg, Kinetic runs out of memory.WWHYYY?

Issue #7124, opened 2024-12-05
STM32F767ZI does not like openblt if in 2mb flash size


rusefillc:
See https://github.com/rusefi/rusefi/issues/7118

rusefillc:
Depends on https://github.com/rusefi/rusefi/issues/7035

PR #7186, opened 2024-12-18
STM32F767ZI does not like openblt if in 2mb flash size #7124 do we expect HW CI to fail?

rusefillc:
#7124


rusefillc:
@FDSoftware F7 HW CI being useful!

PR #7238, opened 2025-01-02
LTFT save to flash

dron0gus:


rusefillc:
@dron0gus curves go into "config" area of configuration, only scalars go into "engineConfiguration" - that's a RAM usage optimization, only one copy of "config" while two copies of "engineConfiguration"

please rebase https://github.com/rusefi/rusefi/commit/4337a5b8d4c298558cd635d46092255a7fc3f036

PR #7248, opened 2025-01-03
how does HW CI handle page size reduction?

rusefillc:
only:config


FDSoftware:
both HWCI are green now :rocket:

Issue #7227, opened 2024-12-29
SD card bit datapoints


rusefillc:
We shall SD card data log single bit fields

rusefillc:
See https://github.com/rusefi/rusefi/commit/7a590c0b4e019122446406261936fa5310892348

the main missing piece here is LogField does not support bit field reference, we need some sort of solution or trick?

Once we have LogField plan we would adjust java code generator. This change could be easily tested without any real hardware since unit tests now use same exact code to produce logs as well see https://github.com/rusefi/rusefi/blob/master/unit_tests/test_results/readme.md

rusefillc:
@kifir23917 it works!

PR #7268, opened 2025-01-08
SD card bit data points #7227

rusefillc:
only:uaefi


rusefillc:
@kifir23917 please adjust unit tests to make branch happy

Issue #1339, opened 2020-04-21
stm32 own build-in watchdog


rusefi:
Are we interested to have stm32 watchdog enabled?

Watchdog impl https://github.com/rusefi/rusefi/pull/5751/
Detect and print Reset cause for STM32 https://github.com/rusefi/rusefi/pull/5783
actually enable watchdog for F4 https://github.com/rusefi/rusefi/pull/5792/

See also https://github.com/rusefi/rusefi/pull/5752
unneeded https://github.com/rusefi/rusefi/pull/5794

rusefillc:
watchdog shall leave a backup RAM bread crumb and reboot if an issue is defected with the firmware
in light of the bread crump we shall software fatal on next boot with "hard fault detected" critical message

rusefillc:
https://github.com/rusefi/rusefi/pull/5792/

rusefillc:
``
set_watchdog_reset 5000
``

rusefillc:
nope, it was not enabled for a year and still is not enabled as of right now @andreika-git

PR #7292, opened 2025-01-12
stm32 own build-in watchdog #1339

rusefillc:
actually enabling for stm32


rusefillc:
@andreika-git how does debug works with debugging? if this PR is merged would this affect our ability to debug?

rusefillc:
just for the record: we have green build!

PR #7297, opened 2025-01-13
FatFS: enable f_expand() function

dron0gus:


dron0gus:
This is going to be used to pre-allocate continuous chain of clusters to avoid too often write of FAT table that increase risk of FS damage in case of sudden power off.

PR #7307, opened 2025-01-14
error reporting: extract all related code to error_handling

dron0gus:
Also:
 -report ChibiOS panic message
 -use bootCount for boot report file naming
 -save few bytes in backup ram by reusing same field for different errors types

### DIFF
--- a/firmware/console/eficonsole.cpp
+++ b/firmware/console/eficonsole.cpp
@@ -260,10 +260,6 @@ void initializeConsole() {
 	addConsoleAction("reset", scheduleReset);
 #endif
 
-  // see https://github.com/rusefi/rusefi/wiki/Resilience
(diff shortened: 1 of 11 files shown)


dron0gus:
```
$ cat fail_report_61.txt 
Reset Cause: Reset from NRST pin
Power cycle count: 61
Last error type firmware cookie 0xcafebabe
firmwareError: 20250113 20250113@uaefi
```

```
$ cat fail_report_62.txt 
Reset Cause: NVIC_SystemReset or by debugger
Power cycle count: 62
Last error type HardFault cookie 0xdeadbeef
type: 0x00000003 addr: 0x20020010 CSFR: 0x00000082
r0  0x08089104
r1  0x0807a920
r2  0x000000c9
r3  0x080497c9
r12 0x00000074
lr (thread)  0x080027a5
pc  0x0804dd5e
xpsr  0x01000000
```

Issue #7291, opened 2025-01-12
TS indicator to leverage hasCriticalReportFile method


rusefillc:
Problem: with progress critical errors would hopefully be less obvious to users #7290

We need a TS indicator. Let's just add a new indicator for now to keep things simpler and more reliable.

We also need a Lua callback to check if SD card contains any critical report files

rusefillc:
random fact: @dron0gus plans to upgrade FatFS but that probably does not matter here

rusefillc:
@dron0gus could you have removed the unfinished implementation of this under https://github.com/rusefi/rusefi/commit/e0c05c47772642c244370ec4a05cb5bf004ae28d#diff-6942f49238cbf6caf68c40d6edb35f2387ddda0ccd16ae67382778267ead8bc0? "findFile" method was intended for this ticket

rusefillc:
@dron0gus skeleton at https://github.com/rusefi/rusefi/commit/65063625ee483a09dc98d9bd9c0c6c95422d0239

dron0gus:
Should we simplify "Some error report files exist on SD card" to "We just booted from error fault"?

One old report file will keep annoying user with active indicator.

dron0gus:
https://github.com/rusefi/rusefi/pull/7347 and https://github.com/rusefi/rusefi/pull/7349

PR #7316, opened 2025-01-16
error_handling: filename depends on problem type

dron0gus:


rusefillc:
@dron0gus see https://github.com/rusefi/rusefi/issues/7291#issuecomment-2598996315

the idea of having key tag in the beginning of the file name was to facilitate "findFile" method

Issue #7163, opened 2024-12-14
Make IGN CLT Correction a 5x5 table


ElDominio:
**Is your feature request related to a problem? Please describe.**
 I'm always frustrated when I want to drop timing at cold idle situations to make the warmup period shorter but the car cannot be driven.

**Would we need new configuration fields?**
Expand the IGN CLT Correction table

**Describe the solution you'd like**
Expand the IGN CLT Correction curve into a 5x5 table

rusefillc:
@ElDominio please be more specific / simpler terms

We have

![image](https://github.com/user-attachments/assets/8887aece-54e6-42a8-88bd-177ef52c9a15)

ElDominio:
![image](https://github.com/user-attachments/assets/05ec5f29-eb41-45f8-a148-716c16f6f45a)

rusefillc:
👁️ 👁️‍🗨️ 👓 🙄

ElDominio:
@FDSoftware any chance you can handle this?

ElDominio:
![Image](https://github.com/user-attachments/assets/ad096a47-57e4-4437-b772-7362feb9c843) not available?

rusefillc:
@ElDominio is this not available in https://rusefi.com/build_server/rusefi_bundle_alphax-2chan.zip?
@ElDominio is this not available in https://rusefi.com/build_server/rusefi_bundle_alphax-4chan.zip?
@ElDominio is this not available in https://rusefi.com/build_server/rusefi_bundle_alphax-8chan.zip?
@ElDominio is this not available in _all_ of https://rusefi.com/build_server/?
@ElDominio is this not available in some SPECIFIC of https://rusefi.com/build_server/?

ElDominio:
Genesis is broken
H154 does not have it (tested)

5 of my other cars are paralela, does this mean paralela will never have it?

rusefillc:
@ElDominio my first attempt to request _exact_ place of so-called `issue` has failed :(

Quck reminder that we have two kinds of Genesis - f4 and f7 are pretty unrelated binaries. 
@ElDominio please spell out which one are you considering to use for this test so that I can manually make sure that that one is fresh?

wider context is that in order to reduce business costs we do not build all binaries on all changes

ElDominio:
Car in the shop is h154 f4

Either that or paralela if possible

rusefillc:
paralela is nightly so you were acting too soon

[...]

PR #7323, opened 2025-01-19
Make IGN CLT Correction a 5x5 table [was 8x8 table]

FDSoftware:
Tested on simulator:
image

note:
Since it is a rebin to a smaller size, the user has to pay attention to how TS alters the table, can this be added as a note to the changelog?

related issue: #7163


rusefillc:
Ticket says table, change touches curve?

@ElDominio is that the right dialog?

ElDominio:
Should be a table like this one: 
image


Where Y axis is load, and X axis is CLT (instead of air temp)

PR #7348, opened 2025-01-23
stm32 watchdog

rusefillc:


rusefillc:
i think this PR does not change anything in terms of actually enabling the watchdog @andreika-git ?

Issue #7376, opened 2025-02-01
knock is broken


rusefillc:
unhooked Proteus on bench, September build - note at least some random numbers

![Image](https://github.com/user-attachments/assets/7be05887-6480-4eb6-8da1-92ac9c969067)

same proteus same bench, current build - nothing :(

![Image](https://github.com/user-attachments/assets/8f9b849c-f953-4da9-9b65-4f4630ecfb17)

rusefillc:
@dron0gus we had touched this area in October around https://github.com/rusefi/rusefi/commit/e06cd82c2667771522b3c46e72eb702c7f998cd4

ElDominio:
I had seen this, but I thought it was my hardware

rusefillc:
@dron0gus still zeros?

![Image](https://github.com/user-attachments/assets/6c04dfd9-db80-43a6-998b-52b599e53b40)

rusefillc:
![Image](https://github.com/user-attachments/assets/4634eaf9-0e47-4b7e-b1f5-3701998e197a)

PR #7377, opened 2025-02-02
stm32: adc v2: fix include order

dron0gus:
First include knock_config.h then use KNOCK_ADC define


dron0gus:
Fixes https://github.com/rusefi/rusefi/issues/7376

PR #7155, opened 2024-12-11
Report analog failure to TS and rise OBD code, use TLS115 PG signal on Hellen boards, use `expected` for adc inputs

dron0gus:
Now we can detect +5VP failure on few Hellen board.
Screenshot from 2024-12-12 01-01-33
```
2024-12-12_00_34_32_433: EngineState: WARNING: P2125: Sensor fault: AcceleratorPedalPrimary has timed out
2024-12-12_00_34_32_445: EngineState: WARNING: P2130: Sensor fault: AcceleratorPedalSecondary has timed out
2024-12-12_00_34_32_447: EngineState: WARNING: P0641: Analog subsystem fault
2024-12-12_00_34_32_447: EngineState: WARNING: P0106: Sensor fault: Map has timed out
```

Also analog inputs depending on this supply is now reported invalid using `expected`. So sensor (or any other analog input user) can propagate sign of error further.
```
2024-12-12_00_35_21_160: EngineState: fast 649421 samples
2024-12-12_00_35_21_161: EngineState:  F ch[ 0] @ PC0 ADC11 12bit=  13 0.000V input 0.000V INVALID
2024-12-12_00_35_21_161: EngineState: slow 32472 samples
2024-12-12_00_35_21_162: EngineState:  S ch[ 0] @ PA0 ADC1 12bit= 189 0.000V input 0.000V INVALID
2024-12-12_00_35_21_163: EngineState:  S ch[ 1] @ PA1 ADC2 12bit= 185 0.000V input 0.000V INVALID
2024-12-12_00_35_21_165: EngineState:  S ch[ 2] @ PA2 ADC3 12bit=1531 1.230V input 2.460V valid
2024-12-12_00_35_21_165: EngineState:  S ch[ 3] @ PA3 ADC4 12bit=  11 0.000V input 0.000V INVALID
2024-12-12_00_35_21_169: EngineState:  S ch[ 4] @ PA4 ADC5 12bit= 112 0.000V input 0.000V INVALID
2024-12-12_00_35_21_170: EngineState:  S ch[ 5] @ PA5 ADC6 12bit=2510 2.016V input 4.033V valid
2024-12-12_00_35_21_171: EngineState:  S ch[ 6] @ PA6 ADC7 12bit=2215 1.779V input 3.559V valid
2024-12-12_00_35_21_175: EngineState:  S ch[ 7] @ PA7 ADC8 12bit=1928 1.548V input 3.097V valid
[...]

PR #7406, opened 2025-02-09
Report analog failure to TS and rise OBD code

dron0gus:
Currently only TLS115 Power Good signal is handled on few Hellen boards.
First step of https://github.com/rusefi/rusefi/pull/7155


dron0gus:
Kinetis and Cypress do not fully regenerate headers?

PR #7440, opened 2025-02-13
refuse to reboot into DFU if IGNVOLTAGE is present, refuse to update …

### LINKED ISSUE #7437: refuse to reboot into DFU if IGNVOLTAGE is present, refuse to update firmware if the same
If proper care is not taken with hardware, random ports can become active, instead of floating on the processor when MCU enters DFU.

this is especially bad if those random ports are connected to pins that are connected to injectors/ignition drivers/other things that should not be on.

Please make the MCU refuse to reboot into DFU if there's IGNVOLTAGE present and refuse to update firmware when there's 12v present, perhaps as a feature flag per board 

-----------

Yes. if user attempts to enter DFU, it could set coils to charge, set injectors to spray, set starts to attempt to start.

Maybe if we want a lockout feature to be configurable at the tune level.

If there's IGNVOLTAGE present, do not enter DFU. refuse to update firmware until that is no longer active.

Remembering to turn off the ignition to remove 12v

rusefillc:
…firmware if the same fix #7437


rusefillc:
@ggurov here you are

ggurov:
This needs to be in the jump_to_bootloader() instead of  reset_and_jump(void)

unless we just stumbled onto another bug...

when there's a  criticalError() raised, "reboot_ecu" button locks the ECU putting it right into the same state DFU would've put it in, with uart2/3 pins high

rusefillc:
@ggurov this ticket is about "reboot into DFU"

Q1: does "reboot into DFU" behave as expected now?
Q2: was a new defect introduces while making this "reboot into DFU" fix here?
Q3: if YES for Q2, does this separate defect has a good separate ticket?

ggurov:
Reboot into DFU with IGN on behaves as expected - ECU stops processing input/shuts down coils/injectors demanding a reboot, code real simple there.

however, when clicking "Reboot ECU", i believe it still calls this same path, and it's refusing a reboot, silently crashing. 

it should only refuse to 
jump to bootloader

`#if EFI_DFU_JUMP
void jump_to_bootloader() {
	// leave DFU breadcrumb which assembly startup code would check, see [rusefi][DFU] section in assembly code

	*((unsigned long *)0x2001FFF0) = 0xDEADBEEF; // End of RAM

	reset_and_jump();
}
#endif`

not refuse to reboot in general.

i tested moving 
`  if (isIgnVoltage()) {
    criticalError("Not allowed with ignition power");
    return;
  }`

to  `jump_to_bootloader()` and it behaves more correct, where clicking "reboot to dfu" raises the error, leaving ECU in a sane state, and clicking "reboot ecu" after that resets the ecu, clearing the error, making it go back to doing ECU things.

rusefillc:
"reboot into DFU" behaves as expected now, no ticket/no steps to reproduce "reboot ecu" issue. this ticket is closed.

@ggurov you are raising a valid point but we have to follow the process. I am not available to jump on the follow-up problem since same as on the original, the ask is getting more complex requiring more efforts.

ggurov:
sounds good, i'll create a bug for this in a few.

PR #7451, opened 2025-02-14
SD in MSD: Speed up a bit

dron0gus:
Read using UAEFI before this patch:
```
$ dd if=test.bin of=/dev/zero status=progress
133939712 bytes (134 MB, 128 MiB) copied, 269 s, 498 kB/s
262144+0 records in
262144+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 269.492 s, 498 kB/s
```

Read using UAEFI with blkbuf1 = 2K bytes:
```
$ dd if=B5A4-271B/test.bin of=/dev/zero status=progress
133677568 bytes (134 MB, 127 MiB) copied, 202 s, 661 kB/s
262144+0 records in
262144+0 records out
134217728 bytes (134 MB, 128 MiB) copied, 202.69 s, 662 kB/s
```

### DIFF
--- a/firmware/hw_layer/mass_storage/mass_storage_device.cpp
+++ b/firmware/hw_layer/mass_storage/mass_storage_device.cpp
@@ -161,7 +161,7 @@ static const scsi_unit_serial_number_inquiry_response_t default_scsi_unit_serial
 };
 
 void MassStorageController::attachLun(uint8_t lunIndex,
-						BaseBlockDevice *blkdev, uint8_t *blkbuf,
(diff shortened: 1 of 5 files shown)


rusefillc:
@dron0gus https://github.com/ChibiOS/ChibiOS-Contrib/pull/418

dron0gus:
Ups. Some boards runs out of RAM. We need to reuse same RAM for SD read/write buffer in MSD mode and for FatFS structures in ECU logging mode. As these functions are used exclusively.

PR #7479, opened 2025-02-21
Gold: some more pinout and board-specific settings

dron0gus:
Screenshot from 2025-02-21 15-32-32


dron0gus:
Screenshot from 2025-02-21 16-40-18

PR #7510, opened 2025-02-25
fix IGN CLT/IAT Correction table issue [float => int16_t with autoscale]

FDSoftware:
~also simplify `setDefaultIatTimingCorrection` / `setDefaultCltTimingCorrection`~ (discarded)


rusefillc:
@FDSoftware please rebase

rusefillc:
this is three PRs bundled together. It's reasonable to sometimes attempt this but in this specific case I definitely do not agree with one of the three changed and not sure about second.

let's please have only one change in this specific PR

Issue #7511, opened 2025-02-25
injector lag is a table


rusefillc:
transition from

![Image](https://github.com/user-attachments/assets/26b2df52-ef1c-4657-aba4-79b5c3a74e91)

to

![Image](https://github.com/user-attachments/assets/c6a4773a-9d31-43fc-9dbb-0a114f9ab45f)

rusefillc:
@FDSoftware let's use 'FuelPressureInjector' for now

FDSoftware:
some notes:

at: `firmware/controllers/algo/defaults/default_fuel.cpp`
`setBosch02880155868` It will break because 8 values ​​are set instead of the 2 that this table will have. How to proceed? Restrict the function to only use the default 2x2?

at: `unit_tests/tests/ignition_injection/test_fuel_map.cpp`
`TEST(misc, testFuelMap) `
This is setting up a custom deadtime table, with these changes I'm going to have to edit this test in addition to the obvious one at `test_injector_model.cpp` / `TEST(InjectorModel, Deadtime)`

at: `firmware/config/engines/mazda/mazda_miata_vvt.cpp` A custom deadtime table is also being set up, will I have to look up some information about the injector used and restrict the table to 2x2?

rusefillc:
can we set both lines of default calibration to same value, with Y axis say 300kpa and 600 kpa?

meaning regardless of pressure defaults would have same exact runtime values in miata and bosch?

PR #7518, opened 2025-02-26
Expose SensorType::FuelPressureInjector on TS

FDSoftware:
added the sensor as "pressureCorrectionReference" inside "injector_model_s" struct for future use inside the deadtime curve to table change (https://github.com/orgs/rusefi/projects/37/views/1?pane=issue&itemId=99441931)

For now the value is not updated at any time, the proposed change for it is the following, but I have doubts if it is the ideal place, so I do not include it in the PR

```c++
float InjectorModelWithConfig::getDeadtime() {
    switch (getInjectorCompensationMode()) {
        case ICM_FixedRailPressure:
            pressureCorrectionReference = getFuelReferencePressure();

        case ICM_SensedRailPressure: {
            auto fps = Sensor::get(SensorType::FuelPressureInjector);
        
            if (!fps) {
                pressureCorrectionReference = getFuelReferencePressure();
            } else {
                pressureCorrectionReference = fps.Value;
            }
        }
    }

    return interpolate2d(
        Sensor::get(SensorType::BatteryVoltage).value_or(VBAT_FALLBACK_VALUE),
        m_cfg->battLagCorrBins,
        m_cfg->battLagCorr
    );
}
```

note:
in the 4 possible cases (without sensor and the 3 sensor variants), the raw value is used without adding the atmospheric pressure as is done for example in `InjectorModelWithConfig::getFuelDifferentialPressure()`

### DIFF
--- a/firmware/controllers/algo/fuel/injector_model.txt
+++ b/firmware/controllers/algo/fuel/injector_model.txt
@@ -3,5 +3,6 @@ float m_deadtime;@@GAUGE_NAME_INJECTOR_LAG@@;"ms",1, 0, 0, 0, 3
 
 float pressureDelta;Fuel: Injector pressure delta;"kPa", 1, 0, -1000, 1000, 1
 float pressureRatio;Fuel: Injector pressure ratio;"", 1, 0, 0, 100, 3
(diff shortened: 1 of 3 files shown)


rusefillc:
part of #7511

Issue #7357, opened 2025-01-25
preserve tune while updating firmware


rusefillc:
At the moment tune is often lost while updating firmware:

At the moment we have java code update firmware and then sometimes (often) new firmware version notices that it's not compatible with tune and wipes to defaults.

instead we shall backup image+.ini pair, and then attempt to apply all/most using new .ini

Yes, that's pretty similar to what TS is doing :(

rusefillc:
@kifir23917 

* red window and error
* complete error text is NOT in the clipboard

```
Console version 20250206
Windows 11 10.0
Bundle BundleInfo{branchName='development', target='uaefi'}
Suspending port scanning...
Port scanning is suspended.
Connecting to port `COM29`...
Performing action on port COM29...
Received a signature rusEFI master.2025.02.07.uaefi.2255628228
```

![Image](https://github.com/user-attachments/assets/d7053310-02e7-4a71-995a-3ac61343e5f4)

rusefillc:
@kifir23917 

```

E 250207 230924.301 [ECU Communication Executor1] CalibrationsHelper - Backing up calibrations failed:
javax.xml.bind.JAXBException: Implementation of JAXB-API has not been found on module path or classpath.
 - with linked exception:
[java.lang.ClassNotFoundException: com.sun.xml.internal.bind.v2.ContextFactory]
	at javax.xml.bind.ContextFinder.newInstance(ContextFinder.java:232)
	at javax.xml.bind.ContextFinder.find(ContextFinder.java:375)
	at javax.xml.bind.JAXBContext.newInstance(JAXBContext.java:691)
	at javax.xml.bind.JAXBContext.newInstance(JAXBContext.java:632)
	at com.rusefi.xml.XmlUtil.writeXml(XmlUtil.java:24)
	at com.rusefi.tune.xml.Msq.writeXmlFile(Msq.java:81)
	at com.rusefi.binaryprotocol.BinaryProtocol.saveConfigurationImageToFiles(BinaryProtocol.java:378)
	at com.rusefi.maintenance.CalibrationsHelper.backUpCalibrationsInfo(CalibrationsHelper.java:144)
	at com.rusefi.maintenance.CalibrationsHelper.lambda$readAndBackupCurrentCalibrations$0(CalibrationsHelper.java:171)
	at com.rusefi.maintenance.BinaryProtocolExecutor.lambda$execute$0(BinaryProtocolExecutor.java:36)
[...]

PR #7520, opened 2025-02-26
injector deadtime as table #7511

FDSoftware:
image

related issue: #7511

### DIFF
--- a/firmware/console/status_loop.cpp
+++ b/firmware/console/status_loop.cpp
@@ -430,8 +430,9 @@ static void updateFuelSensors() {
 
 	engine->outputChannels.flexPercent = Sensor::getOrZero(SensorType::FuelEthanolPercent);
 #if EFI_PROD_CODE
-	// todo: extract method? do better?
+	// todo: extract method? do better? see https://github.com/rusefi/rusefi/issues/7511 for details
 	engine->module<InjectorModelSecondary>()->pressureCorrectionReference = engine->module<InjectorModelSecondary>()->getFuelDifferentialPressure().Value + Sensor::get(SensorType::Map).value_or(STD_ATMOSPHERE);
+	engine->module<InjectorModelPrimary>()->pressureCorrectionReference = engine->module<InjectorModelPrimary>()->getFuelDifferentialPressure().Value + Sensor::get(SensorType::Map).value_or(STD_ATMOSPHERE);
 #endif // EFI_PROD_CODE
 
 	engine->outputChannels.fuelTankLevel = Sensor::getOrZero(SensorType::FuelLevel);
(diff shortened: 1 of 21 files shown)


rusefillc:
@FDSoftware we are soooooo close

my bad. I did not mean 2x2 by default, I meant 2x8 by default - i.e. WITHOUT loosing the precision of xAsix of 8

FDSoftware:
updated:
image

rusefillc:
https://github.com/rusefi/rusefi/issues/7521

mck1117:
do y'all just not care about backward compat for existing tunes?

rusefillc:
@mck1117 I would hope that our actions like https://github.com/rusefi/rusefi/issues/7357 speak for themselves.

FDSoftware:
> do y'all just not care about backward compat for existing tunes?

I understood that all these migrations are done from the Java Tool updater, not on the firmware side where I am modifying.

mck1117:
> @mck1117 I would hope that our actions like #7357 speak for themselves.

I see no code to convert the table format, which has changed both shape and name.

firmware/console/status_loop.cpp
```diff
@@ -430,8 +430,9 @@ static void updateFuelSensors() {
 
 	engine->outputChannels.flexPercent = Sensor::getOrZero(SensorType::FuelEthanolPercent);
 #if EFI_PROD_CODE
-	// todo: extract method? do better?
+	// todo: extract method? do better? see https://github.com/rusefi/rusefi/issues/7511 for details
 	engine->module<InjectorModelSecondary>()->pressureCorrectionReference = engine->module<InjectorModelSecondary>()->getFuelDifferentialPressure().Value + Sensor::get(SensorType::Map).value_or(STD_ATMOSPHERE);
```
mck1117, line 434: if you don't have an SD card and/or TS connected, pressureCorrectionReference is never updated, so this feature is broken

rusefillc: Damn I've even asked myself that question yesterday but made wrong conclusion :( just added extra documentation.

Thank you!

mck1117: in general, two code smells that are good to eliminate:

1. any "load bearing" logic in `status_loop.cpp`
2. anything with the form of `engine->module<MyModule>()->myModuleMember = x`. Do it instead inside a member function of `MyModule`.

rusefillc: right, between mocks/overrides I am often afraid to touch these complex classes

note how https://github.com/rusefi/rusefi/commit/fb8f4486f9966296244ef975cecb67ece68cafeb was not a success requiring https://github.com/rusefi/rusefi/commit/ed4bf68c237793e24c2f2883ff8b2cf11c162095 follow-up

Issue #7515, opened 2025-02-26
TUNING DETECTED indicator and pause STFT if tuning is detected


rusefillc:
problem: STFT negatively affects tuning sessions of new users

proposal: automatically detect that tuning seems to be happening based on tables being changed within last XXX-TBD seconds?

if tuning is detected we light up indicator and pause STFT 

![Image](https://github.com/user-attachments/assets/33470d7e-42c9-4e2c-a7f0-0f848104aaf8)

rusefillc:
@ElDominio @dialedinmotorsports @mck1117 opinions? can you fill in the XXX blank?

ElDominio:
I like this idea, i wonder if it would help tuners or if it would make them angry

ggurov:
this should be toggleable to account for ve live autotune. live ve autotune looks at both EGO corr and current afr

ggurov:
> I like this idea, i wonder if it would help tuners or if it would make them angry

it should be toggleable, default to ON is fine, as long as this sort of forced intervention can be switched off

ElDominio:
> > I like this idea, i wonder if it would help tuners or if it would make them angry
> 
> it should be toggleable, default to ON is fine, as long as this sort of forced intervention can be switched off

I like this, because sometimes i leave STFT on while i tune and watch trim change in real time

rusefillc:
See https://github.com/rusefi/rusefi/issues/7513

kifir23917:
We need to add boolean TUNER DETECTED indicator on STFT dialog.

PR #7528, opened 2025-02-27
Now closed loop fuel correction is not performed when tuning seems to be happening #7515

kifir23917:


rusefillc:
@kifir23917 kin looks like legit failure here, please add relevant guard

Issue #7544, opened 2025-03-01
modeled flow idle


rusefillc:
https://github.com/FOME-Tech/fome-fw/pull/561

FDSoftware:
@rusefillc This can be closed now

PR #7545, opened 2025-03-01
modeled flow idle #7544

rusefillc:
experimental modeled flow idle

* implement basic idle flow model

* grr percent

* experiment with incorporating timing in to modeled idle

* missed a field

* live data, reset, scaling

* livedata


rusefillc:
@FDSoftware this one seems to be a duplicate

Issue #7502, opened 2025-02-25
second units specifically for injector flow


rusefillc:
Harley industry standard for fuel flow is grams per second and every one using this will be familiar with that

not even gallons per... per... what's freedom unit for time?

rusefillc:
@joshImpact25 @dialedinmotorsports forgive me for dumb question what's the conversion rate we will apply for fuel?

rusefillc:
g/s = 0.012 * cc/min ? not sure since I have "990; // 12.6g/sec" and math does not add up

benassi124:
Yes. G/S = cc/min * .012 for quick math I will find full math tomorrow.

rusefillc:
@FDSoftware probably following

```
#if LAMBDA
	gauge2		= lambda1Gauge
#else
	gauge2		= afr1Gauge
#endif
```

pattern. in MS they use this for C vs F maybe, anyway see MS .ini as reference?

FDSoftware:
> not even gallons per... per... what's freedom unit for time?

wikipedia says "Pound per hour (lb/h)" is the unit on US

https://en.wikipedia.org/wiki/Pound_per_hour#:~:text=Pound%20per%20hour%20is%20a,more%20conventionally%20as%20lb%2Fh.

joshImpact25:
For the Harley stuff industry standard is grams per second

rusefillc:
@FDSoftware it's a mess

* we already have a drop-down menu for units - including the desired grams per second
* but it has pretty... unclear naming
* and yes, there is the THIRD set of units which is lb/hr

I believe that minimal improvement is https://github.com/rusefi/rusefi/commit/8f42e81662708017047c78b3936aa3ba75552ce8


with that minimal improvement the whole urgency goes down a bit. Next open question is how to approach all three units

FDSoftware:
> but it has pretty... unclear naming

Yep, I thought that option configured something else, but it's much clearer now with that commit.

reading this on  `status_loop.cpp`

```cpp
static void updateFuelResults() {
#if EFI_VEHICLE_SPEED
	engine->outputChannels.fuelFlowRate = engine->module<TripOdometer>()->getConsumptionGramPerSecond();
	engine->outputChannels.totalFuelConsumption = engine->module<TripOdometer>()->getConsumedGrams();
.....
```
and: `injector_model.cpp`

```cpp
constexpr float convertToGramsPerSecond(float ccPerMinute) {
[...]

PR #7546, opened 2025-03-01
second units for injector flow #7502

FDSoftware:
added pound per hour as unit, keeping gram/sec as default

image
image


FDSoftware:
@ggurov  @ElDominio Does 100lb/h maximum seem good to you? Is it too exaggerated?

rusefillc:
@FDSoftware see https://github.com/rusefi/rusefi/issues/7502#issuecomment-2692285128

rusefillc:
not needed, we already have a confusing but working implementation

Issue #6977, opened 2024-10-09
A/C Idle Adder affects non-idle areas


ElDominio:
**Describe the bug**
A/C Idle adder is adding non-idle areas, affecting driveability. 

Shown at exactly 23.45s, TPS suddenly jumps without a change in PPS, then compressor engages after the chosen wait time.

**Hardware**
 - Board: AlphaX Platinum
 - Engine info LS3

**To Reproduce**
1. Use a non-zero idle value in A/C Idle Adder
2. Drive in non-idle areas

**Expected behavior**
A/C idle adder only affects ETB/Idle control when idleState = 1 (we are actually in idle)

**Additional context**
Tnue and log: https://rusefi.com/online/view.php?log=1427

ElDominio:
Still relevant: AC Idle should only be active when user demand (tps or pps) is less than the idle threshold

ElDominio:
I have fixed this in Lua by watching AC request state, grabbing idle PPS and AC idle extra with getCalibration and doing:
`if isAcRequest  and idlePPS  < accPedal then
			idleAdder = idleAdder - acIdleExtra
end`

Lua should not be the permanent fix for this

ElDominio:
@FDSoftware

ElDominio:
People while driving can hear the load and compensate with their foot. At least on this test car, current functionality makes the car accelerate when compressor is on, since at cruising rpm (about 1300), the additional air affects it quite a bit. 

And actually, on cars with idle  valves this behavior is imperceptible, while on ETB cars it is noticeable. I would think that's because we are  using the main air source for idle control instead of an additional valve that works independently

Maybe this ticket should be the behavior when ETB = true, and the old behavior with cable throttle setups.

PR #7559, opened 2025-03-04
A/C Idle Adder affects non-idle areas #6977

rusefillc:
step one: assert current behaviour


rusefillc:
@FDSoftware this fails as expected. Need your help to have this PR green - i.e. add new coverage but with current (defective) reality expectation

PR #7566, opened 2025-03-04
change idle default positon from slider to curve

FDSoftware:
same commit structure as #7520


FDSoftware:
image
diff report on TS:

Issue #7556, opened 2025-03-03
'-softPrepend`


rusefillc:
in bash script we have 

```-prepend ${BOARD_DIR}/prepend.txt```

which currently required some empty files

we should do smarter and add a feature of '-softPrepend` which would allow for files not to exist

**Comment by rusefillc** [comment]
end goal ``prepend_${SHORT_BOARD_NAME}.txt``

**Comment by rusefillc** [comment]
@FDSoftware can you actually add two different proteus files with only lua script line?
```

! f4 does not like this #define LUA_SCRIPT_SIZE 32000
! TODO what we need is a way to define one LUA_SCRIPT_SIZE for F4 and different one for F7
! todo: maybe some integration between meta_info and prepend?
#define LUA_SCRIPT_SIZE 13500
```

FDSoftware:
> [@FDSoftware](https://github.com/FDSoftware) can you actually add two different proteus files with only lua script line?
 proteus H7 also need the small lua script size, i will add another prepend for that target

FDSoftware:
@rusefillc Can this be closed or do we have to do something extra?

PR #7574, opened 2025-03-06
make prepend.txt optional

FDSoftware:
see: #7556


FDSoftware:
as example this pr also removes `nucleo_f767/prepend.txt`

Issue #7101, opened 2024-11-26
Boost Control open and closed loop Y axis


Artux0418:
**Describe the bug**
I think the closed and open loop use the same Y axis.
If I change the open loop, the values ​​in the closed loop also change. However, as soon as I select a value other than the Y axis for the open loop, it still says TPS is editable.

**Hardware**
 - Board: UAEFI121
 - Engine info VAG 1.8T

**To Reproduce**
1. Go to Advanced
2. Click on Boost Control Open loop
3. Scroll down to Y Axis

**Expected behavior**
I think the closed and open loop use the same Y axis.
If I change the open loop, the values ​​in the closed loop also change. However, as soon as I select a value other than the Y axis for the open loop, it still says TPS is editable.

**Screenshots**
![image](https://github.com/user-attachments/assets/11c6420b-cfd4-4b28-9f45-00567d37c560)

**Additional context**
Sorry if I'm not clear, I used a translation.

rusefillc:
@ElDominio looks like we've broken it around #4778:
yes, we still share axises, while only open look allows to customize what is Y axis

ElDominio:
> @ElDominio looks like we've broken it around #4778: yes, we still share axises, while only open look allows to customize what is Y axis

open looP?
Y axis should be TPS in most cases for **closed loop target**, using modifiers to change said target with other conditions.

FDSoftware:
@Artux0418 || @ElDominio This is already available in the build nightly, please test

rusefillc:
No reply from @Artux0418 @ElDominio unfortunately, have to close blindly

Issue #6843, opened 2024-08-22
separate boost control open loop axis from closed loop axis


rusefillc:
boost control target axis is only KPA
open loop is a bit of a hack with extra flexibility

those two tables have to use separate load axes (yes, 'axes' is plural for 'axis')

FDSoftware:
@rusefillc 

> No reply from @Artux0418 @ElDominio unfortunately, have to close blindly

this one is also the same issue

rusefillc:
I  think @FDSoftware says we can close this one

PR #7580, opened 2025-03-07
separate boost control open loop axis from closed loop axis

FDSoftware:
related issues:

* #7101
*  #6843 

image


ElDominio:
0 is best, values in this table usually lead to broken cars

Issue #7617, opened 2025-03-12
TCP proxy is not reliable


rusefillc:
have ECU on one device, run the following command to launch a proxy
```java -jar rusefi_console.jar local_proxy```

**Comment by rusefillc** [comment]
in the meantime https://sourceforge.net/projects/ser2net/ or https://sourceforge.net/projects/com0com/files/com2tcp/

**Comment by rusefillc** [comment]
@esaulenka see also https://gist.github.com/DraTeots/e0c669608466470baa6c#linux

**Comment by esaulenka** [comment]
It's also not reliable enough with local_proxy replaced by ser2net utility.
Maybe you have some advices with it.

```
I 250319 200052.241 [main] Launcher - rusEFI UI console 20250312
I 250319 200052.260 [main] Launcher - Compiled Wed Mar 19 07:16:27 AMT 2025

...

I 250319 200054.313 [ECU Communication Executor1] BinaryProtocol - rusEFI says tune CRC32 0x2dcfdc48 768597064

I 250319 200054.313 [ECU Communication Executor1] BinaryProtocol - rusEFI says tune CRC16 0xdc48 -9144

I 250319 200054.315 [ECU Communication Executor1] BinaryProtocol - Got configuration from controller 47220 byte(s)
I 250319 200054.356 [AWT-EventQueue-0] BinaryProtocolServer - BinaryProtocolServer on 2390
I 250319 200054.357 [AWT-EventQueue-0] StatusConsumer - ServerSocket 2390 created. Feel free to point TS at IP Address 'localhost' port 2390
I 250319 200056.449 [ECU Communication Executor1] MessagesCentral - postMessage EngineState: USB: got Query command
I 250319 200056.450 [ECU Communication Executor1] MessagesCentral - postMessage EngineState: TS <- Query signature: rusEFI master.2025.03.10.proteus_f7.2433621198
I 250319 200056.450 [ECU Communication Executor1] MessagesCentral - postMessage EngineState: TS <- Get CRC offset 0 count 47220 result 48DCCF2D
I 250319 200056.453 [ECU Communication Executor1] BinaryProtocol - Got livedata
I 250319 200056.454 [ECU Communication Executor1] ConnectionWatchdog - ConnectionWatchdog.reconnectTimer restarting: 1000
I 250319 200056.455 [ECU Commands Queue] BinaryProtocol - Sending [set date 2025-03-19T20:00:54.353982]
[...]

PR #7619, opened 2025-03-12
TCP proxy is not reliable #7617

rusefillc:
minor logging
only:uaefi


rusefillc:
@FDSoftware this could be a red flag

rusefillc:
@kifir23917 can you spend 30 minutes or less on this?

technically there is zero reason to touch other than looking for cleaner code

kifir23917:
> @kifir23917 can you spend 30 minutes or less on this?
Done

PR #7626, opened 2025-03-14
Minor: avoid double precision floats in the embedded part

esaulenka:
due to silly design, float constants without 'f' modifier considered as doubles, which is definitely not a good idea for STM32F4

also changed a few spaces to tabs fo align the code


mck1117:
this doesn't do anything, see `-fsingle-precision-constant` in the makefile

rusefillc:
code formatting https://github.com/rusefi/rusefi/commit/dd310c2de5640f93587fe507632c8e126a82f0bc but I would rather we find a better solution for more automated code formatting enforcement. We cannot afford to have humans deal with code formatting including manual code review of commits where code formatting is mixed with not-code formatting.

PR #7632, opened 2025-03-17
update deprecated java on java_tools/java_console

FDSoftware:
still WIP
notes:
The instability problem seems to be related to the current HTTP client, since it never terminates connections after use; the new method does so automatically, see: [CloseableHttpClient](https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/impl/client/CloseableHttpClient.html).
 I still have exceptions related to writing the .xml file, which seems to be related to an error caused by not having a TS signature on `java_console/inifile/src/main/java/com/rusefi/tune/xml/VersionInfo.java` func "validate".


rusefillc:
reminds me of https://github.com/rusefi/rusefi/pull/7551

FDSoftware:
> reminds me of #7551

the ideal thing would be to resume the progress of that PR, even with this one at least we'll get rid of a couple of Java warnings

FDSoftware:
continued at: https://github.com/rusefi/rusefi/pull/7637

PR #7641, opened 2025-03-20
smaller blocking_factor defect

rusefillc:
…e is not reliable via real network? #7617

only:proteus_f7


rusefillc:
https://github.com/rusefi/rusefi/tree/blocking-smaller
https://github.com/rusefi/rusefi/tree/getBlockingFactor
``int thisWrite = Math.min(remaining, VariableRegistryValues.BLOCKING_FACTOR);``

rusefillc:
@FDSoftware I do not follow, 64K seems sufficient for the proposed 32K packet?

FDSoftware:
> @FDSoftware I do not follow, 64K seems sufficient for the proposed 32K packet?

I was thinking of keeping 4 packets maximum in the buffer (as it happens with the default of 16k) instead of 2, but yes it should still work.

Issue #7616, opened 2025-03-12
taper down transition into idle duration should be configurable


rusefillc:
3 seconds hard-coded into https://github.com/rusefi/rusefi/issues/7612 should become configurable

PR #7655, opened 2025-03-25
taper down transition into idle duration should be configurable #7616

FDSoftware:
image
image

resolves: #7616


firmware/integration/rusefi_config.txt
```diff
@@ -1820,8 +1820,9 @@ uint8_t autoscale knockFuelTrim;Fuel trim when knock, max 30%;"%", 1, 0, 0, 30,
 
 	int8_t[8] airmassToTimingBins;;"mg", 1, 0, -100, 100, 0
 	int8_t[8] airmassToTimingValues;;"deg", 1, 0, -30, 30, 0
+	uint8_t autoscale idleReturnTargetRampDuration;idle return target ramp duration;"seconds", 0.1, 0, 0, 15, 1
```
rusefillc, line 1823: is zero a legit value here?

FDSoftware: @ElDominio may be able to help with the definition. Less than 1 second of ramp doesn't seem like a valid value to me. For now i've updated it to 0.1 and added the validation.

PR #7656, opened 2025-03-25
Simulator won't compile with newer GCC versions #6650

### LINKED ISSUE #6650: Simulator won't compile with newer GCC versions
I'm using GCC 14; if I recall correctly it also doesn't work with GCC 13

```
simulator/can/hal_can_lld.cpp: In function ‘void can_lld_stop(CANDriver*)’:
simulator/can/hal_can_lld.cpp:152:20: error: ignoring return value of ‘constexpr _FIter std::remove(_FIter, _FIter, const _Tp&) [with _FIter = __gnu_cxx::__normal_iterator<CANDriver**, vector<CANDriver*> >; _Tp = CANDriver*]’, declared with attribute ‘nodiscard’ [-Werror=unused-result]
  152 |         std::remove(instances.begin(), instances.end(), canp);
      |         ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/c++/14.1.1/algorithm:61,
                 from simulator/can/hal_can_lld.cpp:35:
/usr/include/c++/14.1.1/bits/stl_algo.h:788:5: note: declared here
  788 |     remove(_ForwardIterator __first, _ForwardIterator __last,
      |     ^~~~~~

```

FDSoftware:
works ok now with:
`gcc versión 14.2.1 20240912 (Red Hat 14.2.1-3) (GCC) `

resolves #6650

### DIFF
--- a/simulator/simulator/can/hal_can_lld.cpp
+++ b/simulator/simulator/can/hal_can_lld.cpp
@@ -149,7 +149,7 @@ void can_lld_stop(CANDriver *canp) {
 
 #if !EFI_SIM_IS_WINDOWS
 	// Remove from the "interrupt handler" list
-	std::remove(instances.begin(), instances.end(), canp);
+	(void)std::remove(instances.begin(), instances.end(), canp);
 
 	// Close the socket.
 	close(canp->sock);


rusefillc:
only manual test?

i guess CI for this would be a matrix on the github action using a few different compilers? but we do not care enough to justify the expense, just saying.

FDSoftware:
> only manual test?
> 
> i guess CI for this would be a matrix on the github action using a few different compilers? but we do not care enough to justify the expense, just saying.

There are workflows that can be configured to be executed manually, it could be tested in one of those as if it were a pre-release step. It would be too expensive in time to have to run the already large array of boards we have and add compiler variants on every commit/pr

PR #7658, opened 2025-03-25
solve some errors on validateConfigOnStartUpOrBurn/setDefaultBaseEngine

FDSoftware:
image

We are missing a couple of defaults that can cause unwanted CRC errors. cc @ggurov for reporting it

### DIFF
--- a/firmware/controllers/algo/defaults/default_base_engine.cpp
+++ b/firmware/controllers/algo/defaults/default_base_engine.cpp
@@ -360,6 +360,10 @@ void setDefaultBaseEngine() {
 	engineConfiguration->minimumOilPressureTimeout = 0.5f;
 	setRpmTableBin(config->minimumOilPressureBins);
 	setRpmTableBin(config->maximumOilPressureBins);
+
+	engineConfiguration->mapExpAverageAlpha = 1;
+	engineConfiguration->alternator_iTermMax = 1000;
+	engineConfiguration->alternator_iTermMin = -1000;
 }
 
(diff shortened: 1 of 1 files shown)


FDSoftware:
> @FDSoftware can you articulate the sequence which causes the issue?

In my case, after recompiling the simulator with pending changes in TS or when trying to save without the generated flashN.bin files, I get a CRC error. The manual solution is to set mapExpAverageAlpha to 1 and then save.

> ohhhh I am starting to wonder if a solution without duplicated constant would be to invoke that validate magic method after 'resetFirmware'?

I think this would solve the problem on real FW, but you could still have the same problem from the simulator side

rusefillc:
@FDSoftware you are explaining how to _reproduce_ the issue but let's have *cause* of issue in writing?

resetTune sets invalid zero value which is not corrected by 'validate' until next power cycler? 'validateConfigOnStartUpOrBurn' says 'onBurn' is it that 'onBurn' is not happening for simulator?

ggurov:
to reproduce: 

connect with ts
click "reset firmware"
file -> offline
file -> online 
accept controller values
change anything, click "burn"
see error

FDSoftware:
> @FDSoftware you are explaining how to _reproduce_ the issue but let's have _cause_ of issue in writing?
> 
> resetTune sets invalid zero value which is not corrected by 'validate' until next power cycler? 'validateConfigOnStartUpOrBurn' says 'onBurn' is it that 'onBurn' is not happening for simulator?

I'm not 100% sure about the entire callstack to be able to answer, I'll leave it in WIP and when I have free time from the rest of the tasks I'll review everything again.

PR #7678, opened 2025-04-04
detect and ignore doubled trigger edges [implement flag for disabling feature] see #4656

FDSoftware:
I found this in the middle of night lurk, I guess it's the correct way to implement this flag that was unused, was added on: ee5c9db [2023]


rusefillc:
https://github.com/rusefi/rusefi/commit/b5c5bca59a94c82c53e2c963fde8d1859f0f1559

rusefillc:
any extra code means extra tech support, extra chance for human error, extra requirement for testing, extra cost when modernizing

if no one can articulate a problem we should not add random code blindly

Issue #7150, opened 2024-12-09
DC motor wastegate position table feed forward open loop table


rusefillc:
4x4 1 byte S08 elements

ElDominio:
any progress on htis?

kifir23917:
@rusefillc, do we need increase this ticket priority?

ElDominio:
a comment was deleted is there anything on this?

ElDominio:
Tune at: https://rusefi.com/online/index.php?vehicleName=NegroWastegate&user_id=675&sorts[uploadedMsq]=-1&sorts[uploadedLog]=-1

Lua is inside tune, but important bit is:

wgPid = Pid.new(.01, 0.1, 0.005, -70, 70)
wgPid : setOffset(0.1)
wgPid : get(10, 10)
wgPid : reset()

startPwm(1, 500, 0.1) --pwm1 is pwm
startPwm(2, 100, 0) --pwm2 is dir
startPwm(3, 1, 0) --pwm3 is dis

wgPos = getSensor("WastegatePosition")
		targetWgPos = table3d(1, RPMread / 4, mapPres)
		-- p, i, d, min, max

		wgPidVal = wgPid : get(targetWgPos, wgPos)
		
		if wgPidVal < 0 then
			setPwmDuty(2, 0)
			wgPidVal = -wgPidVal
		else
			setPwmDuty(2, 1)
		end
		
		setPwmDuty(1, wgPidVal/10)
		wgPid : reset()

FDSoftware:
> 4x4 1 byte S08 elements
> feed forward open loop wastegate position table

@ElDominio I think the interface in TS needs to be made a little more intuitive; but currently, as soon as ETB-Wastegate is selected, the same table is used as feedforward as the PWM output, so there's no need to create a new one just for ETB.



Another additional thing is that we can now select both a PWM and ETB output at the same time to control the same Turbo, and I don't think that's a good idea!
See the images below:
![Image](https://github.com/user-attachments/assets/c50548a9-3c6a-41b3-bd0d-4a15a2b3b82f)

![Image](https://github.com/user-attachments/assets/96abf685-a28e-4d8e-b52d-839a8b4e8e21)

I propose changing the `enable` option from a bool to an enum so that you can select:
* disabled
* enabled (PWM)
* enabled (E-Waste Gate)

[...]

PR #7689, opened 2025-04-07
support for e-waste gate [new enum for boost control mode] see #7150

FDSoftware:


rusefillc:
@FDSoftware at the moment I do not get why do we need a new entity?

if boost solenoid pin is not empty - we are in solenoid PWM mode, would it help to extract helper method?

rusefillc:
@FDSoftware 
```
bool isBoostControlSolenoidMode() {
  return isBrainPinValid(engineConfiguration->boostControlPin);
}
```

FDSoftware:
> if boost solenoid pin is not empty - we are in solenoid PWM mode, would it help to extract helper method?

I'm thinking about changing the bit to an enum so that it's easier for the user to understand what they're configuring.

Only with `enable` as a bit, one can configure TS as follows, which would be "valid"

image

and on the fw side we are only in PWM mode (since there is a pin assigned)

With the enum I can hide the Output/Output mode and the entire e-waste menu depending on the case

FDSoftware:
(but if for now we can accept this edge case as out of scope, yes, we can just extract the method and it is solved)

Issue #7688, opened 2025-04-07
lqfp 208 support


rusefillc:
@ggurov has made some 208 progress which should inhale it

PR #7692, opened 2025-04-08
lqfp 208 support

rusefillc:
https://github.com/rusefi/rusefi/issues/7688


rusefillc:
@dron0gus FYI i've merged that one

Issue #7498, opened 2025-02-24
TDD for smoother idle threshold timing transition


rusefillc:
We have amazing https://github.com/rusefi/rusefi/pull/7045 which does not change existing coverage

let's practice some TDD by first merging coverage into master and then _adjust_ that coverage within https://github.com/rusefi/rusefi/pull/7045 around the affected area

PR #7703, opened 2025-04-11
TDD for smoother idle threshold timing transition see #7498

FDSoftware:
resolves: #7498


FDSoftware:
```
Expected equality of these values:
  10
  getRunningAdvance(rpm, load)
    Which is: 9.99999905
```

How I hate the differences in the implementation of floats, this on Linux compiles just fine ahhaha

I update pr using EXPECT_NEAR for both of these cases

PR #7712, opened 2025-04-14
SD card: show formating indicator while formating

dron0gus:
Also show return values to user.

TODO: store all FatFS returned errors to sd_error
Screenshot from 2025-04-14 15-21-11


dron0gus:
Cypress and Kin not happy because of not outdated `output_lookup_generated.cpp `? Why it is not regenerated?

PR #7713, opened 2025-04-14
binary logging: header size is 32 bit value

dron0gus:


dron0gus:
@rusefillc do we have mlg file format documentation?

PR #7727, opened 2025-04-16
ETB: use measured dt for PID

dron0gus:


dron0gus:
TODO: rework using `Timer`

PR #7733, opened 2025-04-17
ADC: timer HW trigger

dron0gus:
Now ADC is started by timer event without CPU intervention.
This removes 10.000 IRQ/sec.

### DIFF
--- a/firmware/config/stm32f4ems/efifeatures.h
+++ b/firmware/config/stm32f4ems/efifeatures.h
@@ -438,9 +438,9 @@
 
 // TODO: switch to continuous ADC conversion for fast ADC?
 // NOTE: GPT mode triggers ADC convertion through IRQ
-//#define EFI_INTERNAL_FAST_ADC_GPT	&GPTD6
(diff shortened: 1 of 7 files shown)


dron0gus:
HW trigger more disabled by last commit. Will be enabled when fully tested and confirmed no regression.

dron0gus:
@mck1117 can you review this? Any thoughts?

mck1117:
how will you support pin changes at runtime when adcConvert is called once near boot?

dron0gus:
> how will you support pin changes at runtime when adcConvert is called once near boot?

This is not used in current implementation anyway. All fast adc inputs are pre-configured in
https://github.com/rusefi/rusefi/blob/a1d021738567127056a6987bcb92194842a070b0/firmware/hw_layer/adc/adc_inputs.cpp#L207
And then used in:
https://github.com/rusefi/rusefi/blob/a1d021738567127056a6987bcb92194842a070b0/firmware/hw_layer/hardware.cpp#L294
No release or reconfiguration happens for fast ADC inputs.
So support of runtime pin change needs additional efforts in both cases.

Issue #7572, opened 2025-03-06
refactoring: leverage EngineModule more often


rusefillc:
https://github.com/rusefi/rusefi/commit/146522d42a9ec3bd722336a5f5c1373bb2190c45 etc

PR #7739, opened 2025-04-18
refactoring: leverage EngineModule more often [tachometer] see #7572

FDSoftware:
prep + first module


FDSoftware:
I will move the pr to draft, I have to continue having fun with the makefile

FDSoftware:
dropped for a different approach

Issue #7761, opened 2025-04-21
migrate FAST_CALLBACK_PERIOD_MS SLOW_CALLBACK_PERIOD_MS to main_loop.h


rusefillc:
See https://github.com/rusefi/rusefi/commit/e7b1bddc1ec2feb2d06742c13074962260107082 would that also change from 200Hz to 250hz?

FDSoftware:
> See [e7b1bdd](https://github.com/rusefi/rusefi/commit/e7b1bddc1ec2feb2d06742c13074962260107082) would that also change from 200Hz to 250hz?

I don't think the HPFP.PI test likes the change from 200 to 250, it uses the period to obtain the expected result

FDSoftware:
draft here adding 200hz as option & using it on the fast callback
https://github.com/rusefi/rusefi/pull/7763

rusefillc:
@FDSoftware please https://github.com/rusefi/rusefi/issues/7764

FDSoftware:
@rusefillc pls close this & next small step should be migrate SlowAdcController to new main loop?

rusefillc:
Let's have @dron0gus drive anything ADC

Can we mive with not ADC or are we depending?

FDSoftware:
> Let's have [@dron0gus](https://github.com/dron0gus) drive anything ADC
> 
> Can we mive with not ADC or are we depending?

etb & slow callback & fastcallback, but they are more critical, can I add the empty function and a placeholder for the calls? It wouldn't impact anything.

FDSoftware:
> > Let's have [@dron0gus](https://github.com/dron0gus) drive anything ADC
> > Can we mive with not ADC or are we depending?
> 
> etb & slow callback & fastcallback, but they are more critical, can I add the empty function and a placeholder for the calls? It wouldn't impact anything.

done at: https://github.com/rusefi/rusefi/pull/7771

[this is only for easy tracking if someone visits this issue in the future]

PR #7763, opened 2025-04-21
migrate FAST_CALLBACK_PERIOD_MS see #7761

FDSoftware:


firmware/controllers/core/main_loop.h
```diff
@@ -7,6 +7,7 @@ enum class LoopPeriod : uint8_t {
 	Period1000hz = 1 << 0,
 	Period500hz = 1 << 1,
 	Period250hz = 1 << 2,
+  	Period200hz = (1 << 2) + 1, // why not the plain 5?
```
rusefillc, line 10: @FDSoftware note ``inline constexpr LoopPeriod& operator|=`` and ``inline constexpr bool operator&`` I would assume it's something to do with bitmap masks? or is that what you are trying to do here?

i.e. we have to be careful I would assume. Sounds like we need a unit test :)

FDSoftware: > @FDSoftware note `inline constexpr LoopPeriod& operator|=` and `inline constexpr bool operator&` I would assume it's something to do with bitmap masks? or is that what you are trying to do here?
> 

I saw the bitmasks but at the same time this enum ends up being an int8, the rest of the bitmasks also end in numbers that are equivalent to the amount of ms like the previous code, that's why it caused me some confusion

> i.e. we have to be careful I would assume. Sounds like we need a unit test :)

FDSoftware: @rusefillc after using tests from #7765, adding a non bit-shift value like 5ms to the enum breaks the logic (and the new tests) on makePeriodFlag, I am going to review the HPFP.PI test to update it to the new execution rate and revert the 200hz change

rusefillc: @FDSoftware can we add a value between existing ones? Shift 50hz one bit?

FDSoftware: on `HpfpQuantity::calcPI` we have this code using the fast callback define
```c++
float possibleValue = m_pressureTarget_kPa - (engineConfiguration->hpfpTargetDecay *
			(FAST_CALLBACK_PERIOD_MS / 1000.));
```
image

The affected test cases are "Test pressure decay" and "Integral gain", which both depend on the execution time.

@racer-coder do you agree with the changes in the tests?

FDSoftware: > @FDSoftware can we add a value between existing ones? Shift 50hz one bit?

I think I can add one at the end like ` 1 << 4,` now I'm going to try it

FDSoftware: > It's a very very VERY good indication that if an enum has the values specified as `1 << n`, then it's a bit field, not an arbitrary value. Enums for internal use (ie, they don't go in the config) don't really have any other reason to specify values.

thx! I was a little confused about this enum thing.

Issue #7760, opened 2025-04-21
execute critical tasks on one big thread


rusefillc:
we shall take best or most of https://github.com/FOME-Tech/fome-fw/commit/01a397094d3305058966c396e553d0fcab95c11d

PR #7771, opened 2025-04-23
execute critical tasks on one big thread see [preparation] #7760

FDSoftware:
related issue: #7760


FDSoftware:
@dron0gus @rusefillc
The most important thing is in main_loop.cpp see comments marked as "//TODO: main_loop"

PR #7782, opened 2025-04-24
20250424 canbus buttonbox

ggurov:


rusefillc:
https://github.com/rusefi/rusefi/commit/746bc6be10c31d6c538da1aee3130a68b7e767ac

rusefillc:
So what is all this? Why is this not CANbus and/or Lua? TS protocol smells like UART/USB?

ggurov:
> So what is all this? Why is this not CANbus and/or Lua? TS protocol smells like UART/USB?

TS protocol serial. in this case it's serial encapsulated in CAN, which ends up making the exact same `Z` command call.

there's no reason to reinvent THIS wheel to signal the ECU to execute something. this infrastructure already exists in the form of TS communication protocol, the can version just happens to be serial encapsulated in CAN, it also has crc32, and length specified. a few more layers of protection against

ggurov:
button box that speaks TS, has serial, and bluetooth hardware onboard could also be beneficial for no wire button box that uses same implementation over bluetooth, for an even lazier approach that doesn't require any wires for a button box.

rusefillc:
@ggurov are you aware of the sad story of LCD screen and joystick? some painful lessons were learned with that bloat functionality.

I have no capacity to ever support this code, it belongs in Lua at this point.

ggurov:
not aware. did joystick or screen allow modification of runtime variables as is the plan for this, with expansion as simple as make as "if a ts button can do it, so can remote canbus button" ?

ggurov:
lua is out. not user friendly enough.



so where do we go from here ? 

"remote button box" is bloat and does not belong ?
this implementation is not the right way ?

what's the answer here, that's not lua ?

Issue #7776, opened 2025-04-24
maybe a defect around pt2001 timeout to trigger warning for pump


rusefillc:
#7717 

```
2025-04-24_11_40_54_873: EngineState: Time is changed to
2025-04-24_11_40_54_875: EngineState: Current RTC time: 24 Apr 2025 11:40:54
2025-04-24_11_40_54_876: EngineState: confirmation_set date 2025-04-24T11:40:54.3849644:36
2025-04-24_11_41_02_495: EngineState: WARNING: C9013: Injection HPFP pump time excess PT2001 limits time: 31226.1699
2025-04-24_11_41_04_582: EngineState: WARNING: C9002: primary trigger error: too many teeth between sync points: expected 58/0 got 58/0
2025-04-24_11_41_05_247: EngineState: WARNING: C9003: primary trigger error: not enough teeth between sync points: expected 58/0 got 12/0
2025-04-24_11_41_05_654: EngineState: WARNING: C9004: cam VVT B1 Int trigger error: too many teeth between sync points: 8 > 8
2025-04-24_11_41_05_754: EngineState: [err] VVT1 rpm=211 time=274 eventIndex=9 gapIndex=0: n gap=0.436 expected from 0.483 to 0.806 error=Yes
2025-04-24_11_41_05_755: EngineState: [err] VVT1 rpm=211 time=274 eventIndex=9 gapIndex=1: n gap=0.459 expected from 1.167 to 1.944 error=Yes
2025-04-24_11_41_05_866: EngineState: [err] VVT1 rpm=211 time=274 eventIndex=11 gapIndex=0: n gap=0.280 expected from 0.483 to 0.806 error=Yes
```

rusefillc:
@FDSoftware definitely a defect

```engine->module<HpfpController>().unmock().m_deadangle + static_cast<angle_t>(engineConfiguration->hpfpActivationAngle);```

m_deadangle is dead time - that's part of duration
hpfpActivationAngle configuration is POSITION, not duration? i.e. has nothing to do with duration check?

FDSoftware:
> m_deadangle is dead time - that's part of duration

comment on `m_deadangle` says `///< Computed solenoid deadtime in degrees`
for dead time is `deadtime_ms` I assume?
dead angle math is:

`m_deadangle = deadtime_ms * rpm * (360.f / 60.f / 1000.f);`

> hpfpActivationAngle configuration is POSITION, not duration? i.e. has nothing to do with duration check?

[...]

PR #7783, opened 2025-04-24
defect around pt2001 timeout to trigger warning for pump see #7776

FDSoftware:
* ~LimpManager is now mockeable~ now at https://github.com/rusefi/rusefi/pull/7784
* GdiPumpLimit case now uses `m_requested_pump`
* refactor of test & fix false positive (the previous test wasn't actually working correctly, the `finalHpfpPumpAngletime`  was 0! and the test didn't notice)

related issue: #7776

### DIFF
--- a/firmware/controllers/limp_manager.cpp
+++ b/firmware/controllers/limp_manager.cpp
@@ -225,14 +225,15 @@ void LimpManager::updateState(float rpm, efitick_t nowNt) {
 #if EFI_HPFP
 	{
 		// HPFP Fuel cut
-		angle_t m_deadangle = engine->module<HpfpController>().unmock().m_deadangle;
-		angle_t finalHpfpPumpAngle = m_deadangle + static_cast<angle_t>(engineConfiguration->hpfpActivationAngle);
+		angle_t m_deadangle = engine->module<HpfpController>()->m_deadangle;
+		angle_t m_requested_pump = engine->module<HpfpController>()->m_requested_pump;
+		angle_t finalHpfpPumpAngle = m_deadangle + m_requested_pump;
 		float finalHpfpPumpTimeMs = US2MS(static_cast<float>(finalHpfpPumpAngle) * engine->rpmCalculator.oneDegreeUs);
 
 		if (engineConfiguration->tempPumpLimitCheck && isGdiEngine() && finalHpfpPumpTimeMs > engineConfiguration->mc33_hpfp_max_hold) {
 			allowFuel.clear(ClearReason::GdiPumpLimit);
 			warning(ObdCode::CUSTOM_TOO_LONG_FUEL_INJECTION, "Injection HPFP pump time excess PT2001 limits time: %.4fms", finalHpfpPumpTimeMs);
-			efiPrintf("%f %f %f", m_deadangle, (float)engineConfiguration->hpfpActivationAngle, engine->rpmCalculator.oneDegreeUs);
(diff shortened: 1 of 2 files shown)


FDSoftware:
@rusefillc updated pr, only contains changes related to the pt2001 check

firmware/controllers/limp_manager.cpp
```diff
@@ -225,14 +225,15 @@ void LimpManager::updateState(float rpm, efitick_t nowNt) {
...
 		// HPFP Fuel cut
-		angle_t m_deadangle = engine->module<HpfpController>().unmock().m_deadangle;
-		angle_t finalHpfpPumpAngle = m_deadangle + static_cast<angle_t>(engineConfiguration->hpfpActivationAngle);
+		angle_t m_deadangle = engine->module<HpfpController>()->m_deadangle;
+		angle_t m_requested_pump = engine->module<HpfpController>()->m_requested_pump;
+		angle_t finalHpfpPumpAngle = m_deadangle + m_requested_pump;
 		float finalHpfpPumpTimeMs = US2MS(static_cast<float>(finalHpfpPumpAngle) * engine->rpmCalculator.oneDegreeUs);
 
-		if (engineConfiguration->tempPumpLimitCheck && isGdiEngine() && finalHpfpPumpTimeMs > engineConfiguration->mc33_hpfp_max_hold) {
+		if (engineConfiguration->tempPumpLimitCheck && isGdiEngine() && finalHpfpPumpTimeMs > engineConfiguration->mc33_hpfp_max_hold){
```
rusefillc, line 233: grrrr you took my space away from me?!

rusefillc:
@racer-coder does this error detection condition even make sense?

firmware/controllers/limp_manager.cpp
FDSoftware: fixed :joy:

racer-coder:
> @racer-coder does this error detection condition even make sense?

Make sense as in "will this check determine when the PT2001 would trigger an error?"  I think so.  I guess the question is whether the PT2001 timeout is configured correctly.

rusefillc:
> Make sense as in "will this check determine when the PT2001 would trigger an error?" I think so. I guess the question is whether the PT2001 timeout is configured correctly.

@racer-coder "detect that things are so misconfigured that we should prohibit injection with a clear error indicator"

racer-coder:
> > Make sense as in "will this check determine when the PT2001 would trigger an error?" I think so. I guess the question is whether the PT2001 timeout is configured correctly.
> 
> @racer-coder "detect that things are so misconfigured that we should prohibit injection with a clear error indicator"

but this isn't injection.  It's just pressurizing the fuel rail.  At least on my car, there was a physical/mechanical relief valve so the rail wouldn't go over a certain pressure (max running was 200 bar, relief happened at 250 bar).  I could see it trigger when my car sat after running, as the residual heat from the engine would heat up the fuel in the rail, causing the pressure to rise...

The damage here is overheating the solenoid on the HPFP, maybe causing damage to the solenoid and requiring replacement (not cheap part, on my car the whole pump had to be replaced).  But having the timeout in the PT2001 theoretically prevents that, right?

FDSoftware:
> But having the timeout in the PT2001 theoretically prevents that, right?

Yup! What @rusefillc is asking is,if I'm not mistaken, is that with this PR we could have the same check and not have to resort to doing it from PT2001 (since we can't give feedback to the user that there's something "wrong" about the gdi config from the PT2001 side)

rusefillc:
killing the whole pump limp code

PR #7788, opened 2025-04-25
ETB: PID: set iTerm limits on first init too

dron0gus:


rusefillc:
@dron0gus this breaks unit tests

a sad way would be to #ifdef UNIT_TESTS guard some of the code and create a ticket for technical debt follow-up?

rusefillc:
Another option maybe set iTerm limits to zero in unit tests, that might help

rusefillc:
@dron0gus I've addressed kin drama, so if you rebase this would turn green

your call on draft/not draft

dron0gus:
@rusefillc ready.

PR #7790, opened 2025-04-25
main_loop: add ADC to main loop

dron0gus:
Part of 01a397094d3305058966c396e553d0fcab95c11d

### DIFF
--- a/firmware/controllers/core/main_loop.cpp
+++ b/firmware/controllers/core/main_loop.cpp
@@ -45,7 +45,7 @@ void MainLoop::PeriodicTask(efitick_t nowNt) {
 
 #if HAL_USE_ADC
 	if (currentLoopPeriod & ADC_UPDATE_RATE) {
-		// TODO: main_loop adc callback
+		updateSlowAdc(nowNt);
 	}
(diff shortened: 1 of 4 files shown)


dron0gus:
ADC's `updateSlowAdc()` is executed at 500Hz rate.
`MainLoop` executed at 1000 Hz rate.
So each second time execution of other MainLoop task is delayed by `updateSlowAdc()`
Slow ADC conversion is done using `adcConvert()`. This sets task to pause waiting for ADC conversion done.
So this introduce delay for all other tasks in loop. But not each time.
ETB controller expects to be executed evenly. While it is executed at same 500Hz rate - it will be delayed by ADC on each cycle.
But we want 200Hz GDI... This makes things more complex.

rusefillc:
@dron0gus might mean that we have make GDI 250hz?

FDSoftware:
> @dron0gus might mean that we have make GDI 250hz?

The only test that broke on GDI due to the change to 250hz was this one we saw a couple of days ago:

https://github.com/rusefi/rusefi/pull/7763#discussion_r2053320931
image

PR #7803, opened 2025-04-29
7760: mega loop adc prepare

dron0gus:


dron0gus:
@rusefillc I don't know how to fix Cypress. It does not generate some autogenerated headers?

PR #7819, opened 2025-04-30
enable CEL on honda K

ElDominio:


rusefillc:
@ElDominio i assume you know what you are doing so let me just merge these

ElDominio:
I have had complaints of Check Engine light not turning on when key is on which has made me sad

I can manage complaints of check engine lights that come on for other reasons, 99.9999999999% sure without causing additional ruckus

Issue #7832, opened 2025-05-04
TuneCanTool does not produce valid code for GPPWM


rusefillc:
* checkout https://github.com/rusefi/rusefi/tree/canned-tune-tool-issue branch
* open IDEA
* invoke TuneCanTool.java main method

expected: valid code generated in simulator\generated\canned-tunes folder

actual result at least these lines are invalid

```
	copyArray(engineConfiguration->gppwm[3].gppwm[1].loadBins, hardCodedgppwm2_loadBins);
```

FDSoftware:
The problem is related to the "getOriginalArrayName" function (`cName` in the debug view), which is returning both the name of the array and the associated parent, using cf.getName() (`cf.name` in the debug view) solves the problem
![Image](https://github.com/user-attachments/assets/2db37bdd-e143-4ba1-b507-9d02dcb0b034)

FDSoftware:
> `gppwm[3].gppwm1_table` should be `gppwm[3].table` no?

After off-github discussion, the correct return would be the following, since the .msq file affects table number 1 and not 4

![Image](https://github.com/user-attachments/assets/c7bc902e-d4b4-4294-879d-ef308e0709eb)


now much nicer output of the script (pending to send PR, will edit this to the link)

```c++
static void cannedgppwm2_rpmBins() {
	static const float hardCodedgppwm2_rpmBins[8] = {0.0, 1000.0, 2000.0, 3000.0, 4000.0, 5000.0, 6000.0, 7777.0};
	copyArray(engineConfiguration->gppwm[1].rpmBins, hardCodedgppwm2_rpmBins);
}

static void cannedgppwm2_loadBins() {
	static const float hardCodedgppwm2_loadBins[8] = {0.0, 14.0, 28.0, 40.0, 57.0, 71.0, 85.0, 100.0};
	copyArray(engineConfiguration->gppwm[1].loadBins, hardCodedgppwm2_loadBins);
}

static void cannedgppwm2_table() {
	static const float hardCodedgppwm2_table[8][8] = {
{55,000,	55,000,	55,000,	55,000,	11,000,	11,000,	11,000,	11,000,	},
{55,000,	55,000,	55,000,	55,000,	11,000,	11,000,	11,000,	11,000,	},
{55,000,	55,000,	55,000,	55,000,	11,000,	11,000,	11,000,	11,000,	},
{55,000,	55,000,	55,000,	55,000,	11,000,	11,000,	11,000,	11,000,	},
[...]

PR #7836, opened 2025-05-04
TuneCanTool does not produce valid code for GPPWM see #7832

FDSoftware:


FDSoftware:
@rusefillc pr against the issue branch is ok or do I do it against master?

Issue #6711, opened 2024-07-13
Fields.java is a mess


rusefillc:
Fields.java is too much of too many unrelated things mixed together, some sort of decoupling is tempting

rusefillc:
See also https://github.com/rusefi/rusefi/issues/6077

rusefillc:
@ggurov says file size limit is too close!

PR #7838, opened 2025-05-04
only adding blend Y to output_channels.txt breaks HW CI

rusefillc:
only:preparing


rusefillc:
@FDSoftware why is this branch failing? is this a Makefile issue of something not being re-generated or else?

FDSoftware:
> @FDSoftware why is this branch failing? is this a Makefile issue of something not being re-generated or else?

Only the firmware makefile regenerates the output channels correctly, the simulator or the tests usually have problems, for example here: (in the capture, that field ceased to exist because it was renamed in the .txt)

image

rusefillc:
@FDSoftware that screenshot is from older branch right?

please help with exact root cause in this branch with failing HW CI

FDSoftware:
It took me a good couple of tries, but I think I found the problem (although I'm still failing to find the solution :c)
after this commit from 2 months we stop updating `TsOutputs` java code,
https://github.com/rusefi/rusefi/commit/198edefb7efef3b68c5cc6adcb48ac797577c41d
If we add something to output_channels.txt and try to compile it, whether it be tests, simulator, firmware or console, this file is not updated by adding the corresponding output var / offset. After this the test related to VSS cannot obtain the correct value from the sensor and the test fails.

rusefillc:
caused by https://github.com/rusefi/rusefi/issues/6711

@FDSoftware let me fix it, great job pointing at the problem!

@dron0gus we know what the issue is

Issue #7869, opened 2025-05-10
map window averaging does not use crank angle


rusefillc:
start angle zero, width 100 what is this expected to be in engine phase?

for instance below we've used start if 0

![Image](https://github.com/user-attachments/assets/969bee54-d482-4753-8021-db6f87a1af33)

rusefillc:
we suspect multiplication by two?

rusefillc:
this code seems to older when we were scheduling all of these on first tooth?

```
		angle_t offsetAngle = engine->triggerCentral.triggerFormDetails.eventAngles[0];
...
...
			float cylinderStart = start + cylinderOffset - offsetAngle + tdcPosition();

```

while we now schedule in relation to closes tooth?
```
		if (!isPhaseInRange(samplingStart, currentPhase, nextPhase)) {
			continue;
		}
```

so proper logic should NOT have ``eventAngles[0]`` and not have ``tdcPosition``

but probably ``samplingStart - currentPhase``?

rusefillc:
follow up ticket https://github.com/rusefi/rusefi/issues/7880

PR #7879, opened 2025-05-13
fix: "map window averaging does not use crank angle" see #7869

FDSoftware:
related issue: #7869

now using `samplingStart - currentPhase` as map averager start

image

detail on time active:
image

log:
[2025-05-13_01.21.25.mlg.zip](https://github.com/user-attachments/files/20181120/2025-05-13_01.21.25.mlg.zip)

map averager configured as:
image


rusefillc:
https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing#worst

PR #7890, opened 2025-05-15
only:reducing gap between PROD and tests

rusefillc:


rusefillc:
@FDSoftware how hard would be to fix this branch? it would help https://github.com/rusefi/rusefi/pull/7891

rusefillc:
@FDSoftware I've rebases on master with your other fix and we have a different kind of error now, only Windows?

PR #7892, opened 2025-05-15
trigger debug? wow.

rusefillc:
only:dead


rusefillc:
@FDSoftware this depends on https://github.com/rusefi/rusefi/pull/7891

PR #7896, opened 2025-05-15
fix isPinConfigurationChanged on UT case

FDSoftware:


rusefillc:
@FDSoftware on the one hand I've cherrypicked this into master https://github.com/rusefi/rusefi/commits/master/

on the other hand windows is unhappy

PR #7900, opened 2025-05-15
Map revert of revert

rusefillc:
@FDSoftware please branch of this branch, please PR new tests asserting that this branch is broken. New tests should be passing, new tests should be asserting wrong results.


rusefillc:
@FDSoftware do we have a plan for this PR / do we need this PR?

Issue #7700, opened 2025-04-09
Dual bank F7 interrupts execution while erasing/writing flash following "burn" command


ggurov:
**Describe the bug**
Dual bank F7, in dual bank mode, interrupts execution while erasing/writing flash after clicking "burn"

**Hardware**
 - Board: F767ZIT6 Nucleo, F767BIT6   - dual bank 2mb MCU
 - Engine info N/A
 - 
**To Reproduce**
1. Flash latest appropriate F7 code (f7 proteus,  f7 nucleo)
2. Configure any channel for ignition output
3. Connect Logic analyzer to that ignition output
4. Go to ECU Stimulator
5. Start stimulator
6. Start live capture on logic analyzer
7. Change a value, click burn, causing flash to be written
8. Notice a break in ignition signals, and pause in the serial data coming in.

This will likely stall the engine, as ignition and injection go away for about 2 seconds, leaving channels in whatever state they were:

**Screenshots**

![Image](https://github.com/user-attachments/assets/24729fae-a9fa-47d8-a945-f8943ca42236)

**Additional context**

This is pretty dangerous behavior as there could be  1-2 seconds of fuel sprayed, and ignition coils may be left charging for 1-2 seconds.

rusefillc:
@dron0gus do we have/can we have any official F7 burning in background? if not we need code comments why

rusefillc:
@dron0gus see https://github.com/rusefi/rusefi/commit/cccc4cec52ddfcdda920d4d6732a742acc926673

dron0gus:
@ggurov I assume you are testing debug build?
`2mb_flash.mk` is enabled for nuclue_f767. This is dangerous because allows up to 1.5Mb for code. 0.5 of this is located in second bank, that is used for settings store. So when settings are writen to second bank any code execution from this high 0.5 Mb will freeze CPU until write/erase is finished.
Changing "-O0" optimization level to "-Og" (still good for debug) significantly reduces code size and it fits in 1Mb.
With this modification we can remove dangerous `2mb_flash.mk` hack.

Also see https://github.com/rusefi/rusefi/issues/6242 https://github.com/rusefi/rusefi/commit/45ac95e2493b4fffdbdd12b38b74024a354c59b8

ggurov:
i tried a bunch of different combinations, including the proteus build on build server, on a dual bank f7, that also freezes while writing.

[...]

PR #7914, opened 2025-05-19
7700 burn freeze f7

dron0gus:
Do not allow code to occupie second bank on dual-bancked F7 boards. Change '-O0' to '-Og' optimization level to fit code in 1 Mb.
Current binary size is about 700Kb. With 32 Kb reserved for bootloader we still have about 250Kb room.

https://github.com/rusefi/rusefi/issues/7700


dron0gus:
`2mb_flash.mk` is left enabled for Subaru EG33 board because it stores settings in external QSPI flash.

PR #7975, opened 2025-05-30
Disable VVT when the target is close to 0

rusefillc:


firmware/controllers/actuators/vvt.cpp
```diff
@@ -55,6 +55,16 @@ void VvtController::onConfigurationChange(engine_configuration_s const * previou
 	}
 }
 
+static bool shouldInvertVvt(int camIndex) {
+	// grumble grumble, can't do an array of bits in c++
```
awpe, line 59: camIndex can be just part of type so this function at least is a consteval one

PR #7981, opened 2025-05-31
throw an exception if more than one file match pattern,…

### LINKED ISSUE #7883: find file to throw an exception if more than one file match pattern
See todo https://github.com/rusefi/rusefi/commit/e1a1689c22ba8786f17ceffd0fbc0c2a2e737f50

FDSoftware:
resolves #7883


FDSoftware:
@rusefillc  Any hints on how to add a test for this? Should I extract the logic to another function and test that?
now I was testing with this:
https://onlinegdb.com/rreLYsj_r

PR #8002, opened 2025-06-03
fix test_vvt.cpp tests for disable-vvt-close-to-zero

FDSoftware:
we need `m_engineRunningLongEnough` / `m_isRpmHighEnough` flags on true before trying to get vvt target with `getSetpoint`


rusefillc:
picked into https://github.com/rusefi/rusefi/pull/7975

PR #8018, opened 2025-06-05
Haltech can wideband support + TS configs + gcc14

awpe:


rusefillc:
@awpe it's absolutely unacceptable to mix WBO and GCC in on PR. Let's absolutely make sure that next PR does not mix things which are so unrelated

Again I see generated stuff etc etc etc. Please understand my context of very limited availability to code review. Quick reminder that each commit into a PR is an email for me - so ten commits is ten emails, meaning a small but distraction etc etc etc.


Let's absolutely make sure that next PR has only 
* human-produced changes
* focuses on exactly one thing
* does not change more than a dozen of files. that probably means that to begin with we should compromise and not rename the configuration flag.

rusefillc:
@awpe https://github.com/rusefi/rusefi/commit/0734054fd9890e4cd28640d7ebfdc02e90687c29

PR #8024, opened 2025-06-05
Force LogField static const to .rodata to save RAM

awpe:


rusefillc:
@awpe build is red

awpe:
More investigation needed on why next address is not evaluated during compile time:
extern Engine ___engine;
static Engine * const engine = &___engine;
seems this prevent "static const LogField fields[] = {" being automatically put into ro without any hints...

Issue #8035, opened 2025-06-06
Change "base idle valve position"  Y axis to "Final Target Idle RPM" from "RPM"


ElDominio:
**Is your feature request related to a problem? Please describe.**
idle valve position is directly related to the RPM we are trying to acheieve in idle conditions, not the current RPM of the engine (at least in base). Closed Loop brings actual engine RPM into the loop and then modifies this to acheive happy idle

**Describe the solution you'd like**
Change "base idle valve position"  Y axis to "Final Target Idle RPM" from "RPM"

**Additional context**

![Image](https://github.com/user-attachments/assets/0928a1bb-0fa5-4cfd-97fd-7ef19aec01ac)

ElDominio:
Y needs to be Target Idle RPM from table "Target RPM"

ElDominio:
@FDSoftware @rusefillc

rusefillc:
@ElDominio this ticket is closed. what are you trying to say? actual behavior? expected behavior?

ElDominio:
Actual behavior = Y is current engine RPM
expected = Y is Target RPM (as was articulated on OP)

I comment on this ticket since it was referenced on #8142 and it seems misunderstanding this ticket is root cause

rusefillc:
@ElDominio let's start from beginning. What do you call "Y"?

A: "tunerstudio shows a dot and I am sure the dot is what firmware is using"
B: "engine is not controlled right"?

ElDominio:
> [@ElDominio](https://github.com/ElDominio) let's start from beginning. What do you call "Y"?
> 
> A: "tunerstudio shows a dot and I am sure the dot is what firmware is using" B: "engine is not controlled right"?

Y = y axis in the table "idle position by engine temperature"
a: firmware is 20250628
b: idle control Y axis has no point atm

extra: hardware is hellen154hyundai

dialedinmotorsports:
Y-axis currently serves no purpose for the feedforward idle position for idle control, as it follows _ACTUAL ENGINE RPM_.

Y-Axis needs to follow the _TARGET Idle RPM_ , from the table with the same name, for the feedforward to be effective.


Step 1: Take the value input by user in "Idle Target RPM vs Engine Temp" 2D Curve

Step 2: Implement that value as the Y-axis variable in the "Idle Base Position vs Engine Temp" table.

The closed loop algorithm needs to look at what "Idle RPM" is desired by the user, and use it as a variable to lookup the feedforward idle valve position.

[...]

PR #8039, opened 2025-06-07
Change "base idle valve position" Y axis to "Final Target Idle RPM" f…

rusefillc:
…rom "RPM" #8035

only:uaefi


rusefillc:
@ElDominio turned to be a one line change, want to test it with your local build? change is much easier than actually testing change here :(

PR #8052, opened 2025-06-09
-Not yet working for some reason- Gear position sensor for sequential gearboxes

EragonPower:
Addition of gear position sensor for sequential gearboxes, even for motorbike's gearboxes based on a 0-5V GPSwitch.

I've tried to add that sensor as a linear sensor, like the one of oil pressure, also tried to add a gauge, but was unsuccessfull at both, i'd love to receive feedback / critiques / suggestions

### DIFF
--- a/firmware/init/sensor/init_gbps.cpp
+++ b/firmware/init/sensor/init_gbps.cpp
@@ -1,0 +1,42 @@
+#include "pch.h"
+
+#include "init.h"
+#include "adc_subscription.h"
+#include "functional_sensor.h"
+#include "linear_func.h"
+
+static LinearFunc gearboxposSensorFunc;
+static FunctionalSensor gearboxposSensor(SensorType::GearboxPositionSensor, /* timeout = */ MS2NT(50));
+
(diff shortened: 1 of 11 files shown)


firmware/init/sensor/init_gbps.cpp
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  float val1 = cfg.value1;
+  float val2 = cfg.value2;
+
+  // Limit to max given pressure - val1 or val2 could be larger
+  // (sensor may be backwards, high voltage = low pressure)
+  float greaterOutput = val1 > val2 ? val1 : val2;
+
+  // Allow slightly negative output (-5kpa) so as to not fail the sensor when engine is off
```
rusefillc, line 26: 5kpa? https://en.wikipedia.org/wiki/Copy-and-paste_programming?

rusefillc:
I think this should simply be a Lua script? and I would probably stay in voltage domain without the conversion to angular position

one day a somewhat articulated criteria would be nice but as of today I do not see enough bike usage to justify the cost of maintenance of this feature if this was to me merged.

PR #8068, opened 2025-06-13
It is now possible to switch gcc/clang on local dev machine for unit tests builds with coverage

awpe:
Unit tests for Mac OS use mingw to get gcc available in env. However for Mac OS gcc from mingw is a placeholder for llvm/clang thus results of builds are different from pure gcc builds. Sometimes to look into those differences (at least to acquire warnings from clang that are not thrown from gcc) there is a need to build all with clang on local dev machine which is not Mac OS (i.e. on any recent Linux distro you get both gcc and clang). This change allows to run make with CC=clang option so all unit tests are built on your local machine with clang instead of gcc. This change handles coverage compile/linking options to satisfy differences between clang and gcc.


awpe:
All green:
https://github.com/awpe/rusefi/actions

PR #8006, opened 2025-06-03
tests-are-failing-on-windows

rusefillc:
github is decommissioning windows-2019 runners

looks like switching from  windows-2019 to windows-2022 causes our units tests to fail?!

PR #8071, opened 2025-06-13
there is probably signal in the fact that unit tests have started to …

rusefillc:
…fail in Windows somewhere in 2025

only:windows

### DIFF
--- a/.github/workflows/build-unit-tests-windows.yaml
+++ b/.github/workflows/build-unit-tests-windows.yaml
@@ -1,8 +1,8 @@
 name: Unit Tests on Windows
 
 on:
-#  push:
-#  pull_request:
+  push:
+  pull_request:
   workflow_dispatch:
 
 jobs:


rusefillc:
just rebased
@awpe let's see if I get really lucky and this fixes unit tests segfault on Windows

awpe:
Seems no luck so far...

awpe:
Do you know name of test where it fails? Or random?

rusefillc:
> Do you know name of test where it fails? Or random?

image

and then many (all?) tests which read .csv

```
TEST(real4g93, crankingCamOnly) {
	EngineCsvReader reader(1, /* vvtCount */ 0);

	reader.open("tests/trigger/resources/4g93-cranking-cam-only.csv");
```

maybe more than one separate issues?

rusefillc:
@awpe oh we have https://github.com/rusefi/rusefi/pull/8006

awpe:
Will check. It is extremly benefitial to have multiple platforms with multiple compilers doing same things... Would help a lot to see what is going on windows after any code change

awpe:
https://github.com/rusefi/rusefi/blob/master/unit_tests/test-framework/engine_test_helper.cpp#L436
WTF

FDSoftware:
> https://github.com/rusefi/rusefi/blob/master/unit_tests/test-framework/engine_test_helper.cpp#L436 WTF

The objective of that if is to compare that the callbacks are equal, it is also used in:
`EngineTestHelper::assertEvent5` and `EngineTestHelper::assertTriggerEvent`, I just recycled part of "assertTriggerEvent", if there's a saner way I'm totally open to changing it

FDSoftware:
see also usage of the function here:
https://github.com/rusefi/rusefi/pull/8034

awpe:
> > https://github.com/rusefi/rusefi/blob/master/unit_tests/test-framework/engine_test_helper.cpp#L436 WTF
> 
> The objective of that if is to compare that the callbacks are equal, it is also used in: `EngineTestHelper::assertEvent5` and `EngineTestHelper::assertTriggerEvent`, I just recycled part of "assertTriggerEvent", if there's a saner way I'm totally open to changing it

straight from code it does no comparison, only check is for nullptr of callback i.e. it ends up as:

if (reinterpret_cast<void*>(callback) != nullptr)
{
...
}

which is subsititiatable just by:

if (callback){...}

awpe:
I found the problem. There is UB in the code when reinterpret casting void* to uintptr and back while in a meanwhile modifying that pointer. That is not guaranteed to produce correct results as length can be different so manipulated part of memory can be different. So any scheduled event in any test can basically end up with some shifted number to be reinterpreted as memory address pointer.

Check next page to play around different platforms (windows tests now use mingw 12.2). Seems like gcc from 10 till most recent as well as clang from 15 till most recent are ok with that due to Linux being strict in what can be done with addresses (no shifts happen while casting here and there).

https://godbolt.org/z/3dz44eP14

awpe:
What is scary is the fact that gcc for eabi on arm32 gives failures...

awpe:
And arm64 gcc is ok... 

I will add test code for reproduction here in case godbolt removes it.

```
#include <cassert>
#include <cstdint>
#include <cstdio>

struct InjectionEvent {
    int x = 123;
    bool y;
};

struct FuelSchedule {
    int x = 456;
    InjectionEvent arr[12];
    bool y;
};

void test_round_trip_aligned_ptr() {
    FuelSchedule s;
    auto& d = s.arr[1];
    uintptr_t original = reinterpret_cast<uintptr_t>(&d);
    void* asVoid = reinterpret_cast<void*>(original);
    uintptr_t back = reinterpret_cast<uintptr_t>(asVoid);

    printf("original = 0x%016llx, back = 0x%016llx\n", (unsigned long long)original, (unsigned long long)back);
    assert(original == back);
}

void test_round_trip_with_tag() {
    FuelSchedule s;
    auto& d = s.arr[1];
    uintptr_t tagged = reinterpret_cast<uintptr_t>(&d) | 1;
    void* asVoid = reinterpret_cast<void*>(tagged);
    uintptr_t back = reinterpret_cast<uintptr_t>(asVoid);

    printf("tagged = 0x%016llx, back = 0x%016llx\n", (unsigned long long)tagged, (unsigned long long)back);

    assert(tagged == back);
}

void test_tagged_ptr_no_void_star() {
    FuelSchedule s;
    auto& d = s.arr[1];
    uintptr_t base = reinterpret_cast<uintptr_t>(&d);
    uintptr_t tagged = base | 1;
    uintptr_t untagged = tagged & ~1UL;

    printf("base = 0x%016llx, tagged = 0x%016llx, untagged = 0x%016llx\n",
           (unsigned long long)base,
           (unsigned long long)tagged,
           (unsigned long long)untagged);

    assert(base == untagged);
}

struct ActionSLike {
    void* ptr;

    uintptr_t getRaw() const {
        return reinterpret_cast<uintptr_t>(ptr);
    }
};

void test_action_s_behavior() {
    FuelSchedule s;
    auto& d = s.arr[1];
    uintptr_t tagged = reinterpret_cast<uintptr_t>(&d) | 1;

    ActionSLike a = { reinterpret_cast<void*>(tagged) };
    uintptr_t recovered = a.getRaw();

    printf("tagged = 0x%016llx, recovered = 0x%016llx\n", (unsigned long long)tagged, (unsigned long long)recovered);

    assert(tagged == recovered);
}

int main() {
    test_round_trip_aligned_ptr();
    test_round_trip_with_tag();
    test_tagged_ptr_no_void_star();
    test_action_s_behavior();
    return 0;
}

```

awpe:
Unless we are in plain C and want to unload tons of #if macroses around it is better to switch to uintptr_t completely that will remove UB and then such failure can only occur in case of compiler bug.

rusefillc:
> Unless we are in plain C and want to unload tons of #if macroses around it is better to switch to uintptr_t completely that will remove UB and then such failure can only occur in case of compiler bug.

please PR?

rusefillc:
@awpe thank you again!

Issue #8080, opened 2025-06-14
MapAveragingModule_onEnginePhase test


rusefillc:
@awpe reports on discord
```
Anyone eager to check and fix MapAveragingModule_onEnginePhase test?
assertEventExistsAtEnginePhase has typo in condition making it always true 
fixing condition breaks test
```

awpe:
https://github.com/rusefi/rusefi/blob/master/unit_tests/test-framework/engine_test_helper.cpp#L436

awpe:
Check on godbolt.org:
```
int main() {
    if (false, true) {
        return 1;
    } else {
        return 2;
    }
}
```

rusefillc:
@FDSoftware @awpe https://github.com/rusefi/rusefi/pull/8081

PR #8081, opened 2025-06-14
MapAveragingModule_onEnginePhase test #8080

rusefillc:
wow typo?
only:uaefi


awpe:
one test will fail, Im not sure how to fix it, seems like reference data and timings in tests are not aligned

rusefillc:
@FDSoftware @awpe I am confused: looks like we have engine_test_helper.cpp already fix in master as part of the pointer safety? and master is green?

awpe:
it was about test_map_averaging, I disabled one branch in one test after fixing ETH, other tests were still green after ETH fix, but map averaging was failing at the end

rusefillc:
@FDSoftware https://github.com/rusefi/rusefi/pull/8099

FDSoftware:
> @FDSoftware #8099

can we close this PR no? 8099 already has the revert of the commented test and the fix that I uploaded to this branch, so this one is no longer necessary.

PR #8083, opened 2025-06-15
STFT: missed parts of TS template

dron0gus:


dron0gus:
Not sure why we have same defines in `rusefi_config.txt` and `rusefi_config_shared.txt`, but change fuel trim defines in `rusefi_config_shared.txt` too.

PR #8084, opened 2025-06-15
Now modern compilers are happier and actions in scheduler are safe from poiner cast/modify/cast disaster

awpe:
Thanks to the weirdness of MinGw for Windows memory address issue was identified and later confirmed on godbolt.

The main purpose of this PR is to fix action_s usage in scheduler so no UB is present thus allowing many compilers to do builds successfully and apps to not corrupt addresses.

Tested action_s + TaggedPointer on various platforms and compilers using godbolt (gcc 10-15, clang 15-20, mingw, msvc, gcc arm 10 -14, arm32, arm64.

Now action_s has safe and convinient (compile-time) interface and callback name for testing/debugging purposes so you can see callback name in scheduled events container during debug session or add std::cout calls where needed in tests.

Class action_s has been enriched with many compile time checks and helpers so handling of function pointers and object pointers (arguments) is correct from language perspective so no UB unless some specific compiler bug is present. With those checks it is now hard to pass smth fancy to action_s without modifying it as it will not compile. Also in case weird platform is discovered it won't compile either.

Testing ASM output shows no overhead for action_s if compiled with -O1 (on -O0 it is expected to have each and every step in ASM).

Secondary goal happened to be fortification of some also incorrect but yet not fired places which were bloating compile log on any recent compilers and/or platforms and significantly spoiling dev experience.


firmware/controllers/trigger/trigger_central.cpp
```diff
@@ -484,7 +483,7 @@ void hwHandleShaftSignal(int signalIndex, bool isRising, efitick_t timestamp) {
 
 	handleShaftSignal(signalIndex, isRising, timestamp);
 }
-
+extern bool b;
```
rusefillc, line 486: hello bool b, how are you doing?

awpe:
https://github.com/awpe/rusefi/actions/runs/15664609300

awpe:
You can fire unit tests on windows as well here

rusefillc:
let's see how many smaller commits can I grab from this amazing PR

https://github.com/rusefi/rusefi/pull/8085

awpe:
Should I make this one master compliant?

rusefillc:
> Should I make this one master compliant?

let me go over it another nine times later today? please maybe remove 'b' or I can do within https://github.com/rusefi/rusefi/pull/8085?

awpe:
I will update to master and remove that declaration

awpe:
BTW: https://github.com/rusefi/rusefi/commit/c2bf1d4b2cd15cadb6c5cd4e7e689ae54da7a34c

Are you sure original action_s was supporting non arg callbacks? I removed arg from that and other functions with unused args because new action_s has explicit support for that.

awpe:
updated
again

awpe:
Could you please start windows unit tests workflow on that PR?

PR #8086, opened 2025-06-15
todo: document pros and cons

rusefillc:
@awpe please document pros and cons, looks like a three line snippet (hack?) is being replaced with a lot of nicely documented magic?


awpe:
Thats just covering sane warnings from compiler that original check is not doing what it is named to be doing. There are calls in the code where smth like 0xffffffff is converted to float or some EFI_ERROR_STATE or how it is called is converted to float and many implicit int to float conversions to store then that float into 32bit int, original check could have made erroneous assumptions on whether float can pass to int ot not. So I made it bulletproof like with many static_assertions showing and testing corner cases in case compiler/platform is of some weird type.

PR #8111, opened 2025-06-19
WBO: individual FW update

dron0gus:


rusefillc:
@dron0gus tests seem to be red

dron0gus:
@rusefillc should be fixed now.
With new config options added both AFR channel configured to the same WBO index on fresh boards.
Do we care about this?
Screenshot from 2025-06-19 15-49-47

PR #8140, opened 2025-06-23
closed_loop_idle.cpp smoothLtitTable test

FDSoftware:


FDSoftware:
For macOS, `MockIdle` in this test collides with the existing one in `test_ignition_state.cpp`, I'm not sure why, but when editing the ignition_state with the methods I use here, or renaming this one, the test runs correctly, (maybe gmock thinks they are the same and the Mockable<> set doesn't run correctly?)

fix on windows and valgrind fix on ubuntu was the use of ltitTableHelper before being initialized

tested on my fork:
image
https://github.com/FDSoftware/rusefi/pull/22

PR #8153, opened 2025-06-25
Jeep 36-2-2 trigger

dron0gus:


rusefillc:
@dron0gus do you have matching .logicdata or only converted to .csv? tradition is to persist both

dron0gus:
@rusefillc logicdata file added. Still needs to be tested on real car.

Issue #8149, opened 2025-06-24
SmoothedAFRValue and SmoothedAFRValue2 always zero with CAN WBOs


dron0gus:
Both are used only in conjunction with analog AFRs.

PR #8155, opened 2025-06-25
Fix "SmoothedAFRValue and SmoothedAFRValue2 always zero with CAN WBOs"

FDSoftware:
resolves: #8149
depends on: #8154


rusefillc:
@FDSoftware please rebase

PR #8164, opened 2025-06-27
LTFT: real log test

dron0gus:
Test with binary log takes significant time. Should we disable it?
We need better/longer logs for real testing.


rusefillc:
replaced by https://github.com/rusefi/rusefi/pull/8256

PR #8190, opened 2025-07-04
engine module, live data

rusefillc:
only:example


rusefillc:
@FDSoftware ask here is to have one commit with everything needed to add EngineModule with live data

rusefillc:
@FDSoftware wider context that's related to https://github.com/rusefi/rusefi/pull/8195

Issue #7964, opened 2025-05-28
remove some 4cyl engine types


rusefillc:
we have too many hard coded 4 cylinder engines

can we remove FORD_ASPIRE_1996 by adjusting tests? maybe remove some other 4 cyl engines like dodge?

PR #8193, opened 2025-07-04
remove Mitsubishi 4G93 configs

FDSoftware:
related: https://github.com/rusefi/rusefi/issues/7964


FDSoftware:
@rusefillc Maybe I'm missing something, but this has been empty since 2023, so I assume it didn't end up being used?

PR #8195, opened 2025-07-04
Norman ae

rusefillc:


rusefillc:
@ElDominio I see at least one obvious defect: you add new field accelEnrichmentMode but you do not remove the original one. That's between sloppy and dangerous.

rusefillc:
@ElDominio on the bright side it's green - all code compiles and no tests fail. Would be great to add a test @ElDominio 


@FDSoftware only FYI right now

rusefillc:
@ElDominio crazy

```
		float blendDuration = engineConfiguration->mapPredictionBlendDuration;
(blendDuration > 0
```

PR #8204, opened 2025-07-05
remove FORD_ESCORT_GT

FDSoftware:
also removed the `mazda_miata.h/cpp` since no engine was in this files and the function was used only on ford escord
related #7964


FDSoftware:
hwci not entirely happy, I'm going to re-add the files and try to extract as much as possible until the test fails, maybe the issue is near the trigger type?

edit for more context:

here an on the aspire PR i was trying to set a trigger like i do on the normal UT, found this func, maybe i am implementing bad?

`ecu.sendCommand(IoUtil.setTriggerType(com.rusefi.enums.trigger_type_e.TT_FORD_ASPIRE));`

PR #8232, opened 2025-07-08
hardware.cpp: one more hook to board code boardInitHardwareEarly()

dron0gus:
called before any attemp to init storage. Going to be used to prepare SPI interface for flash driver.


rusefillc:
@dron0gus @awpe see https://github.com/rusefi/rusefi/commit/7185c2891c983d32704888547a7accdd8b1f8603

PR #8246, opened 2025-07-09
Testing engine configuration for state injection has a bug visible with recent compilers

awpe:
Test TestEngineConfiguration has an error at:

void TestEngineConfiguration::configureInjectorBattLagCorr(const std::optional<BattLagCorrTable> battLagCorr)...

and same error for secondary injection.

Error is in accessing std::optional internals right after checking there is nothing inside. Older compilers seem to jump over google test macro covered code after such access thus passing over like nothing was done at all. 

The solution so far is to let copy happen if value is present and do nothing if not.

Changes were made by @kifir23917 last time. Maybe he can take a look what was the intent as currently it is not clear what this code was supposed to do if there was no table passed.

I added a comment to the code:

// This whole thing is wrong as you can not access std::optional which was found to be empty.
// On top of that this method is called after explicitly setting test config with test curve data
// (TEST_PRIMARY_INJECTOR_BATT_LAG_CORR_CURVE) with curve filled by 5.6f
// But then it is compared against default config table with very different data from firmware defaults
// in file engine_configuration_defaults.h (INJECTOR_BATT_LAG_CURR)
// { 4.240f, 2.483f, 1.739f, 1.501f, 1.308f, 1.149f, 0.964f, 0.913f },
// { 3.084f, 1.641f, 1.149f, 1.194f, 0.992f, 0.759f, 0.637f, 0.603f },
// Seems like it was hidden by some std::optional+gtest bug...

Newer compilers will fail execution on attempt to access empty std::optional (which is correct behavior)

### DIFF
--- a/unit_tests/tests/binary_log/test_bit_logger_field.cpp
+++ b/unit_tests/tests/binary_log/test_bit_logger_field.cpp
@@ -19,7 +19,7 @@ class BitLoggerFieldTest : public ::testing::Test {
 
         std::unique_ptr<output_channels_s> m_testOutputChannels;
         std::unique_ptr<LogField> m_logField;
-        std::array<char, 6> m_buffer;
+        std::array<uint8_t, 6> m_buffer{};
     };
 
     void BitLoggerFieldTest::SetUp() {
@@ -42,7 +42,7 @@ class BitLoggerFieldTest : public ::testing::Test {
     }
 
     void BitLoggerFieldTest::checkBitLoggerField(const bool expected, const char* const context) {
-        EXPECT_EQ(1, m_logField->writeData(m_buffer.data(), nullptr)) << context;
+        EXPECT_EQ(1, m_logField->writeData(reinterpret_cast<char*>(m_buffer.data()), nullptr)) << context;
         EXPECT_THAT(m_buffer, ::testing::ElementsAre((expected ? 0x01 : 0x00), 0xAA, 0xAA, 0xAA, 0xAA, 0xAA));
     }
 
--- a/unit_tests/tests/util/test_engine_configuration.cpp
+++ b/unit_tests/tests/util/test_engine_configuration.cpp
@@ -335,6 +335,16 @@ void TestEngineConfiguration::configureInjectorBattLagCorr(const std::optional<B
     if (battLagCorr.has_value()) {
         copyTable(engineConfiguration->injector.battLagCorrTable, battLagCorr.value());
     } else {
+        // This whole thing is wrong as you can not access std::optional which was found to be empty.
+        // On top of that this method is called after explicitly setting test config with test curve data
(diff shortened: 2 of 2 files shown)


rusefillc:
@awpe please improve ticket description probably along the lines of https://xyproblem.info/

awpe:
Please check comment in code

awpe:
should I copy it here to pr description?

rusefillc:
@awpe I am failing to comprehend

maybe leverage https://en.wikipedia.org/wiki/Inverted_pyramid_(journalism) - start by four word summary?

rusefillc:
tests are green under https://github.com/rusefi/rusefi/pull/8250

awpe:
> tests are green under #8250

Just to clarify - does it mean bug should be kept until github migrates to gcc 15.1.1 and/or clang 20.1.6 so someone else will have to work it around?

https://en.cppreference.com/w/cpp/utility/optional/operator%252A.html

std::optional<T>::operator->
...
If has_value() is false, the behavior is undefined. (until C++26)

rusefillc:
> Just to clarify - does it mean bug should be kept until github migrates to gcc 15.1.1 and/or clang 20.1.6 so someone else will have to work it around?

definitely not

I am just trying to extract all the possible value from fact that we have first issue specific to newer gcc. See https://github.com/rusefi/rusefi/pull/8251 which is very much not finished yet

awpe:
I don't want to be that toxic guy :-D But honestly there is an obvious UD here which does UD thing as each UD does, the question is does whole case deserve so much attention?

kifir23917:
I would suggest https://github.com/rusefi/rusefi/pull/8255 as a better way to fix a problem with accessing empty optional.

rusefillc:
@awpe is that the one?

```
[----------] 8 tests from FuelDifferentialPressureTest
[ RUN      ] FuelDifferentialPressureTest.checkDefault
/opt/gcc-15/include/c++/15.1.0/optional:1165: constexpr const _Tp* std::optional<_Tp>::operator->() const [with _Tp = std::array<std::array<float, 8>, 2>]: Assertion 'this->_M_is_engaged()' failed.
/home/runner/work/_temp/521529a7-8d46-4712-8df2-21d4017e890e.sh: line 1: 404207 Aborted                 (core dumped) ASAN_OPTIONS=detect_stack_use_after_return=1 build/rusefi_test
```


```
Run gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/opt/gcc-15/libexec/gcc/x86_64-pc-linux-gnu/15.1.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: ../gcc-15-source/configure --prefix=/opt/gcc-15 --disable-multilib --enable-languages=c,c++
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 15.1.0 (GCC) 
```

awpe:
Yes

PR #8251, opened 2025-07-09
Gcc15 tests

rusefillc:


rusefillc:
@awpe @FDSoftware cute - GHA timeouts trying to build GCC?

rusefillc:
```
Run gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/opt/gcc-15/libexec/gcc/x86_64-pc-linux-gnu/15.1.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: ../gcc-15-source/configure --prefix=/opt/gcc-15 --disable-multilib --enable-languages=c,c++
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 15.1.0 (GCC) 
```

success in terms of installing 15

rusefillc:
runtime failure with is unique to gcc15 https://github.com/rusefi/rusefi/pull/8246#issuecomment-3054071077

rusefillc:
> @awpe @FDSoftware cute - GHA timeouts trying to build GCC?

wow sometimes it barely makes it!

image

FDSoftware:
> > @awpe @FDSoftware cute - GHA timeouts trying to build GCC?
> 
> wow sometimes it barely makes it!
> 
> image

Could it be that it's stuck in the installation steps? Normally, apt-get starts in interactive mode, so it may require user confirmation to install a package. Something like this might work better:

`DEBIAN_FRONTEND=noninteractive apt install -y flex bzip2`

rusefillc:
> Could it be that it's stuck in the installation steps? Normally, apt-get starts in interactive mode, so it may require user confirmation to install a package. Something like this might work better:
> 


i was wrong - it does not always timeout. Sometimes it's successful in 1:57 and still has time to run tests

rusefillc:
@awpe with rebase it fails differently

PS: that's not unit test job, that's just random!?

image

FDSoftware:
> > Could it be that it's stuck in the installation steps? Normally, apt-get starts in interactive mode, so it may require user confirmation to install a package. Something like this might work better:
> 
> i was wrong - it does not always timeout. Sometimes it's successful in 1:57 and still has time to run tests

viewing the logs, what seems to take the longest is compiling gcc 15. If we plan to use it for a long time (since there is still no Ubuntu distro with an official package), we can create a base image with gcc15 and then reuse it in the workflow so that it is only compiled once.

PR #8253, opened 2025-07-09
Map estimation tooltip on table editor see #8210

### LINKED ISSUE #8210: MAP Estimation table tooltip
This table represents MAP at a given TPS vs RPM, which we use if our MAP sensor has failed, or if we are using MAP Prediciton. This table should be a direct representation of MAP, you can tune it manually by disconnecting MAP sensor, and filling out the table with values that match an external gauge that shows MAP.

FDSoftware:
image
closes  #8210


rusefillc:
Wow I did not realize tables can have tooltips?!

FDSoftware:
> Wow I did not realize tables can have tooltips?!

It is not documented in the ini file reference, but we are already using them in some places like the idle/injector lag tables

Issue #8249, opened 2025-07-09
let's try listeners for boardInitHardware() and boardInitHardwareExtra()


rusefillc:
https://github.com/rusefi/rusefi/commit/c7e16ad7bc9020db14a7afb77222a09f9128eb69boardInitHardware

FDSoftware:
https://github.com/rusefi/fw-custom-example
https://github.com/rusefi/fw-custom-mega100
https://github.com/rusefi/fw-custom-paralela
https://github.com/rusefi/fw-custom-hellen144-f4
https://github.com/rusefi/fw-custom-mega-legacy
https://github.com/rusefi/fw-custom-test
https://github.com/rusefi/fw-custom-z31-f7
https://github.com/rusefi/fw-custom-z31-h7
https://github.com/rusefi/fw-custom-Cat-GDI
https://github.com/rusefi/fw-custom-core8

All of these do not use `InitHardware` but `setBoardDefaultConfiguration` which is also PUBLIC_API_WEAK

rusefillc:
@FDSoftware https://github.com/rusefi/rusefi/issues/8269

PR #8257, opened 2025-07-09
prep: new board override system for boardInitHardware/boardInitHardwareExtra see #8249

FDSoftware:
I thought about taking advantage of the fact that we are using more c20, we have std::invoke available to be able to invoke functions without having to write a cast

result:
```
[0us]efiPrintfInternal:msg`*** rusEFI LLC (c) 2012-2024. All rights reserved.`
[0us]efiPrintfInternal:msg`*** rusEFI v20250709@1140675965 now=0`
[0us]efiPrintfInternal:msg`*** Chibios Kernel:       7.0.5`
[0us]efiPrintfInternal:msg`*** Compiled:     Jul  9 2025 - 18:26:56`
[0us]efiPrintfInternal:msg`*** COMPILER=14.2.1 20240912 (Red Hat 14.2.1-3)`
[0us]efiPrintfInternal:msg`custom board hello from simulator`
``` 

If we need to invoke a custom board function or a default one, we can also add something like

```c++
std::invoke(custom_board_boardSayHello.value_or(default_board_boardSayHello));
```
related: #8249

### DIFF
--- a/firmware/hw_layer/board_overrides.h
+++ b/firmware/hw_layer/board_overrides.h
@@ -34,7 +34,7 @@ using setup_custom_board_overrides_type = void (*)();
  * Allows boards to perform custom initialization before HAL is initialized
  * Example: special pin setup or hardware initialization needed before HAL
  */
-// extern std::optional<setup_custom_board_overrides_type> custom_board_preHalInit;
+extern std::optional<setup_custom_board_overrides_type> custom_board_preHalInit;
 
 extern std::optional<setup_custom_board_overrides_type> custom_board_boardSayHello;
 
--- a/firmware/hw_layer/board_overrides.h
+++ b/firmware/hw_layer/board_overrides.h
@@ -1,0 +1,47 @@
+/*
+ * @file board_overrides.h
+ * @brief Board-specific override mechanism
+ *
+ * This header provides a centralized way to define and manage board-specific overrides.
+ * Custom boards can implement their own specialized behavior by overriding these function pointers.
+ *
+ * Usage Example:
+ *
+ * 1. Define your custom function:
+ *    void myBoardCustomHello() {
+ *        // Your custom implementation
+ *    }
+ *
+ * 2. Set up the override in your board_configuration.cpp:
+ *    void setup_custom_board_overrides() {
+ *        custom_board_boardSayHello = myBoardCustomHello;
+ *    }
+ *
+ *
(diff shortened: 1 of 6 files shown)


firmware/hw_layer/board_overrides.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @date: jul 09, 2025
+ * @author FDSoftware
+ */
+
+#pragma once
+#include <functional>
+#include <optional>
+
+using setup_custom_board_overrides_type = void (*)();
```
FDSoftware, line 29: this cursed `void (*)()` thing has to be read as "function with no parameters and returning void"

FDSoftware: added comment

firmware/hw_layer/board_overrides.h
```diff
@@ -0,0 +1,48 @@
...
+
+// function with no parameters and returning void
+using setup_custom_board_overrides_type = void (*)();
+
+/**
+ * @brief Pre-HAL initialization override point
+ * Allows boards to perform custom initialization before HAL is initialized
+ * Example: special pin setup or hardware initialization needed before HAL
+ */
+// extern std::optional<setup_custom_board_overrides_type> custom_board_preHalInit;
```
rusefillc, line 37: let's go without example, just convert a few methods and have those act as examples

firmware/main.cpp
```diff
@@ -13,10 +13,21 @@
...
+PUBLIC_API_WEAK void setup_custom_board_overrides() {
+}
+
+// std::optional<setup_custom_board_overrides_type> custom_board_preHalInit;
 
 int main(void) {
+	setup_custom_board_overrides();
 	// Maybe your board needs to do something special before HAL init
 	preHalInit();
+	// call_board_override(custom_board_preHalInit);
```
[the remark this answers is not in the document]
FDSoftware: placeholder, I didn't know if it was right to migrate this or not, I didn't want to make a very large PR

firmware/hw_layer/board_overrides.h
FDSoftware: > let's go without example, just convert a few methods and have those act as examples

done in 68aff94 for nucleo F429
and in c3a2436 for nucleo H743

which are the only ones that use the old `preHalInit`

PR #8262, opened 2025-07-10
better handling of missing TS page

### LINKED ISSUE #8248: better handling of missing TS page
at the moment we throw critical for default page https://github.com/rusefi/rusefi/commit/06f5ed00721475c755edf711cd966242ad25c864 see https://github.com/rusefi/rusefi/pull/8208

we better return protocol error

FDSoftware:
resolves #8248

### DIFF
--- a/firmware/console/binary/tunerstudio.cpp
+++ b/firmware/console/binary/tunerstudio.cpp
@@ -181,7 +181,7 @@ static uint8_t* getWorkingPageAddr(TsChannelBase* tsChannel, size_t page, size_t
 #endif
 	default:
 // technical dept: TS seems to try to read the 3 pages sequentially, does not look like we properly handle 'EFI_TS_SCATTER=FALSE'
-		criticalError("unhandled getWorkingPageAddr");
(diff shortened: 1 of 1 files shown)


firmware/console/binary/tunerstudio.cpp
```diff
@@ -181,7 +181,7 @@ static uint8_t* getWorkingPageAddr(TsChannelBase* tsChannel, size_t page, size_t
 #endif
 	default:
 // technical dept: TS seems to try to read the 3 pages sequentially, does not look like we properly handle 'EFI_TS_SCATTER=FALSE'
-		criticalError("unhandled getWorkingPageAddr");
+		tunerStudioError(tsChannel, "ERROR: page address out of range");
 		return nullptr;
```
rusefillc, line 185: 'return ptr' i have a feeling that it would cause a seg fault? look at getWorkingPageAddr() usages?

rusefillc: > Could it be that we have outdated code in the library used by wideband, and that's where the confusion comes from?

ooops, that's what got me confused

PR #8290, opened 2025-07-14
spi flash dev: w25q, cache workaround

dron0gus:


rusefillc:
@dron0gus tests and mre_legacy are red

Issue #8298, opened 2025-07-15
All boards attempt to write while running


ggurov:
**Describe the bug**
All boards are attempting to write while running ,stalling engines, possibly flooding them. etc.

**Hardware**
 - Board: any

**To Reproduce**
1. load latest daily
2. start engine
3. change anything
4. click burn
5. engine stalls.

**Expected behavior**
writes to be deferred to when engine is not rotating

**Additional context**
https://github.com/rusefi/rusefi/blob/master/firmware/controllers/flash_main.cpp#L77-L100

has no "false" return ever.

PR #8299, opened 2025-07-16
fix "All boards attempt to write while running"

FDSoftware:
see #8298


FDSoftware:
I make this pr based on the comments in the code, I think it might be useful to you @dron0gus

dron0gus:
@FDSoftware thanks for fixing this and thanks for test!

PR #8302, opened 2025-07-16
Farenheit support

rusefillc:


rusefillc:
@FDSoftware I've just rebased this one even while it was only showing two commits

Issue #7947, opened 2025-05-23
A dream about trigger shapes: a tool to locate gap sequence based on logic analyzer file brute force


rusefillc:
problem: at the moment we need a human to suggest gap sequence

proposed solution: if we have a logic analyzer file, and if we know tooth count, it's actually trivial to brute force all gap lengths and and subsets of real life gaps and report all possible trigger sync sequences

PR #8337, opened 2025-07-21
trigger decode, we need to adjust error handling

rusefillc:
…logic analyzer file brute force #7947

only:progress


Sammyhmc1:
I'm having issues with my Tercel 95 I spoke to you already on Messenger yesterday I was able to start the car but was showing trigger error what can I do I believe it's unit issue  I really want to fix this but I don't know anything about tuned  etc etc

PR #8339, opened 2025-07-22
separate "writeOneField" callback from writing scalars/arrays/bits on the final .ini on TsOutput.java

FDSoftware:
same/part of #8313 pr, but separating on small changes, here I only organize code different, I need to call this `writeFieldJob` 2 times for the C/F case, since I need to write the conditional, but don't move the `FieldIterator` size or `tsPosition`, which would happen if I call `writeOneField` recursively


FDSoftware:
@rusefillc Sorry for the wall of commits, but it's the target branch (rusefi:freedom-units) that is outdated with master. By updating that branch, there should be only one commit left 

(alternatively, I can close this pr and create another one based on that branch instead of master)

rusefillc:
@FDSoftware i was force pushing, this shows 48 commits, need your help to clean-up/rebase

FDSoftware:
> @FDSoftware i was force pushing, this shows 48 commits, need your help to clean-up/rebase

gh auto-closed this in middle of the rebase :c, fixed here: https://github.com/rusefi/rusefi/pull/8341

PR #8341, opened 2025-07-22
separate "writeOneField" callback from writing scalars/arrays/bits on the final .ini on TsOutput.java

FDSoftware:
same/part of https://github.com/rusefi/rusefi/pull/8313 pr, but separating on small changes, here I only organize code different, I need to call this `writeFieldJob` 2 times for the C/F case, since I need to write the conditional, but don't move the `FieldIterator` size or `tsPosition`, which would happen if I call `writeOneField` recursively


rusefillc:
@FDSoftware it's asking for rebase

PR #8346, opened 2025-07-24
Subaru ej20g trigger tests

dron0gus:
Still fails when rising edge is not in "right" position.
For VR sensors we should ignore rising edges. https://mcs.woodward.com/content/motohawk/Documentation/MotoHawk2015bSP0/HTML/MotoHawk_topics/VRInterfacing.html
Subaru 6+7 and only 7 triggers defined with `SyncEdge::RiseOnly` . `SyncEdge::FallOnly` was removed in https://github.com/rusefi/rusefi/pull/4621
So I set `invertPrimaryTriggerSignal` and `invertSecondaryTriggerSignal` to true.
But inbvertion first applied at CSV file reader https://github.com/rusefi/rusefi/blob/a8dc3858d334b855783771c576577da9779ddf96/unit_tests/test-framework/logicdata_csv_reader.cpp#L128 and then in trigger central in handleShaftSignal() https://github.com/rusefi/rusefi/blob/70ec6bbf9a0f2060289c7a0856891c85e5efa62b/firmware/controllers/trigger/trigger_central.cpp#L463
So trigger follows only rising edges:

Screenshot from 2025-07-24 12-25-09


rusefillc:
@dron0gus what's the root cause of this whole file replacement again and again? let's start with a separate PR to rename current files

for this PR let's add new files under more detailed names

dron0gus:
Old files are preserved under another name.
Root cause of having new files - I found HW issue causing previous test data was captured with Cam input inverted.

PR #8387, opened 2025-08-03
LTFT: cleanup, custom VE apply

dron0gus:


rusefillc:
@dron0gus let's avoid weak linking it's lame

something like https://github.com/rusefi/rusefi/pull/8386 unless you have better ideas?

dron0gus:
@rusefillc done.

While `weak` is 0 bytes RAM cost and 0 CPU cycles cost :)

PR #8388, opened 2025-08-03
ini: do not expand defines, keep it for TS

dron0gus:


dron0gus:
As example: this drops 60Kb for rusefi_uaefi.ini

rusefillc:
@dron0gus we probably gave a relevant test, see red

rusefillc:
@FDSoftware need your help - I have no idea why are unit tests failing

FDSoftware:
> @FDSoftware need your help - I have no idea why are unit tests failing

`Looks like things were renamed: "TPS" not found in bits, U08, @OFFSET@, [0:5], $gppwm_channel_e_enum`

It's because of the `gppwm_channel_e_enum` remove, tune-can-tools doesn't expand those TS defines on its own, it doesn't find them in the new .ini, so it assumes they were removed, the test itself fails because then the default value can't be assigned to the GPPWM channels

rusefillc:
@FDSoftware I've just merged smaller https://github.com/rusefi/rusefi/pull/8415

Hold on, was it a mistake? is tune-can-tools now failing to handle the stuff I've merged? i.e. what's the happy path forward @FDSoftware do we need to make tune-can-tools smarter or else?

FDSoftware:
> i.e. what's the happy path forward @FDSoftware do we need to make tune-can-tools smarter or else?

yep, the entire ini provider should be able to read these defines correctly, in this specific case it bothers us in tune-can-tools but I suppose it will also be a problem in tune migrations?
edit:
(I'm referring to the migrations between firmware updates that @kifir23917 was working on)

rusefillc:
fixed by https://github.com/rusefi/rusefi/commit/f513f2c78460692ec7b3eba77b0a8932750f54b0

PR #8389, opened 2025-08-03
custom flag for Celsius/Fahrenheit selector

FDSoftware:
replace of `#if CELSIUS` (which defaults to fahrenheit in TS), with new "TEMP_UNITS_CELSIUS" (which defaults to Celsius)

image

settingGroup doesn't support a default value, so I can't give it fancy labels for the user.

### DIFF
--- a/firmware/tunerstudio/tunerstudio.template.ini
+++ b/firmware/tunerstudio/tunerstudio.template.ini
@@ -36,6 +36,8 @@ enable2ndByteCanID = false
 ;    settingOption = CONN_SLOW, "Slower / Wireless"
 ;    settingOption = CONN_FAST, "High Speed / USB"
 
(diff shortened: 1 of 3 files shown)


rusefillc:
@FDSoftware better name needed? USE_METRIC_UNITS maybe?

firmware/tunerstudio/tunerstudio.template.ini
```diff
@@ -238,9 +240,8 @@ page = 3
 	wbo0_hasFault = { enableAemXSeries && ((wb1faultCode >= 1) || !wb1isValid) }
 
 ; calculated temperatures by F/C
-; This is not officially documented in the .ini reference,
```
rusefillc, line 241: Are you moving or killing knowledge here?

Issue #8390, opened 2025-08-03
migrate complete rusefi_enums.h to proper enums


rusefillc:
rusefi_enums.h should be migrated to modern safer enums

rusefillc:
sounds like perfect task for LLM? open question: need proper setup/toolset

@FDSoftware any interest to investigate a proper IDE/plugin/toolset integrated setup, migrate some of the enums and teach me how to migrate the rest?

FDSoftware:
> sounds like perfect task for LLM? open question: need proper setup/toolset
> 
> [@FDSoftware](https://github.com/FDSoftware) any interest to investigate a proper IDE/plugin/toolset integrated setup, migrate some of the enums and teach me how to migrate the rest?

I don't think LLM can determine where the best scope is to put the "using enum" that will be required so as not to have to refactor everything to EnumName::EnumValue, but we can try, now I create a PR migrating a single enum

FDSoftware:
Try of migrating with llm:

> using commits 3cc4cd6 through 8e19add as an example of the changes, I need to replace enum `can_nbc_e`

<img width="1444" height="870" alt="Image" src="https://github.com/user-attachments/assets/ed81e12c-b62d-4283-91c1-f8c658e8be93" />


now with:

> using commits 3cc4cd6 through 8e19add as an example of the changes, I need to replace enum `can_nbc_e`, name should stay the same, also replace all the references to this enum, if using many on a singular scope, is allowed to use `using enum can_nbc_e;`

<img width="823" height="981" alt="Image" src="https://github.com/user-attachments/assets/3ed206bd-753b-4148-8e0f-7e434fcd2750" />

this one is good

<img width="680" height="602" alt="Image" src="https://github.com/user-attachments/assets/4e2eb1c5-b31f-4237-82f9-f2a2f1862503" />

wtf:
<img width="1065" height="145" alt="Image" src="https://github.com/user-attachments/assets/322c2aea-c11d-4bd6-bf75-c7d7bf99a16e" />

eeee, I would give it a 6/10 at least for trying:
<img width="644" height="297" alt="Image" src="https://github.com/user-attachments/assets/00e59f58-e548-4943-a4f5-9ad4d1b9e761" />


[chat-d4bfe602-73e9-4c74-8116-98ff9af550af.txt](https://github.com/user-attachments/files/21609479/chat-d4bfe602-73e9-4c74-8116-98ff9af550af.txt)

[...]

PR #8395, opened 2025-08-04
migrate idle mode typedef enum to enum class

FDSoftware:
related #8390


rusefillc:
@FDSoftware how was this done in terms of toolset? All human or?

FDSoftware:
> @FDSoftware how was this done in terms of toolset? All human or?

all human, 15m total+-, but I think you can use this commit as context for llm so that the next one is more auto-magic
prompt something like "using these commits 3cc4cd6 through 8e19add as an example, I need to replace enum X"

Issue #8279, opened 2025-07-11
spi EGT are broken after Chibi migration


zed65:
If I load a firmware pre Chibi migration the EGT works. F407discovery

rusefillc:
@zed65 what's the statis with your recent change?

zed65:
It works in master but my fw-custom is still broken?

PR #8407, opened 2025-08-06
Potential fix for #8279

zed65:
Don't know if this is the right way but it works when I delete "spiStop(driver);"


rusefillc:
@dron0gus what's your take on this? this is https://github.com/rusefi/rusefi/issues/8279

dron0gus:
Sound strange. But maybe same issue as https://github.com/rusefi/rusefi/blob/bc9f9d715ec6b1933d95aad3665c911b6e18fa97/firmware/hw_layer/drivers/gpio/mc33810.cpp#L231
Looks safe to me. Actually it is not mandatory to call spiStop() after use. Next spiStart() checks for SPI_STOP or SPI_READY state.

dron0gus (approved):

PR #8424, opened 2025-08-12
Add support for additional headers in lua live data access code

awpe:
So some modules that require special access to their internals can define their own headers directly in yaml


rusefillc:
@awpe a change of logic should be accompanied by a change in unit tests?

PR #8442, opened 2025-08-14
Make LogFields immutable and const initialised in PROD

awpe:
@rusefillc Please try this with your private closed modules first, if they put smth to logs that is not compile time resolved (addresses-wise) prod build won't compile. If that is the problem check what was done to ETB code to make sure static storage address is compile-time resolved.


awpe:
I dont understant how this change can break some config java thing for hellen121... Any hint?

rusefillc:
> I dont understant how this change can break some config java thing for hellen121... Any hint?

this change is not breaking hellen121, master was broken

PR #8448, opened 2025-08-17
Fast adc sampling noise

dron0gus:


dron0gus:
From this:
Screenshot from 2025-08-17 10-58-41
To this:
Screenshot from 2025-08-17 10-55-41

Hope noone is using MAP input through slow ADC. At least I was to able to find such users.

rusefillc:
@awpe what is root cause of visibility issue here? See other PRs with experimentation

awpe:
Compiler correctly point to silently hidden "start" name to hint you for possible mistype. To solve that either use different naming or introduce "using ...::start(...)" before that void start() so compiler sees that you really intended to have 2 functions with same name but different signature. This warning exists for a long time and poisons life as many other strangely existent warnings, but in rusefi_rules.mk this warning is explicitly set to produce no errors: RUSEFI_OPT += -Wno-error=overloaded-virtual, question is how this compilation target was left without that option?

rusefillc:
@awpe note how this PR does not seem to be touching anything 'start'?!

awpe:
@dron0gus Is your branch latest master compliant?

awpe:
@rusefillc I'm not saying this PR is done smth wrong, Im questioning why 6 targets have their bootloader built with some includes from firmware while bootloader makefile has different ignorance options (the problematic one is absence of ignorance for overloaded-virtual)

awpe:
Maybe it is same for master currently?

awpe:
I tried master it fails bootloader builds for other boards as well with same error

rusefillc:
see https://github.com/rusefi/rusefi/pull/8453

rusefillc:
@dron0gus @awpe confirming that master is broken https://github.com/rusefi/rusefi/pull/8455

Issue #4788, opened 2022-11-16
Please make UI in Freedom Fahrenheit units Celsius PSI KPA


ElDominio:
Temp, Pressure and Distance be available in Imperial units

mck1117:
This is a way huger task than we're willing to tackle right now, I think

rusefillc:
We used to have it on .ini side, it's just boring labor for each gauge manually right?

rusefillc:
I.e. if we can find commit which killed it we can re-enable for some gauges?

But scales for curves would stay in C :( do we have no cheap solution for for axises?

ElDominio:
But I had a client do it in the INI lol


![image](https://user-images.githubusercontent.com/5685108/202319632-a0922088-5fd5-4478-86ac-89353cb72389.png)


![image](https://user-images.githubusercontent.com/5685108/202319664-1a092e3d-a5b5-4009-b90f-0f9a598508a7.png)

mck1117:
> I think cheapest solution for axes is to build a coolant temp sensor archive (which we should have with the amount of tunes on rusefi online) and forget about writing in F

We already have presets in tunerstudio - trivial to add more if you have suggestions.

rusefillc:
?

I mean "cranking by clt" curve

rusefillc:
#607 would be a typical outcome :(

rusefillc:
https://www.youtube.com/watch?v=JYqfVE-fykk

rusefillc:
@ElDominio got a new complaint in Spanish

ElDominio:
<img width="284" alt="image" src="https://github.com/rusefi/rusefi/assets/5685108/0c072928-f111-49eb-bbda-14232ee78e7f">
 
another one

rusefillc:
@ElDominio we now have freedom temperature

@FDSoftware let's make progress in terms of freedom PSI pressure?

ElDominio:
Definitely showing freedom units:

<img width="1345" height="793" alt="Image" src="https://github.com/user-attachments/assets/fd20d3a9-9e89-401a-adb4-2cb3d7d31f16" />

FDSoftware:
> [@ElDominio](https://github.com/ElDominio) we now have freedom temperature
> 
> [@FDSoftware](https://github.com/FDSoftware) let's make progress in terms of freedom PSI pressure?

next step now that we have psi?

rusefillc:
[...]

PR #8454, opened 2025-08-17
VSSGauge metric/imperial

FDSoftware:
image

related #4788


rusefillc:
@FDSoftware please rebase, we assume master is less red now

Issue #8447, opened 2025-08-17
Torque reduction while using "Clutch Up Switch" trigger is inverted.


ElDominio:
**Describe the bug**
When clutch up is RED (meaning clutch is NOT depressed), this is treated as whatever pin that depends on it being trigger.

**Hardware**
 - Board: H154
 - Engine info BKTheta

**To Reproduce**
1. set trigger condition for Torque Reduction as Clutch Up switch
2. when the user is not stepping on pedal, Clutch Up is RED, and Torque Reduction trigger is also RED. I need that when specifically Clutch Up is RED, Torque Reduction trigger is WHITE. Clutch Down should work like this (when Clutch Down is RED, Torque Reduction trigger is RED.

**Expected behavior**
 I need that when specifically Clutch Up is RED, Torque Reduction trigger is WHITE. 

**Additional context**
Current workaround: set trigger mode to INVERTED DEFAULT. This produces a bold faced LIE that turns the Clutch Up bulb RED when the clutch is DOWN, making it work correctly

rusefillc:
@ElDominio please add some relevant screenshots

ElDominio:
<img width="1053" height="662" alt="Image" src="https://github.com/user-attachments/assets/7c3977ad-5fec-4dad-865f-6b3c4c1c60a7" />

bad behavior, correct setting

<img width="1093" height="697" alt="Image" src="https://github.com/user-attachments/assets/8927205c-1832-4753-a587-3b07e1a6b8e6" />

correct behavior, bad setting

rusefillc:
@ElDominio I am very confused

https://rusefi.com/docs/pinouts/hellen/hellen154hyundai_revD/ has only one clutch input? A vehicle with only one clutch switch probably has clutch DOWN not clutch UP input? Is the input mis-labeled?

rusefillc:
@ElDominio 

<img width="524" height="57" alt="Image" src="https://github.com/user-attachments/assets/cff42b1f-a385-4b38-ab50-dcf459f73827" />

rusefillc:
@ElDominio 

<img width="278" height="75" alt="Image" src="https://github.com/user-attachments/assets/9bee0a4a-980d-47bd-904a-4de3187b4cd0" /> actually no, it has no name. What is this switch - is it clutch UP or clutch DOWN?

rusefillc:
```
engineConfiguration->clutchUpPin = Gpio::H144_IN_RES2;
```

but is that right?

[...]

PR #8459, opened 2025-08-18
Torque reduction while using "Clutch Up Switch" trigger is inverted. …

rusefillc:
…#8447

only:uaefi


rusefillc:
@FDSoftware need help adjusting tests to this change of logic

PR #8465, opened 2025-08-19
Proposal of single point of truth for enums in whole project and all subcomponents

awpe:
Please check overall idea, I'm still thinking how to organize artifacts between folders so it is consistent with other files (see how files are stored in integration, that is temporary)


java_tools/.idea/gradle.xml
awpe, line 1: to be removed

java_tools/configuration_definition_base/src/main/java/com/rusefi/output/ConfigStructureImpl.java
```diff
@@ -70,7 +70,7 @@ public void addAlignmentFill(ReaderState state, int alignment) {
          * we make alignment decision based on C fields since we expect iteration and non-iteration fields
          * to match in size
          */
-        FieldIteratorWithOffset iterator = new FieldIteratorWithOffset(cFields) {
+        FieldIteratorWithOffset iterator  = new FieldIteratorWithOffset(cFields) {
```
awpe, line 73: revert

awpe, line 73: revert

rusefillc:
@awpe overall yes

not sure why this PR is not launching CI?

why so many new files with "super-uaefi"? most/all enums are same for all boards?

awpe:
> @awpe overall yes
> 
> not sure why this PR is not launching CI?
> 
> why so many new files with "super-uaefi"? most/all enums are same for all boards?

This is proposal, I added .h to PR to show how it will look like overall, as stated in description files will be routed to required folders. 

If we don't foresee that some boards will have custom enums only for them then yes no need to create per-board txt/ini/cfg... So are custom per-board enums expected ever?

rusefillc:
> If we don't foresee that some boards will have custom enums only for them then yes no need to create per-board txt/ini/cfg... So are custom per-board enums expected ever?

divide and conquer? we do not have per-board enums at the moment, this PR is definitely not the place to introduce?

PR #8471, opened 2025-08-22
Prepare generated file placeholder as example

awpe:
Please note comments about future removal of redundant data


rusefillc:
@awpe is that legit macos failure or some infrastructure issue?

rusefillc:
@awpe please rebase maybe https://github.com/rusefi/rusefi/pull/8475 would help us see useful messages?

awpe:
I know the problem, will fix later, basically is there an option to bump macos runner/compiler?

rusefillc:
you would have to mess with https://github.com/rusefi/rusefi/blob/master/.github/workflows/build-unit-tests.yaml we have some freedom there. at the moment we use vague 'macos-latest' image

awpe:
@rusefillc github shows that macos should have been migrated already
https://github.com/actions/runner-images/issues/12520
UPD: they started on 4th of August and should finish at 30th of August, clang will be newer

awpe:
@rusefillc looks good now in each compiler/os, I don't know why Simulator failed here, there are no logs for action, though it passed in my pipeline

awpe:
The job failed because it couldn't download https://services.gradle.org/distributions/gradle-8.14-all.zip due to a timeout

PR #8475, opened 2025-08-22
Missing override

rusefillc:
@awpe


awpe:
Overall I dont like to mitigate warnings by shutting them up, can we have an issue please? I think once enums/uego finished I can cleanup it properly without shuting down warnings

rusefillc:
@awpe we do not delete on production code flow, is this even a legit (for us?) issue?

awpe:
I need to check if vtables are generated at all and hide destructors if not  needed or check chains but after...

Issue #6874, opened 2024-08-31
VVL control (switched cam)


ElDominio:
**Is your feature request related to a problem? Please describe.**
I would like to be able to control nitrous systems from the ECU itself, to be able to compensate fuel/timing and control activation from the ECU 

**Would we need new configuration fields?**

1. VVL Output - control the relay which activates the nitrous solenoid
2. Minimum TPS - minimum TPS for Nitrous Activation, set to 0 to disable
3. Minimum CLT - minimum CLT for Nitrous Activation, set to 0 to disable
4. Activation RPM - at the RPM that the VVL system engages when all conditions are met
5. Deactivation RPM -at the RPM that the VVL system engages when all conditions are met
6. Fuel adder % - this is a % of fuel which is added to the base main fuel, since we are basically injecting additional air
7. Ignition retard - this is timing REMOVED from actual final timing (after all corrections) due to additional air

**Describe the solution you'd like**
Add the stuff up there.

**Describe alternatives you've considered**
Using Lua but this allows others to use it

rusefillc:
Variable Valve Lift? Nitrous?

ElDominio:
nitrous references removed, my mistake

ElDominio:
@FDSoftware  very similar to nitrous control

ElDominio:
Not even very similar, it could literally be a clone of Nitrous Control with everything that says Nitrous changed to "VVL" or "Switched Cam"

My FTO has MIVEC + Nitrous and this would make me very happy

ElDominio:
@rusefillc @FDSoftware

FDSoftware:
> Not even very similar, it could literally be a clone of Nitrous Control with everything that says Nitrous changed to "VVL" or "Switched Cam"
> 
> My FTO has MIVEC + Nitrous and this would make me very happy

@ElDominio 
Do you also need to be able to turn it on/off with a digital input/lua?

ElDominio:
Nope @FDSoftware

ElDominio:
Nope

rusefillc:
@ElDominio some progress was made!

ElDominio:
This exists and I have used it and it works perfectly

PR #8481, opened 2025-08-23
proposal: allow engine modules to add/remove fuel/timing

FDSoftware:
Maybe this is better than adding another more conditional on `getRunningAdvance` ?

also prep for #6874

### DIFF
--- a/firmware/controllers/algo/fuel_math.cpp
+++ b/firmware/controllers/algo/fuel_math.cpp
@@ -131,6 +131,8 @@ float getRunningFuel(float baseFuel) {
 	correction *= engine->module<NitrousController>().unmock().getFuelCoefficient();
 #endif
 
+	engine->engineModules.apply_all([&correction](auto & m) { correction *= m.getFuelCoefficient(); });
+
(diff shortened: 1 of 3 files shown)


FDSoftware:
```c++
tests/nitrous_control/test_nitrous_fuel_adder.cpp:32: Failure
The difference between engine->fuelComputer.totalFuelCorrection and expectedTotalFuelCorrection is 0.20088875293731689, which exceeds 0.00001, where
engine->fuelComputer.totalFuelCorrection evaluates to 1.3825875520706177,
expectedTotalFuelCorrection evaluates to 1.1816987991333008, and
0.00001 evaluates to 1.0000000000000001e-05.
All conditions are satisfied
``` 

hmm I'll check it tomorrow, I think it might be a rounding issue?

FDSoftware:
edit: I am dumb, see:

`correction *= engine->module<NitrousController>().unmock().getFuelCoefficient();`

is the same name as "new" on engine.h, gets override by the nitrous one, so I end up applying the same correction twice.

PR #8497, opened 2025-08-24
Introduced enums generator tool to read is not compatible with $ syntax

rusefillc:
minimal change to enable json enum reader

problem: second commit is required to have green build


rusefillc:
success we now have https://github.com/rusefi/rusefi/pull/8497 with clear problem statement

PR #8517, opened 2025-08-28
Hellen and MRE: adcVcc is 3.3V

dron0gus:
Most of these boards uses REF3033 as Vref+ supply. It is 3.3 with 0.2% acuracy.
Some uses LDO with less acuracy, bit still with 3.3V output.

vccAdc was defined as 3.29V for MRE from the begin of ages. And seems was copy-pasted to Hellen.

Maybe @mck1117 knows some story of this.


mck1117:
Somebody measured a 3.3v regulator a long time ago and it was 3.29v. So 3.29v got pasted in there. It's been 3.3v everywhere on FOME for ages.

PR #8550, opened 2025-09-09
TS: UI: show TPS values in V

dron0gus:


rusefillc:
@dron0gus this one is green now, do you have any doubts?

dron0gus:
@rusefillc no seamless update is possible. TS will warn because units for tpsMin, tpsMax and so on is changed from "ADC" to "V".
Merge is we ok with this.

FDSoftware:
> no seamless update is possible. TS will warn because units for tpsMin, tpsMax and so on is changed from "ADC" to "V". Merge is we ok with this.

just linking now related pr:
https://github.com/rusefi/rusefi/pull/8618

PR #8551, opened 2025-09-09
Fix for "Closed loop 1 adder" actually being boostClosedBlend2Table

jankovalski:


rusefillc:
thank you @jankovalski

Issue #2613, opened 2021-05-03
More human approach to Boolean fields true false. "Is that French?"


rusefillc:
We shall avoid "true"/"false" in many dialogs instead going with explicit direct option names for both options

"before", see below for improved "after" version
![image](https://user-images.githubusercontent.com/48498823/116837172-4c183b80-ab97-11eb-8378-3e20927a1662.png)

rusefillc:
See #2571

rusefillc:
After change

![image](https://user-images.githubusercontent.com/48498823/117161137-ad0a6400-ad8f-11eb-9a27-43ec6fd79aae.png)

rusefillc:
@stefanst @OrchardPerformance can I get you guys involved?

there are two steps here
1. identify dialogs where we shall improve above "true"/"false" like on the above examples
2. actually change two relevant text files, I am happy to show how that could be done with just a browser

rusefillc:
was
![image](https://github.com/rusefi/rusefi/assets/48498823/8372f609-40e1-404b-8f9d-c01a717bc087)

rusefillc:
short term fuel trim should be changed to enabled / disabled
enable launch control should be enabled/ disabled
enable shift torque reduction should be enabled/ disabled

rusefillc:
We IT people have left too much true/false visible in TS interface, that's inhumane.

Now that we have our amazing migration framework are we in the position to improve?

rusefillc:
status: we still have about 60 of un-converted field, need to manually review each use-case and decide what field text change is needed/what human readable options are best

technically we do not need configuration migrator for most of the remaining settings

once we have zero of "true"/"false" bits, we should probably change default - open question if yes/no or enabled/disabled should be the default, maybe we should just always require explicit text

rusefillc:
Follow-up ticket https://github.com/rusefi/rusefi/issues/9032

PR #8562, opened 2025-09-13
"less french, more human" add Java migration for specific boolean fields

FDSoftware:
Also `launchControlEnabled` migrated on a separated commit as example
related #2613


FDSoftware:
image

The crash in the tests is because I need to migrate the enums. I'll fix it ~tomorrow.~ edit: fix was easier than i thought lol

PR #8579, opened 2025-09-14
Fix map prediction rpm

ElDominio:
uses the TPS/TPS extra fuel RPM correction as a variable Blend Duration, ignoring this scalar/single value:
image

We should always have varibale duration because we don't need prediction in upper RPMs where airspeed is fast

### DIFF
--- a/firmware/controllers/algo/airmass/speed_density_airmass.cpp
+++ b/firmware/controllers/algo/airmass/speed_density_airmass.cpp
@@ -53,7 +53,7 @@ float SpeedDensityAirmass::getAirflow(float rpm, float map, bool postState) {
 }
 
 float SpeedDensityAirmass::getPredictiveMap(float rpm, bool postState, float mapSensor) {
-	float blendDuration = = interpolate2d(rpm, config->tpsTspCorrValuesBins,
+	float blendDuration = interpolate2d(rpm, config->tpsTspCorrValuesBins,
 						config->tpsTspCorrValues);
 	if ( blendDuration > 2) {
 		blendDuration = 1;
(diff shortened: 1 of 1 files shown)


rusefillc:
@ElDominio what is the plan for "mapPredictionBlendDuration" removal?

ElDominio:
> @ElDominio what is the plan for "mapPredictionBlendDuration" removal?

From how many places does it need to be removed? It won't serve a purpose anymore so it can be removed

i'll do another pr with everything about it removed

ElDominio:
i did it and don't really know how to re-pr? maybe no pr needed?

rusefillc:
> i did it and don't really know how to re-pr? maybe no pr needed?

PR was happily updated, no need to re-PR

rusefillc:
@ElDominio setting was removed but default value assigned stayed and it breaks build now

```
engineConfiguration->mapPredictionBlendDuration = 0.5f; // 500ms blend duration
```

rusefillc:
@ElDominio do now take "duration, seconds" from "multiplier" column of the following?

image

ElDominio:
> @ElDominio do now take "duration, seconds" from "multiplier" column of the following?
> 
> image

Yes, that is how it works. Is that possible to change dynamically?

rusefillc:
> Yes, that is how it works. Is that possible to change dynamically?

Yes it's possible but not trivial. One thing to consider would be separate curve? These are small, very little RAM usage

ElDominio:
I only use table because it already exists, and this was a non-GPT change so i did not build a new curve. As long as it works the same as this curve, it would be fine

rusefillc:
@ElDominio your change is now in master

PR #8628, opened 2025-09-25
less french ep II, migrated more boolean fields - see #2613

FDSoftware:
34 less french fields, related  #2613


FDSoftware:
mr7 crashed on java install for a timeout:
`Error: Exception in thread "main" java.io.IOException: Downloading from https://services.gradle.org/distributions/gradle-8.14-all.zip failed: timeout (10000ms)`

rusefillc:
@FDSoftware we have to do better. I assume that changing boolean names for second time later might become a mess, we have to get it right from first go.

for instance, I believe that "XXX: enabled" is better than "enable xx: yes"

image

image

rusefillc:
https://github.com/rusefi/rusefi/pull/8792

PR #8629, opened 2025-09-25
less french ep III, migrated more boolean fields - see #2613

FDSoftware:
42 fields here, needs #8628


rusefillc:
@FDSoftware please rebase/get it green?

rusefillc:
@FDSoftware do we have failing java tests here?

FDSoftware:
> @FDSoftware do we have failing java tests here?

Yup! I have to fix a couple of assertions, are expecting true/false instaded of yes/no

PR #8639, opened 2025-09-26
added a/c compensation to coasting idle table

ElDominio:
a/c is still in the realm of reality when coasting to idle. we did not compensate for A/C when coasting to idle, making coasting idle table only useful with A/C off. Now A/C compensation is applied when using coasting idle table


firmware/controllers/actuators/idle_thread.cpp
```diff
@@ -459,10 +468,6 @@ void IdleController::onFastCallback() {
 #endif // EFI_SHAFT_POSITION_INPUT
 }
 
-void IdleController::onEngineStop() {
```
rusefillc, line 462: @ElDominio please clarify why are you removing these lines?

PR #8647, opened 2025-09-28
Please make UI in Freedom Fahrenheit units Celsius PSI KPA #4788

rusefillc:
un-doing the change for now, too fragile in terms of existing projects


rusefillc:
@FDSoftware looks like something is not right around TS / existing projects / migration from older versions? we need to disable the change for now

I've commented out java part, next step disabling .ini change. @FDSoftware what's the smallest change to disable feature in .ini? can we go ``#if 1`` instead of ``#if USE_METRIC_UNITS``?

Issue #8631, opened 2025-09-25
next episode of weak override


rusefillc:
Let's address another few of most popular weak overrides similar to https://github.com/rusefi/rusefi/issues/8269

PR #8683, opened 2025-10-05
week override: OnConfigurationChange

FDSoftware:
related: #8631

`call_board_override` now supports args :D (and added notes on a few places where we need a minor refactor)

as always only one board migrated and next PR is the rest on individual commits


rusefillc:
@FDSoftware please rebase - I've cherry picked two commits which seemed separate enough

Issue #8689, opened 2025-10-06
STFT: migrator should work for maxAdd and maxRemove


rusefillc:
Same applies to all four instances of stft_cellCfgs

* all maxAdd type was changed - we should support it
* maxRemove sign was flipped - we should migrate while changing sign

See unit test with hopefully relevant test data?

caused by https://github.com/rusefi/rusefi/pull/8030

PR #8701, opened 2025-10-08
STFT: migrator should work for maxAdd and maxRemove

FDSoftware:
#8689


FDSoftware:
Doing a self-review: both fields are quite similar, I think it's better to make a list with the 8 fields and run them with a foreach, otherwise I'll end up with a lot of duplicate code.

PR #8725, opened 2025-10-14
unit tests: call periodicFastCallback() while advancing time

dron0gus:
This callback calculates new sparkDwell and dwellDurationAngle. Incorrrect spark scheduling happens if dwellDurationAngle is not updated according to RPM.
This cause false Overcharge errors during some trigger tests.

### DIFF
--- a/unit_tests/test-framework/engine_test_helper.cpp
+++ b/unit_tests/test-framework/engine_test_helper.cpp
@@ -22,6 +22,7 @@
 
 bool unitTestBusyWaitHack;
 bool unitTestTaskPrecisionHack;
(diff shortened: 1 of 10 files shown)


dron0gus:
Now we need to fix some tests.

rusefillc:
@dron0gus shall we make "compatibleTechnicalDebtUnitTest" flag to go around the problem for now?

rusefillc:
instead of fixing tests, I suggest we start with just a compatibility flag for existing (legacy) tests?

we already have it for something else. yes, technical debt but such is life :(

dron0gus:
@FDSoftware thanks for fixing this!

PR #8732, opened 2025-10-16
fix migration on CrankingIdle multiplier to position change

### LINKED ISSUE #8730: Cranking Idle Valve Curve Does Not Migrate Probably
Cranking Idle Valve Curve migrator does not account for change of logic

After Migration We Have

Image

This value should be multiplied by the cranking base IAC to give us the proper value shown below

Image

FDSoftware:
resolves #8730


FDSoftware:
@rusefillc I didn't see cases where we changed this table dimensions, do you need this migration to support size changes as well as the AFR one or not?

PR #8744, opened 2025-10-20
unused in board.h

rusefillc:


dron0gus:
These are common macro for GPIO registers bit manipulation. Exist in every STM32's board.h in ChibiOS.
I see no reason in commenting out unused.

rusefillc:
@dron0gus by hiding these I am emphasizing "not expected to be used, you better know what you are doing if you need to use those"

also helps validate consistency between three board.h files (separate issue that we have so much copy-paste between those three?)

Issue #8766, opened 2025-10-25
F4/F7 (H7?) interrupt priority mess


dron0gus:
STM32F4/F7 have some interrupt vectors that shared between few sources:
<img width="941" height="218" alt="Image" src="https://github.com/user-attachments/assets/571c4d91-3b61-4c06-8506-9ad124be3795" />
So conventional method of settings IRQ priority level for TIM9 in PWM mode (for example):
`#define STM32_PWM_TIM9_IRQ_PRIORITY         7`
is disabled in ChibiOS's `stm32_isr.h` as it may conflict with TIM1 IRQ priority defined to some other level.
See `STM32_TIM*_SUPPRESS_ISR` and code like this in all TIM drivers:
```
#if !defined(STM32_TIM1_SUPPRESS_ISR)
      nvicEnableVector(STM32_TIM1_UP_NUMBER, STM32_ICU_TIM1_IRQ_PRIORITY);
      nvicEnableVector(STM32_TIM1_CC_NUMBER, STM32_ICU_TIM1_IRQ_PRIORITY);
#endif
```
Instead `STM32_IRQ_TIM1_BRK_TIM9_PRIORITY` (and similar) defines are used in ChibiOS's stm32_tim1_9_10_11.inc (and similar) files to setup IRQ priority and demux irq handler between several drivers.

Problem: RusEFI's [mcuconf_common_f4_f7.h](https://github.com/rusefi/rusefi/blob/master/firmware/hw_layer/ports/stm32/mcuconf_common_f4_f7.h) still have default values for all shared irq vectors.
While have adjusted priorities like:
`#define STM32_PWM_TIM5_IRQ_PRIORITY         PRECISE_SCHEDULING_TIMER_PRIORITY`
That makes no sense because of `STM32_TIM5_SUPPRESS_ISR` in stm32_isr.h for F4 and F7.

We need to carefully review all non-default IRQ levels for TIMs and UARTs and move from individual level defines to shared https://github.com/rusefi/rusefi/blob/c479e3e52b5110ec4e7729ba6b9081677cf680fe/firmware/hw_layer/ports/stm32/mcuconf_common_f4_f7.h#L137-L150

dron0gus:
We expect TIM5 used for scheduling real-time event in microsecond_timer.cpp to have high priority:
[...]

PR #8770, opened 2025-10-26
ports: F7: deduplicate

dron0gus:
These are already defined in mcuconf_common_f4_f7.h


dron0gus:
A bit related to https://github.com/rusefi/rusefi/issues/8766

PR #8800, opened 2025-11-05
Backport to lts-25jersey: injector lag lookup should be done based on differential pressure

dron0gus:
only:uaefi


dron0gus:
Lets go other way https://github.com/rusefi/rusefi/pull/8802

Issue #8819, opened 2025-11-14
Shift map sampling start angle if converting from jersey


rusefillc:
* if source signature contains lts-25jersey
* if destination signature DOES NOT contain lts-25jersey
* subtract 45 from all map sampling start curve values

<img width="617" height="394" alt="Image" src="https://github.com/user-attachments/assets/bfa33bd4-355f-4cab-9212-37e373372413" />

PR #8820, opened 2025-11-14
map sampling start angle migrator

FDSoftware:
related: #8819


FDSoftware:
@artemrusefi 

Don't you think we should be a bit more relaxed about the prev => updated tune checks inside `BinsIniFieldMigratorStrategy` ? (or add override option inside "TableAddColumnsMigrator" ?)

 I have a bin field `(map_samplingAngleBins)` with `BINS_INI_FIELD_DIGITS` (2) and `BINS_INI_FIELD_TYPE` (F32) that are a bit different from the harcoded types/digits, but that wouldn't have to block the entire bin migration in my opinion [since here the prev/updated digits/types are the same] :c

PR #8830, opened 2025-11-19
OpenBLT: USB: always flush

dron0gus:
Flush USB serial before return from Rs232TransmitPacket() So PC side will see last OpenBLT reply before strarting RE FW.


rusefillc:
@dron0gus has this broken nucleo HW CI compilation?

dron0gus:
@rusefillc https://github.com/rusefi/rusefi/pull/8831

PR #8844, opened 2025-11-23
LTFT: can be disabled at compile time

dron0gus:


rusefillc:
what are the arguments for EFI_LTFT_CONTROL?

i guess maybe specifically EFI_LTFT_CONTROL has merit in light of extra persistence considerations, but why on earth do we have EFI_BOOST_CONTROL?

Issue #5984, opened 2024-02-18
simulator action fails once on meta "add/change" operations


rusefillc:
See https://github.com/rusefi/rusefi/actions/runs/7948960015 and simulator failure within https://github.com/rusefi/rusefi/commit/c8e779753a9f1c4d8dfc3b96c1cea01d21d67123

race condition I believe with config generation pushing to git and fresh tunes pushing to git due to date change or something change?

rusefillc:
@chuckwagoncomputing any chance you can look into this?

chuckwagoncomputing:
I would much prefer to remove all generated files from the repo.

rusefillc:
> I would much prefer to remove all generated files from the repo.

Let me attempt to surprise you with an argument for generated in repo: I totally use git history on generated stuff when troubleshooting. Also valuable to sometimes compare generated of different targets since prepend.txt etc.

chuckwagoncomputing:
That is a very good point. I do that too.

I haven't figured out a good way to generate configs for all boards with make without adding a ton of overhead, and I don't want to have both make and bash scripts generating configs.

Another idea would to make a tool to generate configs for two different revisions in a temporary location and compare, leaving the working directory clean. e.g. call `./bin/config-diff.sh` to compare the current working directory to git HEAD, or pass a commit to compare the working directory to that, or pass two commits to compare them. I think that would only work for configs though; enums and live docs would be tough.

chuckwagoncomputing:
#4302

chuckwagoncomputing:
Have there been any instances of this failure since #6251?

rusefillc:
@chuckwagoncomputing let's see https://github.com/rusefi/rusefi/commit/cd4d9d6402e4a7ba959778914b93ae134d17eb5e

rusefillc:
> @chuckwagoncomputing let's see [cd4d9d6](https://github.com/rusefi/rusefi/commit/cd4d9d6402e4a7ba959778914b93ae134d17eb5e)

so yes https://github.com/rusefi/rusefi/commit/cd4d9d6402e4a7ba959778914b93ae134d17eb5e is red

@chuckwagoncomputing just found a use-case where this issue becomes really annoying: working on branches. On branches we do not push code-generated so branches stay red if meta is changed

[...]

Issue #4302, opened 2022-06-29
reduce generated .h in git


rusefillc:
At least some generated headers should not be checked into git at all

in the meantime proper single line make should be happy meaning that makefile should be managing lazy generation of at least some of the headers?

rusefillc:
At this point I am convinced that this ticker here is more important or at least equally imporant as newparse.

@mck1117 @chuckwagoncomputing any chance anyone would be available to help?

rusefillc:
probably not related to #4454

chuckwagoncomputing:
Is there any reason that Make couldn't handle config generation?

rusefillc:
no reason I am aware of, just a question of identifying next step

rusefillc:
Hello from 2026!

A thing which did not go well would be https://github.com/rusefi/rusefi/issues/8906

chuckwagoncomputing:
## Benefits to removing generated files

- CI pushing to main has always been problematic; it seems to be working now, but there's also reduced contribution volume right now
- Reduced repo size (very rough estimate ~200MB) https://github.com/rusefi/rusefi/issues/8934
- Simpler workflow for contributors who don't have much git experience
- It's just weird to be storing build artifacts in the git repo
- Having inis in the repo could be a source of confusion if new users try to use them with a release build - not sure if this has ever happened

chuckwagoncomputing:
> - if we want to stop polluting git history with crazy CI commits, what's the Megabyte per year cost?
> - even if we stop polluting git with crazy CI commits, what about all the history? maybe shallow clones are an option for CI but is this about CI or else?
> https://github.com/rusefi/rusefi/issues/8934

These aren't "cons", these are avoidance.

chuckwagoncomputing:
Waiting on #8854

PR #8854, opened 2025-11-29
Generate a patch file when generating config/docs

chuckwagoncomputing:
Re: https://github.com/rusefi/rusefi/issues/5984#issuecomment-1951323972
Working towards #4302

The intended use is for developers to generate configs then check the generated patch file after making a change that affects config files. It's also pretty-printed in GHA, so pull requests can be checked easily.
Currently, the file is appended to, so while using locally, you would probably want to remove/empty the batch file between builds. This could be automated, but I haven't done so yet.

## Implementation notes:

_I am not a Java developer_

Overloaded LazyFileImpl constructor to add patchfile name parameter.

In LazyFileImpl, I changed the way the current "Not same at" output works in order to simplify the code and not read the old file twice. Previously, all newlines had been removed before that check, so changes were output as a single line. Now, multiple newlines are squashed to a single newline, and output is on multiple lines if it was in the file. I think this is an improvement, though I'll probably never look at that output if the generated patch file exists.

I used `ConfigDefinitionRootOutputFolder` in ConfigDefinition to locate the generated patch file, but it wasn't available in EnumToString, so I just used the filename by itself, and it works. I think it's fairly safe to assume that EnumToString will always be called from the firmware directory.

### DIFF
--- a/.github/workflows/build-firmware.yaml
+++ b/.github/workflows/build-firmware.yaml
@@ -258,6 +258,10 @@ jobs:
         bash bin/compile.sh ${{env.BOARD_META_PATH}} docs-enums
         bash bin/compile.sh ${{env.BOARD_META_PATH}} config
 
+    - name: Check config diff
+      working-directory: ./firmware/
+      run: colordiff <generated.patch
+
     - name: Building Windows simulator separately just to make github logs more readable
       working-directory: ./firmware/
       run: |
--- a/misc/actions/ubuntu-install-tools.sh
+++ b/misc/actions/ubuntu-install-tools.sh
@@ -5,4 +5,4 @@
 # out of scope: JDK, ARM GCC compiler
 
 apt-get install make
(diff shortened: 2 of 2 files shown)


chuckwagoncomputing:
Failed check was an `apt` error, ignore.

See the "Check config diff" section in any of the board firmware builds for an example.

Once generated files are no longer in the repository, it won't make sense to have that as part of this workflow, but the gen-configs.yaml workflow could be repurposed to run on PR, generate configs before and after a change, and show the diff.

chuckwagoncomputing:
Can this be merged or may I have some feedback regarding why not?

rusefillc:
> Can this be merged or may I have some feedback regarding why not?

I am failing to comprehend the idea/intention big picture, and I am failing to communicate that I would love a conf call to clarify that

chuckwagoncomputing:
[example](https://github.com/rusefi/rusefi/actions/runs/20499035905/job/58903565194?pr=8854)

image

PR #8863, opened 2025-12-02
ETB: calibration, wrong scale 3% calibration error

dron0gus:
During calibration we report TPS1 and TPS2 min/max values back to TS and then TS applys these limits to ECU.
While sending values to TS were using 0.0048828125 scale factor (5/1024) But during TPS sensor initialization TPS_TS_CONVERSION = 200 was used (5/1000) This cause 1024/1000 = 2.4% error.


rusefillc:
We need to clean this up later https://github.com/rusefi/rusefi/issues/8865

PR #8872, opened 2025-12-06
need ETB pins to show for configurability of ETB

ElDominio:
ETB options were not showing, not allowing ETB config.


ElDominio:
image

for documenting

Issue #8774, opened 2025-10-27
Lua lost track of allocated mem


rusefillc:
large script into uaEFI causes reboot

```
#0  chDbgPanic3 (msg=0x809b438 "Lua lost track of allocated mem", file=0x8099b70 "ChibiOS/os/rt/src/chsys.c", line=220) at ./controllers/core/error_handling.cpp:438
#1  0x0800422e in chSysHalt (reason=<optimized out>) at ChibiOS/os/rt/src/chsys.c:220
#2  0x08015fec in luaHeapAlloc (optr=0x2001a318, osize=134847344, nsize=134853688) at ./controllers/lua/lua_heap.cpp:172
#3  0x08074d88 in tryagain (L=L@entry=0x1000e884, block=block@entry=0x2001a318, osize=osize@entry=512, nsize=nsize@entry=308) at ./ext/lua/lmem.c:167
#4  0x08075620 in luaM_realloc_ (L=L@entry=0x1000e884, block=0x2001a318, osize=osize@entry=512, nsize=nsize@entry=308) at ./ext/lua/lmem.c:182
#5  0x08075630 in luaM_saferealloc_ (L=L@entry=0x1000e884, block=<optimized out>, osize=osize@entry=512, nsize=nsize@entry=308) at ./ext/lua/lmem.c:194
#6  0x08075672 in luaM_shrinkvector_ (L=L@entry=0x1000e884, block=<optimized out>, size=size@entry=0x2001b3a4, final_n=77, size_elem=size_elem@entry=4) at ./ext/lua/lmem.c:134
#7  0x08076cac in close_func (ls=ls@entry=0x20003134 <luaThread+3932>) at ./ext/lua/lparser.c:764
#8  0x08010810 in body (ls=ls@entry=0x20003134 <luaThread+3932>, e=e@entry=0x20003080 <luaThread+3752>, ismethod=0, line=line@entry=101) at ./ext/lua/lparser.c:1008
#9  0x080771da in funcstat (ls=ls@entry=0x20003134 <luaThread+3932>, line=line@entry=101) at ./ext/lua/lparser.c:1788
#10 0x08076dea in statement (ls=ls@entry=0x20003134 <luaThread+3932>) at ./ext/lua/lparser.c:1876
#11 0x08076e72 in statlist (ls=ls@entry=0x20003134 <luaThread+3932>) at ./ext/lua/lparser.c:806
[...]

PR #8874, opened 2025-12-07
luaHeapAlloc: try to not use osize

FDSoftware:
https://www.lua.org/manual/5.3/manual.html#lua_Alloc

> When ptr is not NULL, osize is the size of the block pointed by ptr, that is, the size given when it was allocated or reallocated.
> 
> When ptr is NULL, osize encodes the kind of object that Lua is allocating. osize is any of [LUA_TSTRING](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TSTRING), [LUA_TTABLE](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TTABLE), [LUA_TFUNCTION](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TFUNCTION), [LUA_TUSERDATA](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TUSERDATA), or [LUA_TTHREAD](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TTHREAD) when (and only when) Lua is creating a new object of that type. When osize is some other value, Lua is allocating memory for something else.

[...]

> When nsize is not zero, the allocator must behave like realloc. The allocator returns NULL if and only if it cannot fulfill the request. Lua assumes that the allocator never fails when osize >= nsize.

will use a bit more of ram, but `chHeapGetSize` should get us the correct sizes, also this prevents the reboot when we try to use all or more than available ram #8774

before:

```
2025-12-07_00_09_23_371: EngineState: Starting watchdog with timeout 500 ms...
2025-12-07_00_09_23_371: EngineState: Reset Cause: NVIC_SystemReset or by debugger
2025-12-07_00_09_23_371: EngineState: WD resets: 0
2025-12-07_00_09_23_371: EngineState: Power cycle count: 4
2025-12-07_00_09_23_371: EngineState: Last error type ChibiOS panic
2025-12-07_00_09_23_371: EngineState: msg Lua lost track of allocated mem
2025-12-07_00_09_23_371: EngineState: file ChibiOS/os/rt/src/chsys.c
2025-12-07_00_09_23_371: EngineState: line 220
2025-12-07_00_09_23_371: EngineState: SP 0x200099D4
``` 

now:

```
2025-12-07_01_19_25_344: EngineState: TS -> Page 0 write chunk offset 12876 count 832 (output_count=393)
2025-12-07_01_19_25_344: EngineState: confirmation_luareset:8
2025-12-07_01_19_25_344: EngineState: unregistering NONE
2025-12-07_01_19_25_344: EngineState: unregistering NONE
2025-12-07_01_19_25_350: EngineState: unregistering NONE
2025-12-07_01_19_25_350: EngineState: unregistering NONE
2025-12-07_01_19_25_353: EngineState: unregistering NONE
2025-12-07_01_19_25_353: EngineState: unregistering NONE
2025-12-07_01_19_25_355: EngineState: unregistering NONE
2025-12-07_01_19_25_355: EngineState: unregistering NONE
2025-12-07_01_19_25_360: EngineState: LUA: Tearing down instance...
2025-12-07_01_19_25_362: EngineState: LUA loading script length: 6260...
2025-12-07_01_19_25_362: EngineState: LUA ERROR loading script: not enough memory
2025-12-07_01_19_25_362: EngineState: LUA: Tearing down instance...
```

### DIFF
--- a/firmware/controllers/lua/lua_heap.cpp
+++ b/firmware/controllers/lua/lua_heap.cpp
@@ -181,6 +181,7 @@ void* luaHeapAlloc(void* /*ud*/, void* optr, size_t osize, size_t nsize) {
     }
 
     if (optr) {
+    	chDbgAssert(osize <= chHeapGetSize(optr), "Lua lost track of allocated mem");
         // An old pointer was passed in. Only free it if we successfully allocated a new one.
         size_t oldSize = chHeapGetSize(optr);
         if (nptr != nullptr) {


FDSoftware:
@dron0gus can you double-check this a bit to see if am not doing something too dumb? thx :D

dron0gus:
@FDSoftware you are doing everything right. Until we sure Lua is 100% bug-free and we use it 100% correct way :)
Reason I add this `chDbgAssert` comparing Lua's `osize` and ChibiOS `chHeapGetSize()` is because I have seen calls to `luaHeapAlloc()` with unrealistic `osize`.
I suspect without this call we can hide problem under the carpet. And FW will still crash later.
You can add `efiPrintf` (with some delay not to overflow log buffer) at `luaHeapAlloc()` to see all arguments.
During my tests I fount that Lua start making calls to `luaHeapAlloc()` with corrupted data when running out of memory.
AFAIR Lua is doing garbage collecting if it fails to allocate new chunk of memory. And I saw this buggous calls to `luaHeapAlloc()` exactly from garbage collector. May be it is running out of stack (callstack is quite long in this case), may be it corrupting some internal structures - I was to able to track the issue. So I just added this trap.

dron0gus:
> https://www.lua.org/manual/5.3/manual.html#lua_Alloc
> 
> > When ptr is not NULL, osize is the size of the block pointed by ptr, that is, the size given when it was allocated or reallocated.
> > When ptr is NULL, osize encodes the kind of object that Lua is allocating. osize is any of [LUA_TSTRING](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TSTRING), [LUA_TTABLE](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TTABLE), [LUA_TFUNCTION](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TFUNCTION), [LUA_TUSERDATA](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TUSERDATA), or [LUA_TTHREAD](https://www.lua.org/manual/5.3/manual.html#pdf-LUA_TTHREAD) when (and only when) Lua is creating a new object of that type. When osize is some other value, Lua is allocating memory for something else.
>

`chDbgAssert` is in `optr != nullptr` branch, so we should not receive any of special `osize` codes here.

FDSoftware:
> `chDbgAssert` is in `optr != nullptr` branch, so we should not receive any of special `osize` codes here.

Yup, you're right, I didn't realize. I've added it back now thx!. I'm not seeing any weird crashes for now. and it keeps exiting the script without restarting the board in case we run out of ram.

Issue #8665, opened 2025-10-02
ts integation magic classes by version


rusefillc:
See TunerStudioIntegration.java

at the moment magic ProGuarded names are hard-coded for one "3.3.0"

We need a dictionary of version to include previous release https://github.com/rusefi/rusefi_external_utils/blob/master/TunerStudio/TunerStudioMS_Setup_v3.2.03.exe

rusefillc:
this code will be re-used by https://github.com/rusefi/rusefi/issues/8666

FDSoftware:
I need to investigate a bit more, but we're also going to have problems with the Listeners for the wizards with this version of TS:

```
I 251221 015728.091 [pluginFetchThread] TsHeadlessPlugin - Created TsHeadlessPluginImpl instance
E 251221 015728.091 [pluginFetchThread] BackgroundWizard - error on onlineListener com.efiAnalytics.plugin.ecu.ControllerException: Can not subscribe to ECU Configuration AppEvent
 It does not appear to be loaded.
com.efiAnalytics.plugin.ecu.ControllerException: Can not subscribe to ECU Configuration AppEvent
 It does not appear to be loaded.
	at bg.a.subscribe(Unknown Source)
	at com.rusefi.wizard.BackgroundWizard.start(BackgroundWizard.java:46)
	at com.rusefi.ts_plugin.headless.TsHeadlessPluginImpl.run(TsHeadlessPluginImpl.java:26)
	at com.rusefi.ts_plugin.headless.TsHeadlessPlugin.start(TsHeadlessPlugin.java:44)
	at com.rusefi.ts_plugin.TsPluginLauncher.lambda$new$0(TsPluginLauncher.java:37)
	at java.lang.Thread.run(Thread.java:750)
I 251221 015728.091 [wizard-thread] BinaryProtocol - BINARY_IO_TIMEOUT=10000
I 251221 015728.091 [wizard-thread] BinaryProtocol - CONNECTION_RESTART_DELAY=10000
I 251221 015728.123 [AWT-EventQueue-0] TsPluginLauncher - displayPlugin 
I 251221 015728.123 [AWT-EventQueue-0] BackgroundWizard - ECU signature invalid for this plugin: 
```

PR #8897, opened 2025-12-25
background wizard, fix crash on  ts 3.2.03

FDSoftware:
related #8665


FDSoftware:
btw happy christmas :confetti_ball:

PR #8909, opened 2025-12-31
e46 dash unit tests

### LINKED ISSUE #8907: e46 dash unit test
we have no coverage for can_dash.cpp

let's come up with a way to unit test CAN bus transmissions

FDSoftware:
resolves #8907


FDSoftware:
coverage report:

image

PR #8913, opened 2026-01-01
Crop black borders and menu from TS dialog screenshots

chuckwagoncomputing:
This is a little bit hacky because (if I'm understanding correctly) we don't have the real class names of the menu bar. I'm getting "aP.bl" for the menu bar and "aP.hg" for the container. I assume those won't be consistent between TS builds and versions.


rusefillc:
@FDSoftware FYI

PR #8918, opened 2026-01-03
Script to generate SVGs from TS screenshots

chuckwagoncomputing:
This required adding the slice's x origin value to the XML. I could have gotten the y from the file name, but I just went ahead and added the y value while I was at it.

I'm also kind of standardizing on removing units from labels, because the end goal is to put SVGs in the wiki and the links need to target headers, and in my opinion the following header would be acceptable:

## Trigger: VVT Offset Bank 1 Intake

While this header is not to my tastes:

## Trigger: VVT Offset Bank 1 Intake(value)


rusefillc:
nevermind 

pip install Pillow

image

PR #8929, opened 2026-01-06
Make submodule init work properly

chuckwagoncomputing:


chuckwagoncomputing:
Red CI seems to be unrelated, try re-running it I guess?

PR #8972, opened 2026-01-16
GHA to avoid accidentally breaking TS template again

FDSoftware:
image


rusefillc:
let's have this as GHA?

there is no mechanism to enforce pre-commit hooks and I do not want to have python as a local requirement. python as part of GHA I am totally fine with

FDSoftware:
image

https://github.com/rusefi/rusefi/actions/runs/21125142711/job/60744971809?pr=8972

working :D

PR #8974, opened 2026-01-16
Add idle solenoid inversion

drunkbatya:
## About
This PR adds an option to invert the idle valve PWM duty cycle.

## Details
Some idle air control valves (for example, the OSSCA 13450 used on the Opel X20SE engine) are open by default and close as the PWM duty cycle increases. So, the PWM signal needs to be inverted to achieve correct control.

I think the "Idle->Idle hardware->Idle Solenoid output(s) Mode->default inverted" setting should do this job, but it doesn’t. As i understand it correctly, this setting is not intended to work on MicroRusEFI because the PWM signal is generated by the TL8888, not the STM32. I didn’t mess with this LL setting and added a new one that simply inverts the duty cycle.

I found a [mention of a similar case](https://www.facebook.com/groups/rusEfi/posts/1561327974301840) in a 4 years old message from your official Facebook group with no response, i think this PR helps them too.

## Testing
I tested these changes on my X20SE powered by MRE 0.6.0.

## Other
Thank you for your project, it's awesome work!


rusefillc:
This PR cannot proceed as is

MRE is an 0.1% use-case but this PR makes life harder for all the users where existing considerations are sufficient.

Also, why do we even address this on the electrical level, not using 100% table cell for closed IAC and 0% table cell for open IAC?

drunkbatya:
> This PR cannot proceed as is
> 
> MRE is an 0.1% use-case but this PR makes life harder for all the users where existing considerations are sufficient.
> 
> Also, why do we even address this on the electrical level, not using 100% table cell for closed IAC and 0% table cell for open IAC?

> this PR makes life harder for all the users where existing considerations are sufficient

Why it makes life harder? This is another option with proper default value and comment describes why it exists.

> not using 100% table cell for closed IAC and 0% table cell 

Are you about "Base Idle Valve Position" table?

rusefillc:
"Why it makes life harder? This is another option"

Because any option which is not needed for 99% of the users if making everyone lives harder - users need to know to ignore, tech support needs to know to ignore

I am about any solenoid IAC control. I am under the assumption that nowhere we dictate that 0% has to be closed and 100% has to be open?

drunkbatya:
>Because any option which is not needed for 99% of the users if making everyone lives harder - users need to know to ignore, tech support needs to know to ignore

As i see it, there are a lot of options in TS that are useless for 99% of users, but some users will say thanks for their existence. I think this is normal because of the universality of this ECU. Maybe you can advise another location for this setting?

>I am about any solenoid IAC control. I am under the assumption that nowhere we dictate that 0% has to be closed and 100% has to be open?

Yes, there is no open/closed - 0%/100% mapping, assuming all valves are normally closed.

drunkbatya:
@rusefillc what is the proper way to do this? I can limit the visibility of this setting in the UI to MRE only, or maybe do something different?

rusefillc:
@drunkbatya please poke me whatsapp or telegram

drunkbatya:
@rusefillc how i can reach you in TG?

rusefillc:
https://t.me/rusefiecu

rusefillc:
https://github.com/rusefi/rusefi/issues/8976

Issue #8247, opened 2025-07-09
proper disable scatter in simulator


rusefillc:
See https://github.com/rusefi/rusefi/pull/8208

Let's properly disable scatter in simulator: for scatter page we shall return array of zeros, same as we do for missing live data pages

rusefillc:
@FDSoftware is this the root cause of simulator being completely broken?

FDSoftware:
> [@FDSoftware](https://github.com/FDSoftware) is this the root cause of simulator being completely broken?

Is this broken on Windows? I can connect to TunerStudio with current master without problems under linux

<img width="900" height="702" alt="Image" src="https://github.com/user-attachments/assets/e6be277e-86a4-424e-aeca-f404d8cbcd12" />

rusefillc:
@FDSoftware i believe simulator is currently official broken *somehow?*

<img width="581" height="771" alt="Image" src="https://github.com/user-attachments/assets/accff90e-e1b7-4121-b7c0-2d22f5185245" />

can you ask FB people or discord for details?

PR #8982, opened 2026-01-19
Disable system state check in simulator under windows

FDSoftware:
resolves: #8247

image

### DIFF
--- a/simulator/chconf.h
+++ b/simulator/chconf.h
@@ -548,8 +548,15 @@
  *
  * @note    The default is @p FALSE.
  */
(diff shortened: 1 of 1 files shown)


FDSoftware:
We don't have stack trace enabled for the simulator, so I couldn't get much information about the crash. It happens on Windows 11; Windows 10 doesn't seem to have any problems, which makes it even stranger. :c

I could spend a couple more hours trying to find the commit that introduced the error, but I don't think it's worth it, especially for Windows end users.

PR #9005, opened 2026-01-23
WBO: add restart button

dron0gus:


dron0gus:
Screenshot from 2026-01-23 13-48-56

PR #9034, opened 2026-01-29
progress or hallucination?

rusefillc:
only:llm


FDSoftware:
image
looks correct to me, the menu is correctly parsed now

Issue #9029, opened 2026-01-28
re-do freedom units


rusefillc:
Can we do better than 
https://github.com/rusefi/rusefi/issues/8580 https://github.com/rusefi/rusefi/issues/8584

FDSoftware:
We would need to add support for `bitStringValue` and `stringValue` TS functions now that we're using them extensively for the gauges, otherwise we get these errors:

<img width="669" height="646" alt="Image" src="https://github.com/user-attachments/assets/14526f13-18e3-41c2-b101-05d0934ddbd7" />

PR #9035, opened 2026-01-29
freedom units with constants, new migrator & config & sample with gauge

FDSoftware:
related #9029 

image
image


rusefillc:
@FDSoftware hoooold on

do we have drama and risk due to "useMetric = true"? Would this be easier if we go "useImperial"?

rusefillc:
@FDSoftware separate question: what would happen if migrator is NOT used, only say DFU update and TS project update?

FDSoftware:
> @FDSoftware separate question: what would happen if migrator is NOT used, only say DFU update and TS project update?

gauges/table will be showing F units until the user change the setting to metric

PR #9052, opened 2026-02-02
Reset minCrankingRpm

### LINKED ISSUE #2735: minCrankingRpm does not reset between cranking attempts

chuckwagoncomputing:
Fix #2735


chuckwagoncomputing:
Red in firmware build is Gradle download failure

Issue #7256, opened 2025-01-06
Something is not right about EFI_MAIN_RELAY_CONTROL


rusefillc:
What is EFI_MAIN_RELAY_CONTROL about? Seems to be set to 'TRUE' on most of the boards with exception of EG33 somehow?!

rusefillc:
PS: there is also MRE where it's not defined

rusefillc:
Let me at least flip the default to TRUE

FDSoftware:
> PS: there is also MRE where it's not defined

found this:

> at the moment microRusEFI would not be happy with vbatt > 5 since microRusEFI senses main relay output

https://github.com/rusefi/rusefi/commit/d97501b6014ee5efc80905d8eef0040cf663821e

also some info about the relay thing:
https://github.com/rusefi/rusefi/pull/1032

> I'm not sure why that means we should turn off the main relay.

> Currently we don't have a lower-power standby mode in firmware. And we don't want to use an ignition switch voltage as a power source for the ECU, because we need some time for proper firmware shutdown and don't want to just instantly kill the firmware with the turn of the key. We may need time to park the stepper motor, or to keep the ECF running some extra seconds to cool down the hot engine, or just wait for engine to stop and save to flash the settings, etc..
> 
> So the idea behind EFI_MAIN_RELAY_CONTROL is to use the main relay output voltage as the power source for ECU instead ignition switch voltage. And then do a self switch-off when the ignition key voltage drops when a user turns the ignition key, but only after the shutdown is complete - that's why we have isInShutdownMode(). So please note that in EFI_MAIN_RELAY_CONTROL we track vBatt as the ignition key voltage, not the ECU power voltage!

mck1117:
> So the idea behind EFI_MAIN_RELAY_CONTROL is to use the main relay output voltage as the power source for ECU instead ignition switch voltage.

The other purpose of gating main relay enable on ignition voltage is that some cars have the main relay wired such that the ECU can turn it on without the key on. If the main relay was always enabled while the ECU is on, you'd connect USB (key off) and the main relay would turn on!

rusefillc:
> The other purpose of gating main relay enable on ignition voltage is that some cars have the main relay wired such that the ECU can turn it on without the key on. If the main relay was always enabled while the ECU is on, you'd connect USB (key off) and the main relay would turn on!

This part really has to be runtime configuration conditional, not compile time conditional at least on the universal boards?

Issue #2257, opened 2021-01-31
Refactor EFI_MAIN_RELAY_CONTROL


rusefillc:
Too few people underhand EFI_MAIN_RELAY_CONTROL mode. The "EFI_MAIN_RELAY_CONTROL" name is definitely not helping

* do we need more than one main relay strategy? why?
* if we do, what would be a better name for EFI_MAIN_RELAY_CONTROL?

rusefillc:
turns out we actually have three MR inmplementations:

1) EFI_MAIN_RELAY_CONTROL with FSIO
2) EFI_MAIN_RELAY_CONTROL special edition if FSIO is disabled
3) not EFI_MAIN_RELAY_CONTROL

rusefillc:
EFI_MAIN_RELAY_CONTROL is primarily about shutdown

andreika-git:
We can try:
1) Use EFI_MAIN_RELAY_CONTROL branch as default and remove branches;
2) For LE_METHOD_VBATT, we return sensors.vBatt, not getVBatt().
Then, hopefully, the mainRelayLogic (fsio) will work correctly in all cases:
- if ECU is powered from the ignition switch voltage, then it doesn't matter anyway, the firmware dies if the switch is off;
- if vBatt is tied to the battery voltage, then it means the default relay control behavior (always on);
- if vBatt is tied to the ignition switch voltage, and ECU is powered from the main relay output, then is acts as was intended by EFI_MAIN_RELAY_CONTROL...
- if vBatt is not measured yet (no slow ADC yet), then we have "time_since_boot" protection (which should be renamed to time_since_ignition_switch or smth?)
- if vBatt is not enabled at all, then it falls back to "12" value and the main relay is always on (default);
-

rusefillc:
Depends on #2260

rusefillc:
we now have #2260 thank you @mck1117 

this issue is on the critical path of Hellen72

chuckwagoncomputing:
The only functionality of the dead code that's not already implemented is requesting an engine shutdown when ignition voltage is lost.

Issue #3631, opened 2021-11-28
properly handle delayed shutoff


mck1117:
Requires some hardware support.

1. user turns off ignition
2. ECU handles some work before shutdown (park stepper, unmount filesystems, etc)
3. ECU turns off main relay, cutting its own power.

There exists a highly convoluted implementation that doesn't actually work.  See #3180.

NMSTEC:
4. Please add control via LUA for this. Will require graceful CAN shutdown on certain applications, including VAG hellen stuff, and very much Harley stuff.

NMSTEC:
Related to #3660

rusefillc:
https://github.com/rusefi/rusefi/wiki/Delayed-shut-off

matsbloodworks:
![Image](https://github.com/user-attachments/assets/ad9a7d17-cfb6-4553-aa0b-f8959d0286b4)
Hi everyone, I've been testing this circuit and it seems the diode orientation is inverted in the diagram. When I flipped it, it worked as expected, turning the Ignition key to ON supplies power to both "+12V From Main Relay" and "12V Ignition IN".

PR #9061, opened 2026-02-04
Refactor main relay  - Part 1

chuckwagoncomputing:
#2257
#7256

The first commit removes all EFI_MAIN_RELAY_CONTROL defines, and the second removes dead code: a function with its body commented out and another function only called from the first function.

The third commit is the actual change in this pull request, which I want approval for before I go farther.  
Previously, the function `isMainRelayEnabled` always returned true if the firmware was built with EFI_MAIN_RELAY_CONTROL set to false.
With this change, it will always return true if the main relay pin has not been configured.
This function is only used in two places:
1. Currently noop
https://github.com/rusefi/rusefi/blob/d8cca084bb9234ebfd4c91f1a0167167fcbe44fd/firmware/controllers/limp_manager.cpp#L238-L244
2. A somewhat dubious use in stepper control
https://github.com/rusefi/rusefi/blob/d8cca084bb9234ebfd4c91f1a0167167fcbe44fd/firmware/hw_layer/stepper.cpp#L176-L178

### DIFF
--- a/firmware/controllers/algo/engine.cpp
+++ b/firmware/controllers/algo/engine.cpp
@@ -476,12 +476,12 @@ void Engine::onEngineHasStopped() {
 }
 
 bool Engine::isMainRelayEnabled() const {
-#if EFI_MAIN_RELAY_CONTROL
+	// If the relay is unconfigured, we'll just assume it's on because either it's not controlled by the ECU,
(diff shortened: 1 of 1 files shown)


rusefillc:
https://github.com/rusefi/rusefi/issues/3631

PR #9135, opened 2026-02-17
Fix triggers.txt generation bug

### LINKED ISSUE #9132: Some triggers.txt output and therefore trigger images seem incorrect
Example:
Ford Aspire
https://github.com/rusefi/rusefi/blob/7ee45aced6d0e6099e9decd4b19a916b52b1f522/firmware/controllers/trigger/decoders/trigger_mitsubishi.cpp#L18-L31
should work out to:
|channel|value|angle|angle after last|
|--|--|--|--|
|0|0|720|360|
|1|1|52.76|110.86|
|1|0|121.90|69.14|
|1|1|232.76|110.86|
|1|0|301.90|69.14|
|0|1|360|360|
|1|1|412.76|110.86|
|1|0|481.90|69.14|
|1|1|592.76|110.86|
|1|0|661.90|69.14|

So it seems that the first wheel is a half circle, and the second wheel has 4 teeth 180 degrees (110.86 + 69.14 = 180) apart.

However, this is the generated trigger image:

Image

This comes from this part of triggers.txt:
https://github.com/rusefi/rusefi/blob/7ee45aced6d0e6099e9decd4b19a916b52b1f522/unit_tests/triggers.txt#L16-L25

In the same table format:
|channel|value|angle|angle after last|
|--|--|--|--|
|0|1|0|487.24|
|1|1|0|127.24|
|1|0|52.76|52.75|
|1|1|52.76|0|
|1|0|232.76|180|
|0|1|232.76|232.76|
|1|1|412.76|180|
|1|0|412.76|0|
|1|1|592.76|180|
|1|0|592.76|0|

This makes no sense to me.
Both of the first wheel's events coincide with events on the second wheel, which shouldn't be the case.
The first wheel's events are no longer 360 degrees apart.
The second wheel's events are still 180 degrees apart, except for the one which crosses the 720/0 line.

Maybe I'm an idiot and I'm just not understanding something, but this seems very wrong.

chuckwagoncomputing:
Fix #9132

Affects:
TT_FORD_ASPIRE
TT_MAZDA_DOHC_1_4
TT_JEEP_18_2_2_2
TT_JEEP_4_CYL
TT_TRI_TACH
TT_GM_60_2_2_2

### DIFF
--- a/firmware/controllers/trigger/trigger_decoder.cpp
+++ b/firmware/controllers/trigger/trigger_decoder.cpp
@@ -122,8 +122,6 @@ void TriggerFormDetails::prepareEventAngles(TriggerWaveform *shape) {
 	angle_t firstAngle = shape->getAngle(triggerShapeSynchPointIndex);
 	assertAngleRange(firstAngle, "firstAngle", ObdCode::CUSTOM_TRIGGER_SYNC_ANGLE);
 
-	int riseOnlyIndex = 0;
-
 	size_t length = shape->getLength();
 
 	setArrayValues(eventAngles, 0);
@@ -161,9 +159,8 @@ void TriggerFormDetails::prepareEventAngles(TriggerWaveform *shape) {
 
 				// In case this is a rising event, replace the following fall event with the rising as well
 				if (shape->isRiseEvent[triggerDefinitionIndex]) {
-					riseOnlyIndex += 2;
-					eventAngles[riseOnlyIndex] = angle;
-					eventAngles[riseOnlyIndex + 1] = angle;
+					eventAngles[eventIndex] = angle;
+					eventAngles[eventIndex + 1] = angle;
 				}
 			} else {
 				eventAngles[eventIndex] = angle;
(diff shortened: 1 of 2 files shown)


rusefillc:
@chuckwagoncomputing please add code comments explaining the magic?

this PR touches a pretty sensitive area of the code which does not seem to have reasonable unit test coverage, let's be careful/clear/transparent

Would this change adjust sync point location/require "global trigger offset" move for affected triggers?

chuckwagoncomputing:
It seems unlikely to me that any of these triggers could have been working properly, but I suppose if they were, this could potentially change the offset.

chuckwagoncomputing:
It just occurred to me that there could be another bug hiding here. Let me check...

chuckwagoncomputing:
I think there would be a problem if a pattern had:
- two wheels
- rising-edge only
- multiple teeth on both wheels

As of now, we have no pattern with this case. The wheels fixed with this PR match the first two, but not the third.

PR #9139, opened 2026-02-17
Validate TS field expressions to prevent forward references

### LINKED ISSUE #9138: master should be red in light of useMetricOnInterface usages before definition
TS says not to use useMetricOnInterface prior to definition

What's the best place for CI to fail? is that going to be Config reader? is it going to be some later validation

Separate question - have we already implemented runtime logic which survives non-existent useMetricOnInterface? Would we need to adjust?

FDSoftware:
```
Caused by: java.lang.IllegalStateException: Field 'maxAcClt' references 'useMetricOnInterface' which is defined later in the structure (forward reference). Reorder fields in rusefi_config.txt so that 'useMetricOnInterface' is defined before it is used.
        at com.rusefi.output.ConfigStructureImpl.checkForwardReferences(ConfigStructureImpl.java:236)
        at com.rusefi.output.ConfigStructureImpl.checkForwardReferences(ConfigStructureImpl.java:218)
        at com.rusefi.output.ConfigStructureImpl.validateNoForwardReferences(ConfigStructureImpl.java:200)
        at com.rusefi.ReaderStateImpl.handleEndStruct(ReaderStateImpl.java:258)
        at com.rusefi.ReaderStateImpl.readBufferedReader(ReaderStateImpl.java:314)
        at com.rusefi.ReaderStateImpl.doJob(ReaderStateImpl.java:141)
        ... 2 more
``` 

resolves #9138

### DIFF
--- a/java_tools/configuration_definition_base/src/main/java/com/rusefi/ReaderStateImpl.java
+++ b/java_tools/configuration_definition_base/src/main/java/com/rusefi/ReaderStateImpl.java
@@ -253,6 +253,10 @@ public class ReaderStateImpl implements ReaderState {
         if (log.debugEnabled())
             log.debug("Ending structure " + structure.getName());
         structure.addAlignmentFill(this, 4);
+        // Validate only on top-level structs
+        if (isStackEmpty()) {
+            structure.validateNoForwardReferences();
+        }
 
         ConfigStructureImpl existing = structures.put(structure.getName(), structure);
         if (existing != null)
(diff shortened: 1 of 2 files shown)


FDSoftware:
i dont expect an angry ci here?, since makefile has:

```
# CONFIG_DEFINITION is always rebuilt, but the file will only be updated if it needs to be,
# so it won't trigger a config file generation unless it needs to.
.config-sentinel: $(CONFIG_INPUTS) $(CONFIG_DEFINITION_JAR) $(TGT_SENTINEL)
ifneq (,$(CUSTOM_GEN_CONFIG))
	bash $(BOARD_DIR)/$(CUSTOM_GEN_CONFIG)
else
	bash $(PROJECT_DIR)/gen_config_board.sh $(BOARD_DIR) $(SHORT_BOARD_NAME)
endif
	@touch $@
```

FDSoftware:
nvm, CI is pretty angry :D

image

PR #9146, opened 2026-02-20
Fix RiseOnly triggers again

chuckwagoncomputing:
The bug I imagined hiding in #9135 did in fact exist. I thought that no patterns had multiple teeth on both wheels, but because TriTach is FOUR_STROKE_CRANK_SENSOR, it effectively did.
This will fix TriTach and any future cases.

### DIFF
--- a/firmware/controllers/trigger/decoders/trigger_structure.cpp
+++ b/firmware/controllers/trigger/decoders/trigger_structure.cpp
@@ -416,6 +416,17 @@ uint16_t TriggerWaveform::findAngleIndex(TriggerFormDetails *details, angle_t ta
 	return left;
 }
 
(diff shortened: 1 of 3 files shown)


rusefillc:
scary? poor practice? that we fix defects while not having/touching a test? https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing

tests are pretty possible (well, maybe not for non-working/disabled TriTech)

chuckwagoncomputing:
Is that HW CI failure a known issue? It doesn't look related to this.

rusefillc:
both HW CI failing on branch is pretty unusual

just make couple of pushes into master just to compare how it would behave in terms of HW CI

rusefillc:
closing, assume HW CI is a real signal here

Issue #6357, opened 2024-04-11
TriTach trigger


rusefillc:
Two problems
* As is right now firmware could not be configured for that configuration. 
* even if firmware is configured current approach to trigger decoder would consume too much CPU making high RPM operation questionable

rusefillc:
between level of effort and benefits this is low priority.

PR #9172, opened 2026-02-25
Tri Tach

brngates98:
Apologies for vanishing had gotten busy with work and had been traveling, I have returned and finished the tri tach fixes. I redid a lot of my original logic and I think we now have a viable product. I did get a logic analyzer while I was gone so I plan to get some captures from that as soon as I figure out how to use it :)

So far it has tested all build and unit tests and seems to function alright. I have tested this on my 95 urs6 and it seems to work decently.

Going to be playing around with the logic analyzer to see if I can figure out how it works and get more accurate test data. Let me know your thoughts.

https://github.com/rusefi/rusefi/issues/6357


rusefillc:
I am sorry but I strongly suspect that this is AI slop.

Considering the sad history I have to focus on self preservation.

PR #9182, opened 2026-02-27
Set isSecondWheelCam for 1+1

chuckwagoncomputing:
As far as I can tell, this parameter is only used for TriggerImage.
The second wheel's angles would make no sense as a crank trigger, so I'm pretty sure it's meant to be a cam trigger.
Before:
trigger_TT_ONE_PLUS_ONE
After:
trigger_TT_ONE_PLUS_ONE


chuckwagoncomputing:
Cause of redness:
```
Resolving rusefi.com (rusefi.com)... 74.208.236.155
Connecting to rusefi.com (rusefi.com)|74.208.236.155|:443... connected.
OpenSSL: error:0A000438:SSL routines::tlsv1 alert internal error
Unable to establish SSL connection.
```

PR #9207, opened 2026-03-06
Fix gcc warnings

esaulenka:
a set of evident fixes to remove annoying red words from compiler's output


rusefillc:
this PR points at legit issues but I am against hiding things under the carpet

* we have cases of legit unused arguments - arguments should be removed
* we have noise related to conditional compilation - not sure how to address, maybe conditional UNUSED
* we have noise related to legacy "error code as negative value" pattern of error code handling - we need more modern error code patterns
etc

rusefillc:
Many of these were cherry-picked, thank you!

PR #9213, opened 2026-03-09
rusefi.compatibility for older logic on #if / imperial units / etc

### LINKED ISSUE #9211: latest updater crashes because it does not support #if from older board
Image

FDSoftware:
resolves #9211


rusefillc:
@FDSoftware please elaborate. I see a ton of code, I have no idea what is it about. 

In general we might need to be adding more code comments with many/more changes.

FDSoftware:
> @FDSoftware please elaborate. I see a ton of code, I have no idea what is it about.
> 
> In general we might need to be adding more code comments with many/more changes.

going a smaller step at the time with more context here:
https://github.com/rusefi/rusefi/pull/9216

PR #9027, opened 2026-01-28
remove all #if LAMBDA, use new useLambdaOnInterface field

FDSoftware:
image
image

btw there's no need to reload the project, it's all on-fly!

resolves: #8650

PR #9216, opened 2026-03-10
compatibility for older units, IniPreprocessor

FDSoftware:
replaces #9213
progress to https://github.com/rusefi/rusefi/issues/9211

general idea is having a new java package (rusefi.compatibility) so we can move all the relative deadcode from #9027 mantening the compatiblity with older units (evaluating on a simpler way the #if directives), so we can also solve #9028 issue making  `ConfigField`, `IniFileReader`, etc a bit simpler

first PR with the new evaluator without any wiring to actual Ini file read flow

### DIFF
--- a/java_console/inifile/src/main/java/com/rusefi/compatibility/ini/IniPreprocessor.java
+++ b/java_console/inifile/src/main/java/com/rusefi/compatibility/ini/IniPreprocessor.java
@@ -1,0 +1,105 @@
+package com.rusefi.compatibility.ini;
+
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.List;
+
+/**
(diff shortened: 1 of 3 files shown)


rusefillc:
https://github.com/rusefi/rusefi/pull/10007 reminded me about this one

@FDSoftware do we need this PR? what was this mitigating?

FDSoftware:
> #10007 reminded me about this one
> 
> @FDSoftware do we need this PR? what was this mitigating?

this one is for #if/#else/etc era tunes to the new constant ones (mostly the imperial units/afr stuff)

Currently the console does not crash in this case, we only display the message that "your tune cannot be recovered, we will use default values"

rusefillc:
@FDSoftware let me close this one, not crashing and explicitly not supporting old stuff seems like the most reasonable thing for me. Trying to stay lean.

Issue #9144, opened 2026-02-20
diff defect


rusefillc:
I was updating a few units with same instance of console

do we not properly handle some state change when one connected unit is being changed for another unit?

<img width="873" height="536" alt="Image" src="https://github.com/user-attachments/assets/04416260-64aa-4ce7-a552-fd15a6c24c14" />

PR #9224, opened 2026-03-11
diff defect - progress

FDSoftware:
wip, need to test with real target/ecu on windows

related #9144


FDSoftware:
Tested updating 3 different uaefi with diferent versions consecutively, it seems to work well

PR #9236, opened 2026-03-13
Attempt fix of g16b trigger

chuckwagoncomputing:
It passes the AllTriggersFixture test now...

### DIFF
--- a/firmware/controllers/trigger/decoders/trigger_suzuki.cpp
+++ b/firmware/controllers/trigger/decoders/trigger_suzuki.cpp
@@ -38,20 +38,22 @@ void initializeSuzukiG16B(TriggerWaveform *s) {
 	s->initialize(FOUR_STROKE_CAM_SENSOR, SyncEdge::RiseOnly);
 	
 	float w = 5; // width of tooth in degrees
-	
+
 	angle_t teethAngles[] = { 0, 15, 55, 90, 130, 145, 180, 235, 270, 310, 325 }; // angle center of tooth
-	
+
 	int size = sizeof(teethAngles) / sizeof(teethAngles[0]);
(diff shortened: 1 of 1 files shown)


rusefillc:
@chuckwagoncomputing what's the problem statement?

real issue is lack of logic analyzer test data, i am very paranoid about functional changes without test changes :(  https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing

chuckwagoncomputing:
Problem statement is that AllTriggersFixture was failing on this trigger, resulting in no output to triggers.txt and no TriggerImage.

This commit rotates it. This solves two errors:
1. Makes the first event not negative.
2. Makes the last event at 360.

The third error is that the set sync gap was apparently based upon a misunderstanding of how sync gaps work. This meant that the trigger couldn't synchronize when simulated.

I agree that real logic analyzer test data would be useful, but given that we don't have it, we might as well fix broken tests.

rusefillc:
TT_SUZUKI_G16B 95

@chuckwagoncomputing now I see that we do not have "TRIGGERTYPE 94" but how come build is not failing? I would like to fix https://github.com/rusefi/rusefi/issues/9239

rusefillc:
first event not negative should have failed much sooner?!

Makes the last event at 360 I was hoping that we validate that?!

Kami0n:
This PR added an error to the code: that is causing the TDC to be off by 35 degrees.
Because offset is 360 - 325 = 35

As far as I know, offset is not needed.
My mistake is that I started array at 0 instead of ending it with 360.

chuckwagoncomputing:
`tdcPosition` should be relative to the sync point, not to zero as defined. Have you observed that the TDC is now off by 35 degrees, or is that based on your analysis of the code?

Issue #9157, opened 2026-02-23
console acts poorly maybe if no permission to scan for DFU


rusefillc:
DfuFlasher#dfuEnabledInCaseOfError seems to be hiding a problem

new behavior should be to show a red message on the bottom of the frame "Failed to check for DFU devices. Try 'Run as Administrator'" if both conditions are met:
1: ErrorExecutingCommand has happened
2: no serial bus ECU or simulator has been detected within 15 seconds since application start

Graymae reports on discord that he has uaefi in DFU mode, confirmed with device manager

<img width="317" height="117" alt="Image" src="https://github.com/user-attachments/assets/d97a8b1e-d015-47df-92f7-62c7b6e543c0" />

we have two or three problems

* manual DFU is not auto-selected - I suspect failure to execute scanning command?
* if "manual DFU" is auto-selected - it gets un-selected periodically!

rusefillc:
proposal: maybe detect lack of permissions to scan for DFU and suggest that "run as Admin" is required"

byron-dm:
<img width="396" height="303" alt="Image" src="https://github.com/user-attachments/assets/95862640-1e07-4dd9-b8e9-1bd80a25b92e" />

byron-dm:
<img width="883" height="243" alt="Image" src="https://github.com/user-attachments/assets/716114c2-0e48-4802-a012-4375890e3807" />

FDSoftware:
i will try windows 10 in a bit, but seems to be working on my combo of  uaefi target-pc-windows11

FDSoftware:
<img width="883" height="627" alt="Image" src="https://github.com/user-attachments/assets/9e8f5f53-dccc-4515-bd4e-b24df0e16a12" />
windows 10 also ok now :D.

please close @rusefillc?, or we need more work here?

PR #9272, opened 2026-03-20
DFU WMIC race condition fix

FDSoftware:
related #9157


FDSoftware:
I hope the comment is clear enough; it took me a while to find the cause lol

PR #9276, opened 2026-03-22
Improve HW CI

chuckwagoncomputing:
Set environment variables from a file which can be mounted on the docker image at run time. This is used for setting STLink and ECU serials.

Specify STLink serial if it is set in an environment variable, so multiple STLinks can be used on a single computer.


chuckwagoncomputing:
f407-discovery HW CI failure is preexisting

PR #9315, opened 2026-04-03
TunePane: techdebt and new import/export buttons

FDSoftware:
new ui:

image

addded import/export of msq files :D


FDSoftware:
@benassi124 I hope this resolves your question about the import/export msq buttons

PR #9347, opened 2026-04-10
move second tables to a separate TS page

FDSoftware:
related #9335


FDSoftware:
I'll try to separate it into smaller PR's, but overall, does it look okay @dron0gus ?

 I didn't have any problems with uaefi, F7, and H7. I also left the SD card as an option if it's available and the flash fails.

firmware/integration/config_page_4.txt
```diff
@@ -0,0 +1,32 @@
...
+custom pin_input_mode_e 1 bits, U08, @OFFSET@, [0:2], $pin_input_mode_e_enum
+custom gppwm_channel_e 1 bits, U08, @OFFSET@, [0:5], $gppwm_channel_e_enum
+
+struct_no_prefix page4_s
+  ! note this is pretty similar to a blend table!
+  uint16_t[VE_LOAD_COUNT x VE_RPM_COUNT] autoscale secondVeTable;;"%", 0.1, 0, 0, 999, 1
+  uint16_t[VE_LOAD_COUNT] secondVeLoadBins;;{bitStringValue(veLoadUnitLabels, veLoadUnitIdxPcv)}, {!(veOverrideMode == 1 || (veOverrideMode == 0 && fuelAlgorithm == 0)) || useMetricOnInterface ? 1 : @@UNITS_KPA_TO_PSI@@}, 0, 0, {veOverrideMode == 1 || (veOverrideMode == 0 && fuelAlgorithm == 0) ? (useMetricOnInterface ? @@MAP_UPPER_LIMIT@@ : @@MAP_UPPER_LIMIT@@ * @@UNITS_KPA_TO_PSI@@) : (veOverrideMode == 2 ? 100 : @@MAP_UPPER_LIMIT@@)}, 0
+  uint16_t[VE_RPM_COUNT] secondVeRpmBins;;"RPM",	   1, 0, 0, 18000, 0
+
+  switch_input_pin_e secondVeTableInput;Pin that activates the second VE table (hard switch, overrides blend)
```
dron0gus, line 12: Do you think switch input settings should be also on second page?

firmware/controllers/second_tables.cpp
```diff
@@ -0,0 +1,98 @@
+// file second_tables.cpp
+
+#include "pch.h"
+#include "second_tables.h"
+#include "storage.h"
+#include "flash_main.h"
+
+#define PAGE4_DATA_VERSION 1
+
+struct page4_container_s {
```
dron0gus: Someday we should create template for page container?

firmware/integration/config_page_4.txt
FDSoftware: My idea is to treat it as if it were a kind of EngineModules, so it's more similar to what we're going to migrate from EngineModules at some point. That's why I left all the configuration related to the "feature"

firmware/controllers/second_tables.cpp
FDSoftware: done :D

firmware/integration/config_page_4.txt
dron0gus: Ok. Agree.

FDSoftware:
@rusefillc 
I've reorganized the history on this branch, tested it on a couple of boards and it's working fine, so it's ready to be reviewed :D

Issue #9351, opened 2026-04-12
trigger images broken in github master action


rusefillc:
does not look like https://github.com/rusefi/rusefi/commit/89058d6df31f7c0469f5cbc8bb60435a21b618a4 has produced proper images or maybe just at least did not update triggers.txt?

could be https://github.com/rusefi/rusefi/pull/9348

rusefillc:
@chuckwagoncomputing let me try reverting https://github.com/rusefi/rusefi/pull/9348 seems it does not seem to be atomic - it looks like it makes unit tests write text content into yaml file name, there is not CI for .txt to .yaml conversion and it looks like there is manual editing of yaml?

I am overall pretty not convinced about the whole migration, especially since it only seems to produce immediate pain which for me overweighs potential benefits

I guess my take could be summaries with following statement:

"if we are unable to break up huge migration into smaller incremental steps - we probably would not success with huge migration anyway"

if anything I believe that 'change format from txt to yaml" and "change source of truth from C++ to text file" are two separate efforts. It's a red flag that current approach mixes these two things together. I believe that https://github.com/rusefi/rusefi/issues/9350 is a more iterative (meaning SAFER) approach

rusefillc:
current state: even with rollback, triggers.zip artifact does not seem to contain proper ford PIP6 image

chuckwagoncomputing:
I meant it to be atomic - I am not a perfect human and I am sorry for this.

This is meant to be only "change source of truth from C++ to text file". The unit tests can continue to output the old .txt format. It won't be used for anything any more, because the YAML will be the source of truth for generating trigger images as well as the firmware trigger definitions, but it can still exist.

Looking into PIP6...

chuckwagoncomputing:
Hmmm, current master generates PIP6 when run locally.

rusefillc:
@chuckwagoncomputing I've adjusted this ticket summary a bit

I would love automation to work reliably. "[rusEFI validate console, also generate trigger images](https://github.com/rusefi/rusefi/actions/workflows/build-rusEFI-console.yaml)" master artifact does not seem to show correct ford pip6 and pip8, can you please help?

chuckwagoncomputing:
I expect trigger images to be fixed once #9355 is merged.

rusefillc:
> I expect trigger images to be fixed once [#9355](https://github.com/rusefi/rusefi/pull/9355) is merged.

[...]

Issue #9350, opened 2026-04-12
generate triggers.yaml directly with C++


rusefillc:
C++ unit tests produce unit_tests/triggers.txt

triggers_to_yaml.py produces firmware/config/triggers.yaml

i wonder if LLM can help us produce yaml directly with C++ code

chuckwagoncomputing:
I have 2 reasons for basing triggers.yaml on triggers.txt.
1. It's easier/more foolproof to parse triggers.txt.
2. I can close the loop by re-generating triggers.txt based on triggers.yaml, helping to verify that I haven't broken anything.

PR #9355, opened 2026-04-13
Read triggers.yaml

chuckwagoncomputing:
With this commit, triggers.yaml will be used as the source for generating trigger images.

### DIFF
--- a/.github/workflows/build-rusEFI-console.yaml
+++ b/.github/workflows/build-rusEFI-console.yaml
@@ -33,14 +33,14 @@ jobs:
           sudo bash misc/actions/ubuntu-install-tools.sh
           sudo apt-get install xvfb
 
+      - name: Generate Enums & Live Documentation
+        working-directory: ./firmware/
+        run: bash bin/compile.sh config/boards/f407-discovery/meta-info.env docs-enums
+
       - name: Build All Java
         working-directory: .
         run: |
           ./gradlew jar :config_definition_base:shadowJar :enum_to_string:shadowJar :trigger-ui:shadowJar
-
-      - name: Generate Enums & Live Documentation
-        working-directory: ./firmware/
-        run: ./gen_live_documentation.sh
 
       - name: Print GCC version
         run: gcc -v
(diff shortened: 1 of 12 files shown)


rusefillc:
but how does .txt become .yaml?

* can I add new trigger in master and get FW and images automatically?
* what about same with this commit?

chuckwagoncomputing:
After this commit, you can add a new trigger to triggers.yaml and get the image automatically, but it's not added to the firmware yet. That happens in a later commit.

chuckwagoncomputing:
triggers.yaml is not "continuously" generated from triggers.txt because it has manual modifications.
1. Added humanNames
2. triggers.txt has specified operationModes for triggers which don't have a known operationMode - this could be detected by triggers_to_yaml.py, but it would mistakenly be detected for 60_2 (rotary weirdness) so then that one would have to be tweaked or have a hard-coded rule in triggers_to_yaml.py.
3. tri_tach needs manual tweaking because of simultaneous edges.

rusefillc:
3. let's remove/comment out tri_tach since it's the most broken trigger?
1. human names are still duplicated between rusefi_config.txt and trigger something? I would prefer to have some small atomic commit in master addressing human names before the whole huge effort
2. this one we can deal with later :)

chuckwagoncomputing:
3. Sure, whatever, I don't care. It would be pretty cool for some people if it wasn't broken though.
4. humanName duplication is removed in a later commit.
5. What's this one? Rotary?

rusefillc:
Status update:
* I am trying to remove tri_tach https://github.com/rusefi/rusefi/commit/cce201b73f4aec6426b0c367e4a0f16916d3f4c2
* need help fixing master https://github.com/rusefi/rusefi/issues/9351
* I am against transitions from "fully automated master" into "manual activity master", it sounds like either https://github.com/rusefi/rusefi/issues/9350 or https://github.com/rusefi/rusefi/issues/9356 would need to happen next?

chuckwagoncomputing:
Fully automating master now would be a temporary measure that would need to be undone as #9294 is merged.
Implementing #9350 would make C++ the source of truth - in a few more commits, the old C++ is removed.
Implementing #9356 would make triggers.txt the source of truth - triggers.txt is a product of the C++, so in a few commits we'd have a tail-eating snake with no real source of truth.

rusefillc:
@chuckwagoncomputing let me convert this into DRAFT since not fails to meet my CI requirement

PR #9391, opened 2026-04-18
We need to do something about humans

rusefillc:


java_tools/configuration_definition/src/test/java/com/rusefi/test/TSProjectConsumerTest.java
```diff
@@ -149,4 +157,72 @@ public void testReaderDropComments() throws IOException {
...
+
+    @Test
+    public void testDuplicateAcrossPages() {
+        TreeSet<String> usedNames = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+
+        String page1 = "struct page1\n" +
+                "    int field1;PID dTime;\"ms\",      1,      0,       0, 3000,      0\n" +
+                "end_struct\n";
+
+        String page2 = "struct page2\n" +
```
FDSoftware, line 199: I would also check the structure names if there's a case of a duplicate within an inner struct, because you could have something like:


```c++

struct_no_prefix engine_configuration_s

struct vr_threshold_s
	uint8_t[6] autoscale rpmBins;;"rpm", 50, 0, 0, 12000, 0
[....]
end_struct

[....]

struct gppwm_channel
	[...]
	int16_t[GPPWM_RPM_COUNT] rpmBins;;"", 1, 0, -30000, 30000, 0
end_struct


end_struct
```

And it should be valid

(maybe we're already covering this case and I'm just talking nonsense)

rusefillc: @FDSoftware yes, but that would be next PR "one day"

Issue #9232, opened 2026-03-13
Upload Raw Pinouts for custom boards


rusefillc:
https://github.com/rusefi/fw-custom-mega144-f7 needs at least one pin on the pinout

rusefillc:
shall we go with separate per-board files either always or at least for external boards?

```
[MegaTune]
	re_auto_update_root_url = @@RE_auto_update_root_url@@
```

we would also probably need similar for custom URL option

(not sure if that's two or three separate changes at this point)

FDSoftware:
now we have correct yaml for custom boards :D 
<img width="564" height="261" alt="Image" src="https://github.com/user-attachments/assets/8cee9389-c956-48d4-a004-db505f9bfcb7" />

FDSoftware:
> ```
> [MegaTune]
> 	re_auto_update_root_url = @@RE_auto_update_root_url@@
> ```
> 
> we would also probably need similar for custom URL option
> 
> (not sure if that's two or three separate changes at this point)

Do you want me to merge the auto-update urls as another entry within the pinout yaml?

PR #9401, opened 2026-04-18
configurable pinout metadata URL and filename

FDSoftware:
related #9232


FDSoftware:
all the red are timeouts from [gradle-8.14.4-all.zip](https://services.gradle.org/distributions/gradle-8.14.4-all.zip)

Issue #9418, opened 2026-04-20
ethernet F767 nucleo refresh rate


rusefillc:
@FDSoftware reports pretty sad single digit rate at https://github.com/rusefi/rusefi/issues/4665

with some improvement suggestion

rusefillc:
https://github.com/rusefi/rusefi/pull/9420

rusefillc:
@FDSoftware probably related or dup of https://github.com/rusefi/rusefi/issues/10170 ?

FDSoftware:
yup, we fixed the ethernet case a while ago, we can close this one

PR #9420, opened 2026-04-20
F767 ethernet_console: TCP Settings and flush changes

FDSoftware:
old data rate:
image

data rate: yes
image

`+30 KB lwIP static RAM, ~+1.5 KB BSS + ~+120 B text from the coalescer`

related #9418


dron0gus:
WOW!

dron0gus (approved):
A lot of RAM is used for lwIP but it definitely worth it

PR #9425, opened 2026-04-21
Nucleo F767: enable OpenBLT, fix compilation

dron0gus:
Enable OpenBLT for F767 Nucleo,
Fix OpenBLT compilation
Enable Ethernet init errata workaround for both OpenBLT and RE FW.

PR #9438, opened 2026-04-22
OpenBLT: WD counter: zero counter on POR or RST reason

dron0gus:


dron0gus:
Merged as a part of https://github.com/rusefi/rusefi/pull/9425

PR #9443, opened 2026-04-23
ISO-TP frames usually use 8 bytes, even if payload is shorter

esaulenka:
some units may ignore short frames

current version should work well in both cases


rusefillc:
@esaulenka I am afraid to make an incompatible change. Let's have a flag and keep current behavior as default?

esaulenka:
don't panic ;-)
I'm pretty sure that it'll not break anything.
Almost all implementations that I've seen used these additional bytes.

PR #9466, opened 2026-04-26
refactor: extract stepper motors from ETB hardware pool DC

zavdimka:
## Problem

The `DcHardware` pool was shared between ETB/DC-actuator motors and the
H-bridge stepper coil drivers:

```cpp
static DcHardware dcHardware[ETB_COUNT + DC_PER_STEPPER];
//  indices 0-1 → ETB/EWG/DC-idle
//  indices 2-3 → stepper coils A/B
```
This caused a real conflict: using an H-bridge stepper idle valve while
any DC motor function was assigned was silently unsafe. A runtime guard
pickEtbOrStepper() existed only to detect and block this combination.
Additionally, the stepper occupying ETB slots was the root cause of the
hard limit of just 2 DC motor actuators.

## Result
Stepper and DC-actuator motors now use fully independent hardware pools. No behavior change. This is a prerequisite for lifting the 2-slot DC motor limit (allowing ETB + wastegate + DC idle valve simultaneously).

### DIFF
--- a/firmware/controllers/actuators/dc_motors.cpp
+++ b/firmware/controllers/actuators/dc_motors.cpp
@@ -81,7 +81,8 @@
 		}
 	}
 
-static DcHardware dcHardware[ETB_COUNT + DC_PER_STEPPER];
+static DcHardware dcHardware[ETB_COUNT];
+static DcHardware stepperDcHardware[DC_PER_STEPPER];
 
 DcHardware *getPrimaryDCHardwareForLogging() {
 	return &dcHardware[0];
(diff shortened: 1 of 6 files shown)


rusefillc:
to make this PR safer, can we have a separate PR adding code new coverage for current mess? because at the moment too much of this is not covered see https://rusefi.com/docs/unit_tests_coverage/index.electronic_throttle.cpp.be31fe36d59d1a73d6343a64a6587096.html#l1095 etc

no idea if that's even possible or how much sense that would make, but need to ask

rusefillc:
we have to add coverage into existing (legacy, fragile, critical) code first, and only then we would be in the position to re-base this PR

I love the progress here but I am shitless scared of unintended breakage. Manual testing is too expensive :(

PR #9481, opened 2026-04-30
New Cranking Idle Air mode: RPM based

ElDominio:
This new cranking idle air mode uses RPM instead of Duty for cranking purposes. Tuners should be tuning their Idle Duty VS CLT table correctly, in which case we can pull the base idle duty from it, since the purpose of the "Cranking Idle Air" table is to flare RPM at and after cranking, not just "add air". A dropdown menu was added for compatibility purposes

Build_firmware.yaml has cron disabled since i don want to be building nightlies on my branch


rusefillc:
@ElDominio red, please rebase? master is green

ElDominio:
Now green!

rusefillc:
@ElDominio please re-create based on new https://github.com/rusefi/rusefi/wiki/Dev-Quick-Start#pull-requests-basics content

PR #9488, opened 2026-05-01
Cranking idle rpm mode clean

rusefillc:


ElDominio:
is this the expected content of a pull request? i want to know if this is how i should keep doing it

rusefillc:
@ElDominio no

this PR still have a merge commit - i.e. have of commits on this PR should not be here

regardless of GITHUB shit - my preference is separate tables for separate things. Values changing meaning between modes is risky, much smarter to have two separate curves/tables with separate reasonable defaults for each of the modes

ElDominio:
I did not know if it was better to prioritize compatibility? I was afraid of extra tables breaking stuff, but if thatś the case I will do a seperate table

ElDominio:
This can and should be a one value scalar. Closing and will later redo

Just realized pull requests can't be closed?

ElDominio:
Now it's split into two tables, with seperate enables. People can choose to do either legacy (only duty), new (only RPM) and combined (RPM flare + additional air)

In my opinion mot people should only use RPM since the additional air is baked into the RPM request, since our idle duty base table is already considering RPM in the Y axis, but since we ave enables for both users can choose what they prefer.

PR #9525, opened 2026-05-13
lts-25kansas-port: fix CI failures after java backport

FDSoftware:
Follow-up to #9524 

- `config_definition/build.gradle`: add missing `:msq-file` dependency (`TSProjectConsumer` imports `com.rusefi.tune.xml`)
- `ConfigDefinition.java`: restore LTS page numbers (1, 2) instead of master's (2, 3, 4) — LTS firmware only has `config_page_1.txt` and `config_page_2.txt`
- `ConfigFieldImpl.java`: relax `trueName`/`falseName` validation to original LTS behavior — LTS firmware bit fields only specify `trueName`, master's stricter check throws on `falseName==null`
- `inifile/build.gradle`: add `testImplementation mockito` — `IniFileReaderTest` and `TsStringFunctionTest` both use `org.mockito`


FDSoftware:
well, auto-fix with claude is a mess, but was fun to try

Issue #8827, opened 2025-11-18
36-2-1-1 but with terrible compression effects trigger support


ElDominio:
**Is your feature request related to a problem? Please describe.**
Mitsubishi uses the same 36-2-1-1 as 3A92 in the 6G75 (V6 engine). They use the same trigger pattern; but the V6's compression effects are significantly more pronounced

**Describe the solution you'd like**
A way to "see" the -2 trigger gap when crank speed is very not-stable.

**Describe alternatives you've considered**
Not any, really? Aftermarket trigger wheel?

**Additional context**
Tooth log with spark plugs:

[6g75-withsparkplugs-cranking.zip](https://github.com/user-attachments/files/23613768/6g75-withsparkplugs-cranking.zip)
Tooth log without spark plugs:

[6g75-withouutspark-crank.zip](https://github.com/user-attachments/files/23613784/6g75-withouutspark-crank.zip)

TunerStudio logs with plugs:

[withplugslogs.zip](https://github.com/user-attachments/files/23613821/withplugslogs.zip)

Tunerstudio logs without plugs:

[crankingnosparklong.zip](https://github.com/user-attachments/files/23613827/crankingnosparklong.zip)

ElDominio:
**EDIT**

The image i had plugged was for 36-2-1, which is 4g69/4b11. Alldata swears that is 6g75.

Here is real:

<img width="578" height="350" alt="Image" src="https://github.com/user-attachments/assets/95e07b1f-c1a7-4bb0-a231-81227023e509" />

AI upscale for more clarity

<img width="1536" height="1024" alt="Image" src="https://github.com/user-attachments/assets/357de2ad-c3d7-455e-b91d-de5ec22b761b" />
but lost clarity on the -2 :(

ElDominio:
[6g75tune.zip](https://github.com/user-attachments/files/24782645/6g75tune.zip)

rusefillc:
> [6g75tune.zip](https://github.com/user-attachments/files/24782645/6g75tune.zip)

I've failed to load tune without custom ECU .ini

@ElDominio please attach screenshot of "trigger" dialog

ElDominio:
<img width="737" height="550" alt="Image" src="https://github.com/user-attachments/assets/08f82e75-88ce-4e59-b25b-d3e1c2e59e70" />

[miEclipse.zip](https://github.com/user-attachments/files/25499883/miEclipse.zip)

rusefillc:
[...]

PR #9526, opened 2026-05-13
6G75 Trigger fixes

xtitoris:
Updated 36-2-1-1 wheel geometry and sync ratios
Probably need to update tdc offset as well.


xtitoris:
#8827
The issue with V6 is that compression artifacts align perfectly with the gaps in the wheel (6 gaps per 720 degrees and 6 compression cycles)

PR #9548, opened 2026-05-17
PortScanner should not change global LinkManager status on closing ports

FDSoftware:
otherwise we are breaking existing connection in another ports!!

ie: ecu on COM31 auto connected, scanner fails on COM3, both connections were dropped


rusefillc:
@FDSoftware merged but forgot to ask "was there a chance to have automated test coverage here" since important area?

FDSoftware:
I have to mock the portscanner quite a bit, to the point that it's far from reality, but I'll check if I can do it only in the linkmanager.

PR #9549, opened 2026-05-17
Hellen double input voltage divider fix

dron0gus:
Some mega modules route single analog input to few MCU GPIOs to be able to use single ECU as analog or as event input.
UAEFI has IN_PPS2 -> IN_AUX2 routed to PC4 and PE9. PC4 is used in ADC mode, while PE9 is used as EINT.
During startup both pins are configured as inputs with pull-down enabled.
Later PC4 is reconfigured as analog input with no pull-up.
While PE9 is keep untouched with pull-down enabled.
So internal PE9 pull-down affects OpAmp output divider causing divide coefficient higher that 2.
For example when calibrated 2.5V source is connected to PPS2 input ECU measurement shows only 2.36V.
Moreover this internal resistor is not stable and can cause divider coefficient drift over temperature.

WARNING: this can break many tunes!

This fixes only uaefi board.

Screenshot from 2026-05-18 00-08-43


rusefillc:
@dron0gus please s/setuoHellenSharedInputs/setupHellenSharedInputs/

PR #9552, opened 2026-05-18
uaefi brothers: common code

dron0gus:


firmware/config/boards/hellen/hellen_leds_100.cpp
```diff
@@ -1,13 +1,16 @@
 // file hellen_leds_100.cpp
 
-Gpio getCommsLedPin() {
+#include "pch.h"
+#include "hellen_mm100_meta.h"
+
+Gpio __attribute__((weak)) getCommsLedPin() {
```
rusefillc: at some point we've made an explicit effort to kill weak for LEDs

dron0gus: reverted

rusefillc:
https://github.com/rusefi/rusefi/pull/9557

PR #9554, opened 2026-05-18
hide "openblt.bin" into some sub-folder

### LINKED ISSUE #9397: hide "openblt.bin" into some sub-folder
for sake of usability we need to clean-up top level folder of the bundle

let's start by moving "openblt.bin" into "bin/device" subfolder or something like that

* not adding new top level folder into bundle in order to reduce confusion
* rusefi*bin are a bit riskier, we will take care of those later - need backwards compatible logic for autoupdate process

FDSoftware:
resolves #9397


FDSoftware:
blocked by #9805, also related to #9792

PR #9590, opened 2026-05-29
Allow structs to have templates, resolve "targetAfrBlends 0.01 precision"

### LINKED ISSUE #9566: targetAfrBlends 0.01 precision
targetAfrBlends should be 0.01 precision

problem: at the moment all blends share same ``struct blend_table_s`` but AFR is asking for custom higher precision

FDSoftware:
resolves #9566

* now we can add templating to struct on rusefi_config so we dont need to duplicate the entire struct if we only want to tune a single field
* added a migrator for the change of targetAfrBlends from 0.1 to 0.01 precision


rusefillc:
https://github.com/rusefi/rusefi/commit/b68cca4fd98d4591e43243c66b1e58c742ad71d4 hoping to help CI

firmware/controllers/math/engine_math.h
```diff
@@ -69,7 +69,35 @@ struct BlendResult {
...
+
+	float tableValue = interpolate3d(
+		cfg.table,
+		cfg.loadBins, load,
+		cfg.rpmBins, rpm
+	);
+
+	float blendFactor = interpolate2d(value.Value, cfg.blendBins, cfg.blendValues);
+
+	return { value.Value, blendFactor, 0.01f * blendFactor * tableValue, load };
```
rusefillc: what is this 0.01f? is it same as ``blend_table_s<0.01>`` 0.01? if it is, please extract constant into .txt

FDSoftware: added comment & better definition on [51b1cb7](https://github.com/rusefi/rusefi/pull/9590/commits/51b1cb7a390d96d404ceea94c588db19d90f2e1a)

PR #9594, opened 2026-05-29
Adc3 support

ElDominio:
this adds support to have mux channels on adc3, since my s550 board was made with PF analog inputs on STM32F7, and they are muxed. I have tested this on the board and the mux inputs work


rusefillc:
unable to merge

image

maybe because of your choices

image

rusefillc:
why is this here

image

firmware/hw_layer/ports/stm32/stm32_adc_v2.cpp
```diff
@@ -291,7 +388,32 @@ static void slowAdcEndCB(ADCDriver *adcp) {
 		chSysUnlockFromISR();
 	}
 }
-#endif
+
+#ifdef ADC3_SLOW_CHANNEL_COUNT
+static void slowAdc3EndCB(ADCDriver *adcp) {
```
dron0gus: Is there any chance to use same `slowAdcEndCB` for both ADC?
So whole state machine will be in one place easy to read and maintain.

ElDominio: I will be implementing this and testing to see if nothing breaks

rusefillc:
poke @ElDominio

ElDominio:
Have not been able to go to shop, should be tested tomorrow

ElDominio:
Tested in car, working for adc channels on adc3

rusefillc:
@ElDominio can you attempt a (new?) clean PR without libfirmware and merge commits?

ElDominio:
will be happening today!

rusefillc:
let's keep this one open until a better one exists, just in case?

ElDominio:
exploded on proteus_f4 and 8chan f4; master is doing this too; what is the course of action? in my locals i have set 8chan f4 to not comple inis or wideband firmware on board because space

rusefillc:
@ElDominio can you attempt a (new?) clean PR without libfirmware and merge commits?

ElDominio:
I was going to PR, but I got a hunch that something was still up, and talking with my new buddy Antigravity sort of confirmed something. I had not taken time to consider that if we enable ADC3 muxing, the system could possibly also try to mux ADC1. Antigravity said 
"If you enable muxing for ADC3 by defining  ADC_MUX_PIN , ADC1 will still be forced to perform a second set of
  conversions in the muxed state. It will not mess up your physical ADC1 readings, but it will consume extra CPU/DMA
  bandwidth to sample the un-multiplexed ADC1 pins a second time."

so i will be trying to debug this before PR

PR #9596, opened 2026-05-30
removed idle ramp and dashpot, preparation for better idle control

ElDominio:
adding random dashpot air does not make sense, why are we randomly adding air because we're returning to idle? We should be adding an RPM target, to the ramp to idle. This way all our idle modeling would be consistent (no idle aire being added because the engine needs more air because ?, we should add target RPM and then grab the air from the already-calibrated idle duty vs target rpm table)

ramp RPM was removed since this should not just start at whatever upper bound ofidle control we are, we should specifically set an RPM as we arrive at idle (base RPM target + off idle adder), then ramp down to base idle

### DIFF
--- a/firmware/config/engines/canam.cpp
+++ b/firmware/config/engines/canam.cpp
@@ -30,9 +30,6 @@ void setMaverickX3() {
     engineConfiguration->injector.flow = 320;
     engineConfiguration->etbIdleThrottleRange = 10;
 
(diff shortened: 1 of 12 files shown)


ElDominio:
new tested Idle RPM adder depends on this; should I already pull request the RPM adder, or should I wait until a decision is made on this?

PR #9598, opened 2026-05-31
 add ExhaustCutoutController with PWM/H-Bridge, auto modes, more …

ElDominio:
Adds a new ExhaustCutoutController engine module (firmware/controllers/actuators/) that manages an exhaust cutout valve across three hardware output modes:

- Digital: single on/off pin with optional output inversion
- PWM: single pin with configurable open/closed duty cycles and frequency
- H-Bridge: dual-pin motor driver (IN1/IN2) with configurable PWM frequency and duty cycle; de-energizes after the move duration to avoid stalling the motor

Activation sources (exhaustCutoutActivationMode):
- OFF: feature disabled at runtime even if enabled in config
- Switch: reads a hardware input pin (exhaustCutoutSwitchPin)
- LuaGauge: compares a selected Lua gauge value against a configurable threshold

Each activation source independently configures behavior when the input is HIGH vs LOW: Always Open, Always Closed, or Auto.

Auto mode opens the cutout when any of these thresholds are exceeded:
- RPM (exhaustCutoutOpenRpm)
- TPS (exhaustCutoutOpenTps) with a configurable anti-blip hold delay
- MAP/boost (exhaustCutoutOpenMapKpa) All three are OR'd; a configurable hold-open closing delay prevents chatter when triggers clear.

A key-on self-test sequence (optional) cycles the cutout open/close twice when 12V is first detected after power-on to verify the actuator moves. An engine-on test (optional) fires once on the first transition from cranking to running.

An optional LED output mirrors cutout state: off when closed, fast-blinks (3 Hz) while opening, steady when open, slow-blinks (1 Hz) while closing.

Build / platform notes:
- EFI_EXHAUST_CUTOUT defaults to TRUE on STM32F7 (e.g. Proteus F7) and FALSE on STM32F4. The feature requires flash space that does not fit on F4 alongside the other enabled features; this mirrors the same approach used for EFI_DYNO_VIEW and EFI_HPFP.
- proteus_f4: additionally set EFI_EMBED_INI_MSD=FALSE in meta-info-proteus_f4.env, which was the remaining blocker preventing the F4 board build from completing.
- EFI_EXHAUST_CUTOUT is enabled on the simulator and unit test builds.

Configuration changes (rusefi_config.txt):
- 7 bits claimed from unusedBit_Fancy3-9 for cutout enable/invert/test flags.
- Full parameter block appended after rotationalIdleController: output pins and modes, activation and behavior enums, RPM/TPS/MAP thresholds, anti-blip and closing delays, move duration, PWM open/closed duties and frequency, H-Bridge duty cycle and frequency.
- LiveData: isCutoutOpen, isCutoutMoving, targetOpen, isInputHigh, isTriggerRpm, isTriggerTps, isTriggerMap, isTestActive exposed for real-time logging.
- FLASH_DATA_VERSION bumped to 260501.

Unit tests (unit_tests/tests/actuators/test_exhaust_cutout.cpp, 12 cases):
- Disabled / ActivationModeOff: no activation when feature is off
- RpmTrigger: opens above threshold, closes below
- TpsAntiBlip: delay must fully elapse before and after trigger edges
- MapTrigger: MAP threshold behavior
- ClosingDelay: cutout holds open after trigger clears until delay elapses
- MoveDuration: OPENING to OPEN transition gated by motor move timer
- BehaviorAlwaysOpen / BehaviorAlwaysClosed: override modes ignore sensor triggers
- LuaGaugeInput: gauge above/below/invalid threshold routing
- MultiTriggerOr: any trigger opens, all must clear before closing


ElDominio:
out of memory on 8chan f4, disabled included ini; different F4 with different skip rates seem to explode depending on what they are conditionally compiling; somewat seperate topic, should we disable ini embedding on all F4's since they are pretty much legacy, and they would usually rather have features instead of embedded ini?

PR #9607, opened 2026-06-03
chore: prefer generated files from META_OUTPUT_ROOT_FOLDER when set

FDSoftware:


rusefillc:
@FDSoftware do we have "if else" conditional on the generation side of things? Do we really need things to be conditional on the .mk end or can we get away with default empty value of some magic variable or something similar?

firmware/common.mk
```diff
@@ -62,6 +62,11 @@ ifneq ("$(wildcard $(BOARD_DIR)/board_unit_tests.mk)","")
 endif
 LIVE_DATA_GENERATED_DIRS += $(PROJECT_DIR)/live_data_generated
 
+# When META_OUTPUT_ROOT_FOLDER is set prefer generated headers from that location
+ifneq ($(META_OUTPUT_ROOT_FOLDER),)
+	LIVE_DATA_GENERATED_DIRS := $(PROJECT_DIR)/$(META_OUTPUT_ROOT_FOLDER)live_data_generated $(LIVE_DATA_GENERATED_DIRS)
```
rusefillc: @FDSoftware do we need conditional here?

FDSoftware: heh, was updating that over [70a334e](https://github.com/rusefi/rusefi/pull/9607/commits/70a334eec1fb044df51174a724fd4b308476c660)

PR #9608, opened 2026-06-04
added startDisabled() to disable starting when not allowed

ElDominio:
new Lua function startDisabled, since on Mustang I specifically do not want to engage starter until BCM says "ok to start", which only happens after it disables all accesories. If it doesn't wait for that, it could engage the starter while BCM has started shutting down the entire car.

### DIFF
--- a/firmware/controllers/lua/lua_hooks.cpp
+++ b/firmware/controllers/lua/lua_hooks.cpp
@@ -1107,6 +1107,10 @@ extern int luaCommandCounters[LUA_BUTTON_COUNT];
 		engine->module<AcController>().unmock().isDisabledByLua = value;
 		return 0;
 	});
+	lua_register(lState, "startDisabled", [](lua_State* l) {
+		engine->startStopState.startDisabledByLua = lua_toboolean(l, 1);
+		return 0;
+	});
 	lua_register(lState, "getTimeSinceAcToggleMs", [](lua_State* l) {
 		float result = engine->module<AcController>().unmock().timeSinceStateChange.getElapsedSeconds() * 1000;
 		lua_pushnumber(l, result);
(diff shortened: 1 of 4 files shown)


rusefillc:
@ElDominio side effects could occur. Talk to your LLM about https://github.com/rusefi/rusefi/commit/df7ad578d694f0134904e820b6ee3b70154e58a4

ElDominio:
talked to llm and it told me that it we (me and llm)  had not added it to resetLua, causing it to persist even if changed/removed from lua. i have added the changes

firmware/controllers/lua/lua_hooks.cpp
```diff
@@ -1107,6 +1107,10 @@ extern int luaCommandCounters[LUA_BUTTON_COUNT];
 		engine->module<AcController>().unmock().isDisabledByLua = value;
 		return 0;
 	});
+	lua_register(lState, "startDisabled", [](lua_State* l) {
```
rusefillc, line 1110: this is being added between - let's match naming convention with setStartDisabled, similar to setAcDisabled

```
	lua_register(lState, "getGlobalConfigurationVersion", [](lua_State* l) {
		lua_pushinteger(l, engine->getGlobalConfigurationVersion());
		return 1;
	});

	lua_register(lState, "setAcDisabled", [](lua_State* l) {
		auto value = lua_toboolean(l, 1);
		engine->module<AcController>().unmock().isDisabledByLua = value;
		return 0;
	});
	lua_register(lState, "getTimeSinceAcToggleMs", [](lua_State* l) {
		float result = engine->module<AcController>().unmock().timeSinceStateChange.getElapsedSeconds() * 1000;
		lua_pushnumber(l, result);
		return 1;
	});

```

PR #9609, opened 2026-06-04
Add Clutch Down cranking condition option; fix SwitchedState::operato…

ElDominio:
…r bool()

- Adds CC_CLUTCH_DOWN = 3 to cranking_condition_e so start/stop can require the clutch-down switch instead of clutch-up
- Renames the existing "Clutch" label to "Clutch Up" (value 2 unchanged, so existing tunes are unaffected)
- Expands the config bits field from [0:1] to [0:2] to fit four values
- Fixes SwitchedState::operator bool() which returned state != nullptr (always true) instead of dereferencing the pointer; this also corrects the previously broken CC_BRAKE and CC_CLUTCH checks in start_stop.cpp ( tested both clutch/brake to work to start the car)

cars like Mustang have two clutch switches, up and down. it is important to distinct between them to make a decision.

### DIFF
--- a/firmware/controllers/algo/rusefi_enums.h
+++ b/firmware/controllers/algo/rusefi_enums.h
@@ -528,6 +528,7 @@ typedef enum __attribute__ ((__packed__)) {
 	CCNONE = 0,
 	CC_BRAKE = 1,
 	CC_CLUTCH = 2,
+	CC_CLUTCH_DOWN = 3,
 } cranking_condition_e;
 
 /**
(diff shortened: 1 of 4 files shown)


rusefillc:
@ElDominio this should be two separate PRs

to claim a bugfix please follow https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing

rusefillc:
@ElDominio ?

ElDominio:
I'm polishing a few things about misfire detection, i will create a new PR for this with exclusively the clutch up/down options in starter control after I finish work on that branch

PR #9618, opened 2026-06-06
Add ADC3 slow-channel support to stm32_adc_v2

ElDominio:
fixed possible problems of ADCs being declared as the same type while not being true. also only use one function


ElDominio:
I kind of don't want to implement this right now because i found another issue. i will close and keep working on my side.

On my s550 board, PC0 is onboard MAP, PC1 is vbat, pa0/1 are knock2/1. i have overlapping fast adc on map/knock which was causing fast adc to crash; now i need to fix this

PR #9621, opened 2026-06-08
Add EFI_SLOW_ADC and ADC_FAST_DEVICE configurable macros

ElDominio:
Replace hardcoded ADCD1/ADCD2 references in stm32_adc_v2 and stm32_adc_v4 with EFI_SLOW_ADC/ADC_FAST_DEVICE macros (defaulting to the same values), allowing boards to redirect slow/fast ADC to a different peripheral. 

The ADC upgrades have been split into three branches, this one allows ADC peripherals to be assigned to different pins.

BTW, Claude compained about there being a bug in readBatch, and then spoke a lot of things that are considerably over my pay grade:
The bug is that readBatch in stm32_adc_v2.cpp uses three variable names that don't exist (auxConvGroup,
  auxSensorSamples, adcp). If anyone tried to compile with a specific flag set differently, it would just refuse to
  build.

  The TDB rule says: before fixing a bug, first write a test that proves the bug exists — so that when you apply the
  fix, the test changing from "fail" to "pass" is proof the fix actually did something real.

  The problem here is that this bug is a typo in variable names — it's not a logic bug you can observe at runtime, it's
  a "won't compile" bug. You can't really write a Google Test that says "hey, this variable name is wrong."

@dron0gus @FDSoftware ?

### DIFF
--- a/firmware/hw_layer/ports/stm32/stm32_adc_v2.cpp
+++ b/firmware/hw_layer/ports/stm32/stm32_adc_v2.cpp
@@ -8,6 +8,14 @@
 
 #include "pch.h"
 
+#ifndef EFI_SLOW_ADC
+#define EFI_SLOW_ADC ADCD1
+#endif
+
+#ifndef ADC_FAST_DEVICE
+#define ADC_FAST_DEVICE ADCD2
+#endif
+
 #ifdef EFI_SOFTWARE_KNOCK
 #include "knock_config.h"
 #endif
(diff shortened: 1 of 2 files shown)


firmware/hw_layer/ports/stm32/stm32_adc_v2.cpp
```diff
@@ -295,15 +303,15 @@ static void slowAdcEndCB(ADCDriver *adcp) {
...
+	msg_t result = adcConvert(&EFI_SLOW_ADC, &convGroupSlow, b, SLOW_ADC_OVERSAMPLE);
 
 	// If something went wrong - try again later
 	if (result != MSG_OK) {
 		return false;
 	}
 
 	// Temperature sensor is only physically wired to ADC1
-	adcConvert(&ADCD1, &auxConvGroup, (adcsample_t *)auxSensorSamples, auxSensorOversample);
+	adcConvert(&EFI_SLOW_ADC, &auxConvGroup, (adcsample_t *)auxSensorSamples, auxSensorOversample);
```
dron0gus, line 314: Comment above is not true any more. Either remove either add check that ADCD1 is selected as EFI_SLOW_ADC?

rusefillc:
@ElDominio BTW one PR one issue

Quick reminder that humans are VERY expensive. Please be aware of that when passing less-than-on-topic claude opinions unless you are REALLY affected/can explain the REAL damage.

firmware/hw_layer/ports/stm32/stm32_adc_v2.cpp
rusefillc: https://github.com/rusefi/rusefi/commit/24691600fbe39da11fbd40beeadabab1bd8c7caf

Issue #9617, opened 2026-06-06
more nucleo Ethernet


rusefillc:
what about F4? H7?

rusefillc:
@dron0gus error: #error "CH_CFG_USE_DYNAMIC requires CH_CFG_USE_HEAP and/or CH_CFG_USE_MEMPOOLS"

PR #9622, opened 2026-06-08
More ethernet

dron0gus:
https://github.com/rusefi/rusefi/issues/9617


dron0gus:
Nucle F429 is out of flash because configured as 1Mb.

PR #9624, opened 2026-06-08
Replace ADC_MUX_PIN guards with ADC1_SLOW_MUXED boolean macro

ElDominio:
Introduce ADC1_SLOW_MUXED (1 if ADC_MUX_PIN defined, else 0) to replace scattered #ifdef ADC_MUX_PIN checks throughout the slow ADC path in both adc_onchip_slow.cpp and stm32_adc_v2.cpp.


rusefillc:
@ElDominio xyproblem?

PR #9643, opened 2026-06-15
flexCranking: interpolate E0->E100 cranking multiplier across full ethanol range

Matheusschmitz:
## Summary

The second cranking coolant-temperature multiplier table is named `crankingFuelCoefE100` and is described as **E100** in `rusefi_config.txt`, but the firmware blended it in at an **85% (E85)** endpoint. Because `interpolateClamped()` clamps at its endpoints, any flex-fuel reading between 85% and 100% ethanol was pinned to the E100 table value, and the E85→E100 range was never actually interpolated. The TunerStudio label ("Flex Fuel E85") also disagreed with the field name.

This made the behavior inconsistent across three layers:

| Layer | Said it was |
|-------|-------------|
| Config field name | `crankingFuelCoefE100` → **E100** |
| `flexCranking` bit description | **E100** |
| Firmware interpolation endpoint | **85 (E85)** |
| TunerStudio curve label | **E85** |

## Change

Move the blend endpoint to **100%** so the table matches its name:

- E0 uses `crankingFuelCoef`
- E100 uses `crankingFuelCoefE100`
- Intermediate ethanol content (including E85) is now genuinely interpolated between the two

Also renames the local `e85Mult` → `e100Mult` and updates the TunerStudio curve label from "Flex Fuel E85" to "Flex Fuel E100". The persisted config field name is **unchanged**, so tune compatibility/migration is preserved.

## Tests

Adds `FuelMath.flexCrankingE100Interpolation`, covering the no-sensor, E0, E50, E85, E100, and >E100 clamp cases.

## Compatibility note

This changes the meaning of existing tunes that calibrated the second table assuming an 85% endpoint: at E85 they will now receive a slightly lower (interpolated) value rather than the full second-table value. Since the field has always been named `E100`, this aligns behavior with the documented/named semantics.

### DIFF
--- a/firmware/controllers/algo/fuel_math.cpp
+++ b/firmware/controllers/algo/fuel_math.cpp
@@ -69,10 +69,10 @@ float getCrankingFuel3(float baseFuel, uint32_t revolutionCounterSinceStart) {
 	}
 
 	if (engineConfiguration->flexCranking && Sensor::hasSensor(SensorType::FuelEthanolPercent)) {
-		auto e85Mult = interpolate2d(clt, config->crankingFuelBins, config->crankingFuelCoefE100);
+		auto e100Mult = interpolate2d(clt, config->crankingFuelBins, config->crankingFuelCoefE100);
 
-		if (e85Mult <= 0.1f) {
-			warning(ObdCode::CUSTOM_ERR_ZERO_E85_MULT, "zero e85 multiplier");
+		if (e100Mult <= 0.1f) {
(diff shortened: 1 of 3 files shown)


Matheusschmitz:
Closing in favor of #9658, which supersedes this PR. Instead of the 1D E0->E100 blend, #9658 uses full 2D coolant-vs-ethanol tables for both the cranking multiplier and the priming pulse, driven by the single flexCranking toggle. crankingFuelCoefE100 is kept for tune compatibility (hidden in TunerStudio). See #9658 for details and the ethanol cold-start rationale.

PR #9645, opened 2026-06-15
fuel: flex fuel transient compensation (accel enrichment + wall wetting)

Matheusschmitz:
## Summary

Adds optional **flex-fuel compensation for transient fueling** (acceleration enrichment and wall wetting), driven by **CLT × Ethanol%** tables. Gasoline and ethanol behave very differently during transient throttle — especially when cold — so this allows extra enrichment when ethanol content is high and the engine is cold, while staying neutral when warm or on gasoline.

Today the base fuel is compensated for ethanol (`stoichRatioPrimary/Secondary`) and there is `flexCranking`, but transient fueling has no direct `FuelEthanolPercent` compensation.

## What it adds

Three **8×8 CLT (X) × Ethanol% (Y)** multiplier tables with shared axes (`rusefi_config.txt`):

| Table | Applies to |
|-------|------------|
| `flexAeMult` | acceleration enrichment |
| `flexWwTauMult` | wall wetting tau |
| `flexWwBetaMult` | wall wetting beta |

A single helper `getFlexTransientMult()` (`flex_transient.h`) centralizes the gating + lookup:
- returns **1.0 (neutral)** when the new `flexFuelTransientComp` enable bit is off, or when no flex sensor is present;
- treats an **un-initialized (all-zero) table as neutral**;
- a **failed flex sensor defaults to 50% ethanol**, matching `flexCranking`.

### Hooks (3)
- **`accel_enrichment.cpp` `getTpsEnrichment()`** — scales the AE adder. Because the value is unit-agnostic, this covers **both ms-adder and percent-adder** modes with one multiply. `AE_MODE_PREDICTIVE_MAP` returns earlier (AE = 0), so it is unaffected there and relies on the wall-wetting comp instead.
- **`wall_fuel.cpp` `computeTau()` / `computeBeta()`** — scales tau/beta in **both** the basic (constant) and advanced (`complexWallModel`) modes.

### Observability
Output channels `flexAeMultiplier` / `flexWwTauMultiplier` / `flexWwBetaMultiplier` expose the applied multiplier for real-time logging (1.0 = neutral).

### TunerStudio
3 tables + a dedicated dialog/menu under TPS Acceleration Enrichment / Wall Wetting AE, plus the enable checkbox in the Accel/Decel Enrichment dialog.

## Compatibility

- Enable bit **defaults off** and all tables **default to 1.0**, so existing tunes are unchanged.
- `computeBeta()` now clamps to 0..1 in **both** modes (previously advanced-only). `wwaeBeta` is already a 0..1 fraction, so this is a no-op for valid tunes.

## Tests

`unit_tests` cover all three hooks (`test_fuel_wall_wetting.cpp`, `test_accel_enrichment.cpp`): no sensor → neutral, E100 → scaled, **disabled → neutral**. The wall test exposes the protected `computeTau`/`computeBeta` via a small `using`-subclass.

## Validation

Config / `.ini` / output-channel generation was validated locally (config_definition + LiveDataProcessor run clean; the generated struct/`.ini`/output-channels contain the new fields with correct types/scale). C++ compile + unit-test execution rely on CI.

### DIFF
--- a/firmware/console/binary/output_channels.txt
+++ b/firmware/console/binary/output_channels.txt
@@ -96,6 +96,11 @@ int16_t rpmAcceleration;dRPM;"RPM acceleration/Rate of Change/ROC",1, 0, 0, 5, 2
 	uint16_t autoscale wallFuelAmount;@@GAUGE_NAME_FUEL_WALL_AMOUNT@@;"mg",{1/@@PACK_MULT_FUEL_MASS@@}, 0, 0, 0, 3
 	int16_t autoscale wallFuelCorrectionValue;@@GAUGE_NAME_FUEL_WALL_CORRECTION@@;"mg",{1/@@PACK_MULT_FUEL_MASS@@}, 0, 0, 0, 3
 
+! Flex fuel transient compensation multipliers (1.0 = neutral)
+	uint16_t autoscale flexAeMultiplier;Flex: AE multiplier;"mult", 0.001, 0, 0, 5, 3
+	uint16_t autoscale flexWwTauMultiplier;Flex: WW tau multiplier;"mult", 0.001, 0, 0, 5, 3
+	uint16_t autoscale flexWwBetaMultiplier;Flex: WW beta multiplier;"mult", 0.001, 0, 0, 5, 3
+
 	uint16_t revolutionCounterSinceStart;;"",1, 0, 0, 0, 0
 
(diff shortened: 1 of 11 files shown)


rusefillc:
Very nice!

See https://github.com/rusefi/rusefi/commit/08fdc67df94c2cd9ef76f877d0bfbfb59c6085d4

rusefillc:
@Matheusschmitz please rebase to address merge conflict

PR #9646, opened 2026-06-15
sd: conditional SD logging triggers

Matheusschmitz:
## Summary

Adds **conditional SD card logging triggers** (phase 1) so a card captures the interesting events — dyno pulls, launches, intermittent issues — instead of hours of idle/cruise. Today rusEFI logs to SD continuously whenever in ECU logging mode; this lets the log start/stop on configurable conditions.

Default **off**, so existing tunes keep logging unconditionally exactly as before.

## Behavior

A new `SdLogTrigger` module (`controllers/modules/sd_log_trigger`), evaluated in `onSlowCallback`, decides whether to log:

- **RPM start/stop with hysteresis** — start at/above `sdLogStartRpm`; once started, keep logging until RPM stays **below** `sdLogStopRpm` for `sdLogStopDelay` seconds. The dead zone between stop and start holds the current state.
- **Optional secondary conditions** (`sdLogMinTps` / `sdLogMinMap` / `sdLogMinVss`) that gate the **start only** — each ignored when set to `0`. Once logging, only RPM stops it (so a pull that briefly lifts throttle keeps recording).
- **Optional toggle button** (`sdLogTriggerPin`, press on / press off) via `ButtonDebounce`, which takes precedence over the sensor conditions when configured. Its pin is registered in `onConfigurationChange`, so an already-used pin reports the standard pin conflict (`criticalError`), same as the start/stop button.

`mmc_card.cpp` gates the `SD_MODE_ECU` logger on the module: while not logging it closes the current file (so **each session is its own file**) and idles.

## Observability

A new `sdLoggingState` output channel reports **why** logging is or isn't active — `disabled / suppressed / failed / unconditional / active / stop-delay / wait-rpm / wait-cond / button-off` — alongside the existing `sd_logging_internal` bit, and is shown as a readout in the new "SD Conditional Logging" dialog.

## Tests

`test_sd_log_trigger.cpp` covers the state machine: unconditional, RPM hysteresis + stop delay, and entry-gating (secondary condition blocks start but does not stop an active log).

## Validation

- Unit Tests CI: **green** (build + all three `sdLog.*` tests pass).
- Firmware compile/link confirmed on a real STM32 board: module symbols + vtable present in the ELF, `rusefi.bin` produced.

### DIFF
--- a/firmware/console/binary/output_channels.txt
+++ b/firmware/console/binary/output_channels.txt
@@ -453,6 +453,8 @@ float mapFast
 
   uint8_t sd_error;
 
+	uint8_t sdLoggingState;SD: Logging state;"code", 1, 0, 0, 0, 0
+
 	uint8_t fastAdcOverrunCount;ECU: Fast ADC overruns;"", 1, 0, 0, 255, 0
 	uint8_t slowAdcOverrunCount;ECU: Slow ADC overruns;"", 1, 0, 0, 255, 0
 	uint8_t fastAdcLastError;ECU: Fast ADC error type;"", 1, 0, 0, 255, 0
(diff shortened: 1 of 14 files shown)


rusefillc:
@Matheusschmitz do we need a bit of https://github.com/rusefi/rusefi/commit/08fdc67df94c2cd9ef76f877d0bfbfb59c6085d4?

rusefillc:
@Matheusschmitz see https://github.com/rusefi/rusefi/issues/9697

PR #9658, opened 2026-06-17
cranking & priming: flex-fuel 2D coolant x ethanol tables (single flexCranking toggle)

Matheusschmitz:
Supersedes #9643.

Consolidates flex-fuel start enrichment behind the single existing `flexCranking` toggle (plus a present flex fuel sensor):

- **Cranking coolant multiplier** — flex ON: 2D `crankingFuelFlexTable` (coolant x ethanol %, 4-row ethanol axis); flex OFF: the 1D `crankingFuelCoef` curve.
- **Priming pulse mass** — same toggle. flex ON: new 2D `primeFlexTable` (coolant x ethanol %, 4 rows); flex OFF: the 1D `primeValues` curve.

## Why this is needed — ethanol vs gasoline at cold start

Ethanol does not just need "a bit more fuel"; its cold-start behaviour differs from gasoline in a way that a single multiplier (or a two-endpoint E0/E100 blend) cannot represent well:

- **Stoichiometric ratio.** Gasoline is ~14.7:1, E100 is ~9.0:1 (E85 ~9.8:1). For the same air, ethanol needs roughly 30–60% more fuel mass at stoich — a large baseline shift that scales with ethanol content.
- **High heat of vaporization + low cold volatility.** Ethanol's latent heat of vaporization is ~2.5x gasoline's, and its vapor pressure at low temperature is very low. Cold, most of the injected ethanol stays **liquid** (wets the ports/cylinder) instead of forming combustible vapor. To get enough fuel *vapor* to light off, you must inject a large liquid excess.
- **The enrichment is non-linear in temperature, and the curve shape itself changes with ethanol content.** As coolant temperature drops, the required cranking/priming enrichment for high-ethanol blends rises far more steeply than for gasoline. So it is not enough to scale one coolant curve up and down — the *shape* vs coolant differs per ethanol level.

A single coolant curve scaled by ethanol (or a linear blend between only an E0 and an E100 curve) forces intermediate blends like E50 onto a straight line between two endpoints, which does not match the real, non-linear temperature behaviour. A **2D table over (coolant, ethanol %)** lets each ethanol level have its own coolant curve, so the whole E0..E100 range is calibrated directly.

The **priming pulse** wets cold ports before the first crank, so it is governed by the same physics and gets the same 2D treatment.

## Notable changes
- Removes the 1D E0->E100 blend from the firmware path (this is what #9643 implemented). `crankingFuelCoefE100` is **kept in the struct for tune compatibility** but is no longer used and is hidden from TunerStudio (no curve/menu references).
- Drops the interim `crankingUseFlexFuelTable` bit — `flexCranking` alone selects the table.
- 2D tables are `uint8` autoscale to keep RAM tiny (~36 bytes each).
- `defaultsOrFixOnBurn()` seeds each 2D table from its existing 1D curve for tunes that predate it, so enabling flex stays neutral until calibrated.

## Failure behaviour
A configured-but-failed flex sensor falls back to 50% ethanol (`value_or(50)`), matching prior behaviour; with no flex sensor configured the 1D curves are used.

## Tests
Adds/updates unit tests: cranking 2D table + fallback to the 1D curve, and the priming 2D table.

### DIFF
--- a/docs/AI/fueling_system.md
+++ b/docs/AI/fueling_system.md
@@ -36,7 +36,7 @@ crankingFuel = baseCrankingFuel * coolantTemperatureCoefficient * tpsCoefficient
   1. **`flexCranking` off (or no flex sensor)** — a single coolant curve, `crankingFuelCoef`.
   2. **`flexCranking` on, flex sensor present** — `crankingFuelFlexTable`, a coolant (X, shared `crankingFuelBins`) by ethanol-% (Y, `crankingFuelFlexBins`, 4 rows to keep RAM small) `interpolate3d` lookup. The dedicated ethanol axis lets the whole E0..E100 range be calibrated directly.
 
-The legacy 1D `crankingFuelCoefE100` curve (and its E0→E100 blend) is gone from the firmware path; the field is retained for tune compatibility but is no longer used or shown in TunerStudio.
+The legacy 1D `crankingFuelCoefE100` curve (and its E0→E100 blend) is gone from the firmware path. The field is **retired** (renamed with the `unused` prefix so its byte slot stays reserved and the migration framework ignores it); old tunes that calibrated it are carried over to `crankingFuelFlexTable` by `CrankingFlexTableMigrator` (see below).
 
 ## Priming Pulse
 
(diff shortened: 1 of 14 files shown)


rusefillc:
@Matheusschmitz what about legacy configurations?

I am not sure what's the value/long term plan of "retained for compatibility, unused and hidden". Actually see https://github.com/rusefi/rusefi/commit/a02eb0cd8bcac173b5c0b4e61c89f3239d827ba9 for how we handle these things at the moment

PR #9683, opened 2026-06-22
ignition: add dwell duty cycle mode for Ford TFI modules (and others)

ElDominio:
Adds a switch that replaces the RPM/voltage dwell tables with a fixed percentage of the inter-spark interval. Ford TFI (Thick Film Ignition) modules used on Windsor engines expect a 50% duty cycle square wave rather than an absolute dwell time.

When enabled, dwell = (dutyPercent/100) * engineCycleDuration / numSparks, which is the exact inverse of the existing getCoilDutyCycle() function. Cranking still uses ignitionDwellForCrankingMs regardless of this setting.

Tested on a 95 Ford Mustang 5.0 with TFI, auto-calculates correct dwell duty to the setting.
image

### DIFF
--- a/firmware/controllers/algo/defaults/default_ignition.cpp
+++ b/firmware/controllers/algo/defaults/default_ignition.cpp
@@ -114,6 +114,10 @@ void setDefaultIgnition() {
 	setLinearCurve(config->dwellVoltageCorrVoltBins, 8, 15, 0.1);
 	setLinearCurve(config->dwellVoltageCorrValues, 1, 1, 1);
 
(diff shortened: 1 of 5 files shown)


rusefillc:
this change would require new automated coverage

ideally it would be TDD, meaning TWO pull requests: first you would have a test skeleton with first PR asserting current calculation, and then you would be CHANGING a test to show the EFFECT of new mode

rusefillc:
@ElDominio now that base line test was merged, please rebase with test adjustments

ElDominio:
should be updated now

rusefillc:
replaced with https://github.com/rusefi/rusefi/pull/10015

Issue #9714, opened 2026-06-27
One console for all boards from main repo


rusefillc:
if bundle has ``board_compatibility=*`` https://github.com/rusefi/rusefi/commit/f9a932af90623e9f86e275ea3e1f22da483b7def it should be able to handle uaefi, uaefi-pro, 121 and super

rusefillc:
@FDSoftware https://github.com/rusefi/rusefi/issues/9750

rusefillc:
https://github.com/rusefi/rusefi/issues/9753

PR #9748, opened 2026-06-30
One console for all boards from main repo

FDSoftware:
- Implemented `ConnectedEcuTarget` to manage the latest connected ECU target and improve board-specific functionality detection (#9714).
- Integrated firmware prefetching for last connected boards for "Update Firmware".
- Updated logic in DFU and ST-Link flashers to trust connected ECU target over bundled properties.
-  Autoupdate to download firmware on demand for mismatched universal bundles. 
- Added unit tests for board compatibility validation.

related #9714

### DIFF
--- a/java_console/shared_io/src/main/resources/shared_io.properties
+++ b/java_console/shared_io/src/main/resources/shared_io.properties
@@ -1,4 +1,6 @@
-board_compatibility=*
+# #9714: boards a bundle may connect/tune/reflash beyond its own target (case-insensitive).
+# Use '*' for any board.
+board_compatibility=uaefi,uaefi_pro,mre_f4,mre_f7,mre-legacy_f4,nano-tcu-gateway
 auto_update_root_url=https://rusefi.com/build_server
 pinout_base_url=https://rusefi.com/docs/
 pinout_meta_name=boards_meta.yaml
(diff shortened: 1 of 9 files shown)


rusefillc:
@FDSoftware why new GHA?

can we start by allowing uaefi bundle survive uaefi-pro and MRE, and MRE bundle handle nano?

i.e. this PR is probably two PRs mixed together, probably out of order

rusefillc:
@FDSoftware should we introduce 'type of most recent connected board' setting? so that autoupdate would check for update of most recently connected board on start-up?

java_console/shared_io/src/main/resources/shared_io.properties
```diff
@@ -1,4 +1,6 @@
-board_compatibility=*
+# #9714: boards a bundle may connect/tune/reflash beyond its own target (case-insensitive).
+# Use '*' for any board.
+board_compatibility=uaefi,uaefi_pro,mre_f4,mre_f7,mre-legacy_f4,nano-tcu-gateway
```
FDSoftware, line 3: > can we start by allowing uaefi bundle survive uaefi-pro and MRE, and MRE bundle handle nano?


@rusefillc done over here, but ticket #9714 ask for `it should be able to handle uaefi, uaefi-pro, 121 and super`

should i add 121 and super here?

FDSoftware:
> i.e. this PR is probably two PRs mixed together, probably out of order

splited here and in #9751 for when we need to add a bundle without any deps to a board (all fetched from the build server)

java_console/shared_io/src/main/resources/shared_io.properties
rusefillc: huh? why do we need to hard-code a list?

FDSoftware: otherwise * will poke all bundles, so all bundles can connect to any board, this shared_io.properties is on all bundles from this repo

Issue #9754, opened 2026-07-01
MassUpdater


rusefillc:
``java -cp rusefi_console.jar MassUpdater EcuWithOpenblt``

recent change https://github.com/rusefi/rusefi/commit/0010214ba17552515797a8979cfb9c60f4d52472

expected: you can update multiple ECUs but hooking in and out. MassUpdater detects new connection and invokes the job automatically

actual result: does not work in EcuWithOpenblt mode, works only in default OpenBlt target

PR #9758, opened 2026-07-02
MassUpdater: improve auto-detection, fix EcuWithOpenblt flashing, fix…

FDSoftware:
… flash-loop after we leave one board connected post-flash

related #9754

image


FDSoftware:
also this now works under linux!

Issue #9763, opened 2026-07-02
reboot_openblt F7 board is hanging


rusefillc:
hd81 with latest bootloader

reboot_openblt

expected: reboot to BLT
actual: no LED blinking

rusefillc:
@dron0gus https://community.st.com/t5/stm32-mcus-products/bus-stall-on-stm32f767-during-mass-flash-erase-in-dual-bank-mode/td-p/583538

rusefillc:
@dron0gus here is some LLM slop for you

Investigation done. Here's what I found.

## TL;DR

I can't point at one provably-guilty line, but I found a real, F7-specific mechanism that matches "sporadic, same unit, multiple units" — and two genuine firmware bugs around it:

1. **`jump_to_openblt()` fires `NVIC_SystemReset()` with zero interlock against an in-flight background flash write.** On F7 dual-bank 2MB parts (F767 boards) this race is actually reachable, unlike on F4/single-bank where the CPU stalls through the whole flash operation.
2. **On those same F7 dual-bank parts, the IWDG watchdog is *not* extended during flash writes** — and ST has officially confirmed that the Cortex-M7 speculatively prefetches into the bank being erased, stalling the entire bus for up to the full sector-erase time (~0.25–1 s), which is longer than the 500 ms watchdog.

## The mechanism, step by step

- Settings/LTFT burns are **asynchronous**: `writeToFlashNow()` just posts to a mailbox (`firmware/controllers/flash_main.cpp:39`), and the low-priority storage manager thread does the erase+program later (`firmware/controllers/storage.cpp:293`).
- On `DualBank2MB` F7, `mcuCanFlashWhileRunning()` returns true (`firmware/hw_layer/ports/stm32/stm32f7/mpu_util.cpp:53`), so burns happen **any time**, engine running or not, while the firmware keeps servicing USB/CAN. That's the only rusEFI configuration where a `reboot_openblt` command can be *processed concurrently with* an active erase — on single-bank parts the flash controller stalls all code fetches until the erase finishes, so the command can only run before or after, never during.
[...]

PR #9777, opened 2026-07-05
reboot_openblt F7 board is hanging #9763

rusefillc:
#9763


rusefillc:
@dron0gus here's a PR just for fun

problem: this happens on a board where LTFT uses SPI flash, so stm32f7 flash write by LTFT is not possible

dron0gus (approved):

PR #9789, opened 2026-07-07
H7 spi more

dron0gus:
@rusefillc two first commits are harmless.
Last one is questionable.

PR #9816, opened 2026-07-10
stm32h7: mcuconf: switch SPI123 source clock to 52MHz source

dron0gus:
Previously 20MHz source was selected. So maximum SPI clock was 10 MHz. For SD we need something closer to 25 MHz.
After this mode we can get 13MHz clock.

only:uaefi_pro_h7


dron0gus:
Another one byte of https://github.com/rusefi/rusefi/pull/9789

PR #9835, opened 2026-07-13
LLM says that this should make SD mode changes more reliable?

rusefillc:
only:uaefi

### DIFF
--- a/firmware/hw_layer/mmc_card.cpp
+++ b/firmware/hw_layer/mmc_card.cpp
@@ -900,6 +900,11 @@ static int sdModeSwitchToIdle(SD_MODE from)
 		return 0;
 	case SD_MODE_PC:
 		deattachMsdSdCard();
+		// USB-host (MSD) access - especially without a clean eject - can leave the SD/MMC in a state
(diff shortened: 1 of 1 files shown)


firmware/hw_layer/mmc_card.cpp
```diff
@@ -900,6 +900,11 @@ static int sdModeSwitchToIdle(SD_MODE from)
 		return 0;
 	case SD_MODE_PC:
 		deattachMsdSdCard();
+		// USB-host (MSD) access - especially without a clean eject - can leave the SD/MMC in a state
+		// where the next FatFS mount fails with FR_DISK_ERR. Re-connect the block device so whatever
+		// mode we switch into next (typically ECU) mounts from a fresh SD init, like at boot.
+		deinitMmc();
```
dron0gus, line 906: We can have `cardBlockDevice = nullptr` after these calls.
We are not checking for null cardBlockDevice anywhere but only on thread start. And we kill thread if no SD card detected.
This will break whole logic.

Also we are not exposing any API that can change SD card state so hard so it will need re-init. Block read/write should not cause SD card fault.

dron0gus (changes requested):

PR #9840, opened 2026-07-14
sporadic issues with SD RW access #9838

rusefillc:
only:uaefi


dron0gus (commented):
Image

This PR does not implement Reset Recovery logic. So any invalid CBW will lock MSD in STALL state.

BTW I did not observe any timeouts caused by incorrect CBW with current implementation.

PR #9841, opened 2026-07-14
MSD diag #9838

rusefillc:
only:uaefi

Fix 2 is implemented — Bulk-Only Mass Storage Reset now actually resets the transport instead of just ACKing. Proteus F7 builds cleanly, and the only modified files are `mass_storage_device.cpp` and `.h` (I reverted the files the build regenerated).

The reset has three cooperating parts, because the MSD thread can be stuck in different places when Windows gives up and sends the reset:

1. **The EP0 hook does a real reset** (`msd_request_hook_new`). The dead commented-out code is replaced by a call into the controller's new `onBulkOnlyResetIsr()`, which — scoped to the MSD data endpoint only, so the CDC console/TunerStudio endpoints of the composite device are untouched — clears the endpoint's `receiving`/`transmitting` flags (without this, the next re-arm trips ChibiOS's "already receiving" assert) and wakes the MSD thread out of any blocked transfer with `MSG_RESET`, mirroring what `_usb_reset()` does for the whole bus. So a thread stuck waiting for a CBW, for write data, or for a CSW read unblocks immediately.

2. **The abandoned command stops touching USB.** A subtlety I hit while implementing: if the reset lands mid-data-phase, `lib_scsi` ignores transport errors and keeps looping — its leftover data-phase receives would swallow the fresh post-reset CBWs as if they were write data, re-wedging the transport. Since the transport callbacks live in our file, they now check a `m_botResetPending` flag and refuse all USB transfers while a reset is pending, so the SCSI command unwinds quickly without consuming anything from the bus.

3. **The thread skips the stale CSW.** After `scsiExecCmd()` returns, if a reset arrived during execution, `ThreadTask()` does not send the CSW — the host has abandoned that command and a stray CSW would desynchronize the recovered transport. The flag clears when the next CBW arrives, which per the spec only happens after the host considers the reset complete.

One known limitation, inherent to what ChibiOS exposes: if the reset arrives during a double-buffered read with an IN transfer already armed in hardware, that stale data stays armed (killing it needs a low-level endpoint disable + FIFO flush that the HAL doesn't expose per-endpoint). The host's tag check on the following CSW catches this, and the write-hang scenario you're chasing is fully covered.

Together with the stall handling from fix 1 this closes the loop: a transient glitch now costs Windows one retry cycle (milliseconds) instead of a permanent Explorer hang until replug.

### DIFF
--- a/firmware/hw_layer/mass_storage/mass_storage_device.cpp
+++ b/firmware/hw_layer/mass_storage/mass_storage_device.cpp
@@ -34,8 +34,25 @@
 #define MSD_SETUP_INDEX(setup)  MSD_SETUP_WORD(setup, 4)
 #define MSD_SETUP_LENGTH(setup) MSD_SETUP_WORD(setup, 6)
 
+// the one and only mass storage controller, used by the EP0 setup ISR hook and the
+// transport callbacks below which only get a USBDriver/SCSITransport pointer
+static MassStorageController *s_msdInstance = nullptr;
+
+// After a Bulk-Only Mass Storage Reset the host has abandoned the command the SCSI
+// layer is still executing. lib_scsi ignores transport errors and keeps pumping its
+// data phase, so if we kept doing real USB transfers here the leftover data phase
+// would consume the fresh post-reset CBWs as if they were data. Refuse all transfers
+// until the command unwinds; ThreadTask then waits for the next CBW.
+static bool transportAbandoned() {
+	return (s_msdInstance != nullptr) && s_msdInstance->isBotResetPending();
+}
+
 static uint32_t scsi_transport_transmit(const SCSITransport *transport,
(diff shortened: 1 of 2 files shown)


dron0gus (approved):

firmware/hw_layer/mass_storage/mass_storage_device.cpp
```diff
@@ -99,22 +129,69 @@ void MassStorageController::ThreadTask() {
...
+
+				sendCsw(cswStatus, residue);
 			}
 		} else {
 			// ignore incorrect CBW
 		}
 	}
 }
 
+/copyvoid MassStorageController::onBulkOnlyResetIsr(USBDriver *usbp) {
```
dron0gus: Typo?

PR #9861, opened 2026-07-16
MSD: isCommandAbandoned

rusefillc:
The fix is implemented and building cleanly. Here's what changed:

**The mechanism** (both files in `firmware/hw_layer/mass_storage/`):

- **Three timeout-capable USB wait helpers** (`msdUsbTransmitTimeout`, `msdUsbTransmitWaitTimeout`, `msdUsbReceiveTimeout`) — clones of the `hal_usb.c` originals but using `osalThreadSuspendTimeoutS` instead of the untimed suspend. On timeout they clear the endpoint's active flag (same as `onBulkOnlyResetIsr` already does), so the next transfer can re-arm without tripping the "already transmitting" assert.
- **All SCSI data-phase transfers and the CSW send** now go through these with a 10-second timeout. Rationale for 10 s: a live host keeps a data phase moving within milliseconds and usbstor's own give-up is ~20 s, so it can't false-trigger but fires before the host escalates to resetting the device.
- **Expiry is handled exactly like a BOT-reset abandonment**, reusing the existing machinery: a new `m_dataPhaseTimedOut` flag is OR-ed into `transportAbandoned()` (via the renamed `isCommandAbandoned()`), so the rest of the command unwinds with no-op transfers, the CSW is skipped (the host abandoned the command and isn't expecting one), and the thread returns to the CBW wait — re-arming bulk-OUT, which is precisely what breaks the 20-second reset loop that was dropping your CDC console.
- The CBW wait itself stays infinite on purpose — idling forever for the next command is correct.
- **Diagnostics**: `sdinfo` now also prints a `data-phase timeouts` counter, so after this fix your earlier symptom would show as `N data-phase timeouts` instead of `executing opcode 0x28 for 2408898 ms`.

**Validation**: full Proteus F7 firmware build passed, and I confirmed `mass_storage_device.o` was rebuilt from the modified source (this path has no unit-test coverage — it's `EFI_PROD_CODE` + `HAL_USE_USB_MSD` only). The real test is hardware: rerun your SD soak / TS session and watch `sdinfo` — you should see the timeout counter tick instead of a wedge, and no more recurring CDC drops.

Two known residual limitations, both pre-existing behavior I didn't change: an abandoned WRITE(10) still writes stale buffer contents for its remaining blocks (lib_scsi ignores transport errors; fixing that means touching ChibiOS-Contrib), and the OTG TX FIFO isn't flushed on timeout, so the host may read one stale packet at the start of the next data phase — usbstor's normal retry handles that, and the thread is no longer wedged, which is the part that matters. Full details are in `docs/report.md`.


rusefillc:
## 2026-07-15 - usb-disconnect.pcapng analysis: CDC drops are collateral of a wedged MSD interface

What: Analyzed usb-disconnect.pcapng (repo root, USBPcap linktype 249, 5108 packets,
32 s, Windows host) to answer whether it explains the USB CDC disconnects. It does.

Capture contents: dev 6 = rusEFI composite device (EP1/0x81 = MSD bulk, EP2/0x82 =
CDC data with visible TS traffic incl. "rusEfiVersion", EP3/0x83 = CDC interrupt).
dev 1 = unrelated HID device. No re-enumeration during capture (all GET_DESCRIPTOR /
SELECT_CONFIGURATION packets are USBPcap's t=0 snapshot of existing devices).

Timeline of the two disconnect events:
- Pre-capture: a CBW was already pending on MSD bulk-OUT EP1, NAKed by the device.
- t=5.155: Windows usbstor gives up -> cancels ALL pending IRPs on the composite
  device (EP1, EP2/0x82, EP3/0x83 all complete with USBD_STATUS_CANCELED 0xC0010000)
  and resets the device. The CDC IRP cancellation is the console disconnect.
- t=5.34: usbstor re-issues the abandoned command: CBW READ(10), 8 KB, LUN 1 (the
  SD-card LUN). No data, no CSW, no STALL ever comes back. CBW tag 0x4cd72920
  matches the Windows IRP id ...4cd72920 - provably the same request.
- t=5.58: usbser.sys recovers CDC (GET/SET_LINE_CODING succeed), TS traffic resumes
  and runs fine for 20 s (2587 CDC packets between events).
- t=25.35: usbstor's 20 s timeout on the READ(10) expires -> second cancel storm +
  device reset. This time control transfers briefly fail with USBD_STATUS_XACT_ERROR
  0xC0000011 (device not responding on the wire mid-reset) before recovering.
- t=25.54: usbstor downgrades to TEST_UNIT_READY LUN 0 - which ALSO never completes
  within the remaining 6.5 s of capture. CDC traffic never resumes after t=25.86;
  the last CDC control transfer drops DTR (SET_CONTROL_LINE_STATE 0x02) = host app
  closed the port.

Conclusion: the CDC disconnects are collateral damage from Windows mass-storage
reset-recovery on the shared composite USB device (the architecture risk already
documented in CLAUDE.md "CDC console and MSD share ONE composite USB device").
Root cause is firmware-side: the MSD thread is wedged for the entire capture -
bulk-OUT EP1 is never re-armed, so CBWs NAK forever. Consistent with the thread
being stuck inside scsiExecCmd() on an earlier LUN 1 (SD) command: the
onBulkOnlyResetIsr()/MSG_RESET wakeup only aborts waits on the USB data endpoint,
not a wait inside the SD/SPI driver or on the SD mutex.

Validation: parsed with a python/scapy USBPcap header decoder in the session
scratchpad (no tshark on this machine); statuses cross-checked against USBD_STATUS
codes; IRP-id-to-CBW-tag match confirmed the 20 s pending READ(10).

Open follow-ups:
- Instrument/confirm where the MSD thread blocks: m_busyOpcode/m_busySince
  diagnostics (sdinfo / MSD printDiagnostics) will show "executing opcode 0x28 for
  N ms" after such a hang.
- Fix direction: bound LUN 1 SCSI ops (timeout in the block-device read path) or
  report medium-not-present instead of blocking when the SD card is unavailable;
  make bulk-only-reset recovery able to abort a command stuck below the transport
  layer.
- usb-disconnect.pcapng is untracked in the repo root; decide whether to keep it
  (e.g. move under docs/ with a note) or drop it once the bug is fixed.

## 2026-07-15 - Addendum: MSD wedge confirmed on hardware, block point identified

What: Live sdinfo output from the affected ECU confirmed the pcap-based diagnosis and
allowed pinpointing the exact block location in code.

Evidence from sdinfo (2026-07-15 20:08):
- "MSD: executing opcode 0x28 for 2408898 ms" -> the MSD thread has been inside a
  single READ(10) for ~40 minutes, i.e. since before usb-disconnect.pcapng was taken.
  Matches the pre-capture pending CBW seen in the pcap.
- "MSD: 1353 commands, 439 failed (last failed opcode 0x25)" and SD block ops all
  successful (1017 reads, 0 failed) -> the SD/SPI side is healthy; the wedge is not
  in blkRead.
- "SD enabled=Yes status=MSD target mode=ECU/logging auto" on USB-only power
  ("not sending without +12v power").

Block point (code walk):
- lib_scsi data_read10() (firmware/ChibiOS-Contrib/os/various/lib_scsi.c) pumps the
  data phase via scsi_transport_transmit(_wait) -> usbTransmit/usbTransmitWait.
- usbTransmitWait (firmware/ChibiOS/os/hal/src/hal_usb.c:664) parks the thread with
  osalThreadSuspendS and NO timeout. When the host cancels its IN URB mid-data-phase
  and stops issuing IN tokens, the transfer never completes and the thread sleeps
  forever.
- The existing recovery path (m_botResetPending / transportAbandoned() /
  onBulkOnlyResetIsr MSG_RESET wake in mass_storage_device.cpp) only triggers on the
  class-specific Bulk-Only Mass Storage Reset request - the pcap shows Windows
  usbstor does NOT send one in this scenario; it cancels URBs and resets the device,
  which never wakes the suspended thread (no bus-level re-enumeration observed).
- Result: bulk-OUT EP1 is never re-armed, every later CBW NAKs, usbstor resets the
  composite device every ~20 s, each reset cancels the CDC IRPs -> the recurring
  CDC console/TunerStudio disconnects.

Docs updated: CLAUDE.md composite-USB section gained a paragraph describing the
wedge mechanism, the sdinfo diagnostic signature, and the wire-level signature.

Open follow-ups (fix direction, not yet implemented):
- Add a timeout to the MSD data-phase USB waits (or a timeout-capable wait wrapper)
  and treat expiry like transportAbandoned(): unwind the command, skip the CSW,
  return to CBW wait.
- Correction to an earlier draft of this entry: a separate "wake MSD waits from the
  USB RESET/SUSPEND event hook" change is NOT needed. ChibiOS already does this -
  _usb_reset() and _usb_suspend() (hal_usb.c) resume all endpoint-waiting threads
  with MSG_RESET, and every usb* call bails with MSG_RESET while the driver is not
  USB_ACTIVE, so the command unwinds on any real bus reset/suspend. The observed
  wedge is precisely the remaining case: the host cancels its URBs purely host-side,
  the bus stays active, no reset/suspend/BOT-reset ever reaches the device - only a
  device-side timeout can cover it.

## 2026-07-15 - Fix: MSD data-phase timeout so an abandoned command cannot wedge the MSD thread

What: Implemented the fix for the MSD wedge diagnosed earlier today (see the two
entries above): give every MSD data-phase USB wait a timeout and treat expiry like
a Bulk-Only-Reset abandonment.

| File | Change |
|--------------------------------------------------------|--------------------------------------------|
| firmware/hw_layer/mass_storage/mass_storage_device.cpp | Added msdUsbTransmitTimeout / msdUsbTransmitWaitTimeout / msdUsbReceiveTimeout (timeout-capable clones of the hal_usb.c waits, using osalThreadSuspendTimeoutS; on MSG_TIMEOUT the endpoint active flag is cleared, same as onBulkOnlyResetIsr, so re-arm does not assert). SCSI transport wrappers and sendCsw now use them with MSD_DATA_PHASE_TIMEOUT = 10 s and report expiry via onDataPhaseTimeout(). ThreadTask clears the flag on the next valid CBW and skips the CSW for a timed-out command. transportAbandoned() now covers BOT reset OR timeout via isCommandAbandoned(). printDiagnostics gained a data-phase-timeout counter |
| firmware/hw_layer/mass_storage/mass_storage_device.h   | isBotResetPending -> isCommandAbandoned (bot reset OR timeout), new onDataPhaseTimeout(), new m_dataPhaseTimedOut / m_timeoutCount members with rationale comment |

Key decisions:
- Timeout value 10 s: a live host keeps a data phase moving within milliseconds and
  usbstor's own give-up is ~20 s, so 10 s cannot false-trigger on a healthy link but
  fires well before the host escalates to a device reset.
- Expiry reuses the existing m_botResetPending machinery shape (separate flag,
  OR-ed in transportAbandoned()): command unwinds with no-op transfers, CSW is
  skipped (host abandoned the command and is not expecting one), thread returns to
  the CBW wait so bulk-OUT re-arms - which is exactly what breaks the 20 s
  reset-loop that was dropping the CDC console.
- sendCsw also uses the timed transmit: a CSW the host never reads is the same
  wedge; nothing to unwind there, just count and continue.
- The CBW wait in ThreadTask stays infinite on purpose - idling for the next
  command forever is correct.
- OUT-completion resume value semantics preserved: usbReceive-style count return
  in scsi_transport_receive (verified _usb_isr_invoke_out_cb resumes with
  usbGetReceiveTransactionSizeX; MSG_TIMEOUT/-1 and MSG_RESET/-2 handled first).

Validation: full Proteus F7 firmware build (compile_proteus_f7.sh) passed;
mass_storage_device.o confirmed rebuilt from the modified source. No unit-test
coverage possible (EFI_PROD_CODE + HAL_USE_USB_MSD only). Hardware validation
pending: reproduce the host-abandon scenario and check 'sdinfo' shows the new
"data-phase timeouts" counter incrementing instead of "executing opcode 0x28 for
<huge> ms".

Known residual limitations (unchanged behavior, pre-existing):
- lib_scsi data_write10 ignores transport receive errors, so an abandoned WRITE(10)
  still blkWrite()s stale buffer contents for the remaining blocks (same as the
  BOT-reset path today; fixing it means touching ChibiOS-Contrib lib_scsi).
- On timeout the OTG TX FIFO is not flushed, so the host may read a stale packet
  at the start of the next data phase; usbstor recovers via its normal retry, and
  the thread is no longer wedged, which is the part that matters for CDC.

dron0gus (approved):

Issue #9860, opened 2026-07-16
USB serial is not stable


rusefillc:
i have pretty happy USB on my Windows 11 HP desktop

more than one of my Windows 11 Dell laptops have periodic drop of rusEFI console

wireshark attached

[usb-disconnect.zip](https://github.com/user-attachments/files/30069092/usb-disconnect.zip)

rusefillc:
@dron0gus 

**Yes — the capture explains the CDC disconnects, and it points the finger at the mass-storage interface, not CDC itself.** The disconnects are Windows' USB mass-storage error recovery resetting the whole composite device; the CDC console drop is collateral damage.

What the capture shows (USBPcap, 32 s, two disconnect events):

1. **The MSD interface is wedged for the entire capture.** A CBW (SCSI command wrapper) was already sitting on bulk-OUT EP1 before the capture started, NAKed by the ECU — the firmware never re-armed the endpoint. There is not one successful mass-storage exchange in the whole file.

2. **t=5.15 — first disconnect.** usbstor.sys times out, cancels every pending IRP on the composite device — the MSD one *and* the CDC data/interrupt reads (all complete with `USBD_STATUS_CANCELED`) — and resets the device. That IRP cancel storm is exactly what the host-side serial stack experiences as the console dropping.

3. **t=5.34 — the smoking gun.** usbstor re-issues the abandoned command: **READ(10), 8 KB, LUN 1 — the SD-card LUN.** The ECU never answers: no data, no CSW, not even a STALL. The CBW tag `0x4cd72920` matches the Windows IRP ID, so it's provably the same request that sits pending for the next 20 seconds. Meanwhile CDC recovers and TunerStudio traffic flows normally (~2600 packets) between the events.

4. **t=25.35 — second disconnect.** usbstor's 20-second timeout on that READ(10) expires → same cancel storm + reset, this time with control transfers briefly failing `USBD_STATUS_XACT_ERROR` (device not responding mid-reset). The follow-up TEST_UNIT_READY also never completes, and after t=25.9 the host drops DTR and closes the port for good.

[...]

PR #9866, opened 2026-07-16
USB serial is not stable #9860

rusefillc:
To validate after reflashing: boot with the console connected —
  1. sdinfo should show no-data stalls climbing (Windows probes LUN0 Read Capacity at every mount/rescan) while data-phase timeouts stays 0,
  2. volume arrival should be fast (no more ~7 s of 200 ms retries at boot),
  3. a capture should show LUN0 Read Capacity(10) completing as STALL → clear-halt → CSW(failed) instead of USBD_STATUS_BABBLE_DETECTED — and no CDC URB cancellations riding along with disk re-probes.

  That last point is the real test of your CDC reliability complaint: if the babble recovery was the driver, the CDC drops that coincide with disk re-probes should be gone.

### DIFF
--- a/firmware/hw_layer/mass_storage/mass_storage_device.cpp
+++ b/firmware/hw_layer/mass_storage/mass_storage_device.cpp
@@ -20,6 +20,7 @@
 #define MSD_CSW_SIGNATURE               0x53425355
 
 #define CBW_FLAGS_RESERVED_MASK         0b01111111
+#define CBW_FLAGS_DIRECTION_IN          0b10000000
 #define CBW_LUN_RESERVED_MASK           0b11110000
 #define CBW_CMD_LEN_RESERVED_MASK       0b11000000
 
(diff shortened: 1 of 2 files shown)


rusefillc:
@dron0gus 

What changed (mass_storage_device.cpp/.h):
- The SCSI transport wrappers now count data-phase bytes actually moved per command (m_dataPhaseBytes, reset before each scsiExecCmd). I deliberately didn't use lib_scsi's scsiResidue() — it's only written on short transfers and never reset between commands, so it can't distinguish "sent everything" from "sent nothing".
- After a command that declared a data-IN phase completes with zero bytes moved, ThreadTask now stalls bulk-IN before sending the CSW — the BOT case 4/5 handling that was missing. The host clear-halts (ChibiOS hal_usb.c handles that automatically) and then reads the CSW into a proper 13-byte buffer instead of the CSW babbling into the armed 8-byte data URB.
- Failed-command CSW residue is now computed as data_len - bytes moved per spec (previously it reported lib_scsi's stale value).
- sdinfo gains a no-data stalls counter.

Safety properties: an abandoned command (BOT reset / data-phase timeout) still skips both the stall and the CSW; if the endpoint is unexpectedly busy the stall is skipped and behavior is identical to today; any surprise in arm-while-stalled ordering is bounded by the existing 10 s CSW timeout — no new wedge paths.

What I left out on purpose: the symmetric data-OUT stall (host writing to a failed command). Windows never hits it, and if it ever appears it's visible in the invalid CBWs counter.

dron0gus (approved):

PR #9868, opened 2026-07-16
Fix pinout page 1/3 starter control & disable pins

ArendJan:
I was playing around with the pinout and noticed that the pinout page "Starter Relay Pin" was set to the Outputs->"Starter Disable" pin.

This adds the Outputs->"Starter Control Pin" to the pinout page and renames the disable pin to a better describing name.


rusefillc:
Thank you @ArendJan

Issue #9111, opened 2026-02-14
cam profile to VE table wizard


rusefillc:
maybe just five presets of stages

FDSoftware:
i will need a bit more of info on this one, i dont understand how we relate the cam profile to the VE table :c

Issue #9110, opened 2026-02-14
console to have setup wizard


rusefillc:
```
bit wizardNumberOfCylinders,"yes","no"
bit wizardFiringOrder,"yes","no"
bit wizardMapSensorType,"yes","no"
bit wizardCrankTrigger,"yes","no"
bit wizardCamTrigger,"yes","no"
bit wizardInjectorFlow,"yes","no"
```

<img width="531" height="367" alt="Image" src="https://github.com/user-attachments/assets/1fc84d07-7ca0-4c71-a678-5485d1fb4b3c" />

<img width="786" height="593" alt="Image" src="https://github.com/user-attachments/assets/9e0b660f-04a9-4585-b4d9-d3afefffb84c" />

PR #9873, opened 2026-07-19
BaseVeGenerator: factory and first generator, "archetype-base-ve-v1"

FDSoftware:
Why `archetype-base-ve-v1` was created over the available alternatives.

### Internet random A
 has a useful UX pattern (questionnaire -> preview -> explicit apply)
but a weak engine model:

- Output is fixed 16x16;
- Inputs actually used: max RPM, load strategy, aspiration, max boost, then: Displacement,
  cylinders, bore/stroke, compression, and the existing axes are ignored.
- Torque/VE peak is hardcoded at 4500 RPM regardless of engine type.
- The load factor increases above 100 kPa. rusEFI's air-mass equation already multiplies
  by actual MAP, so raising VE above atmospheric in proportion to boost double-counts
  pressure and produces incorrect fuel delivery at boost.

### LLM-assisted generation , Internet random B

- Output is non-deterministic across runs and model versions.  regression tests
  are impossible: the same inputs produce different tables after a model update.
- Requires network access and a model endpoint at generation time. Offline tune editing
  (the primary use case) cannot rely on it.
- The model has no way to enforce rusEFI's physical VE semantics. It may embed target
  lambda, enrichment margins, or MAF-style scaling depending on what training data it
  saw, producing tables that look reasonable but mean something different.
- A "reasoning" explanation from the model is not a substitute for a versioned,
  auditable formula with unit-tested anchor values.

### Mega***** Internet random C

asks for peak torque RPM, peak horsepower RPM, displacement, and boost, then
fits a quadratic WOT curve and applies an empirical MAP factor:

- Torque does not uniquely determine airflow. Two engines with identical torque curves
  but different VE shapes (e.g. one highly tuned, one naturally flat) would receive the
  same table.
- The documented formula includes an AFR/gamma correction (`VE x gamma`) but the
  executable does not implement it
- Its MAP multiplier raises VE above atmospheric in proportion to boost, which
  double-counts pressure in our Speed Density air-mass equation for the same
  reason as Random A.
- A single quadratic cannot represent intake/cam resonance peaks, VVT plateau
  broadening, or the mild dip between low-RPM and mid-RPM that many engines show.
- Boost torque curves often reflect the boost controller strategy rather than the
  engine's physical VE, so the inputs conflate two independent quantities.

## Why `archetype-base-ve-v1`

**Deterministic ** Every constant is named, documented, and tested with
exact values. no random LLM in the middle

**Dimension-agnostic.**  generator takes the ECU's actual RPM and MAP axis arrays
and produces a surface of the correct shape

**No MAP double-counting.** well, that, dont broke the VE table on boosted case

**No hidden enrichment.** VE is physical volumetric efficiency. Target lambda lives in
the lambda table. Mixing enrichment into VE values would misrepresent what the number
means and would interfere with STFT/LTFT learning.

**Extensible.** The `VeGenerator` interface lets a second implementation (e.g. a
measured-airflow recovery mode, a v2 with log-fitted constants, or a board-specific
preset) coexist in the same dialog without touching the v1 code.

**Testable without hardware.** The pure model has no Swing, INI, or protocol
dependencies. All 24 head x cam x VVT combinations can be verified in CI

related #9111 #9110

### DIFF
--- a/java_console/models/src/main/java/com/rusefi/tune/ve/ArchetypeBaseVeV1.java
+++ b/java_console/models/src/main/java/com/rusefi/tune/ve/ArchetypeBaseVeV1.java
@@ -1,0 +1,228 @@
+package com.rusefi.tune.ve;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Archetype Base VE v1 — a transparent heuristic for conventional four-stroke SI engines.
+ * <p>
+ * Profile ID: {@value #PROFILE_ID}.
+ * Constants are part of the version; change only under a new profile ID with updated fixtures.
+ */
+public class ArchetypeBaseVeV1 extends BaseVeGenerator {
+
+    public static final String PROFILE_ID    = "archetype-base-ve-v1";
+    public static final String DISPLAY_NAME  = "Archetype Base VE v1";
+
+    // MAP factor lookup: {absoluteKpa, factor}. Must remain in ascending kPa order.
+    static final double[][] MAP_FACTORS = {
+        {10, 0.60}, {20, 0.65}, {40, 0.72}, {60, 0.82}, {80, 0.92}, {100, 1.00}
+    };
+
+    public enum HeadArchetype {
+        RESTRICTIVE_2V("Restrictive/legacy 2-valve", 85.0),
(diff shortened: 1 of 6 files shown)


FDSoftware:
(btw this is only the generator, we still need a couple of prs for wiring this into existing Tune tab and a new wizard step)

Issue #9853, opened 2026-07-15
writeToothFromCyclicBuffer


rusefillc:
we should always collect basic tooth logger data into BACKGROUND_TOOTH_LOGGER_SIZE buffer if BACKGROUND_TOOTH_LOGGER_SIZE is above zero

we need writeToothFromCyclicBuffer() method which should be invoked for example on sudden cases of zero RPM glitches

rusefillc:
@dron0gus what's the status?

rusefillc:
current status: has to be explicitly selected in settings, that would disable default binary logging

PR #9888, opened 2026-07-21
tooth logger testability

rusefillc:
only:testability

## 2026-07-20 - Review of tooth_logger.cpp PR #9871 (Toothlogger fixes) + coverage assessment

What was done: reviewed recent tooth_logger changes - a403d0b8a61 "Toothlogger
fixes (#9871)" (queue reset/resume on disable/enable, removal of
GetToothLoggerBufferBlocking, composite entry timestamps made relative to
buffer startTime, CSV writer recovers absolute time) plus the earlier CSV
column additions (2e34749 instantMAP, a91e2ae TPS). Assessed unit-test
coverage and identified gaps. Investigation only - no code changes.

Findings:
- Consumer regression risk (main concern): entry->timestamp on the TS wire is
  now an offset from buffer->startTime, so every composite buffer restarts
  near 0. The rusEFI console parser
  (java_console/models/.../composite/CompositeParser.java) assumes
  monotonically increasing timestamps and treats any decrease as a uint32
  overflow, adding 2^32 us (~71.6 min) - with per-buffer-relative stamps it
  will add that on nearly every buffer, wrecking the time axis of
  console composite/VCD/Logicdata output. PR #9871 did not touch the Java
  side. The absolute base (startTime) is not transmitted, so the consumer
  cannot reconstruct real time at all. TS-proper behavior unverified (closed
  source, same wire format). Upside firmware-side: the old absolute-us stamp
  wrapped uint32 at ~71.6 min uptime; relative stamps max out around 5 s
  (buffer flush period), so wrap is gone - the fix direction is right, the
  consumer contract is the problem.
- Re-enable while enabled: TS_SET_LOGGER_SWITCH calls EnableToothLogger()
  directly (e.g. composite enable arriving twice, or Full <-> PrimaryTooth
  switch) without DisableToothLogger(). Mailboxes are then not in reset
  state: resumeX() is a no-op, and the new code no longer drains
  filledBuffers (the old drain loop was removed), while freeBuffers was
  never drained. postI() of all bufferCount buffers into a non-empty free
  mailbox silently drops the overflow and can leave the same CompositeBuffer
  pointer queued in both freeBuffers and filledBuffers -> consumer can read
  a buffer the producer is concurrently refilling. Pre-existing weakness,
  slightly aggravated by dropping the drain. Clean path
  (disable-then-enable) is fine because chMBReset empties the mailbox.
- CSV recovery math checked OK: USF2NT(float) is integer-exact for offsets
  < 2^24 us and buffer lifetime is capped at 5 s, so no precision loss.

Coverage status:
- unit_tests/tests/test_tooth_logger.cpp covers only the CSV path
  (WriteCsvHeader + WriteCsvRows, both pass). WriteCsvRows sets
  startTime = 0, so the new "recover absolute time" logic degenerates to
  identity - a regression back to treating c.timestamp as absolute would
  still pass.
- The code PR #9871 actually changed (enable/disable queue lifecycle,
  findBuffer, production SetNextCompositeEntry, ReturnToothLoggerBuffer) is
  inside the "#else // not EFI_UNIT_TEST" block - compiled out of the test
  binary and replaced by the events-vector stub, so it has zero unit-test
  coverage and cannot gain any without refactoring (mailbox shim or
  extracting buffer management). Only the simulator exercises it.

Recommended coverage additions (in priority order):
1. CSV row test with nonzero startTime (e.g. 100 s) verifying sec.usec output
   equals startTime + offset - pins the new recovery math. Add an offset near
   the 5 s buffer max.
2. Java CompositeParserTest case feeding two consecutive buffers whose
   timestamps restart at 0 - documents/exposes the +2^32 jump and anchors
   whatever consumer-side fix is chosen.
3. Full-buffer CSV test (nextIdx = toothLoggerEntriesPerBuffer) for bounds.
4. Longer term: make the buffer/mailbox lifecycle host-testable (extract from
   the !EFI_UNIT_TEST block behind a small queue abstraction), then test:
   entries relative to startTime, post-on-full, 5 s timeout flush,
   disable/enable reset, and enable-while-enabled not duplicating buffers.

Open follow-ups:
- Decide the consumer-side story for relative timestamps (update
  CompositeParser to accumulate per-buffer, or transmit the base time);
  verify what TS itself does with the composite stream.
- Consider making EnableToothLogger() drain/reset both mailboxes
  unconditionally (resetI + resumeX) to make re-enable safe.

## 2026-07-20 - ToothLogger CSV: regression test for absolute-time recovery

Follow-up to the PR #9871 review above: implemented recommendation 1.

What was done: added ToothLogger.WriteCsvRowsRecoverAbsoluteTime to
unit_tests/tests/test_tooth_logger.cpp. The buffer's startTime is set to
100 s after boot (US2NT(100'000'000), US_TO_NT_MULTIPLIER = 100 on host)
and three entries carry microsecond offsets 0 / 1'000'002 / 4'999'999
(first entry, mid-buffer, and just under the 5 s flush period - the largest
offset a buffer can hold). The test asserts each CSV row starts with the
recovered absolute timestamp: 100.000000 / 101.000002 / 104.999999.

Validation:
- Test passes on current master (3/3 in the ToothLogger suite).
- Negative check performed: temporarily reverted the writer to the
  pre-#9871 code (printing c.timestamp as absolute) - the test failed with
  rows 0.000000 / 1.000002 / 4.999999 as expected, then restored the source.
  Confirms the test pins the recover-time math and would catch a regression
  to treating the stored offset as absolute time.

Open follow-ups: unchanged from the review entry above (Java
CompositeParser multi-buffer test, full-buffer CSV bounds test,
host-testable buffer lifecycle).

## 2026-07-20 - ToothLogger CSV: full-buffer bounds test

Follow-up to the PR #9871 review: implemented recommendation 3.

What was done: added ToothLogger.WriteCsvFullBuffer to
unit_tests/tests/test_tooth_logger.cpp. Fills a CompositeBuffer completely
(nextIdx = toothLoggerEntriesPerBuffer = 250) with entries 20 ms apart
(last offset 4.980000 s, under the 5 s flush period) and asserts:
- exactly 250 CRLF-terminated rows are emitted (one per entry, no
  out-of-bounds extras),
- the returned byte total equals the bytes written,
- first row starts at 0.000000 and the final row is 4.980000 followed by
  the terminating CRLF at end of output.

Validation: full ToothLogger suite passes (4/4).

Build note (tooling quirk, self-inflicted): reverting
firmware/controllers/lua/generated/value_lookup_generated.cpp via git
checkout after a build breaks the next unit_tests build - the checked-in
copy is stale vs the checked-in config (references removed boardUse*
fields), and checkout makes the file newer than its generator inputs, so
make considers it fresh and does not regenerate. Recovery: touch
firmware/integration/rusefi_config.txt (or make clean) to force
regeneration. Leave build-regenerated files modified in the working tree;
just do not commit them.

## 2026-07-20 - Tooth logger buffer lifecycle made host-testable (ToothLoggerBufferPool)

Follow-up to the PR #9871 review: implemented recommendation 4 (long term).

What was done: extracted the production buffer/mailbox lifecycle out of the
"#if !EFI_UNIT_TEST" block in tooth_logger.cpp into a class that compiles in
all builds, and added host unit tests for it.

| File | Change |
|------|--------|
| firmware/console/binary/tooth_logger_buffer.h | new: ToothLoggerBufferPool - free/filled queues, current buffer, big-buffer handle, ready-indication callback; bufferCount now a class constant |
| firmware/console/binary/tooth_logger_buffer.cpp | new: startI/stopI/appendI/returnBufferI/getFilled/fetchFilled/flushCurrentI/hasDataI, logic moved verbatim from tooth_logger.cpp |
| firmware/console/binary/tooth_logger.cpp | prod block now delegates to a static ToothLoggerBufferPool{setToothLogReady}; EFI_UNIT_TEST stub (events vector) unchanged, so EngineTestHelper-based tests are unaffected |
| unit_tests/chibios-mock/mock-mailbox.h | new: host chibios_rt::Mailbox lookalike (postI/fetchI/fetch/resetI/resumeX/getUsedCountI, MSG_OK/MSG_TIMEOUT/MSG_RESET semantics per chmboxes.c; blocking fetch degenerates to non-blocking - tests are single-threaded) |
| unit_tests/tests/test_tooth_logger_buffer.cpp | new: 7 lifecycle tests |
| firmware/console/console.mk, unit_tests/tests/tests.mk | build wiring |

Key decisions and why:
- Class + delegation instead of compiling tooth_logger.cpp's free functions on
  host: EngineTestHelper's ctor calls EnableToothLogger() per test; routing
  that to the real pool would make every test grab the BigBuffer and collide
  with test_big_buffer/perf-trace usage. The stub stays; dedicated tests
  instantiate their own pool.
- The mock mailbox lives in unit_tests/chibios-mock/ (already on the include
  path) and claims the chibios_rt::Mailbox name, so the class header needs no
  ifdef around the member declarations and the mock is reusable for future
  extractions (e.g. loggingcentral).
- One intentional behavior fix: stopI() nulls the current buffer (the old
  DisableToothLogger left currentBuffer dangling into the released big
  buffer, so ToothLoggerHasData stayed true and the SD writer could flush a
  buffer now owned by another BigBuffer user).
- The ready flag (outputChannels.toothLogReady) is a constructor-injected
  callback; production passes setToothLogReady, tests capture into a bool.

Tests (all pass):
- StartStopBigBufferOwnership, StartFailsWhenBigBufferBusy (big-buffer
  arbitration, clean empty state)
- FillPostsBufferWithRelativeTimestamps (250th entry posts; startTime =
  first-event time; entries are us offsets from startTime; ready-flag
  set/clear transitions)
- StaleBufferFlushedAfter5Seconds (partial buffer posted on next append
  after 5 s)
- StopWhileConsumerHoldsBufferThenRestartClean (MSG_RESET tolerated,
  restart is clean and functional)
- EntriesDroppedWhenAllBuffersFull (bufferCount buffers drain full, excess
  entries dropped silently, logging resumes after buffers returned)
- FlushCurrentTakesPartialBuffer (SD idle-flush path)

Validation: full unit test suite passes (1103 tests, 222 suites);
f407-discovery firmware build succeeds (production path compiles+links, so
simulator via the same console.mk should too).

Gotcha found while writing tests: Timer::hasElapsedUs casts the signed
(now - lastReset) delta to uint32_t without a negative guard, so a Timer
reset to a timestamp even slightly in the FUTURE of getTimeNowNt() reads as
"huge elapsed" -> hasElapsedSec(5) true. In tests, event timestamps must not
run ahead of the virtual clock or every append trips the staleness flush
(getElapsedNt clamps negatives to 0; hasElapsedUs does not).
NOTE: commit ff6b9afe5c7 captured EntriesDroppedWhenAllBuffersFull *before*
this fix - the in-tree (uncommitted) version of the test is the passing one.

Open follow-ups:
- Enable-while-enabled remains unhardened (now cleanly testable): startI()
  during an active session fails via getBigBuffer, but the evaluation order
  in 'm_bufferHandle = getBigBuffer(...)' releases the held buffer while
  ToothLoggerEnabled stays true - producers keep writing into a region
  another user may now own.
- Java CompositeParser multi-buffer test (relative-timestamp regression)
  still pending.

### DIFF
--- a/firmware/console/binary/tooth_logger.cpp
+++ b/firmware/console/binary/tooth_logger.cpp
@@ -47,6 +47,8 @@
 	fail("EFI_SHAFT_POSITION_INPUT required to have EFI_EMULATE_POSITION_SENSORS")
 #endif
 
+#include "tooth_logger_buffer.h"
+
 /**
  * Engine idles around 20Hz and revs up to 140Hz, at 60/2 and 8 cylinders we have about 20Khz events
  * If we can read buffer at 50Hz we want buffer to be about 400 elements.
@@ -109,20 +111,10 @@ void DisableToothLogger() {
 
 #else // not EFI_UNIT_TEST
 
-// Multi-buffering between producers and consumers: trigger/coil/injector edge
-// handlers (interrupt context) append composite_logger_s entries into
-// 'currentBuffer'; a buffer that fills up (or goes 5 seconds stale) is posted to
-// 'filledBuffers', where a consumer - the TS composite-log reader or the SD card
-// thread (ToothLoggerWriter) - drains it and returns it to 'freeBuffers'.
-// Buffers live in the shared BigBuffer region, so the tooth logger cannot run
-// concurrently with other BigBuffer users. See docs/AI/sd_card_logging.md
-static constexpr size_t bufferCount = BIG_BUFFER_SIZE / sizeof(CompositeBuffer);
-static_assert(bufferCount >= 2);
-
-static chibios_rt::Mailbox<CompositeBuffer*, bufferCount> freeBuffers;
-static chibios_rt::Mailbox<CompositeBuffer*, bufferCount> filledBuffers;
-
-static CompositeBuffer* currentBuffer = nullptr;
+// The buffer lifecycle itself (free/filled queues, current buffer, 5 second
+// staleness flush) lives in ToothLoggerBufferPool - see tooth_logger_buffer.h.
+// This file owns the enabled flag, the current flag state 'cur', and the
+// TS-visible ready indication.
 
 static void setToothLogReady(bool value) {
 #if EFI_TUNER_STUDIO && (EFI_PROD_CODE || EFI_SIMULATOR)
@@ -130,32 +122,13 @@ static void setToothLogReady(bool value) {
 #endif // EFI_TUNER_STUDIO
 }
 
-static BigBufferHandle bufferHandle;
+static ToothLoggerBufferPool toothBuffers{setToothLogReady};
 
 bool EnableToothLogger(TLmode mode) {
 	chibios_rt::CriticalSectionLocker csl;
 
-	bufferHandle = getBigBuffer(BigBufferUser::ToothLogger);
-	if (!bufferHandle) {
+	if (!toothBuffers.startI()) {
 		return false;
-	}
-
-	CompositeBuffer* buffers = bufferHandle.get<CompositeBuffer>();
-
-	// Reset all buffers
-	for (size_t i = 0; i < bufferCount; i++) {
-		buffers[i].nextIdx = 0;
-	}
-
-	// Reset state
-	currentBuffer = nullptr;
-
-	freeBuffers.resumeX();
-	filledBuffers.resumeX();
-
-	// Put all buffers in the free list
-	for (size_t i = 0; i < bufferCount; i++) {
-		freeBuffers.postI(&buffers[i]);
 	}
 
 	// Enable logging of edges as they come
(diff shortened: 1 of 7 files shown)


rusefillc:
@FDSoftware it's alive!!!

image

FDSoftware:
> @FDSoftware it's alive!!!
> 
> image

cool, i should hide the non-changes areas so is not too spammy

dron0gus:
Totally brakes what I have already done for https://github.com/rusefi/rusefi/issues/9853 .

dron0gus (approved):

PR #9890, opened 2026-07-21
better UX for pin renaming on old tunes

### LINKED ISSUE #9882: 2024 tune failed to load into current uaEFI
Image

FDSoftware:
image

resolves #9882

### DIFF
--- a/java_console/inifile/src/main/java/com/opensr5/ini/field/EnumIniField.java
+++ b/java_console/inifile/src/main/java/com/opensr5/ini/field/EnumIniField.java
@@ -9,25 +9,45 @@ import java.util.*;
 import java.util.regex.Pattern;
 
 public class EnumIniField extends IniField {
+    public enum PinType {
+        NONE(""),
+        GPIO("Pin"),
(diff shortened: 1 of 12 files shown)


rusefillc:
@FDSoftware I am not sure we need this complexity/additional logic

technically we have tune migration which could be used for same - if we cared enough to add tune migrator?

FDSoftware:
> @FDSoftware I am not sure we need this complexity/additional logic
> 
> technically we have tune migration which could be used for same - if we cared enough to add tune migrator?

the tune auto migrator only covers some fields if they match on the labels, this set of commits for example will fail that check:

a90a7d5b1fc	
131f62981b0	
a9eff30c575	
7ed480509fc	
db7d9057770


for this cases we are showing the partial merge warning on the update firmware, with this pr they are showed on the UI of the first comment, if is acceptable that we drop some pinouts during the tune load we can drop this whole pr and use much simpler tune migration

if not acceptable i guess we need to implement this complexity into the firmware update path also?

PR #9949, opened 2026-07-29
flash_main: fix for hybrid INT_FLASH+MFS builds

FDSoftware:


dron0gus (approved):

rusefillc:
@FDSoftware can you please clarify what was broken and what the fix does with by commenting here?

FDSoftware:
Some boards may have both Flash and MFS enabled. Currently, MFS is attempted first, then Flash.
Right now, if a board with this configuration fails or doesn't have MFS, we get an exception and an error that triggers a watchdog timer if we try to write to pages.

To resolve this, we need to rewrite pages that have a shared location into RAM when retrying using the Flash backend (as the comment in the code indicates). 

This is already done in Flash-only configurations, but in the case of Flash+MFS, we were skipping with that #if guard.

dron0gus:
While I agree with this PR and believe that it fixes the issue...
I still don't like that we have total mess in settings/config/storage code. Extra pages over raw flash seems become too complicated and hard to maintain.
Also flash/settings API is overgrown and have total mess in naming. I'm running out of brain RAM when trying to understand whole picture.
I think we should drop it someday and switch to ChibiOS MFS even for internal flash.

FDSoftware:
> I think we should drop it someday and switch to ChibiOS MFS even for internal flash.

Yep, the problem with raw flash pages is mainly the offsets/when/how to save, and Chibios would take care of all that.

rusefillc:
@dron0gus @FDSoftware https://github.com/rusefi/rusefi/commit/35d70ff31c1a571c649a86d4884617bdfe31f9ff

dron0gus:
@rusefillc @FDSoftware 
```
- Exception: STM32F7 DualBank-2MB without `EFI_FLASH_USE_1500_OF_2MB` places
  settings in a region of 16 KB sectors, where offset 72 KB would land
  outside the erased range and inside the backup copy -
  `getExtraPageFlashAddr()` returns 0 (unsupported) and those boards persist
  extra pages on their SD card instead.
```
huh?

rusefillc:
https://github.com/rusefi/rusefi/issues/9953

PR #9954, opened 2026-07-31
add UI overlay when a unsupported board is connected, reject connection to unsupported ecu

### LINKED ISSUE #9945: nicer message when (universal) does not support current board
universal bundle (any bundle?) should say something explicit when an unsupported rusEFI is hooked up

FDSoftware:
also add UI overlay when a unsuported board is connected, takes full screen both on splash or already openened full console

~draft, i will split the UI changes from the PortResult and friends to have this more readable~

image

resolves #9945


FDSoftware:
https://github.com/rusefi/rusefi/pull/9955

Issue #9630, opened 2026-06-11
CI step for stack protection


rusefillc:
### Tools for per-thread (whole call-chain) stack checking in CI

These consume GCC's `-fstack-usage` `.su` files plus the call graph and compute worst-case stack per call tree, so you can fail CI when a thread's deepest path exceeds its `THD_WORKING_AREA`.

#### Ready-made open-source tools

- **`worst-case-stack` / `wcs` (PeterTh/ARM ports)** — Python tools that parse `.su` + ELF `objdump` to produce worst-case stack per root function. Easy to wrap with a threshold check and `exit 1`.
- **`puncover`** (HBehrens/puncover) — reads `.su` + the ELF, builds the call graph, and reports per-function max stack including callees. Has a CLI/JSON mode you can script against thresholds in CI (not just the web UI).
- **`stack-usage` from `eclipse-cdt` / `avstack.pl`** — the classic `avstack.pl` (originally AVR, easily adapted to `arm-none-eabi-objdump`) merges `.su` data with a disassembled call graph and prints per-root worst case. Small, hackable Perl — good fit for a custom CI gate.
- **`mageec`/`stacktool`-style scripts** — various community scripts doing the same `.su` + call-graph merge.

#### Commercial / heavyweight (sound analysis)

- **AbsInt StackAnalyzer** — industry-standard, sound worst-case stack incl. recursion and (annotated) indirect calls; integrates into CI. Overkill unless you need certification (ISO 26262 etc.).
- **GNATstack** (AdaCore) — works on C too via GCC; computes per-entry worst-case and flags unbounded/cyclic paths.

#### LTO caveat for this repo

With `-flto=auto`, `.su` files are emitted at the `ltrans` stage and functions are already inlined/merged, so a tool reading the post-LTO `.su` + final ELF call graph is what you want. The merged frames (like the 2344 B `bmwThread`) already fold callees in, which actually simplifies per-thread summation but means you must analyze the **post-LTO** artifacts, not per-`.cpp` `.su`.

#### Indirect-call caveat

All automatic tools break on function pointers / C++ virtuals (e.g. `CanListener` dispatch, RTOS callbacks) and recursion — they either ignore those edges (under-estimate) or need manual annotations. Budget for a small hand-maintained "extra depth" margin per thread.

### Recommended minimal CI gate for this project

1. Build with the existing `-fstack-usage` (already permanent in `common_private.mk`).
[...]

PR #9976, opened 2026-08-01
stack shortage in secure bootloader

rusefillc:


rusefillc:
we really need https://github.com/rusefi/rusefi/issues/9630

PR #9983, opened 2026-08-03
stack usage report: first iteration of custom tool for one board of F4, F7 and H7

FDSoftware:
related #9630

PR #9998, opened 2026-08-04
StackUsageReport: preparation, fw side hooks

FDSoftware:
related https://github.com/rusefi/rusefi/issues/9630

### DIFF
--- a/misc/installer/uaefi/uaefi-updater.install4j
+++ b/misc/installer/uaefi/uaefi-updater.install4j
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <install4j version="11.0.5" transformSequenceNumber="11">
   <directoryPresets config="../../../../../bundles/rusefi_bundle_uaefi/rusefi.snapshot.uaefi" />
-  <application name="uaEFI Updater" applicationId="9384-2741-6502-8137" mediaDir="." shortName="uaEFI Updater" version="1.0.0.0" allPathsRelative="true">
+  <application name="uaEFI Updater" applicationId="9384-2741-6502-8137" mediaDir="." shortName="uaEFI Updater" version="1.0.0.1" allPathsRelative="true">
     <jreBundles jdkProviderId="Zulu" release="17/17.0.14" />
   </application>
   <files>
@@ -17,7 +17,7 @@
   <launchers>
     <launcher name="uaEFI-Grab-Fresh-And-Update-ECU" id="3">
       <executable name="uaEFI-Grab-Fresh-And-Update-ECU" iconSet="true" executableMode="gui" workingDirectory="console" executionLevel="requireAdministrator" />
-      <java mainClass="com.rusefi.autoupdate.Autoupdate" arguments="basic-ui">
+      <java mainClass="com.rusefi.Launcher" arguments="basic-ui">
         <classPath>
           <archive location="console/rusefi_console.jar" failOnError="false" />
         </classPath>
--- a/misc/installer/universal/universal-updater.install4j
+++ b/misc/installer/universal/universal-updater.install4j
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <install4j version="11.0.5" transformSequenceNumber="11">
   <directoryPresets config="../../../../../bundles/rusefi_bundle_universal/rusefi.snapshot.universal" />
-  <application name="rusEFI Universal Updater" applicationId="5936-1408-7724-2651" mediaDir="." shortName="rusEFI Universal Updater" version="1.0.0.0" allPathsRelative="true" macVolumeId="9af7b89f15108498">
+  <application name="rusEFI Universal Updater" applicationId="5936-1408-7724-2651" mediaDir="." shortName="rusEFI Universal Updater" version="1.0.0.1" allPathsRelative="true" macVolumeId="9af7b89f15108498">
     <jreBundles jdkProviderId="Zulu" release="17/17.0.14" />
   </application>
   <files>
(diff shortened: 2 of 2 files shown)


rusefillc:
human says:
seems to be fragile with too much copy-paste and thread stack meta info detached from stack allocation. Suggest a more rigid pattern 


LLM agrees as LLM does:


Logged. Here's my assessment of PR #9998 and the pattern I'd suggest instead.

## What makes the current pattern fragile

Your instinct is right — the design has four independent hand-maintained encodings of what should be one fact, and none of them are checked by the compiler or linker:

1. **Root identity is a hand-typed demangled string.** `"DcThread::PeriodicTask(long long)"` has no compile-time link to the function. A rename, a signature change (e.g. `efitick_t` ever changing width), or even a demangler-format difference across binutils versions orphans the annotation silently — the report tool just stops finding the root.
2. **Budget is detached from the allocation.** The stack size is repeated at the `RUSEFI_STACK_ROOT` call site, separately from the `ThreadController<TStackSize>` template argument or `THD_WORKING_AREA` that actually allocates it, and it travels through the **section name** (`QUOTE(budget)` → `.rusefi_stack_root.SD__MMC.3*400`), so the tool must parse and evaluate arithmetic out of section names, plus the magic `process` keyword for the main stack.
3. **Data smuggled through identifiers.** Root names need `__` escaping (`SD__MMC`, `main__process`, `lwIP_TCP__IP`), and `board.mk` carries a dot-separated `RUSEFI_STACK_REVIEWED_ROOTS` list that `RUSEFI_STACK_REVIEW_ROOTS()` encodes into yet another section name. That's three stringly encodings of the same identity that must stay in sync by eye.
4. **Per-board review data is positional code.** `RUSEFI_STACK_REVIEW("f407…", "f767…", "h743…")` at ~15 call sites, selected by the `HW_FRANKENSO`/`HW_NUCLEO_*` `#if` chain in `common_headers.h`. Adding a fourth reference board touches every annotated file, the header, and board.mk. Worse: these triples are *hardware measurements* — every soak-test refresh means editing C++ files. And `idle`/`main` roots are already copy-pasted between `main.cpp` and `bootloader_main.cpp`.

## Suggested pattern

**A. Let the compiler carry root identity and budget — metadata in a mangled symbol name, not a string.**

```cpp
// stack_root.h
template <auto Entry, int Budget>
[[gnu::used, gnu::retain]] inline const char stackRoot = 0;

#define RUSEFI_STACK_ROOT(Type, Entry) \
    template const char stackRoot<&Type::Entry, Type::stackSize>;
```

with one line added to `ThreadController` (thread_controller.h:18): `static constexpr int stackSize = TStackSize;`. Usage next to the class:

```cpp
RUSEFI_STACK_ROOT(DcThread, PeriodicTask)
static DcThread dcThread CCM_OPTIONAL;
```

The emitted symbol demangles to `stackRoot<&DcThread::PeriodicTask(long long), 512>` — the tool scans `nm` output for `stackRoot<` and gets everything. Crucially, the NTTP mangling *embeds the exact mangled name of the target function* (`_ZN8DcThread12PeriodicTaskEll` appears verbatim inside the registration symbol), so the tool can extract the call-graph root textually with no demangler round-trip at all. A rename is a compile error; the budget is literally the same constant that sizes the working area, so it cannot drift. For raw-`THD_FUNCTION` threads (storage, mmc), bind via `sizeof`:

```cpp
static THD_WORKING_AREA(storageManagerThreadStack, STORAGE_MANAGER_THREAD_STACK_SIZE);
RUSEFI_STACK_ROOT_FN(storageManagerThread, sizeof(storageManagerThreadStack))
```

Foreign roots you can't take an address of in C++ (`__idle_thread`, `main`, lwIP's internal `tcpip_thread`) keep a string-based fallback — but confined to a single `stack_roots.cpp` per image instead of duplicated across `main.cpp`/`bootloader_main.cpp`, and the main-stack budget should come from the tool reading the `__process_stack_size__` linker symbol rather than a `process` keyword.

Cost: one byte of bss per root (~25 bytes total) instead of zero — negligible, and it avoids all asm/section-name machinery. If zero-image-cost matters, place them in one custom section marked NOLOAD in the linker scripts.

**B. Review measurements are data, not code — move them to per-board files.**

`firmware/config/boards/<board>/stack_review.json` (only the three reference boards need one):

```json
{ "DcThread::PeriodicTask": { "used": 304, "free": 556, "scenario": "normal ETB update" },
  "TunerstudioThread::ThreadTask": { "used": 824, "free": 1324, "scenario": "normal communication" } }
```

The report tool joins ELF roots against the board's file:

- a key that matches no ELF root **fails CI** — this is the rename detection the `board.mk` dot-list was trying to buy, but automatic;
- an absent key simply means "unreviewed" — replacing all the `""` placeholders and the positional argument scheme;
- adding a fourth reference board is one new JSON file, zero C++ edits;
- the tool can grow an `--update-review` mode so refreshed hardware measurements are a regenerate-and-commit, not a 15-file C++ edit.

This deletes `RUSEFI_STACK_REVIEW`, the `HW_*` `#if` chain in `common_headers.h`, `RUSEFI_STACK_REVIEW_ROOTS()`, the `board.mk` defines, and the `__`-escaping convention entirely. What remains in firmware is only what firmware truly owns: *which functions are thread roots and how big their stacks are* — bound to the allocation and compiler-verified.

Two things to verify before committing to this: (1) that an explicit variable-template instantiation with `used`/`retain` survives LTO + `--gc-sections` on the arm toolchain (check `nm` on the final ELF — if not, a `static inline` member of a class template is the fallback); (2) access — `ThreadTask`/`PeriodicTask` overrides that are `protected` need to be made public (harmless; they're only reachable via vtable anyway) or the macro needs a friend hook in the base.

FDSoftware:
> ## Suggested pattern
> **A. Let the compiler carry root identity and budget — metadata in a mangled symbol name, not a string.**
> 

done in this pr

> **B. Review measurements are data, not code — move them to per-board files.**
> 
> `firmware/config/boards/<board>/stack_review.json` (only the three reference boards need one):
> 
> ```json
> { "DcThread::PeriodicTask": { "used": 304, "free": 556, "scenario": "normal ETB update" },
>   "TunerstudioThread::ThreadTask": { "used": 824, "free": 1324, "scenario": "normal communication" } }
> ```
> 
> The report tool joins ELF roots against the board's file:


done in 0676a73ca79c47607621ce7d019e4ce8c578ecc8 of  #9983 PR, but in md format since the list needs to be maintained from time to time and is better for hummans than json

rusefillc:
@FDSoftware please rebase, hopefully I've picked up completely?

Issue #7971, opened 2025-05-29
technical debt: unused class loader magic


rusefillc:
these days autoupdate process invokes new java process, do we still need class loader magic?

Issue #10000, opened 2026-08-04
updater has exception


Arthris:
```
I 260804 111011.121 [main] Autoupdate - Version 20260429
I 260804 111011.128 [main] Autoupdate - Compiled Wed Dec 31 19:00:00 EST 1969
I 260804 111011.128 [main] Autoupdate - Current folder C:\Program Files\Purple Updater\console
I 260804 111011.130 [main] Autoupdate - Source C:\Program%20Files\Purple%20Updater\console\rusefi_console.jar
I 260804 111011.179 [main] PersistentConfiguration - Console configuration from C:\Users\chris\.rusEFI\console_properties.xml
I 260804 111011.179 [main] PropertiesHolder - auto_update_root_url=https://rusefi.com/fw-private
I 260804 111011.179 [main] PropertiesHolder - auto_update_root_url getProperty=https://rusefi.com/fw-private
I 260804 111011.182 [main] BundleInfoStrategy - Snapshot requested
I 260804 111011.182 [main] FindFileHelper - Scanning C:\Program Files\Purple Updater\console\.. for /srec
I 260804 111011.182 [main] FindFileHelper - Searching for *srec in C:\Program Files\Purple Updater\console\..
I 260804 111011.184 [main] FindFileHelper - Located ..\rusefi-development_2026-07-20_purple-gateway_2894741984_4fa31d78a5bf59f20430ae5c3fc09bc7a50871c0_obfuscated.srec
I 260804 111011.403 [main] ConnectionAndMeta - Connecting to https://rusefi.com/fw-private/autoupdate/rusefi_bundle_purple-gateway_obfuscated_public_autoupdate.zip?u=7d83773d-247e-4ae8-a688-61483d937c1d
I 260804 111011.447 [main] ConnectionAndMeta - Request Headers: {Cache-Control=[no-cache, no-store, must-revalidate, max-age=0], User-Agent=[RE-Internal-Sync], Expires=[0], Pragma=[no-cache]}
I 260804 111011.926 [main] ConnectionAndMeta - Response code 200
I 260804 111011.926 [main] Autoupdate - Local file C:\Users\chris\.rusEFI\updates\purple-gateway_development\rusefi_bundle_purple-gateway_obfuscated_public_autoupdate.zip
I 260804 111011.926 [main] Autoupdate - Remote file rusefi_bundle_purple-gateway_obfuscated_public_autoupdate.zip
[...]

Issue #9727, opened 2026-06-28
i have rusefi_console_pending.jar and I see old code running


rusefillc:
i saw the update progress bar and I saw the fancy new console look!

and then I closed and re-started - and I see old splash screen again?

<img width="715" height="143" alt="Image" src="https://github.com/user-attachments/assets/1651543e-a956-45c9-92d3-ecf900b4ef4d" />

[logs.zip](https://github.com/user-attachments/files/29438274/logs.zip)

FDSoftware:
Triaged: this appears to be fixed by https://github.com/rusefi/rusefi/commit/57029bde5e55005ac2d69f25cf49d6d67bb78d4e, committed 2026-06-29 after this report. The fix retries finalizePendingUpdate() while Windows releases the old rusefi_console.jar lock, avoiding the orphaned rusefi_console_pending.jar / old-code restart case described here.

Issue #6777, opened 2024-08-07
autoupdate fails - classloader right after file overwrite


rusefillc:
we crash

```
E 240807 121255.641 [main] Autoupdate - Failed to start
java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at com.rusefi.autoupdate.Autoupdate.startConsole(Autoupdate.java:106)
	at com.rusefi.autoupdate.Autoupdate.autoupdate(Autoupdate.java:85)
	at com.rusefi.autoupdate.Autoupdate.main(Autoupdate.java:42)
Caused by: java.lang.reflect.InvocationTargetException
	at java.desktop/java.awt.EventQueue.invokeAndWait(EventQueue.java:1367)
	at java.desktop/java.awt.EventQueue.invokeAndWait(EventQueue.java:1342)
	at java.desktop/javax.swing.SwingUtilities.invokeAndWait(SwingUtilities.java:1480)
	at com.rusefi.ConsoleUI.startUi(ConsoleUI.java:204)
	at com.rusefi.Launcher.main(Launcher.java:44)
	... 7 more
Caused by: java.lang.IllegalStateException: java.lang.RuntimeException: java.util.zip.ZipException: ZipFile invalid LOC header (bad signature)
	at com.rusefi.ConsoleUI.awtCode(ConsoleUI.java:249)
	at com.rusefi.ConsoleUI.lambda$startUi$2(ConsoleUI.java:204)
	at java.desktop/java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:303)
	at java.desktop/java.awt.EventQueue.dispatchEventImpl(EventQueue.java:770)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:721)
	at java.desktop/java.awt.EventQueue$4.run(EventQueue.java:715)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.security.ProtectionDomain$JavaSecurityAccessImpl.doIntersectionPrivilege(ProtectionDomain.java:85)
[...]

PR #10003, opened 2026-08-04
updater has exception #10000

rusefillc:
only:purple?

### DIFF
--- a/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
+++ b/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/Autoupdate.java
@@ -25,7 +25,10 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
+import java.util.List;
 import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
@@ -74,7 +77,12 @@ import static com.rusefi.core.FindFileHelper.findFirmwareFile;
  *   <li>If auto-update is enabled, {@link #downloadFreshZipFile} downloads the latest
  *       bundle zip (only when server timestamp/size indicate it's newer than the local
  *       copy — see {@link #isNewerAvailable}); otherwise nothing is downloaded.</li>
- *   <li>{@link #unzipAutoUpdate} extracts the bundle over the installation.</li>
+ *   <li>{@link #unzipAutoUpdate} extracts the bundle over the installation. Exception: when
+ *       the updater itself was launched from {@code rusefi_console.jar} (some white-label exe
+ *       launchers are configured that way, #10000) the flow instead reuses the staged
+ *       pending-JAR path of the in-console update ({@link #applyUpdate} +
+ *       {@link #relaunchConsole(String[])}), because extracting over the installation would
+ *       overwrite the JAR this JVM is running from and break lazy class loading.</li>
  *   <li>{@link #prefetchLastConnectedBoardFirmware} (issue #9714) refreshes firmware for
  *       the board the user last connected to, so a later in-console "Update ECU" is
  *       instant.</li>
@@ -261,6 +269,22 @@ public class Autoupdate {
         // Let's try to get console .exe-file name before we rewrite autoupdate .jar file:
         final String consoleExeFileName = new ConsoleExeFileLocator().getConsoleExeFileName();
 
+        if (downloadedAutoupdateFile.isPresent() && isRunningFromConsoleJar()) {
+            // #10000: this launcher runs Autoupdate straight from rusefi_console.jar, so the legacy
+            // unzipAutoUpdate path would overwrite the running JAR on disk and break lazy class
+            // loading (NoClassDefFoundError on any class not loaded yet). Reuse the staged
+            // pending-JAR path of the in-console update instead: the running JAR is never touched,
+            // relaunchConsole() starts the new code from the pending JAR and finalizePendingUpdate()
+            // promotes it to the real name during that startup.
+            try {
+                applyUpdate(downloadedAutoupdateFile.get());
+            } catch (Throwable e) {
+                log.error("Staged update failed, launching existing console", e);
+            }
+            prefetchLastConnectedBoardFirmware();
+            relaunchConsole(args); // does not return - exits this JVM after spawning the console
+        }
+
         // java lazy class-loader would get broken if we replace rusefi_autoupdate.jar file
         // ATTENTION! To avoid `ClassNotFoundException` we need to load all necessary classes before unzipping
         // autoupdate archive
@@ -306,6 +330,14 @@ public class Autoupdate {
      * current JAR path cannot be resolved (e.g. running from an IDE class directory).
      */
     public static void relaunchConsole() {
+        relaunchConsole(new String[0]);
+    }
+
+    /**
+     * Same as {@link #relaunchConsole()} but forwards program arguments to the new console
+     * process - used by the launch-time staged-update path, which must pass its own args through.
+     */
+    public static void relaunchConsole(String[] args) {
         String jarPath = getCurrentJarPath();
         if (jarPath != null) {
             // If a staged update exists, launch from it so the new code is picked up immediately.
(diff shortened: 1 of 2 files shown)


rusefillc:
@FDSoftware where is the real issue? is the problem with purple installer configuration?

## 2026-08-04: Investigation: updater NoClassDefFoundError FindFileHelper$2

What was done:
- Investigated a field failure (Purple Updater bundle, updater version
  20260429, Windows): java.lang.NoClassDefFoundError:
  com/rusefi/core/FindFileHelper$2 at FindFileHelper.findSrecFile
  (FindFileHelper.java:86) called from Autoupdate.unzipAutoUpdate.
  Investigation only, no code changes.

Root cause:
- The updater's own log shows Source = ...console\rusefi_console.jar: in
  this bundle the exe4j launcher runs Autoupdate FROM rusefi_console.jar.
- The legacy silent-update path (Autoupdate.autoupdate ->
  unzipAutoUpdate) overwrites rusefi_console.jar on disk FIRST
  (unzipFreshConsole, needed historically to prepare a classloader when the
  updater ran from rusefi_autoupdate.jar, see #6777), then calls
  findSrecFile(false).
- FindFileHelper$2 is the anonymous AdditionalFileHandler on the
  "Second choice: current folder" branch (FindFileHelper.java:86). It was
  never loaded before the overwrite: the startup scan found the srec at the
  first location, so that branch never executed. findSrecFile(false) ALWAYS
  reaches that branch (findXNumberOfFile with keepOneFile=false
  unconditionally returns null), so the first post-overwrite call forces a
  lazy class load from a jar whose bytes were just replaced under the live
  classloader -> stale URLClassPath zip directory -> ClassNotFoundException
  -> NoClassDefFoundError. Same failure family the code already documents
  for rusefi_autoupdate.jar ("invalid LOC header", #6777) and the todo in
  unzipAutoUpdate references (#7971). tryInstallTsPlugin catching
  NoClassDefFoundError is prior whack-a-mole evidence.
- Standard bundles launch the updater from rusefi_autoupdate.jar where the
  console-jar overwrite is safe; the crash needs a launcher that runs
  Autoupdate from the console jar itself.

Collateral on the affected machine:
- The exception fired after the console jar was replaced and after the old
  srec was moved to .rusEFI\older-fw\, but BEFORE the rest of the bundle
  (fresh srec etc.) was unzipped and before the console was started. A
  re-run will likely skip re-unzip (cached bundle zip is up-to-date ->
  doDownload returns empty), leaving no srec next to the console; recovery
  = delete the cached zip under .rusEFI\updates\<target>_<branch>\ to force
  a full re-apply, or unzip the firmware artifacts from it manually.
  findForConnectedBoard can also still recover the displaced srec from
  older-fw at flash time.

Recommended fixes (not implemented):
- Proper: when the running jar IS the console jar, route the startup silent
  update through the pending-jar path that already exists for in-console
  updates (applyUpdate: archive srec BEFORE unzip, unzip everything except
  the console jar, stage rusefi_console_pending.jar, promote via
  finalizePendingUpdate on next start). Retires the unzipFreshConsole-first
  ordering; closes #7971.
- Cheap hardening: replace the two anonymous AdditionalFileHandler classes
  in findSrecFile with FindFileHelper::moveFile method references (as
  archiveFirmwareFiles already does) - lambdas/method refs need no jar
  entry at runtime, removing this specific lazy-load landmine.

## 2026-08-04: Fix: updater NoClassDefFoundError - cheap hardening (#10000)

What was done (committed as 77b83bff0b9):

| File | Change |
|-----------------------------------------------------------------|--------------------------------------------|
| java_console/shared_io/.../core/FindFileHelper.java | findSrecFile: two anonymous AdditionalFileHandler classes -> FindFileHelper::moveFile method references, with a comment explaining the lazy-load hazard |
| java_console/autoupdate/.../autoupdate/Autoupdate.java | AUTOUPDATE_VERSION -> 20260804 (change runs inside the updater) |
| java_tools/version/.../UiVersion.java | CONSOLE_VERSION -> 20260804 |

Key decision and why:
- Method references (like archiveFirmwareFiles already used) compile to
  invokedynamic against already-loaded code - no FindFileHelper$N jar entry
  to lazily load - so findSrecFile can no longer crash after the running
  rusefi_console.jar has been replaced on disk. This removes the specific
  landmine only; the underlying overwrite-the-running-jar ordering in the
  legacy unzipAutoUpdate path (#7971) remains an open follow-up.

Validation:
- ./gradlew :shared_io:test :autoupdate:compileJava :version:compileJava
  passes; shared_io suite 40 tests, 0 failures (includes FindFileHelperTest
  and FindFileHelperTargetSelectionTest which exercise findSrecFile).

Open follow-ups:
- Proper fix per investigation above: route the startup silent update
  through the pending-jar path when the updater runs from the console jar.

## 2026-08-04: Fix: updater self-overwrite - staged pending-JAR routing (#10000)

What was done (implements the "proper fix" from the investigation above):

| File | Change |
|-----------------------------------------------------------------|--------------------------------------------|
| java_console/autoupdate/.../autoupdate/Autoupdate.java | autoupdate(): when an update was downloaded AND the updater runs from the console jar, apply via the staged path (applyUpdate) and launch via relaunchConsole(args) instead of legacy unzipAutoUpdate + startConsoleAsANewProcess |
| java_console/autoupdate/.../autoupdate/Autoupdate.java | new relaunchConsole(String[] args) overload forwarding program args; no-arg version delegates to it |
| java_console/autoupdate/.../autoupdate/Autoupdate.java | new isRunningFromConsoleJar()/isConsoleJarFileName()/consoleJarFileName() helpers (white-label aware, matches pending jar too) |
| java_console/autoupdate/.../autoupdate/Autoupdate.java | unzipFreshConsole(): defense-in-depth guard - refuses to overwrite the jar the JVM is running from |
| java_console/autoupdate/.../autoupdate/Autoupdate.java | class javadoc section 3 documents the two launch-time routes |
| java_console/autoupdate/src/test/.../ConsoleJarDetectionTest.java | new unit test for the detection seam (staged with git add) |

Key decisions and why:
- Reused the existing in-console staged machinery (applyUpdate ->
  extractConsoleJarAsPending -> relaunchConsole -> finalizePendingUpdate)
  rather than inventing a second mechanism: applyUpdate already archives
  the srec BEFORE unzipping and never touches the running jar;
  rusefi_console.jar's manifest Main-Class is com.rusefi.Launcher, which
  calls finalizePendingUpdate() early, so java -jar on the pending jar
  promotes the new code on its own startup.
- Detection compares the running jar's file name against the resolved
  consoleJarZipEntry base name, NOT a hardcoded rusefi_console.jar:
  getRusEfiConsoleJarName() is white-label aware (prefix derived from the
  code-source jar), so hardcoding would break white-label bundles. The
  pending jar name also counts as "console jar".
- The staged branch triggers only when an update was actually downloaded;
  the no-download launch path is unchanged (console exe as before).
- If applyUpdate throws, the error is logged and the existing console is
  launched anyway - same degrade-gracefully spirit as the legacy path.
- AUTOUPDATE_VERSION / CONSOLE_VERSION already bumped to 20260804 earlier
  today - no further bump needed.

Validation:
- ./gradlew :autoupdate:test - 18 tests, 0 failures (includes new
  ConsoleJarDetectionTest; first version wrongly hardcoded
  rusefi_console.jar and failed under gradle where the resolved white-label
  name is shared_console.jar - fixed by asserting against
  consoleJarFileName()).
- ./gradlew :ui:compileJava passes (StartupFrame/MainFrame/
  StartupUpdateActions still use the no-arg relaunchConsole()).

Open follow-ups:
- Field-verify on the affected Purple Updater installation: next bundle
  update should stage rusefi_console_pending.jar and relaunch cleanly
  instead of crashing with NoClassDefFoundError.
- The legacy unzipAutoUpdate path (updater running from
  rusefi_autoupdate.jar) is intentionally untouched (#7971 remains open
  for retiring it entirely).

FDSoftware:
we have the same configuration as the universal bundle, so we have the same issue over there, seems to be related to the `PENDING_CONSOLE_JAR` that i fixed i while ago for other issue https://github.com/rusefi/rusefi/pull/9743 #9727

(still checking if we are missing some other edge case with this fix)

rusefillc:
@FDSoftware TL,DR? do I create new installer? or TBD?

FDSoftware:
> @FDSoftware TL,DR? do I create new installer? or TBD?

new installer + needs manual test for non-installer bundles, but PR seems ok

rusefillc:
@FDSoftware this PR does not seem right - do we really need conditional isRunningFromConsoleJar? this would be a mess to maintain

PR #10005, opened 2026-08-04
update install4j profiles for crash autoupdate fix

FDSoftware:
related #10003  #10000


rusefillc:
@FDSoftware please elaborate what's going on here

please elaborate relation to questionable https://github.com/rusefi/rusefi/pull/10003

FDSoftware:
this change is for "Grab Fresh And Update ECU" so it on launchs enter `rusefi_console.jar` through
`com.rusefi.Launcher` rather than `Autoupdate.main`, fixing the entire crash,

is alternative to #10003 questionable fix, but still we need to fortify better the auto-update procces so this dont happen again (working on that now)

Issue #9073, opened 2026-02-09
Check Engine phase one


rusefillc:
#7332
#7382

olokos:
I see part of the reason why CEL is not in outputs at all:
https://github.com/rusefi/rusefi/blob/e8aea488c310bfe0fd7a6e065cd8273fdc37312f/firmware/config/boards/prometheus/board_configuration.cpp#L158
Here prometheus board does have CEL assigned to a pin in board config

Hellen GM E67 has it commented out:
https://github.com/rusefi/rusefi/blob/e8aea488c310bfe0fd7a6e065cd8273fdc37312f/firmware/config/boards/hellen/hellen-gm-e67/board_configuration.cpp#L30

While on rusefi uaEFI board, it does not seem to exist at all in board configuration:
https://github.com/rusefi/rusefi/blob/e8aea488c310bfe0fd7a6e065cd8273fdc37312f/firmware/config/boards/hellen/uaefi/board_configuration.cpp

Looks like the core enablement of CEL is here:
https://github.com/rusefi/rusefi/blob/e8aea488c310bfe0fd7a6e065cd8273fdc37312f/firmware/controllers/gauges/malfunction_indicator.cpp#L117

By checking if the pin is defined.

It is defined in uaEFI board header file here:
https://github.com/rusefi/rusefi/blob/e8aea488c310bfe0fd7a6e065cd8273fdc37312f/firmware/controllers/generated/engine_configuration_generated_structures_uaefi.h#L2476

---

I tried `set malfunction_indicator_pin PA8`
Which returns  `EngineState: confirmation_set malfunction_indicator_pin PA8:33`

But then 
`get malfunction_indicator_pin` gives

`2026-06-20_23_41_45_543: EngineState: Invalid Parameter: malfunction_indicator_pin`
`2026-06-20_23_41_45_543: EngineState: confirmation_get malfunction_indicator_pin:29`

Which leads me to believe, the first step into getting CEL to work, would be to assign some arbitrary unused pin for `malfunction_indicator_pin` and then it might become visible in outputs again.

Can somebody verify this?

FDSoftware:
after #10070

phase 2 can be:

- CLT, EGO, MAF, EGT, flex, and sync-loss checks.
- Individual enable switches and startup delays.
- Sensor fluctuation/rate checks.
- Lua-defined CEL conditions.

phase 3 can be:
* storing/reading DTC
* Total/stored/pending DTC displays.
[...]

PR #10044, opened 2026-08-10
CheckEngine: phase one

FDSoftware:
* new engine module for handling the error codes from the sensors
* updated the TS UI:

image

(will be more fancier in latter pr's, just a minimal-working version)

* refactored the MIL logic for using onSlowCallback, now is a engine module! (still pending to move into /controllers/modules folder)

MIL module can be disabled, to having only the CEL one, saving 1k of flash (we can also disable CEL, but we can't disable CEL and enable MIL, added a makefile guard for that)

#9073


rusefillc:
@FDSoftware please rebase to grab master fixes

PR #10051, opened 2026-08-10
DTC manager

dron0gus:


FDSoftware:
If it's not too much trouble, could you please move `dtc_manager.h‎` to an engine module so we're in line with the other PR? (#10044)[MIL/generic Check Engine things]

dron0gus:
DTC manager becomes part of Engine.
Engine is located in CCM memory of F4.
CCM is overflown now.

F

PR #10054, opened 2026-08-11
fix "Air fuel ratio gauge is in lambda"

### LINKED ISSUE #9988: RE console: Air fuel ratio gauge is in lambda
AFR is selected why is it displaying lambda?
Image

FDSoftware:
now we can have happy dynamic AFR/Lambda gauge using the existing "lambda display" setting, without reloading the dashboard :D

image

image

resolves #9988

part of the fix was error on firmware, another one on the java side, and another bigger one on the .ini template

depends on #10053


rusefillc:
"this bug is in fact three different bugs"?!

FDSoftware:
> "this bug is in fact three different bugs"?!

is the same big bug (afr is mixed up) but affects 3 different places

firmware: We calculated the AFR incorrectly.
ini: We displayed the labels incorrectly in the AFR case.
console: We didn't parse the label correctly in any case.

Issue #8359, opened 2025-07-25
rusefi_updater on Linux, no way to invoke


rusefillc:
rusefi_updater.exe is the simpler UI with smart configuration logic

problem: we have no linux script to invoke rusefi_console.jar with "basic-ui"

PR #10081, opened 2026-08-14
Autoupdate: find the bundle launcher on Linux/macOS (#8360)

### LINKED ISSUE #8360: rusefi_updater on Linux, autoupdater would not invoke new process
a bit different from https://github.com/rusefi/rusefi/issues/8359

part of autoupdate we invoke freshly downloaded console as a new java process, this probably does not work on linux

TokenGoblin:
Fixes #8360. Also removes the main obstacle behind #8359.

## Problem

The autoupdater could not re-invoke the console as a new process on Linux. Digging in, there are **two** independent defects in the launcher lookup, and the second one is not Linux-specific.

**1. Wrong suffix.** `ConsoleExeFileLocator` built its default as `<prefix>_updater.exe` unconditionally. The Linux/macOS bundle ships `<prefix>_updater.sh` instead — see `ROOT_FOLDER_SOURCES` in `firmware/bundle.mk` and `firmware/bin/assemble_universal_bundle.sh`.

**2. Wrong directory.** That name was returned bare and resolved against the current working directory. But both launchers chdir into the bundle's `console` folder before starting the JVM — launch4j `<chdir>console</chdir>` in `rusefi_updater.xml`, and the `cd` in `rusefi_updater.sh` — while the launchers themselves live at the **bundle root**, one level up.

Verified with a real JVM whose working directory was a bundle's `console` folder:

```
Files.exists("rusefi_updater.exe") = false
Files.exists("rusefi_updater.sh")  = false
```

So `Files.exists()` in `startConsoleAsANewProcess()` was returning false for the Windows launcher too, not only on Linux. Fixing only the suffix would have left the bug alive.

**3. Not executable.** `misc/console_launcher/rusefi_updater.sh` was committed as mode `100644`, so it shipped without the executable bit and could not be run without a manual `chmod +x`. It also dropped the arguments that `startConsoleAsANewProcess` passes to it, while launch4j already forwards them on Windows.

## Approach

Deliberately conservative, so nothing that works today can regress:

- `installation.properties` `console-exe-file` still wins, untouched. Installers write it, so installed layouts keep exactly their current behavior — only the fallback default changed.
- Search order is CWD first, then the parent. That is a strict superset of the old lookup.
- When nothing is found, the bare file name is still returned, so the existing "file to launch isn't found" dialog is unchanged.
- `resolveDefaultLauncher()` takes the platform flag and working directory as parameters instead of reading `os.name`/CWD, so the tests assert both the Windows and the Unix rule on any host. No reflection, per the Java testing rules in `CLAUDE.md`.

Also fixes a `` `$s` `` → `` `%s` `` format typo in the launch-failure log message, which was printing the literal `$s` instead of the command.

## Changes

| File | Change |
|---|---|
| `ConsoleExeFileLocator.java` | OS-aware launcher suffix; resolve against CWD then the bundle root; new `resolveDefaultLauncher` test seam |
| `Autoupdate.java` | Best-effort executable bit for non-Windows launchers; `$s` → `%s` log fix; bump `AUTOUPDATE_VERSION` |
| `ConsoleExeFileLocatorTest.java` | 7 new OS-independent tests |
| `misc/console_launcher/rusefi_updater.sh` | mode 100755; forwards `"$@"`; `exec`; exits if the `cd` fails |
| `docs/report.md` | dated entry per `CLAUDE.md` |

`CONSOLE_VERSION` already read today's date (20260813), so per `CLAUDE.md` that bump is a no-op.

## Validation

- `gradlew :autoupdate:test :shared_io:test` green — 21 autoupdate tests (7 new) and 41 shared_io tests. Green baseline captured before any edit.
- End-to-end check with a real JVM in a simulated bundle layout: `resolveDefaultLauncher` returns the absolute path of the bundle-root launcher for both the Windows and the Unix case, where the old bare-name lookup found nothing.

**Not verified on real Linux hardware** — no Linux host was available. The executable-bit runtime path in particular rests on reasoning plus the git mode change, so a Linux smoke test before merge would be worthwhile.

`:ui:shadowJar` could not be built in my environment (`:ecu_io` fails on a missing `peak.can.basic` package); confirmed pre-existing by reproducing it on a clean tree.

## Notes for reviewers

- Happy to split the `rusefi_updater.sh` changes (items 3) into a separate PR against #8359 if you'd prefer. I kept them together because the #8360 fix does nothing useful if the script still cannot be invoked.
- The `NoClassDefFoundError: com/rusefi/autoupdate/TsPluginInstaller` reported in the #8360 thread is a separate packaging problem and is **not** addressed here.

### DIFF
--- a/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/ConsoleExeFileLocator.java
+++ b/java_console/autoupdate/src/main/java/com/rusefi/autoupdate/ConsoleExeFileLocator.java
@@ -1,6 +1,7 @@
 package com.rusefi.autoupdate;
 
 import com.devexperts.logging.Logging;
+import com.rusefi.core.OsUtil;
 import com.rusefi.core.net.JarFileUtil;
 
 import java.io.InputStream;
@@ -13,12 +14,31 @@ import java.util.Properties;
 
 import static com.devexperts.logging.Logging.getLogging;
 
+/**
+ * Finds the bundle launcher which {@link Autoupdate#startConsoleAsANewProcess} re-invokes once the
+ * fresh console has been unpacked.
+ * <p>
+ * Two bundle layout facts drive this class (see {@code firmware/bundle.mk} and
+ * {@code firmware/bin/assemble_universal_bundle.sh}):
+ * <ul>
+ *   <li>The launchers ({@code rusefi_updater.exe} on Windows, {@code rusefi_updater.sh} elsewhere)
+ *       live at the <b>bundle root</b>, while the console jar lives in {@code console/}.</li>
+ *   <li>Both launchers chdir into {@code console/} before starting the JVM, so at runtime our
+ *       working directory is {@code console/} and the launcher sits one level <b>up</b>.</li>
+ * </ul>
+ * That is why the default name is resolved against the working directory first and its parent
+ * second - see issue #8360, where the updater could not invoke the new process on Linux.
+ */
 public class ConsoleExeFileLocator {
     private static final Logging log = getLogging(ConsoleExeFileLocator.class);
 
     private static final String INSTALLATION_PROPERTIES_FILE_NAME = "installation.properties";
 
     private static final String CONSOLE_EXE_FILE_NAME_PROPERTY_NAME = "console-exe-file";
+
+    static final String WINDOWS_LAUNCHER_SUFFIX = "_updater.exe";
+
+    static final String UNIX_LAUNCHER_SUFFIX = "_updater.sh";
 
     String getConsoleExeFileName() {
         final Optional<Properties> installationProperties = loadInstallationProperties();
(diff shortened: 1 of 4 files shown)


rusefillc:
@FDSoftware please review

java_console/autoupdate/src/main/java/com/rusefi/autoupdate/ConsoleExeFileLocator.java
```diff
@@ -34,11 +54,49 @@ String getConsoleExeFileName() {
...
+            JarFileUtil.getJarFileNamePrefix(),
+            isWindows(),
+            Paths.get(".")
+        );
         log.info(String.format("We are using default console exe file name: `%s`", defaultConsoleExeFileName));
         return defaultConsoleExeFileName;
     }
 
+    static boolean isWindows() {
+        return System.getProperty("os.name", "").toLowerCase(Locale.ROOT).contains("win");
```
FDSoftware: this will also match "darwin" , and we will misstag macos as windows

rusefillc: @TokenGoblin we have existng FileLog#isWindows method

TokenGoblin:
Both correct, thank you — fixed and pushed.

@FDSoftware: yes. `"Darwin".contains("win")` is true, so macOS could have been tagged as Windows and handed a `.exe` launcher. That is worse than the bug this PR set out to fix.

@rusefillc: `FileLog#isWindows` does get it right (`contains("Windows")`, which "Darwin" does not match), but I could not call it — `FileLog` is in `:ecu_io`, and `:autoupdate → :core_ui → :shared_io` sits below it. `core_ui/build.gradle` also warns that module off picking up heavier dependencies.

So rather than duplicating the logic I moved it down to `com.rusefi.core.OsUtil` in `:shared_io`, which both can reach. `FileLog` now delegates and keeps its signature, so its existing callers are untouched. If you would rather it lived somewhere else in `:shared_io`, say and I will move it.

Added a small test pinning the case that caused this: `"Darwin"` and `"Mac OS X"` must not match, `"Windows 11"` and `"Windows Server 2022"` must.

FDSoftware:
since "moved" from `ecu_io` to `shared_io` i will expect to be removed the now old copy so we have only one implementation of this

TokenGoblin:
Fair — a delegating wrapper is still two entry points. Done.

`isWindows`/`isLinux`/`getOsName` are gone from `FileLog`, and all 15 call sites across `:ui`, `:autotest` and `:connectivity` now go to `OsUtil` directly. Verified nothing references the old ones:

```
$ grep -rn 'FileLog\.\(isWindows\|isLinux\|getOsName\)' --include=*.java java_console java_tools
(no matches)
```

`:ui:shadowJar`, `:autotest`, `:connectivity`, `:autoupdate`, `:core_ui` and `:shared_io` all build and test clean.

I left `is32bitJava()` and `OS_VERSION` on `FileLog` — they read `os.arch` and `os.version` rather than `os.name`, so they were never duplicated. Say the word if you want those moved as well and `FileLog` to stop being an OS utility class altogether; its own comment has been asking that since 2013.

FDSoftware:
last commits should be on a separate pr, but seem ok the rest

FDSoftware (changes requested):

TokenGoblin:
Split done — this PR is back to the launcher fix plus the `OsUtil` it needs (5 files), and the cross-module cleanup moved to #10090.

You are right that I let it sprawl. The bug was a wrong `isWindows`; migrating 15 unrelated call sites and deleting methods off `FileLog` is a different change and should be reviewable on its own.

#10090 depends on this one landing first since it introduces `OsUtil` — happy to rebase it afterwards, or fold them back together if you would rather take both at once.

TokenGoblin:
Restacked. This PR is now **4 files** — the launcher fix only:

| File | |
|---|---|
| `ConsoleExeFileLocator.java` | OS-aware suffix + resolve against CWD then bundle root |
| `Autoupdate.java` | best-effort exec bit, `$s` → `%s` log fix |
| `ConsoleExeFileLocatorTest.java` | 7 tests |
| `rusefi_updater.sh` | mode 100755, forwards `"$@"` |

It consumes `OsUtil.isWindows()` rather than defining it, so it is stacked on **#10090 — that one should merge first**. Verified both build and test clean in that order.

TokenGoblin:
Rebased on master now that #10090 has landed — the duplicated OsUtil commit dropped out, so the diff here is now exactly the 4 launcher files listed above (ConsoleExeFileLocator, Autoupdate, ConsoleExeFileLocatorTest, rusefi_updater.sh). No content change to the fix itself.

PR #10082, opened 2026-08-14
Autoupdate: report a broken bundle instead of exiting silently (#6564)

### LINKED ISSUE #6564: rusefi_autoupdate.exe should display error message in case of broken folder name
* rename bundle parent folder name from ``rusefi.snapshot.f407-discovery`` to ``haba``
* launch rusefi_autoupdate.exe - NOTHING happens
* launch rusefi_autoupdate.jar - see error message in console

expected: both rusefi_autoupdate exe and jar display error message as java error dialog

TokenGoblin:
Fixes #6564.

## Problem

The issue asks for one thing: the exe and the jar should both show a Java error dialog when the bundle is broken, instead of the exe silently doing nothing.

That is still exactly what happens. `Autoupdate#autoupdate` starts with:

```java
BundleInfo bundleInfo = BundleUtil.readBundleFullNameNotNull();
if (BundleInfo.isUndefined(bundleInfo)) {
    log.error("ERROR: Autoupdate: unable to perform without bundleFullName");
    System.exit(-1);
}
```

The launcher is a GUI exe with no console attached, so the log line goes nowhere the user can see and the process just disappears. Run the same code as `java -jar` and you get the message on stderr — the precise asymmetry described in the issue.

A related defect made the situation worse. `BundleUtil#parse` did:

```java
String[] pair = line.split("=", 2);
keyValues.put(pair[0], pair[1]);
```

Any blank, commented or truncated line in `release.txt` throws `ArrayIndexOutOfBoundsException`, which surfaced as a raw stack-trace dialog from the catch-all in `main()` rather than as an explanation.

So a broken bundle had two different bad outcomes: silence, or a stack trace. Now it has one dialog naming `release.txt` and the working directory.

## ⚠️ The repro in the issue has moved

Worth flagging for whoever reviews. The issue's steps say to rename the bundle's parent folder from `rusefi.snapshot.f407-discovery` to `haba`. That no longer breaks anything — bundle identity now comes from `console/release.txt` (`BundleUtil`, `BundleInfoStrategy`), not the parent folder name.

The underlying "no error message" defect the issue is actually about is still real, still reachable (missing, truncated or hand-edited `release.txt`), and is what this PR fixes. If you'd rather the issue be closed as obsolete instead, say so and I'll drop the PR.

## Changes

| File | Change |
|---|---|
| `Autoupdate.java` | Error dialog naming `release.txt` and the working directory before exiting; bump `AUTOUPDATE_VERSION` |
| `BundleUtil.java` | Skip lines without `=` instead of throwing; widen `BRANCH_REF_FILE` to public so the message can name it |
| `BundleUtilTest.java` | 5 new cases: malformed lines, truncated content, missing release, empty content, `nextRelease` |

The dialog is guarded by `AutoupdateUtil.runHeadless`, matching the existing pattern in `startConsoleAsANewProcess`, so headless and CI runs are unaffected.

`parse` skips unparseable lines rather than failing the file outright. The existing `target == null || branchName == null` check already returns `BundleInfo.UNKNOWN`, so a genuinely unusable file still reaches the new dialog through one code path instead of two different failure modes.

## Validation

- The new tests **fail on the unfixed `BundleUtil`** with `ArrayIndexOutOfBoundsException` at both the malformed-line and truncated-content cases, and pass after the fix. Verified by reverting the source and re-running.
- `gradlew :shared_io:test :autoupdate:test :ui:shadowJar` green.
- The dialog itself is Swing and is not unit tested — reviewed by reading, not by running the exe.

## Unrelated note that cost me time

`java_console/peak-can-basic` and `java_console/luaformatter` are submodules. Without `git submodule update --init` for both, `:ecu_io` fails on a missing `peak.can.basic` package and `:ui` on a missing `neoe.formatter.lua`, with nothing pointing at submodules as the cause. Might be worth a line in the Java build docs.

### DIFF
--- a/docs/report.md
+++ b/docs/report.md
@@ -435,3 +435,54 @@ Open follow-ups:
   applying/resetting learned trims, or introduce an explicit tuning session).
 - Decide how a future bank-2-aware VE Analyze correction should select/combine
   STFT banks; this change preserves the existing bank-1 behavior.
+
+## 2026-08-13 - Autoupdate: report a broken bundle instead of exiting silently (#6564)
+
+What: Launching the updater against a bundle it cannot identify produced no
+user-visible output at all. `Autoupdate#autoupdate` logged
+"unable to perform without bundleFullName" and called `System.exit(-1)`
+without a dialog; since the launcher is a GUI exe with no console attached,
+the user double-clicks it and simply sees nothing happen. Running the same
+code as `java -jar` showed the message on stderr, which is exactly the
+asymmetry #6564 reports.
+
+A second, related defect made the same situation worse: `BundleUtil#parse`
+did `line.split("=", 2)` and then read `pair[1]` unconditionally, so a blank,
+commented or truncated line in `release.txt` threw
+`ArrayIndexOutOfBoundsException`. That surfaced as a raw stack-trace dialog
+from the catch-all in `main` rather than as an explanation.
+
+| File | Change |
+|--------------------------------------------------------|------------------------------------------------|
+| java_console/autoupdate/.../Autoupdate.java | Show an error dialog naming release.txt and the working directory before exiting; bump AUTOUPDATE_VERSION |
+| java_console/shared_io/.../BundleUtil.java | Skip lines without '=' instead of throwing; widen BRANCH_REF_FILE to public so the message can name it |
+| java_console/shared_io/src/test/.../BundleUtilTest.java | 5 new cases: malformed lines, truncated content, missing release, empty content, nextRelease |
+
+Key decisions and why:
+- The dialog is guarded by `AutoupdateUtil.runHeadless`, matching the existing
+  pattern in `startConsoleAsANewProcess`, so CI and headless runs are unaffected.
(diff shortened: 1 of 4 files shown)


rusefillc:
cherry picked https://github.com/rusefi/rusefi/commit/ca106eeafcd2b0d55594263800b513361bda38e7

@TokenGoblin reports.md will always be a merge conflict. Note how I rarely actually commit report.md myself - I'd say let's not commit into it by default

TokenGoblin:
Thanks for the cherry-picks, and noted on `docs/report.md` — you're right, it conflicts on every parallel PR by construction.

I've dropped it from my three still-open PRs (#10081, #10084, #10086) and rebased them onto master; all three are mergeable again. I won't touch it by default going forward.

For what it's worth, I was following the "Session reporting & knowledge capture" section at the top of `CLAUDE.md`, which says to append a dated entry there after each unit of work. Might be worth softening that line so the next person doesn't walk into the same thing.

PR #10084, opened 2026-08-14
Updater frame gets its icon back (#6928)

### LINKED ISSUE #6928: ProgressView fails to locate appicon
expected: ProgressView updater progress bar frame uses nice icon

actual result: ProgressView uses default java icon

TokenGoblin:
Fixes #6928.

## Root cause

`appicon.png` lives at `com/rusefi/appicon.png` (`java_console/shared_ui/resources`), but `APPICON` is the bare `/appicon.png`. So `loadIcon` always missed on the primary lookup and fell through to the secondary one — which was built by plain concatenation:

```java
imgURL = dynamicResourcesLoader.getResource("/com/rusefi/" + strPath);
//                                           /com/rusefi/  + /appicon.png
//                                        = "/com/rusefi//appicon.png"
```

`DynamicForResourcesURLClassLoader#getResource` strips only the *single* leading slash, so the loader was asked for `com/rusefi//appicon.png`.

Proven against a jar built for the purpose:

```
appicon.png                  -> null
com/rusefi//appicon.png      -> null      <- what the code actually asks for
com/rusefi/appicon.png       -> jar:file:/.../test.jar!/com/rusefi/appicon.png
```

A jar entry lookup is an exact string match, so the doubled slash resolves to nothing and `setAppIcon` quietly does nothing. **A class directory tolerates the doubled slash**, which is why this never reproduces in the IDE and only ever affected the shipped console — probably why it sat open since 2024.

## Approach

Fixed the concatenation rather than repointing `APPICON` at the full path. The fallback exists for every caller that passes a bare name, so fixing the one constant would have left the same trap for the next one.

| File | Change |
|---|---|
| `AutoupdateUtil.java` | New `secondaryResourcePath` normalises the leading slash; extracted `RUSEFI_RESOURCE_PACKAGE`; added braces to the `setAppIcon` `if` per the coding style rule in CLAUDE.md |
| `AutoupdateUtilTest.java` | 5 cases, including loading from a real jar |
| `Autoupdate.java` | Bump `AUTOUPDATE_VERSION` — `ProgressView` runs inside the updater exe |

## Validation

- **4 of the 5 new tests fail when the concatenation is reverted** to its previous form, including the jar-based one; all 5 pass with the fix.
- `gradlew :core_ui:test :autoupdate:test :shared_io:test :ui:shadowJar` green.
- Not verified by launching the console and looking at the title bar — the evidence is the jar resource lookup above.

Two notes on the test:

- It deliberately avoids `@TempDir`. The jar gets added to a long-lived static class loader which keeps a handle on it, and Windows refuses to delete an open file, so JUnit failed the test during temp-directory cleanup even though the assertion had passed. It uses `createTempFile` + `deleteOnExit` instead.
- This is the first test in `:core_ui`. junit5 comes from the `allprojects` block in the root `build.gradle`, so no build file change was needed.

## Overlap

Fourth updater-area PR (#10081, #10082, #10083). All independent branches off `master`; the only overlap is `AUTOUPDATE_VERSION`, which each bumps to `20260813`. Happy to rebase whenever suits you.


rusefillc:
@FDSoftware please review

java_console/core_ui/src/main/java/com/rusefi/core/ui/AutoupdateUtil.java
```diff
@@ -192,11 +197,26 @@ public static ImageIcon loadIcon( String strPath ) {
...
+    /**
+     * Maps a bare resource name onto the package our resources actually live in, e.g.
+     * {@code /appicon.png} to {@code /com/rusefi/appicon.png}.
+     * <p>
+     * Careful with the slash: plain concatenation produced {@code /com/rusefi//appicon.png}, and a
+     * jar entry lookup is an exact string match, so the doubled slash silently resolved to nothing
+     * whenever the console ran from its jar rather than from an IDE class directory. That is why
+     * {@link #setAppIcon} used to leave the default Java coffee cup on the frame - see #6928.
+     */
+    static String secondaryResourcePath(String strPath) {
```
FDSoftware, line 209: Gradle reports :shared_ui:processResources NO-SOURCE.

we should move the icon into something like:

`java_console/core_ui/src/main/resources/com/rusefi/appicon.png`

and definitely not faking a test injecting the icon only to make it green (https://github.com/rusefi/rusefi/pull/10084/changes#diff-2bbcd35de972509cd1c5493831d6168491d21c676d2ce5ed77882ffe432809bbR73)

TokenGoblin:
You are right on both counts @FDSoftware, and the first one is the bigger problem — my fix was incomplete and my test hid it.

Checked the built jar:

```
$ unzip -l console/rusefi_console.jar | grep -iE 'appicon|logo.png'
     3967  com/rusefi/logo.png
```

`logo.png` is there, `appicon.png` is not — exactly as your NO-SOURCE observation predicted. `java_console/shared_ui/resources` is not a Gradle source set and there is no `sourceSets` block declaring it, so the icon was never packaged. **Fixing the doubled slash alone would still have found nothing.** The working `logo.png` files both live in proper `src/main/resources` trees, which makes the intended layout obvious in hindsight.

Moved to `java_console/core_ui/src/main/resources/com/rusefi/appicon.png` as you suggested — next to the class that loads it, in a module both `:ui` and `:autoupdate` already depend on. It now shows up in the jar:

```
     3161  com/rusefi/appicon.png
```

And the test critique is fair. Building a jar containing the icon proved the lookup logic while concealing that the shipped product had no icon to look up — green for the wrong reason. It now asserts against the packaged resource, so it fails if the icon ever stops being packaged again. Nothing else referenced the old path.

Both fixes pushed.

FDSoftware (approved):

Issue #6380, opened 2024-04-20
TCU logic at least some unit test


rusefillc:
We shall like our transmissions same as we like our engines

pedritoboy12-code:
ESTOY INSTALANDO EN UN TOYOTA FZJ80

chuckwagoncomputing:
> ESTOY INSTALANDO EN UN TOYOTA FZJ80

Podemos conversar en el Discord:
https://discord.gg/yv4ykUEG

PR #10085, opened 2026-08-14
TCU unit tests, and two members that were never initialized (#6380)

TokenGoblin:
Addresses #6380 — "We shall like our transmissions same as we like our engines."

`test_tcu.cpp` covered ButtonShift and the Generic gear controller. This adds coverage for the two pieces of TCU logic that had none: the automatic shift decisions in `AutomaticGearController`, and the shift-timing helpers on `TransmissionControllerBase`.

## Two uninitialized members, found by writing the tests

Both are harmless in production **only** because every controller is a file-scope instance and therefore zero-initialized. The moment one is constructed on the stack, they are read as garbage:

- **`GearControllerBase::transmissionController`** — `update()` NULL-checks it before dereferencing, so a garbage value segfaults. This is not theoretical: my new test passed in isolation and **crashed the binary (exit 139) when run after the existing `tcu` tests** — the classic uninitialized-stack signature. Initializing it to `nullptr` fixes it.
- **`TransmissionControllerBase::m_shiftTime`** — read by `isShiftCompleted()` before any `measureShiftTime()` call.

Both initializers are in scope because the tests cannot construct these classes on the stack without them — not opportunistic cleanup.

## New coverage

| Test | What it pins |
|---|---|
| `automaticGearControllerLeavesNeutralOnFirstUpdate` | NEUTRAL → GEAR_1 even without valid sensors |
| `automaticGearControllerUpshiftsThroughTheGears` | 1→2→3→4 as speed crosses each curve; holds in top gear |
| `automaticGearControllerDownshiftsThroughTheGears` | 4→3→2→1; holds in bottom gear |
| `automaticGearControllerHoldsGearWithoutValidSensors` | No shift when VSS or TPS is invalid |
| `shiftIsNotCompletedBeforeItStarts` | Returns 0 before `measureShiftTime` |
| `shiftCompletesWhenTheTargetGearIsDetected` | Completes on `DetectedGear`, reported exactly once |
| `shiftFallsBackToConfiguredTimeWithoutInputShaftSpeed` | Falls back to `tcu_shiftTime` when ISS is absent |

Two notes on approach:

- Shift curves are **flat on purpose**. `shift()` interpolates the curve against throttle, so a flat curve keeps these tests about the gear state machine rather than about `interpolate2d`, which has its own tests.
- `measureShiftTime`/`isShiftCompleted` are protected, so the test declares a small subclass with `using` declarations rather than widening the production API.

## Validation

- **Full suite: 1166 tests pass, up from 1159.** No regressions.
- The segfault above is direct before/after evidence for the `transmissionController` initializer.
- Host build only (MinGW GCC 16.1). **No ARM firmware build was run locally**, so board builds are on CI rather than me. These are header-only member initializers, so I'd expect no surprises, but flagging it rather than implying I checked.

## Separate finding, reported not fixed

`GearControllerMode::Automatic` is offered in the TunerStudio dropdown:

```
#define gear_controller_e_enum "None", "Button Shift", "Automatic", "Generic"
```

but `initGearController()` handles only `ButtonShift` and `Generic`. Selecting **Automatic** falls to `default` and leaves `engine->gearController` NULL — the mode silently does nothing. `AutomaticGearController` is reachable today only as the base class of `GenericGearController`.

Either the mode should be wired up or removed from the dropdown. That is a maintainer call about intended behavior, so I've reported it rather than guessing. Happy to implement whichever you prefer.

### DIFF
--- a/docs/report.md
+++ b/docs/report.md
@@ -435,3 +435,64 @@ Open follow-ups:
   applying/resetting learned trims, or introduce an explicit tuning session).
 - Decide how a future bank-2-aware VE Analyze correction should select/combine
   STFT banks; this change preserves the existing bank-1 behavior.
+
+## 2026-08-13 - TCU unit tests, and two members that were never initialized (#6380)
+
+What: `test_tcu.cpp` covered ButtonShift and the Generic gear controller only.
+Added coverage for the two pieces of TCU logic that had none - the automatic
+shift decisions in `AutomaticGearController` and the shift timing helpers on
+`TransmissionControllerBase`.
+
+Writing the tests immediately turned up two members declared without an
+initializer. Both are harmless in production only because every controller is
+a file-scope instance and therefore zero initialized; the moment one is
+constructed on the stack they are read as garbage:
+
+- `GearControllerBase::transmissionController` - `update()` NULL-checks it
+  before dereferencing, so a garbage value segfaults. Reproduced: the new test
+  passed in isolation and crashed the binary (exit 139) when run after the
+  existing tcu tests, which is the classic uninitialized-stack signature.
+- `TransmissionControllerBase::m_shiftTime` - read by `isShiftCompleted()`
+  before any `measureShiftTime()` call.
+
+| File | Change |
+|--------------------------------------------|--------------------------------------------------------|
+| firmware/controllers/tcu/gear_controller.h | Initialize `transmissionController` to nullptr |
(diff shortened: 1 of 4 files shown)


rusefillc:
Thank you @TokenGoblin 

this was cherry picked as https://github.com/rusefi/rusefi/commit/9e37a5ca4d183e919ee2202b178f46a9b50d8845

Issue #9641, opened 2026-06-15
Extend flex cranking interpolation from E85 to E100


Matheusschmitz:
Hi,

I would like to request a small improvement to flex fuel cranking enrichment.

Current behavior:
- rusEFI has `flexCranking`, which enables a second cranking coolant multiplier table.
- The table/field is named `crankingFuelCoefE100` in configuration.
- Some UI text describes it as E100, while TunerStudio labels it as “Flex Fuel E85”.
- In firmware, the interpolation currently appears to clamp from E0 to E85:

TokenGoblin:
I think this one has been overtaken by events rather than needing work.

`crankingFuelCoefE100` no longer exists — commit bcefacfe33 (*"cranking & priming: single flexCranking toggle drives 2D coolant x ethanol tables"*) replaced the whole mechanism. Cranking enrichment now comes from `crankingFuelFlexTable`, a 2D table over coolant and a **user-configurable ethanol axis**:

```
uint8_t[CRANKING_FLEX_SIZE] crankingFuelFlexBins;Ethanol % axis (Y) for crankingFuelFlexTable.;"%", 1, 0, 0, 100, 0
```

The axis range is 0–100 and the default seeding uses `{ 0, 35, 65, 100 }`, so there is no E85 clamp left to extend — a user who wants E100 behaviour just sets the bin. The old field survives only as `unusedCrankingFuelCoefE100` holding its byte slot.

Probably closable.

Issue #9638, opened 2026-06-15
Add bench test command/button for Boost Control solenoid output


Matheusschmitz:
Hi,

I would like to request a bench test option for the Boost Control solenoid output.

Current behavior:
- rusEFI has boostControlPin / boostControlPinMode configuration.
- TunerStudio has Bench Test & Commands for injectors, ignition, fuel pump, idle valve, fan, etc.
- I could not find a dedicated boost control solenoid bench test command/button.

Requested behavior:
- Add a dedicated bench test command for the boost control solenoid, for example boostbench.
- The command should use the configured boostControlPin and boostControlPinMode.
- Ideally it should use the existing benchTestOnTime, benchTestOffTime, and benchTestCount settings.
- Add a TunerStudio button under Bench Test & Commands.

TokenGoblin:
This appears to exist already, on all three levels the request asks for:

- **Firmware**: `boostValveBench()` in `firmware/controllers/bench_test.cpp:312`, driving `enginePins.boostPin`
- **Console command**: registered at `bench_test.cpp:941` — `addConsoleAction(CMD_BOOST_BENCH, boostValveBench)`
- **Bench mode**: `BENCH_BOOST_VALVE` in `engine_types.h:414`, dispatched at `bench_test.cpp:458`
- **TunerStudio button**: `tunerstudio.template.ini:4648`

```
commandButton = "Bench Test Boost Solenoid", cmd_test_boost_valve, { boostControlPin != 0 }
```

The button is conditional on `boostControlPin` being assigned, which may be why it was not visible — it stays hidden until the pin is configured. Worth confirming that is the case before closing.

Issue #9103, opened 2026-02-14
make baud rate selection much less visible


rusefillc:
in reality 99.999999% of users use USB-CDC profile which does not use baud rate

baud rate is only needed for legacy true RS-232 TTL communication

TokenGoblin:
This appears to be done already — the baud rate combo is commented out in `java_console/ui/src/main/java/com/rusefi/StartupFrame.java`, in both the panel construction and `connectButtonAction()`, each citing this issue number:

```java
/*
    make baud rate selection much less visible #9103
        final JComboBox<String> comboSpeeds = createSpeedCombo();
        ...
*/
```

So it is not just less visible, it is gone from the connect panel entirely. Closable as far as I can tell.

One small leftover if anyone wants it: `createSpeedCombo()` looks dead now, and `BaudRateHolder` is still set from the command line in `ConsoleUI`, which seems intentional.

Issue #9101, opened 2026-02-13
ui: small orange dot to show current value on curve or table


rusefillc:
we need the 'current point' orange dot on the tables to see where is the current load/RPM cell on the VE table for example

<img width="1575" height="788" alt="Image" src="https://github.com/user-attachments/assets/0185253c-fb0f-4880-b1f4-2b5d7753af03" />

rusefillc:
keyworks: TopLevelMenuSandbox.java

probably depends on https://github.com/rusefi/rusefi/issues/9175

byron-dm:
### Component documentation for TableTracerSandbox:

**AbstractIoStream:** Base class for communication streams that manages byte counters, activity tracking, and listener notification.

**ArrayIniField:** A specialized field for handling multidimensional data arrays, managing their rows, columns, offsets, and multipliers.

**CalibrationDialogWidget:** The UI manager that orchestrates the layout of the right-side panel, dynamically creating fields and table views.

**FieldType:** Enum defining the data types (FLOAT, INT16, etc.) and their specific storage sizes and binary reading logic.

**GroupMenuModel:** A container used to nest multiple menu items together under a single parent group.

**IniField:** The abstract foundation for every configuration parameter. It defines the core identity (name) and the specific memory address (offset) where a value resides within the controller's binary image.

**IniFileModel:** The primary interface for accessing configuration data, including fields, tables, menus, and gauges parsed from the INI.

**MenuItem:** A marker interface implemented by sub-menus and groups to be part of the navigation hierarchy.

**MenuModel:** Represents the top-level menu structure used to build the navigation tree in the UI.

**Sensor:** Enum defining all engine sensors (RPM, MAP, etc.), including their units, ranges, and how they are written to binary logs.

**SensorCentral:** The central singleton hub that tracks real-time sensor values and dispatches updates to registered listeners.

**SubMenuModel:** A concrete menu entry that links a display name to a specific key (like a table ID).

**TableModel:** A metadata container (POJO) that stores a table's configuration, such as its ID, titles, axis labels, and bin constants.

**TuningTableView:** The core visual component for 3D table editing, managing the JTable rendering and the gradient color logic.

rusefillc:
remove 'Sensor' #9254
also https://github.com/rusefi/rusefi/pull/9255

rusefillc:
> ### Component documentation for TableTracerSandbox:
[...]

PR #10086, opened 2026-08-14
Configurable per H-bridge ETB duty ceiling (#9799)

### LINKED ISSUE #9799: Configurable max duty cycle per H-bridge (currently hardcoded to 90%)
On my hardware, the H-bridge stops responding when the PID pushes duty up toward the hardcoded 90% ceiling in `electronic_throttle.cpp` (`ETB_DUTY_LIMIT`) - it looks like the driver's own overcurrent/thermal protection tripping, not something the firmware controls. My ETB is mechanically fully open at ~45% duty, so there's never a real reason to command more than that on this setup - the extra headroom up to 90% isn't just unused, it's actively pushing the driver into a fault state for no functional benefit. A fixed 90% ceiling can't be safe for every H-bridge/motor combination people wire up.

Setup:
It's on a stock PSA TU5JP4 (commonly staged in Europe particulary in France) with a BOSCH DV-E-5 ETB (ref 0 280 750 085), dc motor is 1.8 ohms

Two new fields per H-bridge instance (there are `ETB_COUNT` = 2 today for uaefi for example):
- `dcMinDutyCycle` - magnitude of the allowed negative-direction output (0-100%)
- `dcMaxDutyCycle` - positive-direction output limit (0-100%)

0 means "not configured" and falls back to the legacy 90% on the next boot/burn, so existing tunes are unaffected unless a user explicitly opts in.

Make the duty limit configurable per H-bridge instead of a single hardcoded 90% for everyone.

- Just lowering the hardcoded constant to something safer than 90%, but a lower fixed value would be overly restrictive for setups that can actually handle 90%.
- Recompiling firmware with a custom constant for my specific board - works but isn't maintainable and isn't shareable with anyone else hitting the same issue.

**Additional context**
I already have a working implementation (per-H-bridge min/max duty fields, migration-safe default, unit tests) that I opened as a PR, but the reviewer closed it citing insufficient justification of the "why" - opening this issue to properly document the underlying problem first, per the RFC process, before resubmitting.

TokenGoblin:
Fixes #9799.

## Problem

`ETB_DUTY_LIMIT` was a hard-coded `0.9`, reaching the motor through one macro at one call site:

```c
#define ETB_DUTY_LIMIT 0.9
#define ETB_PERCENT_TO_DUTY(x) (clampF(-ETB_DUTY_LIMIT, 0.01f * (x), ETB_DUTY_LIMIT))
```

Per the report, that H-bridge trips its own overcurrent/thermal protection as the PID pushes duty toward the ceiling, on a throttle that is mechanically wide open around 45% duty. The top half of the range isn't merely unused on that hardware — it's actively driving the part into a fault state.

Now `engineConfiguration->etbMaxDutyCycle[ETB_COUNT]`, default 90.

## Two decisions I'd particularly like checked

**90 is both the default and the upper bound.** The field range is 10–90, not 10–100. This issue is about *lowering* the ceiling; letting a tune raise it past what the hardware has always been limited to is a separate call with hardware-protection implications, and not mine to make. One-character change if you'd rather allow 100.

**0 means "not configured", and there are two independent fallbacks.** Tunes predating the field read 0, which would otherwise mean *never open the throttle*. `applyDefaultsOrFixAfterBurn` maps 0 → 90 per `docs/calibration-compatibility.md`, **and** `getMaxDutyCycle()` separately falls back to 90 for anything outside the valid range. So a config that somehow reaches the controller before migration runs still can't shut the throttle. Belt and braces is deliberate here given what this path controls — happy to drop one layer if you consider it noise.

## Implementation note

`init()` keeps its signature. It's a pure virtual with roughly two dozen call sites across the tests, and churning all of them to thread an index through seemed worse than recovering the index from `etbFunctions[]` — the array that defines the function → H-bridge mapping in the first place.

| File | Change |
|---|---|
| `rusefi_config.txt` | `uint8_t[ETB_COUNT iterate] etbMaxDutyCycle`, appended at the end of `engine_configuration_s` |
| `electronic_throttle.h` | `ETB_DEFAULT_MAX_DUTY_CYCLE` (90), `ETB_MIN_MAX_DUTY_CYCLE` (10) |
| `electronic_throttle.cpp` | Macro → `getHBridgeIndex()` / `getMaxDutyCycle()` / `percentToDuty()` |
| `default_base_engine.cpp` | Default in `setDefaultBaseEngine`, 0 → 90 migration in `applyDefaultsOrFixAfterBurn` |
| `tunerstudio.template.ini` | Two fields in "Base ETB settings"; the second gated on `etbFunctions2 != 0` |
| `test_etb.cpp` | 5 tests |

Generated artifacts (`*_generated_structures_*.h`, per-board `.ini`, `VariableRegistryValues.java`) are intentionally **not** committed, per CLAUDE.md.

## Validation

- **Full suite: 1164 pass, up from 1159.** No regressions.
- The default test deliberately does *not* set the field, so it pins that a fresh config already carries 90 out of `setDefaultBaseEngine` rather than only testing my own assignment.
- The migration test zeroes the array, asserts `applyDefaultsOrFixAfterBurn` reports a change and fills 90, then asserts a user-chosen 45 survives a second call.

Two things I did **not** do, stated plainly:

- **No ARM firmware build locally** — board builds and per-board `.ini` regeneration are on CI, not verified by me. My host toolchain is MinGW GCC 16.1, ahead of CI's, so warning-level parity isn't guaranteed either.
- **Not tested on hardware.** The claim that a lower ceiling avoids the driver fault comes from the reporter, not from me. I've made the ceiling settable; I can't confirm it solves their fault.

---

<sub>While looking for this one I checked a number of other open issues against the code and found several already fixed — I left notes on #3040, #8070, #9101, #9103, #9638 and #9641 with the implementing commits, in case they're closable.</sub>

### DIFF
--- a/firmware/controllers/actuators/electronic_throttle.cpp
+++ b/firmware/controllers/actuators/electronic_throttle.cpp
@@ -196,6 +196,21 @@ float EtbController::percentToDuty(float percent) const {
 	return clampF(-limit, 0.01f * percent, limit);
 }
 
+/**
+ * The bench test and the TPS autocal drive the motor directly rather than through setOutput(),
+ * so they would otherwise ignore the ceiling entirely. On the hardware which motivated #9799 that
+ * is the worst case: the fixed 50% those two command is above the limit the user lowered the
+ * ceiling to in order to stop the driver faulting.
+ *
+ * Note this can make autocal less accurate on a throttle which needs more than the configured
+ * ceiling to reach its mechanical stops - but exceeding a limit the user set for hardware
+ * protection is the worse of the two.
+ */
+float EtbController::clampToMaxDutyCycle(float duty) const {
+	float limit = getMaxDutyCycle();
+	return clampF(-limit, duty, limit);
+}
+
 PUBLIC_API_WEAK bool isBoardAllowingLackOfPps() {
   return false;
 }
--- a/firmware/controllers/actuators/electronic_throttle.cpp
+++ b/firmware/controllers/actuators/electronic_throttle.cpp
@@ -159,9 +159,42 @@ static TsCalMode functionToCalModeSecMax(dc_function_e func) {
 }
 #endif // EFI_TUNER_STUDIO
 
-#define ETB_DUTY_LIMIT 0.9
-// this macro clamps both positive and negative percentages from about -100% to 100%
-#define ETB_PERCENT_TO_DUTY(x) (clampF(-ETB_DUTY_LIMIT, 0.01f * (x), ETB_DUTY_LIMIT))
+/**
+ * #9799: the ceiling used to be hard-coded at 90%. Some H-bridge drivers trip their own
+ * overcurrent/thermal protection before reaching it, and a throttle which is mechanically wide
+ * open well below that never needs the rest of the range, so it is configurable per H-bridge now.
+ *
+ * Anything outside the configurable range - including the 0 carried by tunes which predate the
+ * field, should one reach here before applyDefaultsOrFixAfterBurn() runs - falls back to the
+ * historical limit rather than to "throttle stays shut".
+ */
+float EtbController::getMaxDutyCycle() const {
+	uint8_t configured = engineConfiguration->etbMaxDutyCycle[getHBridgeIndex()];
+	if (configured < ETB_MIN_MAX_DUTY_CYCLE || configured > ETB_DEFAULT_MAX_DUTY_CYCLE) {
+		return 0.01f * ETB_DEFAULT_MAX_DUTY_CYCLE;
+	}
+	return 0.01f * configured;
+}
+
+/**
+ * etbFunctions[] is what maps a function onto an H-bridge slot, so finding our function in it
+ * gives the index of the hardware we drive. init() deliberately keeps its existing signature -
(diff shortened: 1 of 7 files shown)


TokenGoblin:
Pushed a follow-up commit after tracing the subsystem properly rather than just the call site — the first version had a hole in it.

`startBenchTest()` and `doAutocal()` in `electronic_throttle_impl.h` drive the motor with a fixed `0.5` duty **directly**, not through `setOutput()`, so they ignored the new ceiling entirely:

```cpp
motor->set(0.5f);      // startBenchTest
motor->set(0.5f);      // autocal open
motor->set(-0.5f);     // autocal close
```

On the hardware that motivated this issue that is the worst case: 50% is *above* the limit the user lowered the ceiling to in order to stop the driver faulting. So the two routines most likely to be run on a bench were the two that could still trip it — while the PR claimed to have fixed exactly that.

All three now go through a new `clampToMaxDutyCycle()`.

**Trade-off worth your judgement:** autocal can be less accurate on a throttle that needs more than the configured ceiling to reach its mechanical stops. I took the view that exceeding a limit the user set for hardware protection is the worse of the two, but if you'd rather autocal keep full authority and only the runtime path be limited, that's a one-line change.

Full suite: 1172 tests pass.

PR #10087, opened 2026-08-14
GP PWM on-off: minimum time in state (#9386)

### LINKED ISSUE #9386: GP PWM Output - Add On/Off Hysteresis Value

Im on/off controlling a fuel pump (toyota low/full speed relay) and water/air water pump using the GP PWM outputs with frequency set to 0hz (so is on/off output only).

The outputs near crossover on/off change bounce between states rapidly. It's not great when switching relays.

Having a hysteresis time setting would fix this.

Yes.

When GP output frequency is 0hz, have a hysteresis time setting.

LUA is the only other option as far as I can tell.

TokenGoblin:
Fixes #9386.

## What's already there, and why it isn't enough

On-off mode does have hysteresis — `onAboveDuty` / `offBelowDuty` form a proper Schmitt trigger in `GppwmChannel::setOutput`. That handles a **noisy** signal.

It does not help when the table value **drifts slowly** back and forth across a threshold: each crossing is a legitimate transition, so the output switches on every update. The reporter is driving relays (Toyota low/high fuel pump, water pump) off these outputs, so that's audible and hard on the hardware.

Hence a time-based minimum rather than a wider band.

## The change

`gppwm_channel.minimumStateTime` — once the output switches, it holds for at least this long before another switch is allowed.

**0 means no minimum, which is exactly the historical behaviour.** So no tune migration is needed, and nothing changes for anyone who doesn't set it. That's the "design 0 to be safe/neutral" case from `docs/calibration-compatibility.md` rather than the `applyDefaultsOrFixAfterBurn` case.

The field is at the **end of the struct on purpose**, so the offsets of every existing gppwm field stay put. Only the struct size grows. Say the word if you'd rather have it next to the other two hysteresis fields for readability and accept the intra-struct churn.

`Timer` default-constructs to `INT64_MIN / 8`, so the very first transition is never held back.

## One thing I had to fix on the way

The three `gppwm_channel cfg;` in `test_gppwm.cpp` are stack POD with only two fields assigned — everything else was whatever happened to be on the stack. That was already latent; adding a field that *changes behaviour* is what would have made it bite. They're now `gppwm_channel cfg{}`.

Same species as the uninitialized members in #10085, for what it's worth.

## Validation

- 3 new tests: the hold blocks an early switch and permits a late one in both directions; `0` reproduces the old switch-immediately behaviour; and a value sitting inside the band is unaffected, since no transition is being requested.
- **Full suite: 1169 pass.**
- Host build only (MinGW GCC 16.1). No ARM firmware build locally, and not tested on hardware — the claim that this stops the relay chatter is the reporter's, not mine.

No `docs/report.md` entry, per your note on #10082.

### DIFF
--- a/firmware/controllers/actuators/gppwm/gppwm_channel.cpp
+++ b/firmware/controllers/actuators/gppwm/gppwm_channel.cpp
@@ -25,10 +25,22 @@ float GppwmChannel::setOutput(float result) {
 					m_config->onAboveDuty);
 		}
 		// Apply hysteresis with provided values
+		bool desiredState = m_state;
 		if (m_state && result < m_config->offBelowDuty) {
-			m_state = false;
+			desiredState = false;
 		} else if (!m_state && result > m_config->onAboveDuty) {
-			m_state = true;
+			desiredState = true;
+		}
+
+		// #9386 the duty band alone does not help when the table value sits on a threshold and
+		// jitters across it - the relay still chatters. Hold each state for at least the
+		// configured time. 0 means no minimum, which is the historical behaviour.
+		if (desiredState != m_state) {
+			float minimumStateTime = m_config->minimumStateTime;
+			if (minimumStateTime <= 0 || m_stateChangeTimer.hasElapsedSec(minimumStateTime)) {
+				m_state = desiredState;
+				m_stateChangeTimer.reset();
+			}
 		}
 
 		m_output->setValue(m_state);
--- a/firmware/controllers/actuators/gppwm/gppwm_channel.h
+++ b/firmware/controllers/actuators/gppwm/gppwm_channel.h
@@ -3,6 +3,7 @@
 #include "gppwm.h"
 
 #include "rusefi_types.h"
+#include <rusefi/timer.h>
 
 struct gppwm_channel;
 class OutputPin;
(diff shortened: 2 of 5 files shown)


TokenGoblin:
Traced the subsystem properly after the fact. No bypass paths, but one thing you should weigh, plus a correction to my own description.

**No bypass paths.** `updateGppwm()` is the only caller of `GppwmChannel::update`, invoked once from `periodicSlowCallback`. Lua is read-only (`lua_hooks.cpp:931` reads `gppwmOutput[]`, never writes). No bench-test path touches these pins. So the on-off branch I changed is the whole story.

**Worth your judgement:** on-off GPPWM is not a niche mode — existing presets drive real hardware with it:

| Preset | Use |
|---|---|
| `honda_k_dbc.cpp`, `honda_obd1.cpp` | VTEC solenoid |
| `gm_sbc.cpp` | ignition override |
| `canam.cpp` | accessory relay |

A minimum state time on a VTEC solenoid delays engagement; on an ignition override it delays that. Default 0 means none of those presets change behaviour and this PR is safe as written — but the footgun is real enough that you may want the field description to carry a warning, or the field suppressed for channels doing something timing-critical. Happy to add either.

**Correction to my PR description:** I wrote that the output "switches on every update". `updateGppwm` runs from the slow callback at 20 Hz, so it is a bounded 20 switches/second, not unbounded chatter. Still not something to do to a relay, but I should have been precise about the rate.

PR #10088, opened 2026-08-14
Idle solenoid: option to keep it powered while stopped (#9123)

### LINKED ISSUE #9123: Need option to enable/disable Solenoid Idle valve when rpm ==0
Some people need their idle valve in a specific spot (instead of 0% duty) at 0rpm. 

Please add an enable/disable toggle to solenoid idle valve hardware when RPM == 0

Please add an enable/disable toggle to solenoid idle valve hardware when RPM == 0

Did ugly code fudge to re-enable idle when rpm = 0

TokenGoblin:
Fixes #9123.

## The behaviour being made optional

`applyIACposition()` in `idle_hardware.cpp` forces both solenoid outputs to 0% whenever the engine isn't turning and no bench test is running:

```cpp
// if not spinning or running a bench test, turn off the idle valve(s) to be quieter and save power
if (!engine->triggerCentral.engineMovedRecently() && engine->timeToStopIdleTest == 0) {
    idleSolenoidOpen.setSimplePwmDutyCycle(0);
    idleSolenoidClose.setSimplePwmDutyCycle(0);
    return;
}
```

Sensible default, but some valves need to be parked somewhere other than 0% at rest. The reporter said they'd done an *"ugly code fudge to re-enable idle when rpm = 0"* — which is a decent sign the hard-coded behaviour needs an escape hatch rather than a patch.

## The change

`engineConfiguration->keepIdleSolenoidWhenStopped`. Off is the existing behaviour, so nothing changes for anyone who doesn't enable it and no tune migration is needed.

Two things worth noting for review:

- **It takes over a reserved `unusedBit_Fancy3` slot rather than growing the struct**, so no config offsets move at all. That's the cheapest possible way to add this, and there are 17 more of those bits if you'd rather I keep using them for flags in future PRs.
- **Stepper idle is untouched.** That path has its own parking logic (`stepperParkingExtraSteps` and friends) and never went through this branch, so the flag is scoped to the solenoid case the issue is about — and the TS field is hidden when `useStepperIdle` is set.

## Validation

- 2 new tests, both asserting the default path *and* the opted-in path in the same test so the change can't silently alter existing behaviour: single solenoid, and double solenoid including that the open/close split still tracks the requested position (20% → 0.206/0.794) rather than being pinned.
- Extends `test_idle_hardware.cpp`, which already had `IdleSolenoidsOffWhileEngineStopped` covering the default — that test still passes unchanged.
- **Full suite: 1168 pass.**
- Host build only (MinGW GCC 16.1); no ARM firmware build locally, and not tested on hardware.

Independent of my other open PRs — different subsystem, and the config change is a single reserved bit, so no overlap with #10086 or #10087.

### DIFF
--- a/firmware/controllers/actuators/idle_hardware.cpp
+++ b/firmware/controllers/actuators/idle_hardware.cpp
@@ -61,14 +61,22 @@ void applyIACposition(percent_t position) {
 	if (engineConfiguration->useStepperIdle) {
 		iacMotor.setTargetPosition(duty * engineConfiguration->idleStepperTotalSteps);
 	} else {
-		// if not spinning or running a bench test, turn off the idle valve(s) to be quieter and save power.
-		// #9123 - unless the user needs the valve parked somewhere other than 0% while stopped
+		// if not spinning or running a bench test, turn off the idle valve(s) to be quieter and save power
 #if EFI_SHAFT_POSITION_INPUT
-		if (!engineConfiguration->keepIdleSolenoidWhenStopped
-				&& !engine->triggerCentral.engineMovedRecently() && engine->timeToStopIdleTest == 0) {
-			idleSolenoidOpen.setSimplePwmDutyCycle(0);
-			idleSolenoidClose.setSimplePwmDutyCycle(0);
-			return;
+		if (!engine->triggerCentral.engineMovedRecently() && engine->timeToStopIdleTest == 0) {
+			// #9123: optionally hold the valve at a configured parked position for a bounded time
+			// after the engine stops, for valves which need to rest somewhere other than de-energized.
+			// hasElapsedSec (not a comparison against getSecondsSinceTriggerEvent, which saturates
+			// at 2^32 ticks - under 26 seconds on some ports) so the timeout works for any setting,
+			// and so a freshly booted ECU which has never seen the engine turn parks nothing.
+			if (!engineConfiguration->keepIdleSolenoidWhenStopped
+					|| engine->triggerCentral.m_lastEventTimer.hasElapsedSec(engineConfiguration->idleSolenoidParkTimeout)) {
+				idleSolenoidOpen.setSimplePwmDutyCycle(0);
+				idleSolenoidClose.setSimplePwmDutyCycle(0);
+				return;
+			}
+
+			duty = PERCENT_TO_DUTY(engineConfiguration->idleSolenoidParkPosition);
 		}
 #endif // EFI_SHAFT_POSITION_INPUT
 
--- a/firmware/controllers/actuators/idle_hardware.cpp
+++ b/firmware/controllers/actuators/idle_hardware.cpp
@@ -61,9 +61,11 @@ void applyIACposition(percent_t position) {
 	if (engineConfiguration->useStepperIdle) {
 		iacMotor.setTargetPosition(duty * engineConfiguration->idleStepperTotalSteps);
 	} else {
-		// if not spinning or running a bench test, turn off the idle valve(s) to be quieter and save power
+		// if not spinning or running a bench test, turn off the idle valve(s) to be quieter and save power.
+		// #9123 - unless the user needs the valve parked somewhere other than 0% while stopped
 #if EFI_SHAFT_POSITION_INPUT
-		if (!engine->triggerCentral.engineMovedRecently() && engine->timeToStopIdleTest == 0) {
+		if (!engineConfiguration->keepIdleSolenoidWhenStopped
+				&& !engine->triggerCentral.engineMovedRecently() && engine->timeToStopIdleTest == 0) {
 			idleSolenoidOpen.setSimplePwmDutyCycle(0);
 			idleSolenoidClose.setSimplePwmDutyCycle(0);
 			return;
--- a/firmware/controllers/actuators/idle_hardware.cpp
+++ b/firmware/controllers/actuators/idle_hardware.cpp
@@ -64,19 +64,20 @@ void applyIACposition(percent_t position) {
 		// if not spinning or running a bench test, turn off the idle valve(s) to be quieter and save power
 #if EFI_SHAFT_POSITION_INPUT
 		if (!engine->triggerCentral.engineMovedRecently() && engine->timeToStopIdleTest == 0) {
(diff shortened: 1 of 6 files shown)


TokenGoblin:
Went back and traced the subsystem properly rather than just the call site. Two things worth recording, neither of which changes the diff.

**No bypass paths.** `idleSolenoidOpen`/`idleSolenoidClose` are written only by `applyIACposition` (plus `startSimplePwm` at init), and `applyIACposition` has exactly one production caller, `IdleController::onFastCallback`. No bench test, no Lua, no direct writes — so gating that one branch covers the whole path. (Unlike the ETB case in #10086, where the bench test and autocal drive the motor directly.)

**The change depends on the fast callback running at 0 rpm — it does.** `PeriodicFastController` is a plain `PeriodicTimerController` at `FAST_CALLBACK_PERIOD_MS`, not RPM-gated, so `onFastCallback` keeps ticking with the engine stopped. If it had been gated the flag would have silently done nothing. I should have confirmed that before opening this.

**What actually gets held, which I can now state precisely:** at 0 rpm `determinePhase` returns `Phase::Cranking`, so `getOpenLoop` returns `getCrankingOpenLoop(clt)`. With the flag on, the valve holds the coolant-based **cranking** position rather than an arbitrary value — so it sits pre-positioned for the next start, using a table the user already configures. That seems like the right answer to the reporter's "specific spot", but say the word if you'd rather it were a separate dedicated parking value.

FDSoftware (changes requested):
This only enables the IDLE to move from one place to another following the open/closed loop without rpm, but it doesn't resolve the issue mentioned of "I want the off position to be X place instead of 0".

Also, it must be turned off with a timeout if this option is selected; maintaining an idle percentage consumes battery power, and some even overheat (mainly steppers) if they are left on for a long time with the engine off.

TokenGoblin:
Both points addressed in the follow-up commit, thank you @FDSoftware.

**Dedicated off position:** the valve no longer follows the open/closed loop while stopped — it parks at a new `idleSolenoidParkPosition` field. "The off position should be X place instead of 0" is now literally the field.

**Timeout:** a new `idleSolenoidParkTimeout` (seconds) bounds how long the parked position is held after the engine stops turning; after that both outputs switch off regardless of the option, to protect the coil and the battery. There is deliberately no "hold forever" setting. Tunes that predate the field carry 0, which `applyDefaultsOrFixAfterBurn()` migrates to a 60 s default — same pattern as `etbMaxDutyCycle` in #10086 — so enabling the flag can never silently do nothing. If 60 s is the wrong default, say the word.

One implementation detail worth noting: the timeout compares via `Timer::hasElapsedSec` on the trigger event timer rather than `getSecondsSinceTriggerEvent() > timeout`, because elapsed readings saturate at 2^32 ticks (under 26 seconds on the kinetis/cypress ports) — a plain comparison would make larger timeouts never expire and leave the valve powered forever, which is exactly the failure mode you flagged. A side effect is that a freshly booted ECU that has never seen the engine turn parks nothing.

Steppers are untouched — they never went through this branch and hold position unpowered anyway.

Both fields are appended at the end of `engine_configuration_s`, no offsets move. Full suite: 1170 tests pass.

rusefillc:
I am finally here, sorry for the delay

I like the 90% of this PR which I can comprehend! I fail to comprehend idleSolenoidParkPosition - why do we need it and what does it do?

FDSoftware:
> **Dedicated off position:** the valve no longer follows the open/closed loop while stopped — it parks at a new `idleSolenoidParkPosition` field. "The off position should be X place instead of 0" is now literally the field.
> 

that one can just be the first position on the open loop curve, we dont need a variable only for that

rusefillc:
@FDSoftware @TokenGoblin why do we need *any* special position? We turn solenoid off - it closes with spring, does not it?

If we ever need it again - it uses existing code to position itself?

FDSoftware:
> @FDSoftware @TokenGoblin why do we need _any_ special position? We turn solenoid off - it closes with spring, does not it?
> 
> If we ever need it again - it uses existing code to position itself?

generally yes, but ticket from @ElDominio says they want a special position:

> Some people need their idle valve in a specific spot (instead of 0% duty) at 0rpm.

TokenGoblin:
@rusefillc @FDSoftware — you are both right, and both new fields are now gone.

**What `idleSolenoidParkPosition` did, and why it should not have existed.** With it removed, what the option holds is the position the idle controller had *already* computed. At zero RPM `rpmCalculator` is not running, so `determinePhase()` returns `Phase::Cranking`; `getClosedLoop()` returns 0 for any non-`Idling` phase, and `getOpenLoop()` returns `getCrankingOpenLoop(clt)` = `interpolate2d(clt, cltCrankingCorrBins, cltCrankingCorr)`. So the held position **is** the cranking curve at the current coolant temperature — exactly @FDSoftware's "the first position on the open loop curve", read from a table the user already calibrates. I had added a second way to say something the calibration already said.

**@rusefillc on "why do we need *any* special position".** For an ordinary 2-wire solenoid, none — off is spring-closed, and that stays the default for everyone. #9123 asks for a toggle ("Please add an enable/disable toggle to solenoid idle valve hardware when RPM == 0") for valves where resting de-energized is not wanted. This is now exactly that toggle and nothing more.

**`idleSolenoidParkTimeout` is gone too.** The bounded hold @FDSoftware asked for stays, but as a constant `IDLE_SOLENOID_HOLD_TIMEOUT_SEC = 60` rather than a field: it protects the coil and the battery rather than tuning anything, and a config field would cost flash on every board including the ones that will never turn this on. With no field there is also nothing to migrate, so `applyDefaultsOrFixAfterBurn()` and `setDefaultBaseEngine()` go back to untouched.

Net config cost is now **one bit** (`unusedBit_Fancy3`), no `TOTAL_CONFIG_SIZE` change. The whole diff is the gated branch in `applyIACposition()`, that bit, one TunerStudio field, and two unit tests.

If you would rather the hold time be tunable after all, or would rather the option did not exist, say the word — happy to add the field back or to close this.

Verified: `unit_tests` 1170 pass on MinGW GCC 16.1. Not verified locally: any ARM board build, the simulator, ASan (no libasan on this machine), and nothing was tried on hardware or in TunerStudio.

Issue #10023, opened 2026-08-06
Add a load-axis override to the second VE table


Matheusschmitz:
Alpha-N where MAP has no resolution, SD elsewhere. With ITBs or a big cam, MAP at small throttle collapses into a narrow noisy band while TPS still resolves. veOverrideMode is global — you can't say "TPS below X kPa, MAP above".

MAP-sensor limp-home. secondVeTableInput is a hard pin switch, ideal for a fallback table — but useless for MAP failure, since the fallback is also indexed by MAP.

Evaluating another load axis. Flipping veOverrideMode reinterprets the primary table: veLoadBins are kPa and get re-read as percent, destroying the calibration.

PR #10089, opened 2026-08-14
Second VE table: its own load axis (#10023)

TokenGoblin:
Part of #10023 - this covers the load-axis half of the issue. It does **not** change the air model: `fuelAlgorithm` still picks one globally, so in speed density the airmass is still computed from the measured MAP. Blending SD and Alpha-N air models is a separate change (see the discussion below), so the issue stays open.

## The constraint being removed

`AirmassVeModelBase::getVe()` resolves the load axis **once, globally**:

```cpp
load = getVeLoadAxis(engineConfiguration->veOverrideMode, load);
```

and then uses that same `load` for the primary table *and* the second table — hard switch and blend alike. So the second table can never be indexed by anything other than whatever the primary is using.

That blocks both cases in the issue: speed density on MAP with a second table on TPS for the small-throttle region where MAP collapses into a narrow noisy band on a cammed or ITB engine, and MAP-sensor limp-home, where a pin-switched fallback table indexed by MAP is no help at all.

## Why `VE_None` is the right default, not a new sentinel

My first instinct was that this needed a distinct "same as primary" enum value, because defaulting to `VE_None` would silently change behaviour for anyone already using `veOverrideMode`.

Tracing the function showed that's already solved. `idleVeOverrideMode`, twelve lines below, is passed the **already-overridden** `load`:

```cpp
idleVeLoad = getVeLoadAxis(engineConfiguration->idleVeOverrideMode, load);
```

so `VE_None` there means "inherit whatever the primary resolved to". Using the identical pattern gives `0` → inherit → **bit-identical behaviour for every existing tune**, no new enum, and no migration. It also keeps the second table consistent with how the idle table already behaves.

## Changes

| File | Change |
|---|---|
| `rusefi_config.txt` | `ve_override_e secondVeOverrideMode`, appended at the end of `engine_configuration_s` |
| `airmass.cpp` | Separate `secondVeLoad` used by both second-table lookups; `load` itself untouched, so the primary table, `veBlends`, the idle table and `veTableYAxis` are all unaffected |
| `config_page_4.txt` | `secondVeLoadBins` axis label/scale now follows the second table's own mode, falling back to the primary's when None |
| `secondary_panels.ini` | Field in the existing "Second VE Table" dialog |

The `.ini` axis change matters: without it the axis would keep claiming kPa while the table was indexed by TPS.

I put the scalar in `engine_configuration_s` rather than `page4_s` because page 4 requires custom types to be re-declared locally, and because the other two override modes already live there. Happy to move it if you'd rather it sat with the rest of the second-table config.

## Validation

3 new tests, using the existing `MockAirmass` harness. The second table is filled so the value it returns **is** the load it was indexed with, which makes the axis directly observable rather than inferred:

- inherit: None/None sees the passed load; primary→MAP then primary→TPS are both inherited by the second table (this is the regression guard for existing tunes)
- override: MAP primary + TPS second, and the reverse
- the same applies through the blend path, not just the hard switch

**Full suite: 1169 pass.**

Host build only (MinGW GCC 16.1); no ARM firmware build locally, and not tested on an engine. This is a fuel path, so the inherit-by-default behaviour is the part I'd most want a second pair of eyes on.

### DIFF
--- a/firmware/controllers/algo/airmass/airmass.cpp
+++ b/firmware/controllers/algo/airmass/airmass.cpp
@@ -121,6 +121,9 @@ float AirmassVeModelBase::getVe(float rpm, float load, bool postState) const {
 		engine->engineState.veTableIdleYAxis = idleVeLoad;
 #if EFI_PROD_CODE || EFI_UNIT_TEST
 		engine->engineState.isSecondVeTableActive = switchTableActive;
+		// #10023 the second table has its own load axis, so the primary's veTableYAxis does not
+		// say what this table was actually indexed with - publish it separately
+		engine->engineState.veTableSecondYAxis = secondVeLoad;
 #endif
 	}
 
--- a/firmware/controllers/algo/airmass/airmass.cpp
+++ b/firmware/controllers/algo/airmass/airmass.cpp
@@ -37,11 +37,17 @@ float AirmassVeModelBase::getVe(float rpm, float load, bool postState) const {
 	const bool pinActive = isBrainPinValid(secondTablesGetState()->secondVeTableInput) &&
 		efiReadPin(secondTablesGetState()->secondVeTableInput, secondTablesGetState()->secondVeTableInputMode);
 
+	// #10023 the second table may be indexed by a different load axis from the primary one - useful
+	// where MAP has no resolution at small throttle, or as a fallback for MAP failure, where a
+	// MAP-indexed second table would be no help. VE_None inherits the primary's already-overridden
+	// load, exactly like idleVeOverrideMode below, so existing tunes are unaffected.
+	float secondVeLoad = getVeLoadAxis(engineConfiguration->secondVeOverrideMode, load);
+
 	if (pinActive) {
 		// Hard switch: pin overrides everything, replace VE entirely
 		ve = interpolate3d(
 			secondTablesGetState()->secondVeTable,
-			secondTablesGetState()->secondVeLoadBins, load,
+			secondTablesGetState()->secondVeLoadBins, secondVeLoad,
 			secondTablesGetState()->secondVeRpmBins, rpm
 		);
 		switchTableActive = true;
@@ -50,7 +56,7 @@ float AirmassVeModelBase::getVe(float rpm, float load, bool postState) const {
 			secondTablesGetState()->secondVeBlendParameter,
 			secondTablesGetState()->secondVeBlendBins, secondTablesGetState()->secondVeBlendValues,
 			secondTablesGetState()->secondVeTable,
-			secondTablesGetState()->secondVeLoadBins, load,
+			secondTablesGetState()->secondVeLoadBins, secondVeLoad,
 			secondTablesGetState()->secondVeRpmBins, rpm
 		);
 		if (result.Bias > 0) {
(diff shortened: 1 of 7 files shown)


Matheusschmitz:
I already did some tests with 2 tables, blending alpha N + SD

Two things that look missing.

**This does not change the air model.** It only changes which axis the VE table is
read with. The air model is still picked once by `fuelAlgorithm`. In speed density the
airmass is calculated from the **measured MAP**, in Alpha-N from a fixed 101.3 kPa.
So primary on MAP + second on TPS is not "SD plus Alpha-N" — it is still speed density,
still multiplied by MAP, just with the second table addressed by TPS.

**No output channel.** The primary and idle axes are visible as `veTableYAxis` and
`veTableIdleYAxis`. There is nothing for the second table, so you cannot log or check
which load it actually used.

**The TS editor still uses the primary axis.** `tunerstudio.template.ini` is untouched,
so the second VE table editor still has `xyLabels` on `veLoadSrc` and `yBins` on
`veTableYAxis`. With primary on MAP and second on TPS the firmware is right, but the
editor shows the wrong axis label and puts the cursor at the primary's load — you would
be tuning the table with the crosshair on the wrong axis. It needs a `secondVeLoadSrc`
resolved from `secondVeOverrideMode` (falling back to `veOverrideMode` when None), and
`yBins` bound to a new second-table axis channel.

TokenGoblin:
Thank you — points 2 and 3 were real gaps and both are fixed in the follow-up commit. Point 1 is right, and I have corrected the framing rather than the code.

**Air model.** Correct, and worth stating plainly: `fuelAlgorithm` still picks one air model globally, so in speed density the airmass comes from the measured MAP and this PR only changes which *axis* the second table is read with. It is not "SD plus Alpha-N". Real blending between the two air models is a different change — it has to happen where the airmass is computed, not where VE is looked up — and I would rather not smuggle it into this PR. Since the opening line of #10023 does ask for that, I have changed the description from "Fixes #10023" to "Part of #10023" so the issue stays open for the air-model half.

**Output channel.** Added. `veTableSecondYAxis` in `engine_state.txt`, published from `getVe()` next to `isSecondVeTableActive`; the generator picks it up as a datalog entry ("VE: Second table load axis"), so the load the second table was actually indexed with sits next to `veTableYAxis` in a log instead of having to be inferred.

**TS editor.** You are right about where it mattered. I had changed `secondVeLoadBins` in `config_page_4.txt`, but not the `table = secondVeTableTbl` block in `tunerstudio.template.ini` — which is what drives the axis label and where the crosshair sits — so with the primary on MAP and the second on TPS the editor was still labelled and positioned from the primary. Now there are new PcVariables `secondVeLoadSrc` and `secondVeLoadUnitIdx`, following the same `maintainConstantValue` pattern as `veLoadSrc`/`veLoadUnitIdx`: they resolve `secondVeOverrideMode` and fall back to the primary's resolution when None, which is the same inherit rule the firmware uses. `xyLabels` now reads `secondVeLoadSrc` and `yBins` reads `veTableSecondYAxis`.

Checked in the generated ini: `veTableSecondYAxis = scalar, U16, 1516, "%", 0.01, 0` in `[OutputChannels]`, aligned, with the datalog entry. `unit_tests` 1169 pass on MinGW GCC 16.1. Not verified locally: any ARM board build, and I have not opened this in TunerStudio — the ini expressions are mirrored from the existing `veLoadSrc` ones rather than tested.

If you have the two-table setup running on a car, I would value knowing whether the axis override alone covers your small-throttle case or whether the air model really does need blending — that is the part worth capturing in #10023.

PR #10090, opened 2026-08-15
Remove the FileLog OS helpers, leaving one implementation

TokenGoblin:
Split out of #10081 at @FDSoftware's request — that PR should stay a Linux launcher fix.

> last commits should be on a separate pr, but seem ok the rest

**Depends on #10081**, which introduces `com.rusefi.core.OsUtil` in `:shared_io`. Happy to rebase once that lands, or to fold this in if you would rather take them together.

## What this does

#10081 added `OsUtil` and left `FileLog#isWindows`/`isLinux`/`getOsName` delegating to it, so existing callers were untouched. @FDSoftware's point was that a delegating wrapper is still two entry points:

> since "moved" from `ecu_io` to `shared_io` i will expect to be removed the now old copy so we have only one implementation of this

So this deletes those three methods and moves all 15 call sites onto `OsUtil` directly:

| Module | Files |
|---|---|
| `:ui` | `DfuFlasher`, `DriverInstall`, `MaintenanceUtil`, `ProgramSelector`, `PortsComboBox`, `StartupFrame`, `ConsoleTools`, `TunerStudioHelper`, `UiHelper`, `StatusPanel`, `DevicePane`, `DevicePaneTest` |
| `:autotest` | `IoUtil`, `SimulatorExecHelper` |
| `:connectivity` | `SerialPortScanner` |

Verified nothing references the old ones:

```
$ grep -rn 'FileLog\.\(isWindows\|isLinux\|getOsName\)' --include=*.java java_console java_tools
(no matches)
```

`:ui:shadowJar`, `:autotest`, `:connectivity`, `:autoupdate`, `:core_ui` and `:shared_io` all build and test clean.

## What I deliberately left alone

`FileLog` keeps `is32bitJava()` and `OS_VERSION`. Those read `os.arch` and `os.version` rather than `os.name`, so they were never duplicated and moving them is a separate decision. That does leave `FileLog` as a two-member class whose own comment has been asking *"what the hell is this anyway? rename this utility class?"* since 2013 — say the word and I will finish the job in another PR rather than deciding that for you here.

### DIFF
--- a/java_console/autotest/src/main/java/com/rusefi/IoUtil.java
+++ b/java_console/autotest/src/main/java/com/rusefi/IoUtil.java
@@ -1,5 +1,7 @@
 package com.rusefi;
 
+import com.rusefi.core.OsUtil;
+
 import com.devexperts.logging.Logging;
 import com.rusefi.config.generated.Integration;
 import com.rusefi.core.*;
@@ -119,7 +121,7 @@ public class IoUtil {
 
     public static void connectToSimulator(LinkManager linkManager, boolean startProcess) throws InterruptedException {
         if (startProcess) {
-            if (FileLog.isWindows()) {
+            if (OsUtil.isWindows()) {
                 // this check seems not to work on Linux
                 if (!TcpConnector.getAvailablePorts().isEmpty())
                     throw new IllegalStateException("Port already binded on startup?");
--- a/java_console/autotest/src/main/java/com/rusefi/SimulatorExecHelper.java
+++ b/java_console/autotest/src/main/java/com/rusefi/SimulatorExecHelper.java
@@ -1,5 +1,7 @@
 package com.rusefi;
 
+import com.rusefi.core.OsUtil;
+
 import com.devexperts.logging.Logging;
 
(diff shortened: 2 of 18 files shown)


FDSoftware:
huh, this pr should be only the work on OsUtil, like a preparation pr, without any of the code for fixing the launcher things from the other pr

TokenGoblin:
You're right — I based it on the launcher branch, so its diff dragged that work along. Rebuilt from master.

#10090 is now the preparation PR only: **18 files, zero launcher code**. Create `OsUtil` in `:shared_io`, point the 15 existing `FileLog` call sites at it, delete the three methods from `FileLog`. No behaviour change, nothing gained or lost by any caller.

```
$ git diff --name-only origin/master | grep -E 'ConsoleExeFileLocator|rusefi_updater.sh|Autoupdate.java'
(no matches)
```

#10081 is now stacked on top of this and contains only the launcher fix (4 files) — it consumes `OsUtil` rather than creating it, so **this one wants merging first**. Both build and test clean in that order.

FDSoftware (approved):

Issue #8345, opened 2025-07-23
Single Cylinder Batch  vs Sequential not doubling number of pulses but doubles/halves the fuel


ggurov:
**Describe the bug**
on single cylinder configuration, injection is always in sequential pattern, switching to batch mode halves the fuel, but does not double the number of injection events

**Hardware**
 - Board: any

**To Reproduce**
1. Set up one cylinder engine in sequential injection mode
2. Stare at logic analyzer
3. Switch to batch injection mode
4. See that number of pulses did not double, but fuel pulse length was cut in half

**Expected behavior**
Number of pulses doubles on a 1 cylinder engine, like it does on a 2 cylinder

**Screenshots**
https://youtu.be/WJPuQXsXoik  - 1 cylinder

two cylinder version of the same works fine:

https://youtu.be/VdpHw4U1WQc

**Additional context**
European mopeds are suffering.

rusefillc:
@ggurov please provide tune and log

PR #10092, opened 2026-08-15
One cylinder batch injection runs lean (#8345)

TokenGoblin:
Addresses the fuel half of #8345. **Deliberately does not change the pulse count** — reasoning at the bottom.

## Why one cylinder gets one pulse

Batch pairs each injector with the cylinder 360° later in the firing order, `InjectionEvent::update()`:

```cpp
int secondOrder = (ownIndex + (engineConfiguration->cylindersCount / 2)) % engineConfiguration->cylindersCount;
int secondIndex = getCylinderNumberAtIndex(secondOrder);
secondOutput = &enginePins.injectors[secondIndex];
```

With `cylindersCount == 1` that is `(0 + 0) % 1 == 0`, so `secondIndex` resolves to the **same injector** — `outputs[0]` and `outputs[1]` are the same pin.

And `FuelSchedule` holds exactly one event per cylinder — `InjectionEvent elements[MAX_CYLINDER_COUNT]`, with both `addFuelEvents()` and `onTriggerTooth()` bounded by `cylindersCount`.

So a single-cylinder engine gets **one pulse per cycle in every mode**.

## The bug

`getInjectionModeDurationMultiplier()` returned `0.5f` for batch unconditionally, and `getNumberOfInjections()` returned `2`. Both assume a second pulse that never happens on one cylinder, so switching sequential → batch **halved the fuel and ran the engine lean**, while the duty cycle estimate read double.

That matches the report exactly: *"switching to batch mode halves the fuel, but does not double the number of injection events"*.

Both now account for the single pulse. Multi-cylinder is untouched.

## Validation

Wrote the tests **first**, against unfixed code, and they failed on the real values — `0.5` where 1 is correct, `2` injections where 1 happens:

```
tests/ignition_injection/test_one_cylinder_logic.cpp:72: Failure
    Which is: 0.5
tests/ignition_injection/test_one_cylinder_logic.cpp:75: Failure
    Which is: 2
```

- `oneCylinderBatchDeliversWholeCycleFuel` uses `GY6_139QMB`, a real single-cylinder engine type, and asserts batch matches sequential.
- `multiCylinderBatchStillHalvesFuel` pins 2 and 4 cylinders at `0.5` / `2 injections`, so the halving cannot be lost for engines that genuinely get two pulses.
- **Full suite: 1168 pass.**

`getInjectionModeDurationMultiplier()` gains a header declaration so it can be tested — it had none and was only used inside `fuel_math.cpp`.

## What I did not do, and why

The report's other half is that the pulse count should double. That needs a second injection event per cycle, which means breaking the one-event-per-cylinder invariant in `FuelSchedule` — a second element with its own 360° phase offset, `ownIndex` semantics, `WallFuel` instance and stage-2 outputs. That is a structural change to the injection scheduler, and the reporter's own method was *"stare at a logic analyzer"*. I have no hardware and no way to validate the resulting pulse train, so I have not guessed at it.

Happy to implement it if you tell me the shape you want. In the meantime this removes the lean condition, which is the part that damages engines.

Host build only (MinGW GCC 16.1); no ARM build locally, not tested on hardware.

### DIFF
--- a/firmware/controllers/algo/fuel_math.cpp
+++ b/firmware/controllers/algo/fuel_math.cpp
@@ -256,7 +256,11 @@ int getNumberOfInjections(injection_mode_e mode) {
 	case IM_SINGLE_POINT:
 		return engineConfiguration->cylindersCount;
 	case IM_BATCH:
-		return 2;
+		// Batch pairs each injector with the cylinder 360 degrees later in the firing order, so
+		// every injector fires twice per cycle. A single cylinder has no distinct twin -
+		// InjectionEvent::update() resolves secondIndex to the same injector - and FuelSchedule
+		// holds one event per cylinder, so exactly one pulse is produced. #8345
+		return engineConfiguration->cylindersCount == 1 ? 1 : 2;
 	case IM_SEQUENTIAL:
 		return 1;
 	default:
@@ -283,7 +287,9 @@ float getInjectionModeDurationMultiplier() {
 	case IM_SINGLE_POINT:
 		return 1;
 	case IM_BATCH:
-		return 0.5f;
+		// See getNumberOfInjections(): a single cylinder only gets one pulse per cycle, so halving
+		// the fuel for a second pulse which never happens leaves the engine running lean. #8345
+		return engineConfiguration->cylindersCount == 1 ? 1 : 0.5f;
 	default:
 		firmwareError(ObdCode::CUSTOM_ERR_INVALID_INJECTION_MODE, "Unexpected injection_mode_e %d", mode);
 		return 0;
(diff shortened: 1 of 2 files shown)


rusefillc:
tests first, fix later says https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing

https://github.com/rusefi/rusefi/commit/dc07f366421500f3ba229e64a15bb1d50eea6747 is the test part of this fix, please rebase to show coverage being flipped

TokenGoblin:
Rebased onto master as asked. The PR is now the fix only, on top of your test commit dc07f3664 — the diff shows the coverage flip directly:

- `oneCylinderBatchCurrentlyHalvesFuel` (expects 0.5 / 2 pulses) becomes `oneCylinderBatchDeliversWholeCycleFuel` (expects 1 / 1 pulse)
- `multiCylinderBatchHalvesFuel` is untouched and keeps passing, so the halving stays pinned for engines that really do get two pulses
- the `fuel_math.h` declaration hunk is gone from this PR since master already carries it

Full suite on the rebased branch: 1168 tests pass.

PR #10095, opened 2026-08-15
Fix: bound the ISO-TP single-frame length (out-of-bounds read)

TokenGoblin:
Second of two, per [TDB](https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing). **Merge #10101 (coverage) first** — this PR is stacked on it and currently carries its commit too; once #10101 lands I will rebase so this is the fix alone.

## The fix

Bound the single-frame data length by what the frame actually carries, using the same `minI()` that the FIRST and CONSECUTIVE branches already apply to their own reads:

```cpp
int payloadBytes = minI(rxmsg.DLC, (int)sizeof(rxmsg.data8)) - 1 - (int)isoHeaderByteIndex;
numBytesAvailable = minI(rxmsg.data8[isoHeaderByteIndex] & 0xf, payloadBytes);
```

**Both bounds are needed:**
- `DLC` alone is not enough — it is a 4-bit field, so a malformed frame can claim up to 15 while `data8[]` is 8 bytes wide (the second coverage case).
- `sizeof(data8)` alone is not enough — a short frame carries fewer bytes than the array is wide.

Taking the minimum of the two, minus the PCI byte and any vendor header byte, is the number of payload bytes that really exist. `payloadBytes` cannot go negative: `receiveFrame()` already returns early unless `DLC >= 1 + isoHeaderByteIndex`.

## Coverage adjustment (the proof)

| coverage test | #10101 asserts | after this fix |
|---|---|---|
| `singleFrameLength...ByDlc` | 15 copied | **7** copied |
| `singleFrameLength...ByPayloadSize` | 15 copied | **7** copied |

Renamed `IsNotBoundedBy...` → `IsBoundedBy...` accordingly.

## Safety / regression

No engine-control path — this is the diagnostic/TS transport. The clamp only ever *reduces* the copied length, and only for frames that were already malformed: well-formed single frames always satisfy `SF_DL <= DLC - 1`. The six pre-existing `testCanSerial` round-trip tests (`test1Frame`, `test2Frames`, `testIrregularSplits`, `testLongMessage`, `test64_7Message`, `test3_64_4Message`) pass unchanged, which is the evidence for that.

The FIRST branch (offset `2 + header`, capped at `6 - header`) and CONSECUTIVE branch (offset `1 + header`, capped at `7 - header`) stay within the array by construction and are untouched.

## Validation

```
cd unit_tests && make -j12 && ./build/rusefi_test.exe    1170 tests pass
```
Firmware/simulator builds left to CI (no 32-bit MinGW/ARM toolchain locally).

### DIFF
--- a/firmware/controllers/can/isotp/isotp.cpp
+++ b/firmware/controllers/can/isotp/isotp.cpp
@@ -104,11 +104,18 @@ int CanStreamerState::receiveFrame(const CANRxFrame &rxmsg, uint8_t *destination
 	int numBytesAvailable, frameIdx;
 	const uint8_t *srcBuf;
 	switch (frameType) {
-	case ISO_TP_FRAME_SINGLE:
-		numBytesAvailable = rxmsg.data8[isoHeaderByteIndex] & 0xf;
+	case ISO_TP_FRAME_SINGLE: {
+		// The low nibble is the single-frame data length and can encode up to 15, but the frame
+		// only carries DLC bytes - and DLC is itself a 4-bit field, so a malformed frame can claim
+		// more than data8[] is wide. Bound the length by both before copying, the same way the
+		// FIRST and CONSECUTIVE branches below bound their own reads. Well-formed single frames
+		// always satisfy SF_DL <= DLC - 1, so they are unaffected.
+		int payloadBytes = minI(rxmsg.DLC, (int)sizeof(rxmsg.data8)) - 1 - (int)isoHeaderByteIndex;
+		numBytesAvailable = minI(rxmsg.data8[isoHeaderByteIndex] & 0xf, payloadBytes);
 		this->waitingForNumBytes = numBytesAvailable;
 		srcBuf = rxmsg.data8 + 1 + isoHeaderByteIndex;
 		break;
+	}
 	case ISO_TP_FRAME_FIRST:
 		this->waitingForNumBytes = ((rxmsg.data8[isoHeaderByteIndex] & 0xf) << 8) | rxmsg.data8[isoHeaderByteIndex + 1];
 		this->waitingForFrameIndex = 1;
--- a/unit_tests/tests/controllers/can/test_can_serial.cpp
+++ b/unit_tests/tests/controllers/can/test_can_serial.cpp
@@ -230,12 +230,9 @@ TEST(testCanSerial, test64_7Message) {
  *
  * A single frame carries its payload length in the low nibble of the PCI byte, so it can encode up
  * to 15 - but a classic-CAN frame only delivers DLC bytes and CANRxFrame::data8 is 8 bytes wide.
- * receiveFrame() takes that nibble at face value and copies that many bytes from data8 + 1, so a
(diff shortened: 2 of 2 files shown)


rusefillc:
@TokenGoblin please apply https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing methodology same as we had with 1 cylinder batch fuel

TokenGoblin:
Applied, thank you — restructured into Coverage-first / Fix-second per the TDB page.

**Commit 1 — [Coverage](https://github.com/rusefi/rusefi/pull/10095/commits/075488ad0b):** green against unfixed master, asserting the bad behaviour. Two cases, because there are two independent ways to exceed the payload:

| case | frame | asserts today |
|---|---|---|
| `singleFrameLengthIsNotBoundedByDlc` | DLC 8, SF_DL 15 | 15 bytes copied (8 past the payload) |
| `singleFrameLengthIsNotBoundedByPayloadSize` | DLC 15, SF_DL 15 | 15 bytes copied |

The second case exists because `DLC` is a **4-bit field**, so a malformed frame can claim more than `data8[]` is wide. **I had missed that in the first version of this PR** — the original fix clamped to DLC only, which would still have allowed a read past the array for DLC > 8. Writing the coverage first is what exposed it, which I think rather makes your point.

Note these tests deliberately provoke the over-read, so the frame is embedded in a padded wrapper and the read lands in memory the test owns — otherwise it is a stack-buffer-overflow and ASan (on by default here) aborts the run. `assertOverreadStaysInsidePaddedFrame()` proves the padding actually covers the largest over-read, so it fails loudly instead of corrupting the stack if that ever stops holding.

**Commit 2 — [Fix](https://github.com/rusefi/rusefi/pull/10095/commits/50c6fc8daa):** bounds the length by both DLC and `sizeof(data8)`, and adjusts both expectations 15 → 7.

The six pre-existing `testCanSerial` round-trip tests pass unchanged throughout, which is the evidence that well-formed frames are unaffected. Full suite: 1170 pass.

Happy to split the coverage commit into its own PR if you would rather merge it separately.

TokenGoblin:
Split into two PRs as the updated page asks.

- **#10101 — Coverage** (merge first): green on unfixed master, asserts 15 bytes copied in both the DLC and the payload-size cases.
- **This PR is now the Fix**, retitled, stacked on #10101. It still carries the coverage commit so it can build; once #10101 lands I will rebase so this is the fix alone.

TokenGoblin:
Rebased now that #10101 has landed — this PR is the fix alone, two files:

```
firmware/controllers/can/isotp/isotp.cpp                 | 11 +++++++++--
unit_tests/tests/controllers/can/test_can_serial.cpp     | 21 +++++++++------------
```

The test diff is purely the expectation adjustment: `singleFrameLengthIsNotBoundedByDlc` / `...ByPayloadSize` (15 bytes copied) become `singleFrameLengthIsBoundedByDlc` / `...ByPayloadSize` (7 bytes copied).

Full suite on the rebased branch: 1172 tests pass.

PR #10096, opened 2026-08-15
Fix: findBaudIndex restores the configured console speed

TokenGoblin:
Second of two, per [TDB](https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing). **Merge #10102 (coverage) first** — this PR is stacked on it and currently carries its commit too; once #10102 lands I will rebase so this is the fix alone.

## The fix

The recovery that restores the configured console/TS speed is unreachable: it sits inside `for (baudIdx = 0; baudIdx < efi::size(baudRates); baudIdx++)` guarded by `baudIdx == efi::size(baudRates)`, which is never true inside the loop body. On a total probe failure the function fell through to a bare `return 255` with the channel still started at the last probe rate (57600).

Moved to after the loop, where it is reached: stop the channel and restart it at `engineConfiguration->tunerStudioSerialSpeed` before returning 255.

## Impact

A failed `bluetooth_<module>` setup attempt (wrong module, module not answering AT, already-configured module in a silent mode) left the serial console/TS link at 57600 instead of the configured speed, until the ECU was rebooted.

## Coverage adjustment (the proof)

`findBaudIndexLeavesChannelAtLastProbedRate` (asserts 57600 in #10102) becomes `findBaudIndexRestoresConsoleSpeedWhenNoModuleResponds` (asserts 250000, the configured speed).

## Safety / regression

No engine-control path. The success paths are unchanged — when a baud rate answers, the channel is still left at the found rate for the caller, exactly as before. The only behavioural change is on the total-failure branch of a manually-invoked setup command, and it restores the documented intent.

## Validation

```
cd unit_tests && make -j12 && ./build/rusefi_test.exe    1169 tests pass
```
Firmware builds left to CI (no ARM toolchain locally).

### DIFF
--- a/firmware/console/binary/bluetooth.cpp
+++ b/firmware/console/binary/bluetooth.cpp
@@ -236,12 +236,6 @@ uint8_t findBaudIndex(SerialTsChannelBase* tsChannel) {
 		tsChannel->stop();
 		chThdSleepMilliseconds(10);	// safety
 
-		if (baudIdx == efi::size(baudRates)) {
-			efiPrintf("Failed to find current BT module baudrate");
-			tsChannel->start(engineConfiguration->tunerStudioSerialSpeed);
-			return 255;//failed
-		}
-
 		efiPrintf("Restarting at %lu", baudRates[baudIdx].rate);
 		tsChannel->start(baudRates[baudIdx].rate);
 		chThdSleepMilliseconds(10);	// safety
@@ -269,6 +263,16 @@ uint8_t findBaudIndex(SerialTsChannelBase* tsChannel) {
 		}
 		/* try next baudrate */
 	}
+
+	// None of the known baud rates responded. The channel is currently running at the last probe
+	// rate, so restore the configured console/TS speed before giving up - otherwise a failed setup
+	// attempt leaves the serial link at the wrong baud until reboot. This recovery used to sit
+	// inside the loop above, guarded by baudIdx == efi::size(baudRates), which the loop condition
+	// makes unreachable.
+	efiPrintf("Failed to find current BT module baudrate");
+	tsChannel->stop();
+	chThdSleepMilliseconds(10);	// safety
+	tsChannel->start(engineConfiguration->tunerStudioSerialSpeed);
 	return 255;//failed
 }
 
--- a/unit_tests/tests/test_bluetooth.cpp
+++ b/unit_tests/tests/test_bluetooth.cpp
@@ -56,16 +56,12 @@ public:
 /**
  * Coverage for findBaudIndex() when no Bluetooth module answers at any known baud rate.
  *
- * The function is supposed to put the serial link back to the configured console/TS speed before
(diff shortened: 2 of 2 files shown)


rusefillc:
@TokenGoblin please attempt https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing methodology

TokenGoblin:
Applied, thank you — restructured into Coverage-first / Fix-second per the TDB page.

**Commit 1 — [Coverage](https://github.com/rusefi/rusefi/pull/10096/commits/8c6118748f):** `findBaudIndexLeavesChannelAtLastProbedRate`, green against unfixed master. It drives `findBaudIndex()` with a mock `SerialTsChannelBase` that records the last baud it was started at and never answers the AT probe, and asserts what happens today: returns 255, and the channel is left at **57600** (the last entry of `baudRates[]`) while `tunerStudioSerialSpeed` is 250000 — i.e. the restore never ran, because the branch that does it sits inside the probe loop guarded by `baudIdx == efi::size(baudRates)`, which the loop condition makes unreachable.

That commit also carries the test-enablement: `bluetooth.cpp`'s guard widened to `EFI_BLUETOOTH_SETUP || EFI_UNIT_TEST` (the pattern `serial_can.cpp` already uses here) plus two unit-test-only shims for the ChibiOS primitives the host build omits.

**Commit 2 — [Fix](https://github.com/rusefi/rusefi/pull/10096/commits/fd685ffa48):** moves the recovery after the loop where it is reached, and adjusts the expectation 57600 → 250000 (test renamed to `findBaudIndexRestoresConsoleSpeedWhenNoModuleResponds`).

I verified the coverage commit is green on its own checkout, not just at the tip. Full suite: 1169 pass.

Happy to split the coverage commit into its own PR if you would rather merge it separately.

rusefillc:
@TokenGoblin I've just improved https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing to more explicitly request two separate PRs

TokenGoblin:
Split into two PRs as the updated page asks — thank you for making it explicit.

- **#10102 — Coverage** (merge first): green on unfixed master, asserts the channel is left at 57600.
- **This PR (#10096) is now the Fix**, retitled, stacked on #10102. It still carries the coverage commit so it can build; once #10102 lands I will rebase so this is the fix alone.

Same split applied to the ISO-TP pair: **#10101** (coverage) → **#10095** (fix).

TokenGoblin:
Rebased now that #10102 has landed — this PR is the fix alone, two files:

```
firmware/console/binary/bluetooth.cpp | 16 ++++++++++------
unit_tests/tests/test_bluetooth.cpp   | 18 +++++++-----------
```

The test diff is purely the expectation adjustment: `findBaudIndexLeavesChannelAtLastProbedRate` (57600, the last probed rate) becomes `findBaudIndexRestoresConsoleSpeedWhenNoModuleResponds` (250000, the configured speed).

Full suite on the rebased branch: 1173 tests pass.

PR #10097, opened 2026-08-15
Bluetooth: set up the module from TunerStudio, not just a console command

TokenGoblin:
## Problem

Configuring a serial Bluetooth module (HC-05/06, BK, JDY-33/31) required the **rusEFI Console** command `bluetooth_<module> <baud> <name> <pin>`. TunerStudio — where every other setting lives — had no interface for it, and nothing indicated the capability existed. The project's own wiki states: *"There is currently no TunerStudio interface for setting up the module — it is initialised from the rusEFI Console."*

## Change

A **"Bluetooth Setup"** dialog (under Settings, next to Connection) with module-type, baud, name and PIN fields, plus an **"Apply Bluetooth Setup"** button and a **Cancel** button. Apply runs the exact same AT sequence the console command does.

Reuses existing machinery end-to-end:
- New persisted config fields `btModuleType` / `btBaudRate` / `btName` / `btPinCode`, **appended at the end of `engine_configuration_s`** so no existing field offset moves (tune migration is by field name; `TOTAL_CONFIG_SIZE` 15804 → 15832, append-only).
- The button sends the standard `TS_IO_TEST_COMMAND` with a new subsystem `TS_BLUETOOTH_SETUP` (index 0 = apply, 1 = cancel) — the same button mechanism the spark/injector/wideband test buttons use. Generated: `cmd_bluetooth_apply = "Z\x00\x28\x00\x00"`.
- `executeTSCommand()` dispatches it to a new `bluetoothStartFromConfiguration()`, which reads the config fields and calls the existing `bluetoothStart()`. **The console commands are untouched and still work.**

`bluetooth_module_e` moved from `bluetooth.h` to `engine_types.h` so it can serve as the `btModuleType` field type; a new `bluetooth_baud_e` maps the dropdown to `btBaudRateValues[]`. Both `__packed__` to stay one byte (U08).

## Design notes

- The firmware only ever **configures** the module (name/PIN/baud/SPP mode) — it has no live enable/disable — so this is deliberately a *set-fields-then-Apply* flow, not an on/off switch.
- As with the console command, the AT sequence runs once the UART goes quiet (TS disconnected). The dialog and the `btBaudRate` description both warn the baud **must match the serial speed of the port Bluetooth uses** (secondary TTL if the board has one, else primary) — the single most common setup mistake.
- Applying with empty/invalid name or PIN is rejected by the existing `bluetoothStart()` validation, so unconfigured defaults are safe.

## Safety impact

None on engine control. This is the diagnostic/comms setup path; no timing or scheduling change, and the on-hardware AT sequence is byte-identical to the existing console path.

## Validation

- Unit build regenerates the config cleanly: the two enum dropdowns, the two string fields, the dialog/menu entry, and the `TS_BLUETOOTH_SETUP` command constant (40 = 0x28) all generate as intended.
- Full unit suite: **1168 tests pass** (config-layout change did not disturb persistent-config tests).
- The `EFI_BLUETOOTH_SETUP`-gated additions (`bluetoothStartFromConfiguration`, `bluetoothCancelSetup`, the `executeTSCommand` case) were compile-checked with the flag forced on — clean, no warnings. A full firmware/board build is left to CI (no ARM/32-bit MinGW toolchain locally).

## Regression risk

Low. Existing console commands and the AT sequence are unchanged; the new path is additive and gated by `EFI_BLUETOOTH_SETUP` exactly like the console commands. The config change is append-only.

### DIFF
--- a/firmware/console/binary/bluetooth.cpp
+++ b/firmware/console/binary/bluetooth.cpp
@@ -334,6 +334,38 @@ void bluetoothStart(bluetooth_module_e moduleType, const char *baudRate, const c
 	btSetupIsRequested = true;
 }
 
+// Ascending, indexed by bluetooth_baud_e (the btBaudRate config field).
+static const uint32_t btBaudRateValues[] = { 2400, 4800, 9600, 19200, 38400, 57600, 115200 };
+
+void bluetoothStartFromConfiguration() {
+	uint8_t baudIndex = (uint8_t)engineConfiguration->btBaudRate;
+	if (baudIndex >= efi::size(btBaudRateValues)) {
+		efiPrintf("Invalid Bluetooth baud selection %d", baudIndex);
+		return;
+	}
+
+	char baudStr[8];
+	chsnprintf(baudStr, sizeof(baudStr), "%lu", (unsigned long)btBaudRateValues[baudIndex]);
+
+	// The config strings are fixed-width and are not guaranteed to be null-terminated when fully
+	// used, so copy into a local buffer with room for the terminator before handing them off.
+	char name[BT_NAME_SIZE + 1];
+	char pin[BT_PIN_SIZE + 1];
+	memcpy(name, engineConfiguration->btName, BT_NAME_SIZE);
+	name[BT_NAME_SIZE] = '\0';
+	memcpy(pin, engineConfiguration->btPinCode, BT_PIN_SIZE);
+	pin[BT_PIN_SIZE] = '\0';
+
+	bluetoothStart(engineConfiguration->btModuleType, baudStr, name, pin);
+}
+
+void bluetoothCancelSetup() {
+	if (btSetupIsRequested) {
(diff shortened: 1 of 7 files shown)


rusefillc:
Considering how unhealthy/unreliable/broken is Bluetooth connectivy, and considering pretty real flash shortage on 1MB - I believe that's it NOT a good idea to invest even 10 byte of flash into this feature.

TokenGoblin:
Understood — closing.

That is your call to make and the reasoning is sound: `engine_configuration_s` is shared by every board, so the config fields cost flash on the 1MB targets too, whether or not the board has a module on it. Spending that on a feature you consider unreliable is a bad trade, and it is not something I could engineer away.

Thanks for the two Bluetooth *fixes* you took today (#10102/#10096) — the dead-code recovery path was the part that actually mattered. I will keep future Bluetooth work to defects rather than surface area.

PR #10102, opened 2026-08-15
Coverage: findBaudIndex leaves the link at the last probed baud rate

TokenGoblin:
First of two, per [TDB](https://github.com/rusefi/rusefi/wiki/TDB-Test-Driven-Bugfixing). **Coverage only — green, asserting the bad behaviour.** No behavioural change to `findBaudIndex()` in this PR. The fix (and the adjustment of this expectation) follows in a second PR.

## What is asserted

`findBaudIndex()` probes each known baud rate looking for a Bluetooth module. When none answers it is supposed to put the serial link back to the configured console/TS speed before giving up. That recovery exists, but it sits **inside** the probe loop:

```cpp
for (uint8_t baudIdx = 0; baudIdx < efi::size(baudRates); baudIdx++) {
    ...
    if (baudIdx == efi::size(baudRates)) {     // never true inside this loop
        tsChannel->start(engineConfiguration->tunerStudioSerialSpeed);
        return 255;
    }
```

so it never runs. `findBaudIndexLeavesChannelAtLastProbedRate` drives the function with a mock `SerialTsChannelBase` that records the last baud it was started at and never answers the AT probe, and asserts what happens today: returns 255, and the channel is left at **57600** — the last entry of `baudRates[]` — while `tunerStudioSerialSpeed` is 250000.

## Test enablement (why bluetooth.cpp is touched)

`findBaudIndex` sits behind `EFI_BLUETOOTH_SETUP`, which is off in the host unit-test build, so `bluetooth.cpp` compiled empty there and the function had no coverage at all. This PR widens that guard to `EFI_BLUETOOTH_SETUP || EFI_UNIT_TEST` — the same pattern `serial_can.cpp` already uses in this subsystem — and adds two unit-test-only shims for the ChibiOS primitives the host build omits (`chThdSleepMilliseconds`, `TIME_MS2I`). `getBluetoothChannel()` / `getTsSignature()`, which need real serial ports, are stubbed in the test.

That is the only production-file change here, and it is compile-gating only — no logic is altered.

## Validation

```
cd unit_tests && make -j12 && ./build/rusefi_test.exe
  1169 tests pass (green on unfixed master, as coverage should be)
```

### DIFF
--- a/firmware/console/binary/bluetooth.cpp
+++ b/firmware/console/binary/bluetooth.cpp
@@ -20,7 +20,7 @@
 #include <stdio.h>
 #include <ctype.h>
 
-#if EFI_BLUETOOTH_SETUP
+#if EFI_BLUETOOTH_SETUP || EFI_UNIT_TEST
 
 #ifndef EFI_BLUETOOTH_SETUP_DEBUG
 #define EFI_BLUETOOTH_SETUP_DEBUG TRUE
@@ -359,4 +359,4 @@ void bluetoothSoftwareDisconnectNotify(SerialTsChannelBase* tsChannel) {
 	}
 }
 
-#endif /* EFI_BLUETOOTH_SETUP */
+#endif /* EFI_BLUETOOTH_SETUP || EFI_UNIT_TEST */
--- a/unit_tests/chibios-mock/mock-threads.h
+++ b/unit_tests/chibios-mock/mock-threads.h
@@ -27,6 +27,15 @@ static inline systime_t chTimeAddX(systime_t systime,
   return systime + (systime_t)interval;
 }
 
+// There is no tick clock on the host, so an interval is just the millisecond count itself.
+#ifndef TIME_MS2I
+#define TIME_MS2I(ms) (ms)
+#endif
+
+// Nothing to sleep on: host tests drive time explicitly (setTimeNowUs/advanceTimeUs) rather than
+// waiting, so sleeping is a no-op.
+static inline void chThdSleepMilliseconds(uint32_t /*milliseconds*/) { }
+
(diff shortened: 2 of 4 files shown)


firmware/console/binary/bluetooth.cpp
```diff
@@ -20,7 +20,17 @@
 #include <stdio.h>
 #include <ctype.h>
 
-#if EFI_BLUETOOTH_SETUP
+#if EFI_BLUETOOTH_SETUP || EFI_UNIT_TEST
+
+#if EFI_UNIT_TEST
+// The ChibiOS timing primitives are compiled out of the host unit-test build, so provide
```
rusefillc: I could be wrong but this does not look right

we have some chibios mocking in more designated files, this place is too random this test fixtures

TokenGoblin:
You're right, and thank you — that was me putting test scaffolding somewhere convenient rather than somewhere correct. Fixed and pushed.

`TIME_MS2I` and `chThdSleepMilliseconds` now live in **`unit_tests/chibios-mock/mock-threads.h`**, next to the existing `systime_t` / `sysinterval_t` / `chTimeAddX` / `CH_CFG_ST_FREQUENCY` mocks, which is where ChibiOS mocking belongs. Nothing else in the unit-test build defines either symbol, so there is no clash.

That leaves the production diff in this PR as just the compile gate, nothing else:

```diff
-#if EFI_BLUETOOTH_SETUP
+#if EFI_BLUETOOTH_SETUP || EFI_UNIT_TEST
```

(plus the matching `#endif` comment) — the same pattern `serial_can.cpp` already uses in this subsystem to make itself testable. No logic changed.

Still green: 1169 tests pass.

Issue #9673, opened 2026-06-20
Would h-bridge be useable at least as one GPIO?


rusefillc:
older non-ETB vehicle, uaEFI

https://www.youtube.com/watch?v=p0VtJTGlQac is rightfully complaining about output limits. Can we use both h-bridge to make at least two GPIOs? With what limitations?

dron0gus:
<img width="728" height="809" alt="Image" src="https://github.com/user-attachments/assets/1deb3fa8-0de7-4195-8ad7-b05fde925d81" />

Possible. But:
- Only one output from each H-bridge.
- Second output of the same H-bridge should not be connected as it will also change its state!
- Internal freewheeling diodes with no option to remove it.
- Chance to burn everything in case of incorrect connection.

FDSoftware:
> Can we use both h-bridge to make at least two GPIOs? With what limitations? 


yup, works! 

https://github.com/user-attachments/assets/ff6127a6-d4b1-4c7f-957f-d05cda3c644d

with this and a few more pr's:
https://github.com/rusefi/rusefi/pull/10103

FDSoftware:
tested!, this one can be closed :D

PR #10104, opened 2026-08-17
h-bridge as gpio

FDSoftware:
related #9673


firmware/config/boards/hellen/uaefi/connectors/uaefi-A.yaml
```diff
@@ -10,12 +10,12 @@ pins:
     color: orange
 
   - pin: A1
-    function: DC1- (usually electronic throttle body)
+    function: DC1- (ETB/stepper; disconnect when A5 is used as an output)
```
rusefillc: "disconnect when A5 is used as GPIO output" maybe?

rusefillc (changes requested):

PR #10105, opened 2026-08-17
bundle: always create bin/device directory

dron0gus:
I'm not a Makefile guru, but looks like bin/device directory is not created during local build.
@FDSoftware can you review?
```
BOARD_DIR=../././config/boards/proteus BOARD_META_PATH=../ SHORT_BOARD_NAME=proteus_f4 TGT_SENTINEL=.././.target-sentinel WHITE_LABEL=rusefi make -C bootloader -r SUBMAKE=yes TS_PAGE_GUARD_DEFS=""
ln -rfs build/rusefi.srec rusefi.snapshot.proteus_f4/rusefi_development_yymmdd_proteus_f4_354763248_local_update.srec
cd ./.. && flock -o /tmp/java.lock ./gradlew :ui:shadowJar
make -C ../simulator -r OS="Windows_NT" SUBMAKE=yes TS_PAGE_GUARD_DEFS=""
make[1]: Entering directory '/home/work/rusefi/rusefi/simulator'
OS is [Windows_NT]
make[1]: Entering directory '/home/work/rusefi/rusefi/firmware/bootloader'
if [ "$(cat ../.target-sentinel 2>/dev/null)" != proteus_f4 ]; then \
echo proteus_f4 >../.target-sentinel; fi
Creating blbuild/openblt_proteus_f4.srec
make[1]: Leaving directory '/home/work/rusefi/rusefi/firmware/bootloader'
ln -rfs bootloader/blbuild/openblt_proteus_f4.bin rusefi.snapshot.proteus_f4/bin/device/openblt_development_yymmdd_proteus_f4_354763248_local.bin
ln -rfs deliver/rusefi.bin rusefi.snapshot.proteus_f4/rusefi_development_yymmdd_proteus_f4_354763248_local.bin
ln: failed to create symbolic link 'rusefi.snapshot.proteus_f4/bin/device/openblt_development_yymmdd_proteus_f4_354763248_local.bin': No such file or directory
make: *** [bundle.mk:212: rusefi.snapshot.proteus_f4/bin/device/openblt_development_yymmdd_proteus_f4_354763248_local.bin] Error 1
make: *** Waiting for unfinished jobs....
if [ "$(cat .os-sentinel 2>/dev/null)" != Windows_NT ]; then \
echo Windows_NT >.os-sentinel; fi
make[1]: Leaving directory '/home/work/rusefi/rusefi/simulator'
Reusing configuration cache.
```


dron0gus:
Nope, this does not help on clean build.

dron0gus:
https://github.com/rusefi/rusefi/blob/b04b60bf18db004ab596018f12b881c6d6217378/firmware/bundle.mk#L266

rusefillc:
@dron0gus xyproblem? who cares why when?

dron0gus:
> @dron0gus xyproblem? who cares why when?

I care. Local build either fails either final zip does not contain bootloader. Don't know why gh is not affected

rusefillc:
@FDSoftware @dron0gus true or false:

On local bundle builds (per the PR description), one of two failures happens — either:

ln: failed to create symbolic link 'rusefi.snapshot.proteus_f4/bin/device/openblt_..._local.bin': No such file or directory

or the build "succeeds" but the final zip silently ships without the bootloader. GitHub Actions builds are unaffected.

The root cause is an interaction between three rules (all visible in the current bundle.mk):

1. $(BIN_FOLDER): .FORCE | $(FOLDER) — because of .FORCE, this runs on every build, and its recipe starts with rm -rf $@. So <staging>/bin — including bin/device — is wiped and recreated each run (this is the "Line 266 rm -rf" dron0gus pointed at in the discussion).
2. $(DEVICE_BIN_FOLDER): | $(BIN_FOLDER) with recipe mkdir -p $@. Since BIN_FOLDER is only an order-only prereq here, its forced rebuild never marks bin/device stale. On an incremental local rebuild the directory existed when make started, make's cached stat says "target exists, no normal prereqs are newer → up to date", and the mkdir -p is skipped — even though step 1's rm -rf just deleted the directory out from under it.
3. The bootloader symlink rule then either runs ln -rfs into the now-missing directory (the ln error), or — if make also judged the pre-existing symlink up to date — skips the recipe entirely while the symlink has actually been deleted by the wipe, producing a zip with no bootloader.

CI never hits this because GHA always builds in a fresh workspace: no pre-existing staging folder means no stale stat cache, and the clean-build path (mkdir actually runs) works fine. Only incremental local rebuilds, where rusefi.snapshot.*/bin/device survives from the previous run, trigger the skip.

Why this particular fix

Making the directory a normal prerequisite pulls its timestamp into the dependency calculation: after the forced wipe-and-recreate of bin/, the directory is fresh (or missing), so make actually processes it and considers the bootloader symlink out of date — guaranteeing both the mkdir -p and the ln -rfs run, in order, on every build. Hence the PR title, "always create bin/device directory."

Context from the discussion: this staging layout came from PR #9554 ("hide openblt.bin into some sub-folder", by FDSoftware, whom dron0gus tagged for review). A suggested alternative during review didn't survive a clean build ("Nope, this does not help on clean build"). When you asked whether it was an XY problem, dron0gus's justification was the concrete impact: local builds either fail outright or quietly produce bundles missing the bootloader — the latter being the nastier failure since a "successful" bundle flashed to a blank ECU would lack OpenBLT.

rusefillc:
Short answer: PR #10105's one-liner doesn't actually fix the bug — I reproduced the failure in a minimal makefile mirroring the three bundle.mk rules, and it still fails every other run with the PR's change applied. The     
  proper fix is to make the bootloader symlink rule forced and self-sufficient:                                                                                                                                                   
                                                                                                                                                                                                                                  
  $(BOOTLOADER_BIN_OUT): $(BOOTLOADER_BIN) .FORCE | $(BIN_FOLDER)                                                                                                                                                                 
        mkdir -p $(dir $@)                                                                                                                                                                                                        
        ln -rfs $< $@                                                                                                                                                                                                             
                                                                                                                                                                                                                                  
  and delete the $(DEVICE_BIN_FOLDER): | $(BIN_FOLDER) rule entirely (grep confirms nothing else uses it as a target).                                                                                                            
                                                                                                                                                                                                                                  
  Why the PR doesn't help. The real mechanism, confirmed by the repro: $(BIN_FOLDER) is .FORCEd and starts with rm -rf, so every build wipes bin/ including bin/device. When make evaluates the $(DEVICE_BIN_FOLDER) rule, it has 
  already stat'ed the directory as existing (make caches mtimes and can't see deletions performed mid-build by other recipes), so it decides the target is up to date and skips the mkdir — even though the wipe just removed it. 
  Changing how BOOTLOADER_BIN_OUT refers to the directory (order-only vs. normal prerequisite) doesn't touch that skip at all; in the repro the pre-PR and post-PR behavior are identical: fail, succeed, fail, succeed (after a  
  failed run the dir is genuinely missing at startup, so the next run works — which is probably why the PR looked like a fix when tested once). GHA is immune only because it always builds in a fresh workspace, never           
  incrementally.                                                                                                                                                                                                                  
                                                                                                                                                                                                                                  
  Why this fix works (verified deterministic across clean and incremental builds, -j1 and -j12, 7 consecutive runs):                                                                                                              
  - mkdir -p $(dir $@) inside the recipe recreates the directory after the wipe, in the same rule that needs it — immune to make's stat cache. This kills the ln: No such file or directory failure.                              
  - .FORCE means the symlink can never be judged "up to date" while the wipe has just deleted it — this kills the second, nastier symptom (a "successful" zip silently missing the bootloader). Cost is one ln per build, which is
  nothing.                                                                                                                                                                                                                        
  - | $(BIN_FOLDER) keeps the ordering so the link is created after the wipe, never before it.                                                                                                                                    
                                                                                                                                                                                                                                  
  The rm -rf itself is worth keeping — the openblt symlink name embeds date/SHA, so stale-name cleanup is exactly what that wipe is for (commit b684315f6d0). But it violates make's contract (a recipe deleting other rules'     
  outputs), so any target staged under bin/ must follow this same forced-and-self-sufficient pattern; a comment next to the rm -rf warning about that would prevent a recurrence.                                                 
                                                                                                                                                                                                                                  
  I've logged the investigation in docs/report.md. Since this is your PR review call, I haven't changed bundle.mk — happy to apply the fix to the working tree, or draft a PR comment for dron0gus with the repro, if you want    
  either.

rusefillc:
@dron0gus @FDSoftware machine suggests https://github.com/rusefi/rusefi/commit/6c466326d63762f72233d60532ab30e1f590a9db

it's in master @dron0gus please give it a try?

dron0gus:
> it's in master @dron0gus please give it a try?

I'm awk/awc for few days. :(

FDSoftware:
> @dron0gus @FDSoftware machine suggests [6c46632](https://github.com/rusefi/rusefi/commit/6c466326d63762f72233d60532ab30e1f590a9db)
> 
> it's in master @dron0gus please give it a try?

working ok with that commit now

PR #10115, opened 2026-08-19
Fix fan PWM output during bench test (#10109)

### LINKED ISSUE #10109: fan pwm output not working
The fan's PWM output doesn't work, and the test button in the console only triggers the output in on/off mode. The same pin as gppwm works fine.

**Hardware**
 uaefi rev:a with m52b25
 mre 0.6.0 testing on bench 

Fan controll:
Image
Image
output waveform on the oscilloscope : 
Image
testing the same pin as gppwm:
Image
Image

thiagomucci:
## Description
Fixed an issue where running a bench test on the fan would ignore isPwmEnabled() and return early, causing the system to trigger a direct digital output instead of driving the m_pwm structure.
## Testing
Ran local ./test.sh. I don't have the physical hardware or an oscilloscope to capture the waveform directly on the board, so validation on a physical bench setup would be appreciated!
## Format
Formatted the code using clang-format. I only committed fan_controller.cpp to keep the PR clean and avoid formatting changes across unrelated repository files.

Fixes #10109

### DIFF
--- a/firmware/controllers/algo/defaults/default_base_engine.cpp
+++ b/firmware/controllers/algo/defaults/default_base_engine.cpp
@@ -267,6 +267,16 @@ bool applyDefaultsOrFixAfterBurn(const engine_configuration_s* previousConfigura
 		changed = true;
 	}
 
+	if (engineConfiguration->fan1PwmFrequency == 0) {
+        engineConfiguration->fan1PwmFrequency = 250;
+        changed = true;
+    }
+
+    if (engineConfiguration->fan2PwmFrequency == 0) {
+        engineConfiguration->fan2PwmFrequency = 250;
+        changed = true;
+    }
+
 	if (get_board_override_result(custom_board_fix_configuration, false, previousConfiguration)) {
 		changed = true;
 	}
--- a/firmware/controllers/modules/fan_control/fan_control.cpp
+++ b/firmware/controllers/modules/fan_control/fan_control.cpp
@@ -79,8 +79,12 @@ void FanController::initPwm() {
 	if (!isBrainPinValid(getConfigPin())) {
 		return;
 	}
-	startSimplePwm(&m_pwm, "Fan PWM", &engine->scheduler, &getPin(), getPwmFrequency(), 0);
-	m_pwmInitialized = true;
+	
+	float freq = getPwmFrequency();
+    if (freq > 0) {
+        startSimplePwm(&m_pwm, "Fan PWM", &engine->scheduler, &getPin(), freq, 0);
+        m_pwmInitialized = true;
+    }
 #endif
 }
 
--- a/unit_tests/tests/actuators/test_fan_control.cpp
+++ b/unit_tests/tests/actuators/test_fan_control.cpp
@@ -2,6 +2,8 @@
 
 #include "fan_control.h"
 
+#include "defaults.h"
+
 static void updateFans() {
 	engine->module<FanControl1>()->onSlowCallback();
 }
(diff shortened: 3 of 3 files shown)


rusefillc:
human says: please do not mix code formatting and real changes into same commit and same review. Reminder that human review is very expensive and no one is excited to decipher which lines are real changes and which are code formatting changes

LLM says: the fix doesn't work as intended and introduces a regression

The PR changes `FanController::onSlowCallback()` so that during a bench test with PWM enabled it calls `initPwm()` and forces `m_pwm.setSimplePwmDutyCycle(1.0f)` instead of returning early.

### Finding 1 (blocker): any unrelated bench test now forces the fan to 100%

`isRunningBenchTest()` is a **global** flag — it's true during *any* bench test: injector, coil, fuel pump, AC relay, starter, etc. (`bench_test.cpp:43`). With this PR, running an injector bench test on a car with fan PWM enabled will spin the radiator fan at full duty for the duration of the test, every slow callback. The old code deliberately touched nothing ("let's not mess with bench testing"); the new code actively actuates hardware that isn't under test. If the intent is "while *the fan* is on the bench", the check must be `getOutputOnTheBenchTest() == &getPin()`.

### Finding 2 (blocker): during an actual fan bench test, the new duty write never reaches the pin

While a bench test runs on a pin, `OutputPin::setValue()` silently drops all non-forced writes to it (`efi_gpio.cpp:607`: `if ((isHwQcMode() || getOutputOnTheBenchTest() == this) && !isForce) return;`), and the soft-PWM generator writes non-forced (`pwm_generator_logic.cpp:403`, default `isForce = false`). So during a fan bench test the wire is still driven exclusively by the bench's forced on/off toggling — exactly as before the PR. The only observable effect of this change is the Finding 1 regression on *other* pins' bench tests. Net: the PR does not change fan-bench behavior at all.

### Finding 3: this almost certainly doesn't fix #10109

The issue's primary complaint is that **normal-mode** fan PWM produces no waveform (while GP-PWM on the same pin works). This PR only touches the bench branch. A likely real root cause sits in `initPwm()` (`fan_control.cpp:74-85`):

- `startSimplePwm()` bails out with a `warning()` when `frequency < 1` (`pwm_generator_logic.cpp:320-323`), but `initPwm()` sets `m_pwmInitialized = true` unconditionally afterward. From then on every duty write goes into a `SimplePwm` whose executor was never started — output stays dead forever, with only a one-time "low frequency 0.00 Fan PWM" console warning.
- `fan1PwmFrequency`/`fan2PwmFrequency` are new fields; the 250 Hz default is applied only via `setDefaultConfiguration()`. A tune predating the field carries **0**, and there is no migration in `applyDefaultsOrFixAfterBurn()` (grep shows no other reference to these fields). That matches the reporter's setup (existing uaefi/MRE tune) perfectly.

Suggested actual fix: (a) only latch `m_pwmInitialized` when `startSimplePwm` really started, and (b) fix up `fanNPwmFrequency == 0` in `applyDefaultsOrFixAfterBurn()` (per project convention, config mutation belongs there, not in validation). Also note the reported "stuck fully on" symptom is reproduced whenever CLT is invalid: `onSlowCallbackPwm` then commands `maxPwm` (default 100%), and 100% duty is a flat high line on a scope — indistinguishable from "not PWMing" on a bench with no CLT sensor.

### Minor notes

- The nested `#ifndef EFI_UNIT_TEST` is redundant — the whole branch is already inside `#if EFI_PROD_CODE`, and the two flags are mutually exclusive.
- Hardcoded `1.0f` ignores `getMaxPwm()`; if bench-driving the PWM were ever made to work, it should respect the configured ceiling.
- The whitespace fixes (2-space → tab) in `getState()` are unrelated churn in a functional PR, though at least confined to the same file.
- Out of scope but spotted nearby: `getOutputPinForBenchMode()` maps `BENCH_FAN_RELAY_2` to `&fanRelay` instead of `&fanRelay2` (`efi_gpio.cpp:391`) — the TS fan-2 bench button tests the wrong pin. Worth a separate issue.

**Recommendation**: close or rework the PR. The safest bench-branch behavior is the existing early return; the real #10109 fix belongs in `initPwm()` plus a frequency migration for old tunes.

I couldn't post this to GitHub — `gh` returned "Bad credentials" (`gh auth login` needed). If you re-authenticate (type `! gh auth login -h github.com` to run it in this session), I can post the review as a PR comment.

thiagomucci:
Updated the PR with the root cause fix and added unit tests, the zero-frequency migration for legacy tunes is handled in applyDefaultsOrFixAfterBurn, and PWM initialization guard is in place. All unit tests are passing locally

rusefillc:
I would have to close this PR.

In this repo expectation is that default values are not changed under "Fix fan PWM output during bench test" title

The idea of PR discipline was brought up when code formatting has leaked into first edition. That message was not sufficient. Second level of escalation is closed PR.

rusefillc:
partially picked into https://github.com/rusefi/rusefi/commit/e77eff54d517c1c54ec43bb5a53afee9518eccd9

Issue #10117, opened 2026-08-20
reduce clicks to switch from ETB to cable TPS, and back


rusefillc:
at the moment we have to touch four windows to go from default uaEFI into cable TPS configuration

and similar four windows to go back

rusefillc:
there are two potential vectors for improvement
* we can re-align menues
* but probably better add "Switch to cable TPS" button similar to existing "stepper mode" on Base tunes? Just place it not on base tunes?

rusefillc:
@Matheusschmitz what do you think?

FDSoftware:
> there are two potential vectors for improvement
> 
> * we can re-align menues
> * but probably better add "Switch to cable TPS" button similar to existing "stepper mode" on Base tunes? Just place it not on base tunes?

<img width="474" height="519" alt="Image" src="https://github.com/user-attachments/assets/ec52ba02-e7fa-4e23-a5ce-0c56f2e90fd8" />


I think it might be easier to find if we put it here in a menu called "Basic Configurations."

Then we put:

- Switch ETB to Stepper
- Switch Stepper to ETB

- Switch to Cable TPS
- Switch to PPS TPS

rusefillc:
@FDSoftware let's do it!

Matheusschmitz:
I’ve had a few ideas, but I don’t have a car with an electronic throttle body to test them on.

I’m going to buy a throttle body from a scrapyard to test how it behaves in practice.

Matheusschmitz:
@rusefillc @FDSoftware , what do you think about changing the "IDLE" menu to "Air Control"?

Grouping all menus related to engine air intake there?

Something like this:

├──  **Throttle Body Setup**
│   ├── Throttle Mode & Hardware (Cable vs DBW / ETB 1 / ETB 2)
│   ├── Pedal & TPS Calibration (PPS 1/2 & TPS 1/2 Calibration)
│   ├── Drive-by-Wire Pedal Maps (Pedal-to-TPS 3D Table)
│   └── ETB Tuning (Bias Curve Feed-Forward & PID)
│
├──  **Idle Control**
│   ├── Idle Setup & Actuator Type (ETB / PWM Valve / Stepper Motor)
│   ├── Target RPM (Target RPM vs CLT Temp Curve)
│   ├── Open-Loop / Closed-Loop PID Tuning
│   ├── Idle Cranking & Warmup Air
│   └── Idle Compensations (A/C, Electric Fan, Alternator)
│
[...]

PR #10120, opened 2026-08-20
reduce clicks to switch from ETB to cable TPS, and back

FDSoftware:
depends on #10119
image

related #10117


rusefillc:
@FDSoftware do we have some of these right on TPS dialog?

PR #10126, opened 2026-08-21
TPS dialog: add switch to Cable and PPS buttons

FDSoftware:
image

related #10117 #10127


rusefillc:
@FDSoftware let's try conditional visibility to preserve screen space? pps1==pps2==none means cable?

FDSoftware:
image
image

PR #10139, opened 2026-08-22
Fan PWM: guard initialization against non-positive frequency

### LINKED ISSUE #10109: fan pwm output not working

thiagomucci:
### Problem
When fan PWM frequency is unconfigured or zero (such as on legacy tunes or during bench tests), calling `startSimplePwm` with a non-positive frequency causes division-by-zero/assertion issues in the scheduler.

### Solution
Add a guard in `FanController::initPwm()` to only initialize and start simple PWM when `getPwmFrequency() > 0`.

Closes #10109

### DIFF
--- a/firmware/controllers/modules/fan_control/fan_control.cpp
+++ b/firmware/controllers/modules/fan_control/fan_control.cpp
@@ -79,8 +79,12 @@ void FanController::initPwm() {
 	if (!isBrainPinValid(getConfigPin())) {
 		return;
 	}
-	startSimplePwm(&m_pwm, "Fan PWM", &engine->scheduler, &getPin(), getPwmFrequency(), 0);
-	m_pwmInitialized = true;
+
+	float freq = getPwmFrequency();
+    if (freq > 0) {
+        startSimplePwm(&m_pwm, "Fan PWM", &engine->scheduler, &getPin(), freq, 0);
+        m_pwmInitialized = true;
+	}
 #endif
 }
 


rusefillc:
> When fan PWM frequency is zero , calling `startSimplePwm` with a non-positive frequency causes division-by-zero/assertion issues in the scheduler.

Does it? That's not what I see in the code. Please clarify.

```
void startSimplePwm(SimplePwm *state, const char *msg,
		Scheduler *executor,
		OutputPin *output, float frequency, float dutyCycle, pwm_gen_callback *callback) {
	efiAssertVoid(ObdCode::CUSTOM_ERR_PWM_STATE_ASSERT, state != NULL, "state");
	efiAssertVoid(ObdCode::CUSTOM_ERR_PWM_DUTY_ASSERT, dutyCycle >= 0 && dutyCycle <= PWM_MAX_DUTY, "dutyCycle");
	if (frequency < 1) {
		warning(ObdCode::CUSTOM_OBD_LOW_FREQUENCY, "low frequency %.2f %s", frequency, msg);
		return;
	}
```

PR #10147, opened 2026-08-26
Fix ProtectedResource condition wait context

Matheusschmitz:
Fixes a ChibiOS API contract violation in ProtectedResource::free(): the mutex is acquired at thread level, but the old code waited with the S-class condition API. The timeout path now restores mutex ownership before MutexLocker unwinds. Tests: focused regression test reproduced the old failure, then passed after the fix; full unit suite passed 1195/1195.


Matheusschmitz:
Closing this draft; the work will remain only on the contributor fork.

PR #10152, opened 2026-08-27
Codex/fix resource protector condvar

Matheusschmitz:
I experienced issues with the UAEFI losing LTFT table data when connecting a USB device...

Needs review

AI says:

Directly fixes the previously reported ChibiOS panic:
- ChibiOS panic
- msg SV#11
- chsys.c line 220
A protected resource was waiting for a condition using a ChibiOS "S-class" function, which requires the system to be locked beforehand via chSysLock(). This function was being called within a standard thread context.
Additionally, when the wait ended due to a timeout, the mutex could remain unlocked. The destructor would attempt to release it again, even though the thread did not hold the mutex.
The fix:
- Uses the appropriate wait function for the thread context.
- Ensures the mutex is reacquired after a timeout.
- Prevents the destructor from releasing a mutex not owned by the thread.
- Adds tests for timeouts and normal signaling.
This eliminates a genuine cause of the SV#11 error, which could trigger an ECU reboot or freeze.

### DIFF
--- a/docs/report.md
+++ b/docs/report.md
@@ -576,3 +576,27 @@ Open follow-ups:
   picks up this change.
 - Optional future hardening: same-evaluated-unit check for expressions
   once an expression evaluator with ini context is available.
+
+## 2026-08-26 - Fix ProtectedResource condition-variable wait context
+
+What was done:
+- Replaced the S-class condition-variable wait in ProtectedResource::free()
+  with the matching thread-level ChibiOS API.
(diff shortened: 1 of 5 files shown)


Matheusschmitz:
#10149

rusefillc:
Does https://github.com/rusefi/rusefi/pull/10149 address the concern?

Matheusschmitz:
> Does #10149 address the concern?
Yep

PR #10153, opened 2026-08-27
fix(sensors): defer "grab baro from MAP" until MAP has a valid sample

### LINKED ISSUE #9744: Grab baro from MAP can latch 101.325 kPa before MAP has a valid sample
### Problem

When "Grab baro value from MAP" (`useFixedBaroCorrFromMap`) is enabled, barometric pressure can still be latched at `101.325 kPa`, even though the MAP sensor later reports the correct atmospheric pressure.

Example:

- Key-on / engine-off MAP reads about `95 kPa`
- Barometric pressure remains around `101 kPa`
- "Grab baro value from MAP" is enabled

Expected behavior:

- Barometric pressure should be fixed to the actual startup MAP value, for example `95 kPa`, assuming it passes validation.

Matheusschmitz:
Fixes #9744.

LLM says:

With useFixedBaroCorrFromMap enabled, barometric pressure was latched at 101.325 kPa even though the MAP sensor reported the real atmospheric pressure a moment later.

initMapDecoder() read Sensor::get(SensorType::MapSlow).value_or(STD_ATMOSPHERE). In commonInitEngineController(), initNewSensors() only subscribes slowMapSensor to the ADC; initSensors() -> initMapDecoder() runs three lines later on the same thread, so no slow-ADC callback has fired yet and MapSlow is always invalid. The .value_or() therefore returned STD_ATMOSPHERE, validateBaroMap() accepted it (plausible range is 60..110 kPa), and Sensor::setMockValue() latched it for good.

initMapDecoder() now only arms the request; the new updateFixedBaroFromMap(), called from Engine::periodicSlowCallback(), performs the grab:

- one shot per valid sample - validate once, then latch or disable, never retry (retrying would re-run validateBaroMap() at 20Hz and spam warning())
- give up if the engine starts turning, since MAP no longer reads atmosphere
- give up after a 3s timeout with a single OBD_Barometric_Press_Circ warning, covering an unconfigured or faulted MAP channel

On every failure path BarometricPressure is left unregistered rather than mocked to STD_ATMOSPHERE, matching the pre-existing "fixed baro correction will be disabled" branch: getBaroCorrection() returns 1 when the sensor is absent, so the up-to-50ms delay before the first grab is neutral.

Adds unit_tests/tests/sensor/test_baro_from_map.cpp (5 tests) covering the deferred latch, the disabled case, an implausible reading, the engine-turning guard and the timeout.


firmware/controllers/sensors/impl/map.cpp
```diff
@@ -63,23 +63,64 @@ static void printMAPInfo() {
...
 		}
+		return;
 	}
 
+	// whatever this sample says, we only get one shot at it
+	baroFromMapPending = false;
+
+	efiPrintf("Get initial baro MAP pressure = %.2fkPa", mapKPa.Value);
+	// validate if it's within a reasonable range (the engine should not be spinning etc.)
+	float storedInitialBaroPressure = validateBaroMap(mapKPa.Value);
```
dron0gus: Making validateBaroMap() return SensorResult will simplify further code.

firmware/controllers/sensors/impl/map.cpp
```diff
@@ -63,23 +63,64 @@ static void printMAPInfo() {
...
+	if (!std::isnan(storedInitialBaroPressure)) {
+		efiPrintf("Using this fixed MAP pressure to override the baro correction!");
```
dron0gus: Print actual value instead of "this"?

dron0gus (changes requested):

dron0gus (approved):

PR #10154, opened 2026-08-27
Potential memleak fixed for lua handle, explicit constructor was adde…

alex-aparin:
This PR is related with small improvements for LuaHandle:

1. Usually it's good idea to have explicit constructor in order to prevent any accidential captures of native handles. 
2. Move assingment had potential memory leak. Currently this operator is not used in codebase, but I decided to fix it just in case.

P.S. These fixes lead me to the thought to create special allocator to check potential mem leaks in unit tests. Is it good idea to cover such cases? I may add it later in separate PR

### DIFF
--- a/firmware/controllers/lua/lua.cpp
+++ b/firmware/controllers/lua/lua.cpp
@@ -64,12 +64,12 @@ static void loadLibraries(LuaHandle& ls) {
 }
 
 static LuaHandle setupLuaState(lua_Alloc alloc) {
-	LuaHandle ls = lua_newstate(alloc, NULL);
+	LuaHandle ls(lua_newstate(alloc, NULL));
 
 	if (!ls) {
 		criticalError("Failed to start Lua interpreter");
 
-		return nullptr;
+		return ls;
 	}
 
 	lua_atpanic(ls, [](lua_State* l) {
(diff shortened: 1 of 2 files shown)


rusefillc:
@alex-aparin I am sorry but do not understand this change, partially because I am a very naive C++ developer. Please articulate the exact problem this PR is looking to address.

alex-aparin:
@rusefillc Without problems. Actually potential memory leak is in original move assignment constructor. If we have code like this 

```cpp
LuaHandle luastate1{lua_newstate(...)};
// after this lua_close for luastate1's orinigal native handle will not called
// it contains new state, moved object contains null
luastate1 = LuaHandle{lua_newstate(.. .)};
```

My fix is just to swap handles without overwite, destructor of moved object correctly close handle. Other changes are not related with mem leak.

Explicit constructor is just syntax hint to avoid pitfalls when you can accidentally capture and destroy handle. For example:
```cpp
void foo(LuaHandle);
void bar(lua_State*);
lua_State* a = ...
foo(a);
bar(a); // error! a handle is released
```


`if this == &rhs` is just check for code like `luahandle a; a = std::move(a)` it can be freely omitted at all, because at the moment lua handle does not contain any complex logic and this example looks synthetic and buggy, check is not required. If you have additional questions or examples feel free to ask

rusefillc:
* resource to review are limited
* any change is risk
* I still have no idea what this change is about - there is no  https://en.wikipedia.org/wiki/Inverted_pyramid_(journalism) and https://en.wikipedia.org/wiki/BLUF_(communication) 

with that my choice is to close this PR. Please be aware of limited resources for code review and consider costs/benefits ratio in your future contributions.

PR #10166, opened 2026-09-02
Fix #10144: Bind modal dialogs to active parent window to prevent focus loss on Linux

### LINKED ISSUE #10144: Console: alt+tab hides pop-up dialog forewer
OS: Ubuntu Linux
Start with fresh HW with only OpenBLT flashed, try to flash FW, got this message:

Image

Press Alt+Tab twice:

Image

No way to get back to dialog, no way to close Console.

thiagomucci:
### Problem Description
Resolves **#10144**. On Linux window managers, error dialogs sink behind the main application window or disappear when switching windows using `Alt+Tab`.

### Root Cause
`ErrorMessageHelper` created unparented "ghost" `JFrame` instances for popups. Linux window managers require a parent-child relationship (`WM_TRANSIENT_FOR`) to stack modal dialogs correctly above parent windows. Without it, the main window covers the dialog on window focus changes.

### Key Changes
* **Active Frame Discovery:** Updated `ErrorMessageHelper.showErrorDialog()` to scan `Window.getWindows()` for an active `JFrame` and set it as the parent for `JOptionPane`.
* **Resource Cleanup:** Added `createdNewFrame` tracking so temporary fallback frames are explicitly hidden and disposed (`dispose()`) upon dismissal, preventing memory leaks and orphaned window handles.

### Testing
* Verified build compilation (`./gradlew classes -x test`).
* Confirmed proper modal z-ordering and `Alt+Tab` retention on Linux.

### DIFF
--- a/java_console/core_ui/src/main/java/com/rusefi/core/ui/ErrorMessageHelper.java
+++ b/java_console/core_ui/src/main/java/com/rusefi/core/ui/ErrorMessageHelper.java
@@ -2,6 +2,7 @@ package com.rusefi.core.ui;
 
 import org.jetbrains.annotations.NotNull;
 
+import java.awt.*;
 import javax.swing.*;
 
 public class ErrorMessageHelper {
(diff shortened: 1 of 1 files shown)


rusefillc:
thank you! this has reminded me that we have already faced same problem and already have exactly the code your PR was suggesting https://github.com/rusefi/rusefi/commit/f1239aeccf84100ddb364128a27bd6e1dabbb726

PR #10196, opened 2026-09-08
VE table generator panel: add engine fields

### LINKED ISSUE #10195: wizard should ask engine displacement

FDSoftware:
resolves #10195 

image


rusefillc:
@FDSoftware let's fork displacement into explicit separate screen

We need to bring attention at displacement as required entry field.

All other options on the left of VE wizard are somewhat optional.

PR #10205, opened 2026-09-10
I2c preparations

dron0gus:


dron0gus:
Should not affect anything but premium board only

PR #9626, opened 2026-06-09
Logging central memory footprint

dron0gus:
DO NOT MERGE

Before:
```
Lua total heap(s) usage: 0 / 31488 bytes = 0.0%
Lua optional heap usage: 0 / 0
Lua CCM heap usage: 8 / 5256
Common ChibiOS heap: 0 bytes free
ChibiOS memcore usage: 0 / 26232
confirmation_heapinfo:8
```

After:
```
Lua total heap(s) usage: 0 / 42824 bytes = 0.0%
Lua optional heap usage: 0 / 0
Lua CCM heap usage: 8 / 10256
Common ChibiOS heap: 0 bytes free
ChibiOS memcore usage: 0 / 32568
confirmation_heapinfo:8
```

Issue #7578, opened 2025-03-07
Rusefi broadcast on Can2 just hammers the bus with messages


rusefillc:
I think there is a bug with 0x200 broadcast on 8chan. Broadcasting on bus 2 it just hammers the bus, busload 99%. Increasing the data period variable makes it worse, almost like inverse is happening. You request longer data period and bus traffic increases. I'v found lowest bus traffic at 10ms data period, which is still way too fast. With only the 8chan connected bus traffic load is over 50%.

dron0gus:
Is this issue taken into account?
<img width="371" height="162" alt="Image" src="https://github.com/user-attachments/assets/0efef138-45d9-4116-bf8c-1a55592a7b89" />
After swapping 16A and 24A in J3 I got normal CAN traffic with 7% load
` can0@500000   258   35780  16512   7% |X...................|`
RusEFI data repeated every 50mS as configured.
```
 (000.000000)  can0  200   [8]  84 36 59 08 02 00 00 00
 (000.000233)  can0  201   [8]  00 00 3A 02 00 00 00 00
 (000.000482)  can0  202   [8]  9C FF 9C FF 9C FF 9C FF
 (000.000731)  can0  203   [8]  00 00 3B 00 28 28 50 00
 (000.000980)  can0  204   [8]  00 00 00 00 28 28 7F 32
 (000.001243)  can0  205   [8]  D5 00 00 00 00 00 00 00
 (000.001493)  can0  206   [8]  00 00 00 00 00 00 00 00
 (000.001732)  can0  207   [8]  00 00 00 00 00 00 00 00
 (000.001983)  can0  208   [8]  00 00 00 00 00 00 00 00
 (000.002231)  can0  209   [8]  00 00 00 00 00 00 00 00
 (000.002482)  can0  20A   [8]  00 00 00 00 00 00 00 00
 (000.003869)  can0  00770000   [8]  00 00 00 00 00 00 00 00
 (000.004123)  can0  00770001   [8]  00 00 00 00 00 F1 F1 70
 (000.004373)  can0  00770006   [8]  00 00 00 00 00 00 00 00
 (000.050000)  can0       200   [8]  86 36 59 08 02 00 00 00
 (000.050230)  can0       201   [8]  00 00 3A 02 00 00 00 00
 (000.050483)  can0       202   [8]  9C FF 9C FF 9C FF 9C FF
 (000.050732)  can0       203   [8]  00 00 3B 00 28 28 50 00
[...]

PR #10225, opened 2026-09-14
fix "one enabled but disconnected CAN bus stalls the healthy/connected one"

FDSoftware:
related #10223

Now each CAN TX queue has its own thread, so that no bus stall or lag can affect the other bus.

Also, frames are no longer sent the moment the object is destroyed; instead, they are queued, and the worker handles sending them in the correct order.

also verbose CAN now shows TX dropped frames due timeout or bus errors


FDSoftware:
also related / fixes? [Rusefi broadcast on Can2 just hammers the bus with messages](https://github.com/rusefi/rusefi/issues/7578)
 #7578

FDSoftware:
HW CI fails because this uses 3.8kb of ram :c, now I'll see how I can reduce it, at least for F4 builds

dron0gus:
> HW CI fails because this uses 3.8kb of ram :c, now I'll see how I can reduce it, at least for F4 builds

https://github.com/rusefi/rusefi/pull/9626 huh?

firmware/hw_layer/drivers/can/can_hw.cpp
```diff
@@ -326,10 +326,8 @@ void initCan() {
 		}
 	}
 
-	// fire up threads, as necessary
-	if (engineConfiguration->canWriteEnabled) {
-		canWrite.start();
-	}
+	// ISO-TP/Lua may transmit even when the periodic CAN writer is disabled.
+	canWrite.start();
```
dron0gus: Separate PR?

dron0gus (changes requested):

firmware/hw_layer/drivers/can/can_hw.cpp
FDSoftware: test over: https://github.com/rusefi/rusefi/pull/10229 ,now I'm reverting the fix from this side to put it on another pr

edit: done

FDSoftware:
> > HW CI fails because this uses 3.8kb of ram :c, now I'll see how I can reduce it, at least for F4 builds
> 
> #9626 huh?

I ended up using a single thread; I liked the idea of ​​bus-separated workers better, but well, this way I save 2.7 kb of ram

dron0gus (approved):

PR #10244, opened 2026-09-16
Fix fans PWM mode

dron0gus:
Due to incorrect preprocessor conditions pwm was never started if PWM mode is selected for Fan 1 or Fan 2.


dron0gus:
@FDSoftware can you review?

FDSoftware (approved):

Issue #6399, opened 2024-04-23
technical debt: ADC fast not ready


rusefillc:
What's the root cause of unexpected ADC state?
```
		engine->outputChannels.fastAdcErrorsCount++;
		// todo: when? why? criticalError("ADC fast not ready?");
```

dron0gus:
High CPU load/too long critical section with disabled IRQ. Both can cause delays in ADC start.
Currently GPT -> IRQ -> start ADC conversion path is used. We can use HW trigger from timer to start ADC and skip one SW step.

PR #10301, opened 2026-09-22
ADC fast error investigation: three separate warning codes for three separate conditions

rusefillc:
733 +| C6730 / CUSTOM_FAST_ADC_DMA_COMPLETION_PENDING | DMA transfer complete flag is pending while the driver is still active. |

    734 +| C6731 / CUSTOM_FAST_ADC_CONVERSION_BUSY | DMA remains enabled with transfers outstanding and no pending completion or hardware error. |
    735 +| ISR snapshot and deferred reporting | Read-only DMA/ADC snapshot in the rejected-start branch; coalesced reason bits are drained under lock and warnings published outside the interrupt from the main loop. |
    736 +| Existing C6720 fallback | Unexpected driver states, pending hardware errors and ambiguous completion transitions are not mislabeled as one of the timing cases. |


rusefillc:
@dron0gus https://github.com/rusefi/rusefi/issues/6399

dron0gus (approved):

